PR #4, opened 2024-08-22
Deletes auth token on logout

dwwoelfel:
Deletes the auth token on logout and unhides the logout button while we work on the hamburger.


stopachka (approved):
🚢 Awesome!

PR #5, opened 2024-08-22
[Readme] Wrap`stars` with links

stopachka:
Context: 

If you click the 'stars' button, it opens up a new tab unnecessarily. I wrapped it with an `a` tag. This makes a click no-op.


dwwoelfel (approved):
That was annoying me too!

PR #6, opened 2024-08-22
Destroy token on app logout

dwwoelfel:
Destroys the user's token on `auth.signOut()`.

Changes the signature of `signOut` to return a promise, so might require a version bump.


nezaj (approved):
Nice!

PR #12, opened 2024-08-23
Add admin page that shows paid apps

dwwoelfel:
Shows a list of paid apps, with how much they're paying, how long they've been paying and their usage.

### DIFF
--- a/server/src/instant/stripe.clj
+++ b/server/src/instant/stripe.clj
@@ -10,9 +10,10 @@
    [instant.model.instant-user :as instant-user-model]
    [instant.util.tracer :as tracer]
    [instant.util.exception :as ex])
-  (:import (com.stripe Stripe)
+  (:import (com.stripe Stripe StripeClient)
            (com.stripe.model Event)
-           (com.stripe.net Webhook)))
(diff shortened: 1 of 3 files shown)


server/src/instant/stripe.clj
```diff
@@ -117,6 +118,60 @@
...
+  (let [params (-> (SubscriptionListParams/builder)
+                   (.addExpand "data.discounts")
+                   (.build))]
+    (-> (StripeClient. (config/stripe-secret))
+        (.subscriptions)
+        (.list params (RequestOptions/getDefault))
+        (.autoPagingIterable)
+        (.iterator)
+        (iterator-seq)
+        (#(map format-subscription %)))))
```
stopachka, line 171: 🫡 Love it. Great java wrangling

stopachka (approved):
🚢🚢🚢

server/src/instant/stripe.clj
dwwoelfel: This was painful--there must be a better way than all of the `clojure.reflect/reflect`s I was doing.

PR #15, opened 2024-08-23
Log session origin for better error tracking

dwwoelfel:
Logs the `cf-connecting-ip` for the request so that when we can see if errors for sessions are coming from multiple users or just the same user reconnecting.


stopachka (approved):
TIL cf-connecting-ip!

PR #16, opened 2024-08-23
Fix some grammar in the tutorial

stopachka:
We got a light note from a user on HN: 
- Missing the word 'how' 
- The sentence 'Check out the walkthrough below on your left with the full code and preview on the right.' wasn't as clear on mobile 

I went ahead and: 
- Added 'how'
- Made it so in mobile it says 'Check out the walkthrough below, and the full code example right after.'


nezaj (approved):
Great!

dwwoelfel (approved):
Looks good to me!

PR #23, opened 2024-08-23
Better error logging in cli pull

markyfyi:
We're logging status codes and HTTP messages, but not the error response body, which would be very helpful. :)


stopachka (approved):
Nice!

PR #35, opened 2024-08-25
Track how long things are queued

dwwoelfel:
Adds metrics to track how long items are in the receive queue and in the worker queue.

Also tracks the receive-q length in instant.gauges. To support this without cyclic-dependency errors, you use `gauges/add-gauge-metrics-fn` to register a function that will run whenever the gauges are calculated and should return a list of `:path, :value` maps. It returns a cleanup function you can call on `stop`.


markyfyi (approved):

stopachka (commented):
Awesome idea!

PR #37, opened 2024-08-25
[tutorial] add use client directive to the tutorial example code

stopachka:
Feedback from an Instant Hacker: 

> Hey there - really enjoying InstantDB - quick call out, on your [https://instantdb.com/tutorial](https://t.co/JPqEwHoxcs) page you may want to add in "use client"; at the top of the page.tsx example on the right. 

This makes sense; since we are explaining this is a NextJS app, it would be good to include the 'use client' directive. 


dwwoelfel (approved):
LGTM!

PR #39, opened 2024-08-26
Cancel all pending operations in `on-close`

dwwoelfel:
Keeps track of all the outstanding `event-fut` from `handle-receive` so that we can cancel them in `on-close`.

This will prevent races where a session joins a room after the on-close is called, leaving an orphaned session-id in the room that never gets removed.

Canceling all of the pending futures can be noisy, because any in-progress sql queries will throw interrupt exceptions. To prevent a bunch of errors from showing up in the logs, I added a new `tracer/with-exceptions-silencer` that we can use to prevent spans from being marked as exceptions. It will add a `silenced-exception` attribute so that we have some ability to debug if the silenced exceptions are causing problems.

### DIFF
--- a/server/.clj-kondo/config.edn
+++ b/server/.clj-kondo/config.edn
@@ -3,4 +3,5 @@
            :unresolved-var {:exclude [ring.util.http-response/set-cookie
                                       amazonica.aws.s3]}}
  :ignore [:inline-def]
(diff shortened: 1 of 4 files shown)


stopachka (approved):
Nice!

PR #40, opened 2024-08-26
Store inferred types on attrs

dwwoelfel:
Stores inferred_types on the attrs, based on the types of the triples when we add them.

We'd like inferred_types to be a set, but Postgres doesn't have a set type. The options I'm aware of are arrays, hstore, and bitset. I chose bitset because it seems like the easiest and most efficient to update. The downside is that you have a fixed number of options (32 in our case) when you define the field (but you could probably modify it to be a bigger value later).

I went with only the base types we can infer from JSON (`:string`, `:number`, `:boolean`, plus `:json` for everything else) to keep it simple. Someone may want to make an argument for supporting composite types, e.g. `array<string>`, but that gets complicated fast. We also may want to support more base types, like dates, uuids, floats, integers, etc. 

Some limitations:
- The set of types only grows--if you save a number in a field, then replace it with a string, we'll say that the type is `[:number, :string]`, even though the data is really only `[:string]`. Not sure if that's a feature or a bug--if you've set it to a number in the past, who's to say you won't try it again in the future.
  - We could try storing counts to solve this problem, ala https://github.com/cristianoc/REInfer, but I'm not sure we can reliably detect when data has been deleted.

There is a migration in `instant.db.model.triple` that will page through all of the triples and update the types for all of their attrs. It's not the most efficient thing--I tried to make use of the indexes in the query, but an explain analyze shows that postgres still has to do a lot of filtering. But I think it's fine at the current count of triples.

To support the migration, I added a new `config` table to the database that stores k/v pairs. I use the new table to keep track of the migration state so that it can be restarted if it gets interrupted. I'll wait to run the migration until after this is fully deployed so that that we get full coverage of the triples table.


stopachka (approved):
This is going to be such a huge win!

server/src/instant/jdbc/sql.clj
```diff
@@ -53,9 +53,11 @@
   [^PGobject v]
   (let [type (.getType v)
         value (.getValue v)]
-    (if (#{"jsonb" "json"} type)
-      (when value (<-json value))
-      value)))
+    (when-not (nil? value)
+      (case type
+        ("json" "jsonb") (<-json value)
```
stopachka, line 58: TIL: Did not know you could group like this in a case statement

server/test/instant/db/transaction_test.clj
```diff
@@ -1567,5 +1575,95 @@
...
+                      {:id attr-id
+                       :forward-identity [(random-uuid) "namespace" "field"]
+                       :value-type :blob
+                       :cardinality :one
+                       :unique? false
+                       :index? false}]
+                     [:add-triple target-eid attr-id value]])
+                   (catch Exception e
+                     (is (not e))))
+              (testing "deep-merge accepts top-level scalar values"
```
stopachka: nit: I think this message needs to be changed

stopachka (approved):
Beautiful!

PR #41, opened 2024-08-26
docker compose makefile improve

nascarsayan:
While running the docker compose application, if `docker-compose` command is not found, use `docker compose`.


dwwoelfel:
Thanks!

Issue #42, opened 2024-08-26
`lookup` issue


ayoung19:
I'm getting this error:
```
Unhandled Runtime Error
Error: playerId is not a unique attribute.

Source
src/components/App.tsx (43:8) @ transact

  41 |
  42 |   useEffect(() => {
> 43 |     db.transact([
     |        ^
  44 |       tx.movements[lookup("playerId", user.id)].update({
  45 |         x: movement.x,
  46 |         y: movement.y,
```

trying to do this write operation ([link](https://github.com/ayoung19/instant-ctf/blob/e730c2666dbff6606272aa52cf71fa652ff52ee8/src/components/App.tsx#L43-L48)):
```
    db.transact([
      tx.movements[lookup("playerId", user.id)].update({
        x: movement.x,
        y: movement.y,
      }),
    ]);
```

with the following schema ([link](https://github.com/ayoung19/instant-ctf/blob/e730c2666dbff6606272aa52cf71fa652ff52ee8/instant.schema.ts)):
```
import { i } from "@instantdb/core";

const graph = i.graph(
  "c7a8aae8-5bfc-43e7-9be1-6f439475a114",
  {
    movements: i.entity({
      playerId: i.string().unique().indexed(),
      x: i.number(),
      y: i.number(),
    }),
  },
  {}
);

export default graph;
```

What am I doing wrong here? I'm simply trying to update the movements entity by its unique `playerId` attribute. As a sanity check, here are the attribute settings in the dashboard: 

<img width="614" alt="image" src="https://github.com/user-attachments/assets/9d1f3c61-3ca8-42d0-88a6-8eed1d89c9f2">

dwwoelfel:
The underlying issue is that the `db.transact` happens before the client has fetched its attributes and so it doesn't know that it's a unique attribute.

Working on a fix.

dwwoelfel:
This should be fixed in 0.12.9. Please reopen and ping me if you're still having issues.

Zeb88:
> This should be fixed in 0.12.9. Please reopen and ping me if you're still having issues.

This issue seems to persist in  "@instantdb/react": "^0.12.14".  Here's my query - 

```
instantDb.transact([
    tx.readings[lookup('positionId', reading.positionId)].update(reading).link({ jobs: reading.jobId }),
 ]);
```

dwwoelfel:
@Zeb88 what's the error that you're seeing here?

dwwoelfel:
[...]

PR #46, opened 2024-08-26
Generate lookup attrs optimistically

dwwoelfel:
Fixes a problem where you can get an error when the attrs have not finished initializing before you call `transact` with a lookup (https://github.com/instantdb/instant/issues/42).

This change will optimistically generate an attr for the lookup if we don't have it locally. Just like the regular attrs, it will be


stopachka (approved):
Love it!

Issue #36, opened 2024-08-25
`id` from `@instantdb/admin` doesn't have a type unless I import `@types/uuid`


ayoung19:
This is just a minor problem and I'm not sure how it can be fixed. Maybe just adding a note in the docs when it instructs you to install `@instantdb/admin` or `@instantdb/react`?

dwwoelfel:
Thanks for reporting!

This should be fixed in `0.12.9`. Please reopen the issue if you're still hitting the issue.

PR #48, opened 2024-08-26
Wrap `id` so that it exports the type

dwwoelfel:
Wraps the uuid generator with our own function so that users don't have to import `@types/uuid` to get a type for `id`.

Fixes https://github.com/instantdb/instant/issues/36


stopachka (approved):
Yes!

PR #51, opened 2024-08-26
Remove current-user when the token is bad on init

dwwoelfel:
If a user's token was revoked, we would never clear the `user` object out of localstorage so it would look like they were still logged in.

With this change, we will unset the current user if `init` returns a user not found error.


stopachka (approved):
Heck yeah!

PR #53, opened 2024-08-26
Revamped CLI login

markyfyi:
Rewrote CLI login due to Safari's and other browsers's restrictions on our localhost method.

A high-level overview:

1. CLI asks server for a token pair: a ticket (shared with the browser) and a secret (stored in-memory and only shared with the server).
2. CLI opens a browser window with the ticket in the URL.
3. User authenticates as normal.
4. Once the user authenticates, they claim the ticket, associating it to their ID.
5. CLI polls server with the secret until it's associated ticket has been claimed.  Once it's claimed, the poll returns an authentication token that the CLI can use for subsequent requests.

The user has 2 minutes to complete the flow before the token pair expires.  A claimed token can only be read once.

In order to accomplish this, I tweaked our Oauth flow, adding a `ticket` column to `instant_oauth_redirects` so that we're able to remember it across all the redirects.

I also added some UX to Dash, rendering a modal that informs the user that they've successfully authenticated and can return to the CLI.

@stopachka @dwwoelfel


client/packages/cli/index.js
```diff
@@ -92,15 +92,44 @@ program.parse(process.argv);
...
+    {
+      method: "POST",
+      headers: {
+        Accept: "application/json",
+      },
+    },
+  );
+
+  if (verbose) {
+    console.log(
```
stopachka, line 106: aside: makes me think we may want to just `wrap` fetch, so all fetch's handle the verbose flag by default

client/sandbox/cli/instant.schema.ts
```diff
@@ -0,0 +1,53 @@
+// foo
```
stopachka: nit: I am not sure what this was meant to be

client/sandbox/cli/instant.schema.ts
```diff
@@ -0,0 +1,53 @@
...
+// http://localhost:3000/dash?s=main&t=home&app=b3b4bbd7-0bb9-4f83-95d4-e4d488921c75
+
+import { i } from "@instantdb/core";
+
+const INSTANT_APP_ID = "b3b4bbd7-0bb9-4f83-95d4-e4d488921c75";
```
stopachka: nit: Let's use dot-env.

client/sandbox/cli/instant.schema.ts
```diff
@@ -0,0 +1,53 @@
...
+
+// Example entities and links (you can delete these!)
```
stopachka: I am not 100% sure I understand what you mean by this comment

client/sandbox/cli/package.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "sandbox-cli",
+  "private": true,
+  "dependencies": {
+    "@instantdb/admin": "workspace:*",
+    "@instantdb/core": "workspace:*",
+    "instant-cli": "workspace:*"
```
stopachka: nit: I would add a `README.md`. Perhaps we should also name this with the same nomenclature as the other sandbox apps?

client/www/components/dash/Auth.tsx
```diff
@@ -164,13 +171,23 @@ export default function Auth(props: { emailOnly?: boolean; info?: ReactNode }) {
...
-        isLoading: false,
-        error: errorFromVerifyMagicCode(err),
-      }));
-    });
+    verifyMagicCode({ email, code }).then(
+      ({ token }) => {
+        const ticket = props.ticket;
+        if (!ticket) return;
+
+        claimTicket({ ticket, token }).then(() => {
```
stopachka: I notice we have the `claimTicket` already in the Dash Root. Is it necessary to call it here?

My thinking is, it should only be called there, as it doesn't matter from 'where' the ticket comes

client/www/pages/dash/index.tsx
```diff
@@ -135,20 +174,31 @@ function isTabAvailable(tab: Tab, role?: Role) {
...
   } | null>(null);
 
   const dashResponse = useAuthedFetch<DashResponse>(`${config.apiURI}/dash`);
 
+  // new CLI login flow
+  useEffect(() => {
+    if (!token) return;
+    if (!ticket) return;
+    claimTicket({ ticket, token }).then(() => {
+      onClaimTicket();
```
stopachka: Should we handle the failure case? For example, if the user starts the auth process, but closes their cli, in that case, this 'ticket' would fail?

client/packages/cli/index.js
```diff
@@ -637,25 +666,35 @@ async function saveConfigAuthToken(authToken) {
...
-        res.statusCode = 200;
-        res.end();
-        server.close();
+function sleep(ms) {
+  return new Promise((resolve) => setTimeout(resolve, ms));
+}
+
+async function waitForAuthToken({ secret }) {
+  for (let i = 1; i <= 24; i++) {
+    await sleep(i * 5000);
```
stopachka: Do I understand correctly, that the flow is: 

```
Wait 5 seconds, then check
If nothing, wait 10 seconds and check
If nothing, wait 15 seconds and check
```

I worry 5 seconds may be a bit too long even in the happy case, and that afterwards the experience will feel slower.

stopachka (commented):
Awesome push!

server/resources/migrations/26_cli_auth.up.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE instant_cli_auth (
+  id uuid PRIMARY KEY,
+  secret uuid NOT NULL,
+  used boolean NOT NULL DEFAULT false,
+  created_at timestamp NOT NULL DEFAULT now(),
+  user_id uuid REFERENCES instant_users(id) ON DELETE CASCADE
```
dwwoelfel, line 6: Can you add an index on `user_id`? That way when a user is deleted, postgres doesn't have to do a full table scan on the `instant_cli_auth` table.

server/resources/migrations/26_cli_auth.up.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE instant_cli_auth (
```
dwwoelfel: Might want to call this `instant_cli_auths` to keep with the local custom.

server/src/instant/dash/routes.clj
```diff
@@ -1090,6 +1095,38 @@
...
+    (instant-cli-auth-model/create!
+     aurora/conn-pool
+     {:secret secret
+      :ticket ticket})
+    (response/ok {:secret secret :ticket ticket})))
+
+(defn cli-auth-claim-post [req]
+  (let [{user-id :id} (req->auth-user! req)
+        ticket (ex/get-param! req [:body :ticket] uuid-util/coerce)]
+    (println "claiming ticket" ticket)
```
dwwoelfel: debug code

server/resources/migrations/26_cli_auth.up.sql
```diff
@@ -0,0 +1,11 @@
...
+  id uuid PRIMARY KEY,
+  secret uuid NOT NULL,
```
dwwoelfel: The database itself should probably store a hash of the `secret` to prevent timing attacks. Not super likely here, since you'd have to have the ticket id too, but good to keep in the habit.

Example for how we store the oauth codes (the column in the table is bytea instead of uuid): https://github.com/instantdb/instant/blob/main/server/src/instant/model/instant_oauth_code.clj#L16

server/resources/migrations/26_cli_auth.up.sql
markyfyi: Nice!  Went with `instant_cli_logins`

client/packages/cli/index.js
markyfyi: Dropped to 1

client/sandbox/cli/instant.schema.ts
markyfyi: This is generated by `instant-cli`

client/sandbox/cli/instant.schema.ts
markyfyi: App name, generated by `instant-cli`

server/resources/migrations/26_cli_auth.up.sql
markyfyi: Thanks, PTAL!

client/packages/cli/index.js
markyfyi: Yeah this file is due for some cleanup/helpers

stopachka:
Great work @markyfyi!

client/packages/cli/index.js
```diff
@@ -92,15 +92,44 @@ program.parse(process.argv);
...
+    `This will open Instant in your brower (${authUrl}), OK to proceed?`,
   );
 
   if (!ok) return;
 
   open(authUrl);
 
-  const { token, email } = await execMagicLocalhostCallback(magicLocalhostPort);
+  console.log("Waiting for authentication...");
+  const { token, email } = await waitForAuthToken({ secret });
```
dwwoelfel: What happens if `waitForAuthToken` doesn't get a token after 120 seconds?

dwwoelfel (approved):
Beautiful!

PR #56, opened 2024-08-27
Fix storage whitelist query

reichert621:
Wrap storage whitelist query in try catch in case namespace or attrs do not exist


dwwoelfel (approved):
LGTM!

PR #57, opened 2024-08-27
[index] reduce jitter from SSR

stopachka:
There were to 'jittery' bugs:
- We didn't have widths on our logo and yc icon. This led to a horizontal jitter
- I also realized our blurup for the video was too dark. It's because it was using an image from the wrong thumbnailTime. I updated it. 

**Before**

https://github.com/user-attachments/assets/b7a06393-ee50-4e16-976d-2b8c2fd4eedc

**After**

https://github.com/user-attachments/assets/d68396f9-c357-4e46-ad7c-ccd3f8777b47


dwwoelfel (approved):
Sweet!

PR #61, opened 2024-08-27
Fix `type(data.property)` in cel

dwwoelfel:
This fixes a bug where `type(data.property)` would throw an exception when the property was an integer.

cel-java doesn't know how to determine the type of integers, it only knows longs. So we convert integers to longs when we retrieve them from the map.

Here's where the type check happens: https://github.com/google/cel-java/blob/dae82c6d10114bb1da643203569f90a757c6c5e6/runtime/src/main/java/dev/cel/runtime/StandardTypeResolver.java#L73-L105

As a bonus, it also fixes checking if something is false. Previously evaluating `data.bool_prop == false` would return false if it got `data={bool_prop: false}` because the `or` would return `NULL_VALUE` on false.

### DIFF
--- a/server/src/instant/db/cel.clj
+++ b/server/src/instant/db/cel.clj
@@ -171,7 +171,18 @@
   (proxy [java.util.AbstractMap] []
     ;; If a value is not found, we must return a null value that CEL understands
     (get [k]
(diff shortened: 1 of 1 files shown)


stopachka (approved):

stopachka:
Niice!

PR #62, opened 2024-08-27
Fix tutorial Monaco model error

markyfyi:
Monaco internals were throwing, not entirely sure why, but giving each model a path resolved the issue.  Also, DRY'd up some code.


dwwoelfel (approved):
LGTM!

PR #69, opened 2024-08-27
Fix bug where duplicated attrs would cause transaction to fail

dwwoelfel:
Fixes a bug where duplicate `add-attr` tx-steps would cause the transaction to fail.

The bug was introduced yesterday when I fixed the problem with `lookup` not working until `init` finished: https://github.com/instantdb/instant/commit/ef8b6b5811b99cfb31480d7d6ea59dd4de5e90a3

Now we deduplicate those on the server.


stopachka (approved):
Love it!

PR #70, opened 2024-08-27
CLI auth confirmation step

markyfyi:
Adds an additional layer of security by prompting the user to allow/deny CLI auth in the browser.


stopachka (approved):
SGTM!

PR #71, opened 2024-08-27
Add support for `config/overlay.edn` for local overrides

dwwoelfel:
If you create a `server/resources/config/overlay.edn`, we'll merge the values there with the values we normally pick up from config. Useful for adding small local-only changes to the config.


stopachka (approved):
Love it!

PR #72, opened 2024-08-27
[cli][schema][attrs] Fix parsing of user schema defs into instant schema

markyfyi:
We forgot to update the backend after renaming all of the keys! 😱


stopachka (approved):
🚢🚢🚢

PR #73, opened 2024-08-27
[schema][cli] don't codegen id attr

markyfyi:
`id` is implicit and managed at the system-level.  Users shouldn't have to define it in the schema file, and we shouldn't generate it for them, either. ;)


stopachka (approved):
🫡

PR #74, opened 2024-08-27
Fix query validation errors

dwwoelfel:
Fixes the query validation error where the user expected a link in the query, but the attr is not a ref, e.g.

```
{users: { handle: {} }}
```

It also makes the server return a validation error in this case instead of treating it like a attr that hasn't been defined yet.


stopachka (approved):
Nice!

PR #75, opened 2024-08-27
[cli][schema] Skip id attr name when applying schema

markyfyi:
`id` attr is automatically generated, so omit adding it twice if the user explicitly defined it


stopachka (approved):
🚢🚢🚢

PR #77, opened 2024-08-27
Persist selected namespace in dashboard explorer

reichert621:
Right now if you select a namespace in the dashboard explorer, and then refresh the page, it reverts to selecting the first namespace in the app db.

This PR adds a query param that persists the selected namespace id as `ns`.


stopachka (approved):
Yeess!

Honestly makes me think: Maybe we should even persist namespace in localStorage too. This way when loading dash, you always load your 'favorite' app of sorts

reichert621:
good idea @stopachka! can handle that in a followup PR :)

PR #83, opened 2024-08-28
Reconnect the wal listener on connection errors

dwwoelfel:
Adds reconnect logic when the wal listener connection fails.

We don't want to miss any updates while the connection is down, so this PR switches our replication slot from temporary to non-temporary.

When the slots were temporary, they automatically got cleaned up when the connection closed. We need to make sure that we clean up old slots so that postgres can delete the wal logs.

There are two things we do to keep on top of inactive slots:

1. There is a new shutdown handler that will call `stop` on the invalidator, which will shutdown the wal listener and drop the slot. It works if you hit Ctrl+c locally, but I'll have to monitor to see if it actually works in elastic beanstalk.
2. There is a background process that will check for inactive slots every hour and drop them if they're still inactive 5 minutes later.

Other changes:
1. Uses `read` on the stream instead of `readPending`. `read` is blocking, so we won't have to do the `Thread/sleep` and we should have slightly less latency on handling updates. I'm not sure why it was switched to `readPending` before, so I might be missing something important here.
2. Better logging of lists in the logging_exportert (no more `list=clojure.lang.LazySeq@7861`)

Tested by running `select pg_terminate_backend(active_pid) from pg_replication_slots` and checking that the invalidator still worked afterwards.

### DIFF
--- a/server/src/instant/core.clj
+++ b/server/src/instant/core.clj
@@ -25,6 +25,7 @@
    [instant.util.tracer :as tracer]
    [instant.reactive.store :as rs]
    [instant.reactive.session :as session]
+   [instant.jdbc.wal :as wal]
    [instant.reactive.invalidator :as inv]
    [instant.reactive.ephemeral :as eph]
    [instant.session-counter :as session-counter]
@@ -105,6 +106,20 @@
   (stop)
   (start))
 
+(defn add-shutdown-hook []
+  (.addShutdownHook (Runtime/getRuntime)
+                    (Thread. (fn []
+                               (tracer/record-info! {:name "shut-down"})
+                               (tracer/with-span! {:name "stop-server"}
+                                 (stop))
+                               (tracer/with-span! {:name "stop-invalidator"}
+                                 ;; Hack to get the invalidator to shut down in
+                                 ;; dev. Otherwise the stream takes forever to close.
+                                 (when (= :dev (config/get-env))
+                                   (future (Thread/sleep 100)
+                                           (wal/kick-wal aurora/conn-pool)))
+                                 (inv/stop))))))
+
 (defn -main [& _args]
   (let [{:keys [aead-keyset]} (config/init)]
     (crypt-util/init aead-keyset))
@@ -130,9 +145,11 @@
   (stripe/init)
   (session/start)
   (inv/start)
+  (wal/init-cleanup aurora/conn-pool)
   (ephemeral-app/start)
   (session-counter/start)
   (when (= (config/get-env) :prod)
     (log/info "Starting analytics")
     (analytics/start))
-  (start))
+  (start)
+  (add-shutdown-hook))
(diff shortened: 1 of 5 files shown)


server/src/instant/core.clj
```diff
@@ -105,6 +106,20 @@
...
+(defn add-shutdown-hook []
+  (.addShutdownHook (Runtime/getRuntime)
+                    (Thread. (fn []
+                               (tracer/record-info! {:name "shut-down"})
+                               (tracer/with-span! {:name "stop-server"}
+                                 (stop))
+                               (tracer/with-span! {:name "stop-invalidator"}
+                                 ;; Hack to get the invalidator to shut down in
+                                 ;; dev. Otherwise the stream takes forever to close.
+                                 (when (= :dev (config/get-env))
```
stopachka, line 118: Question: should we do this in prod too? 

My thinking is there could be some situation where Instant goes down, and all database writes stop

stopachka (approved):
Super thoughtful PR! So darn happy to kick this error's bum

server/src/instant/jdbc/wal.clj
```diff
@@ -159,16 +202,45 @@
 
    We do some book-keeping for the replication stream, by recording the LSN
    for the last record that was pushed to `to`."
-  [stream to shutdown?]
-  (while (not @shutdown?)
-    (let [buffer (.readPending stream)]
+  [stream to close-signal-chan]
+  (loop []
+    (let [buffer (.read stream)]
```
stopachka, line 207: Some context on why this was `.readPending` before: 

https://stackoverflow.com/questions/76074375/safely-stopping-pgreplicationstream 

I wasn't able to interrupt the thread when it was blocked on `.read`. I think `kick-wal` is a great goaround for this though! 

I remember searching all over github for good examples of PGReplicationStream. Glad we get to contribute this back into the collective conciousness.

dwwoelfel: Interesting--I was always able to get `.close` to eventually complete, it just took a while (~20 seconds max) when there was no activity on the wal. Let me experiment a little more, though.

server/src/instant/core.clj
dwwoelfel: (discussed in person, but will recap here for posterity)

In prod, it should be fine if the close gets stuck because elasticbeanstalk will force it to shut down eventually after it sends the initial shutdown signal (assuming it sends one we can respond to and doesn't just send SIGKILL). Before that happens, eb will drain the connections and we call server/stop to force anyone who is connected to reconnect to a new server.

server/src/instant/jdbc/wal.clj
dwwoelfel: `.close` always eventually completed for me, but the max was a little over a minute. 

If we set up `component` or something similar and we're doing a ton of restarting in development, I'll set up `kick-wal` to run whenever we restart. I updated `kick-wal` to just do `pg_notify`, which seems to work and is less dumb.

PR #85, opened 2024-08-28
Allow empty `:in` in the where clause

dwwoelfel:
Allows a query like `{:users {:$ {:where {:id {:in []}}}}}` to succeed (with an empty set of join rows).

It would be more efficient to skip the trip to the db here, but I worry about how complicated that will make queries where a nested ref query has an empty `in` array or where there are multiple sibling top-level queries.


stopachka (approved):
Nice!

PR #86, opened 2024-08-28
Strip newlines when logging to stdout in production

dwwoelfel:
This should make it easier to search and view the logs in production because the exceptions won't span multiple separate log lines.


stopachka (approved):
Love it!

PR #87, opened 2024-08-28
Fix for duplicate ref attrs

dwwoelfel:
Fixes a bug on the client where we would add duplicate ref attrs if you tried to add the reverse and forward ref in the same transaction, here's a simple example:

```js
const userId = id();
const bookId = id();
db.transact([
  tx.users[userId].update({}).link({books: bookId}),
  tx.books[bookId].update({}).link({users: userId})
]);
```

We would try to create two ref attrs for that transaction, which would throw an error because two idents tried to be `users.books` (or `books.users`).

Now we use a reduce to build up the attrs and we don't run into that problem.

Fixes https://github.com/jsventures/instldraw/issues/3


nezaj (approved):
Nice!

PR #89, opened 2024-08-29
Strong init: support schema in the core libraries

markyfyi:
Allows supplying `instant.schema.ts` definition to `init`, in both core and react.

- `useQuery` will typecheck the InstaQL param and provide a strongly typed response.
- Adds `db.tx`, similar to global `tx`, but with strongly typed namespaces, plus `update` and `link` inference.
- `useQuery` will infer cardinality from the schema, both statically and at runtime, in both directions.  e.g., if there's a link with forward cardinality = :one, `useQuery` will yield a single item instead of an array.

The only changes required to get this working are to supply a schema to `init`, and use `db.tx` instead of importing `tx`.

```ts
// in instant.schema.ts...
const schema = i.graph(/* ... */);

// in your app's setup...
const db = init({
  appId,
  schema, // <-- your schema here
});

// later...
tx.foos[id()].update({ title: 'foo!' }); //  ❌ don't use global tx anymore
db.tx.foos[id()].update({ title: 'foo!' }); // ✅ use db.tx instead of importing global tx
```


client/www/lib/hooks/explorer.tsx
```diff
@@ -38,7 +38,6 @@ export function useNamespacesQuery(
       ? {
           [selectedNs.name]: {
             $: {
-              // @ts-expect-error: admin-only feature
```
markyfyi, line 41: Not sure why this stopped throwing. ¯\\_(ツ)_/¯  Intellisense still works here, it just doesn't complain about aggregate anymore

client/packages/admin/src/index.ts
```diff
@@ -389,7 +391,7 @@ class InstantAdmin<Schema = {}> {
    *   )
    */
```
stopachka, line 538: nit: would we want to add `deprecated` + a note about these functions?

client/packages/core/__tests__/src/instaql-query.test.js
```diff
@@ -0,0 +1,112 @@
...
+    {
+      id: ids.profile1,
+      user: {
+        id: ids.user1,
+      },
+    },
+  ],
+};
+
+test("Simple Query", () => {
```
stopachka: filename nit: afaik we generally use camelCase for file names. I know the modern js convention is kebab case, and I'm down, but would be nicer if we did a rename in one go. 

filename nit 2: I would consider a different name, as instaql.test is the same as `instaql-query` really

I know it's on the plate, but It would be great if we expanded this to show all the different permutations of [one, many] | [one, many]

client/packages/core/src/instaql.js
```diff
@@ -106,7 +106,9 @@ function refAttrPat(makeVar, store, etype, level, label) {
 
   const nextEtype = fwdAttr ? revEtype : fwdEtype;
 
-  return [nextEtype, nextLevel, attrPat];
+  const dir = fwdAttr ? "forward" : "reverse";
```
stopachka: nit: I would consider having this be an `isForward` variable, as the string implies an enum of more than 2.

client/packages/core/src/instaql.js
```diff
@@ -258,21 +260,33 @@ function extendObjects(makeVar, store, { etype, level, form }, objects) {
...
         });
-        return { [label]: child };
+
+        let children = childrenArray;
+        if (store.schema) {
+          const isForwardCardinalityOne =
+            dir === "forward" && attr.cardinality === "one";
+          const isReverseCardinalityOne = dir === "reverse" && attr["unique?"];
+
+          if (isForwardCardinalityOne || isReverseCardinalityOne) {
```
stopachka: nit: I would call this `isReverseUnique`. I do think `cardinality` is a good name for what we want, but would be nice to keep it with it's current meaning for now.

client/packages/core/src/instaql.js
```diff
@@ -258,21 +260,33 @@ function extendObjects(makeVar, store, { etype, level, form }, objects) {
...
+        const childrenArray = queryOne(store, {
           etype: nextEtype,
           level: nextLevel,
           form: form[label],
           join,
         });
-        return { [label]: child };
+
+        let children = childrenArray;
+        if (store.schema) {
```
stopachka: One thing I realize: 

What if someone _doesn't_ want to use schema yet, but _does_ want this behavior?

I guess we could have another flag. 

Perhaps it's not a problem though 🫡

client/packages/core/src/instatx.ts
```diff
@@ -6,7 +8,10 @@ type LookupRef = [string, any];
 type Lookup = string;
 export type Op = [Action, EType, Id | LookupRef, Args];
 
-export interface TransactionChunk {
+export interface TransactionChunk<
+  S extends InstantGraph<any, any>,
```
stopachka: nit: would use more verbose variable names. Also surprised about the `any` here.

client/packages/core/src/instatx.ts
```diff
@@ -15,7 +20,18 @@ export interface TransactionChunk {
    *  const goalId = id();
    *  tx.goals[goalId].update({title: "Get fit", difficulty: 5})
    */
-  update: (args: { [attribute: string]: any }) => TransactionChunk;
+  update: (
+    args: {
+      [K in keyof S["entities"][E]["attrs"]]?: S["entities"][E]["attrs"][K] extends DataAttrDef<
```
stopachka: nit: would abstract the types for `update` `link` and `unlink`, to as they may look a bit unwieldy in intellisense. 

One stackoverflow that may be of interest to you: 

https://stackoverflow.com/questions/77867172/is-it-possible-to-produce-outputt-which-visually-omits-some-keys-in-t

client/packages/react-native/src/index.ts
```diff
@@ -36,10 +36,15 @@ import { RoomSchemaShape, id, tx } from "@instantdb/core";
...
-  config: Config,
-) {
-  return new InstantReactNative<Schema, RoomSchema>(config);
+function init<
+  Schema = {},
+  RoomSchema extends RoomSchemaShape = {},
+  Config_ extends Config | ConfigWithSchema<any> = Config,
+>(config: Config_) {
+  return new InstantReactNative<
+    Config_ extends ConfigWithSchema<infer CS> ? CS : Schema,
```
stopachka: daam

stopachka (commented):
So exciting!

client/packages/admin/src/index.ts
markyfyi: Note: will deprecate global `tx` down the road

client/packages/core/src/instatx.ts
markyfyi: This is a pattern I see a lot in TS. `any` acts as a placeholder, kinda like to `InstantGraph<_, _>`.  We _could_ substitute it with the types the generic param is extending (i.e. `InstantGraph<EntitiesDef, LinksDef<EntitiesDef>>`), but `InstantGraph` already constrains it.

markyfyi: Improved the generic param names tho!

client/packages/admin/src/index.ts
```diff
@@ -235,6 +348,18 @@ function init<Schema = {}>(config: Config) {
   return new InstantAdmin<Schema>(config);
 }
 
+function init_experimental<
```
stopachka, line 351: nit: we can land in follow-on PR, but i.m.o would add docstring

client/packages/admin/src/index.ts
```diff
@@ -244,12 +369,19 @@ function init<Schema = {}>(config: Config) {
...
-class InstantAdmin<Schema = {}> {
+class InstantAdmin<Schema extends i.InstantGraph<any, any> | {} = {}> {
   config: FilledConfig;
   auth: Auth;
   storage: Storage;
   impersonationOpts?: ImpersonationOpts;
 
+  public tx =
+    txInit<
+      Schema extends i.InstantGraph<any, any>
```
stopachka, line 380: woohoo!

client/packages/core/__tests__/src/instaqlInference.test.js
```diff
@@ -0,0 +1,210 @@
...
+    user1: id(),
+    profile1: id(),
+  };
+
+  const { result } = queryData(
+    { inference: false },
+    [
+      {
+        id: ids.usersProfilesLink,
+        "value-type": "ref",
```
stopachka, line 148: nit: (not necc for this PR, but i.m.o later we could abstract these into quick 'attr' creation functions

client/packages/core/src/instaql.js
```diff
@@ -258,21 +260,34 @@ function extendObjects(makeVar, store, { etype, level, form }, objects) {
...
```
stopachka: cc @markyfyi instead of using schema as the flag, I would use cardinalityInference directly. 

Would also extract this bit into a helpful function

stopachka (approved):
Ship it!

Issue #45, opened 2024-08-26
Feature Request: Stripe Connect but for Instant DB


vikeri:
# Feature request: Stripe Connect but for Instant DB
## BLUF
We’d like to provision Instant DB accounts and apps on behalf of our users, similar to how Squarespace can provision Stripe accounts and payments on behalf of their users. 
## Need
[GPT Engineer](https://gptengineer.app) lets you build and deploy a complete frontend app using only prompts (powered by LLMs). For adding persistence we today support a Supabase integration. This requires the user to go into supabase, create an account, and configure the schema and then use the API keys to Connect to GPT Engineer. This is great for users that already use Supabase or who want a lot of control. But many users don’t care about what a persistence layer looks like as long as it gets the job done and would prefer minimal friction. We believe that Instant DB could unlock this low friction use case if there was a way for us to programmatically create Instant DB apps on behalf of our users. **We don't want our users to have to go to instantdb.com and sign up, we'd want to sign up for them, ideally without them needing to do anything.**
## Proposed solution
A couple of API endpoints (not exactly how they should look but something to start the discussion):
### `POST /users`
```
curl https://api.instantdb.com/v1/users \
  -u "our_secret_api_key" \
  -d email=test@example.com
  -d password=long_difficult_password

Returns
{
  id: string;
  email: string;
}
```
### `GET /users/<id>`
```
curl https://api.instantdb.com/v1/users/<id> \
  -u "our_secret_api_key" \
  -d email=test@example.com

Returns
{
  id: string;
  email: string;
}
```
### `POST /app`
```
curl https://api.instantdb.com/v1/app \
  -u "our_secret_api_key" \
  -d user_id="user_id" \
  -d name="cool_app"

Returns
{
  name: string;
  id: string;
  public_api_key: string; // The user's client side public API key
}
```
### `GET /app/<id>`
```
curl https://api.instantdb.com/v1/app/<id> \
  -u "our_secret_api_key"

Returns
{
  name: string;
  id: string;
  public_api_key: string; // The user's client side public API key
}
```

### Validating users
One possible concern here would be the validation of users and their emails. We naturally have our users already validated so either we establish trust that way or we device some other way of establishing trust. Ideally we'd like to avoid the user having to re-validate their email again for example.
### Giving users access to their Instant DB dashboards
[...]

PR #93, opened 2024-08-29
[superadmin] CRUD apps

reichert621:
Some user’s want the ability to spin up apps and add collaborators to apps programmatically. Basically, we want to expose a subset of dashboard actions in a documented and easy-to-use API. 

This PR sets up the boilerplate required for creating apps within an account programmatically.

Related: https://github.com/instantdb/instant/issues/45

Dashboard:
CleanShot 2024-09-09 at 12 48 18@2x

**Fast-follows** 

We still need to implement a 'transfer' app flow. After that, we can add schema and perms push / pull to the admin API. Once that's done, we're at a good point for the platform API!


server/src/instant/superadmin/routes.clj
```diff
@@ -0,0 +1,60 @@
...
+            [instant.util.uuid :as uuid-util]
+            [instant.model.app :as app-model]
+            [instant.model.instant-user :as instant-user-model]
+            [instant.util.exception :as ex]
+            [instant.util.http :as http-util])
+
+  (:import
+   (java.util UUID)))
+
+(defn req->auth-user! [req]
```
nezaj: Nit. I wonder if we should make shared module under `/auth` since this is now duplicated in `/dash/routes`

nezaj (approved):
Heck yeah!

server/src/instant/superadmin/routes.clj
reichert621: (now using personal access tokens, so this is no longer duplicated)

server/src/instant/model/app.clj
```diff
@@ -36,6 +36,38 @@
 (defn get-by-id! [params]
   (ex/assert-record! (get-by-id params) :app {:args [params]}))
 
+(defn list-account-apps
+  ([user-id] (list-account-apps aurora/conn-pool user-id))
+  ([conn user-id]
```
stopachka, line 41: nit: this would only list the apps where the user is the creator. afaik this is good, but would just call it `list-created-apps` with `creator-id`. 

Since we are in the `app` namespace too, maybe we can it.

`list-by-creator-id`

server/src/instant/model/app.clj
```diff
@@ -36,6 +36,38 @@
...
+               ["SELECT a.*
+                FROM apps a
+                WHERE a.creator_id = ?::uuid"
+                user-id])))
+
+(comment
+  (def u "6412d553-2749-4f52-898a-0b3ec42ffd28")
+  (list-account-apps u))
+
+(defn get-account-app-by-id
```
stopachka: nit: maybe would call this `get-one-by-creator-id`. Reasoning: right now it sounds like it is returning an `account`, but it's actually 'get one app by id and creator'

stopachka (approved):
Let's ship it! 🚢

server/src/instant/model/app.clj
reichert621: what about `get-creator-app-by-id`? basically i want "get-app-by-id" for a specific creator... do you still like `get-one-by-creator-id` for this?

stopachka: Maybe 

`get-by-creator-and-app-id`? 

Admittedly it is looking verbose

Right now the heuristic is something like 

`get-{name-of-model, but removed for brevity}-{qualifier}`

For example, there's a `get-by-id` in there. 

cc @dwwoelfel if you have thoughts too!

reichert621: how about... `get-by-id-and-creator`? 😛

stopachka: Love it!

PR #96, opened 2024-08-29
Admin storage SDK

reichert621:
- Adds storage functions to `@instantdb/admin` package
  - `db.storage.upload`
  - `db.storage.list`
  - `db.storage.delete`
  - `db.storage.deleteMany`
- Renames browser SDK `put` to `upload` (`put` still supported, but deprecated)
- Updates storage docs


server/src/instant/admin/routes.clj
```diff
@@ -340,6 +344,61 @@
                      :headers {"authorization" (str "Bearer " admin-token)
                                "app-id" (str counters-app-id)}}))
 
+
+;; ---
+;; Storage
+
+(defn signed-download-url-get [req]
+  (let [{app-id :app_id} (req->admin-token! req)
+        filename (ex/get-param! req [:params :filename] string-util/coerce-non-blank-str)
```
stopachka, line 351: Nit: Looks like apart form the perm check, these are going to be very similar to the client endpoints. i.m.o it may make sense to have them as general-purpose functions, that access some kind of 'perms' object. This way, we can use them in both places

stopachka (approved):
Excited!

server/src/instant/util/storage.clj
```diff
@@ -0,0 +1,207 @@
...
+                          {"auth" (cel/->cel-map (<-json (->json current-user)))
+                           "data" (cel/->cel-map {"path" filepath})})))))
+
+
+(defn upload-image-to-s3 [app-id filename image-url]
+  (let [object-key (->object-key app-id filename)]
+    (storage-beta/assert-storage-enabled! app-id)
+    (s3-util/upload-image-to-s3 object-key image-url)))
+
+(defn create-signed-download-url!
```
reichert621, line 56: lmk if the multi-arity situation here is getting a bit too wonky!

basically wanted to support:
- browser client use case (refresh token needed)
- admin/dashboard use case (no refresh token needed to verify permissions)
- testing use case (mock rules)

is this the best way to handle?

reichert621: (cc @stopachka @dwwoelfel)

client/packages/admin/src/index.ts
```diff
@@ -556,6 +558,142 @@ class Auth {
...
+  /**
+   * Retrieves a download URL for the provided path.
+   *
+   * @see https://instantdb.com/docs/storage
+   * @example
+   *   const url = await db.storage.getDownloadUrl('photos/demo.png');
+   */
+  getDownloadUrl = async (pathname: string): Promise<string> => {
+    const { data } = await jsonFetch(
+      `${this.config.apiURI}/admin/storage/signed-download-url?app_id=${this.config.appId}&filename=${pathname}`,
```
dwwoelfel: Will we ever need to urlencode the `pathname` or is it always url-safe?

reichert621: good point -- probably better to encode it! :+1:

server/src/instant/storage/routes.clj
```diff
@@ -23,63 +22,57 @@
...
-       ;; deny access by default if no permissions are currently set
-       false
-       ;; otherwise, evaluate the permissions code
-       (cel/eval-program! program
-                          {"auth" (cel/->cel-map (<-json (->json current-user)))
-                           "data" (cel/->cel-map {"path" filepath})})))))
+(defn req->app-file! [req params]
+  (let [id (ex/get-param! params [:app_id] uuid-util/coerce)
+        {app-id :id} (app-model/get-by-id! {:id id})
+        filename (ex/get-param! params [:filename] string-util/coerce-non-blank-str)
```
dwwoelfel: Might validate the `filename` param before fetching the app, so that you don't have to do the extra work of fetching the app if the filename param is invalid.

shelwinsunga:
very excited

server/src/instant/storage/routes.clj
```diff
@@ -23,63 +22,57 @@
...
+        refresh-token (http-util/req->bearer-token req)]
```
dwwoelfel: It seems like we should validate the refresh-token has access to the app before going any further. Someone who comes to this codebase later might assume that it's already validated because that's the behavior of similar functions in e.g. dash/routes.

server/src/instant/util/storage.clj
```diff
@@ -0,0 +1,207 @@
...
+            [instant.storage.s3 :as s3-util]
+            [instant.storage.beta :as storage-beta])
+
+  (:import
+   (java.util UUID)))
+
+
+;; scopes filename to app-id directory
+(defn ->object-key [app-id filename]
+  (str app-id "/" filename))
```
dwwoelfel, line 18: I wonder if we'll run into problems with rate limits for partitioned prefixes: https://docs.aws.amazon.com/AmazonS3/latest/userguide/optimizing-performance.html

We could probably ask the user to add their own prefixes if we run into that problem, though.

server/src/instant/util/storage.clj
```diff
@@ -0,0 +1,207 @@
...
+                       (app-user-model/get-by-refresh-token {:app-id app-id
+                                                             :refresh-token refresh-token}))
+
+        program (rule-model/get-program! rules "$files" action)]
+    (ex/assert-permitted!
+     :has-storage-permission?
+     ["$files" action]
+     (if-not program
+       ;; deny access by default if no permissions are currently set
+       false
```
dwwoelfel, line 44: :heart:

server/src/instant/util/storage.clj
dwwoelfel: I would just make different functions? 

Here's my concern with the mutli-arity--the function you call during a test when you don't need permissions is the same function you call when you do need permissions. It seems like we're teetering on the edge of a security issue when someone accidentally calls a function with the wrong airty.

If you have different functions with different names, then it will be much more obvious when you do that.

server/src/instant/util/storage.clj
```diff
@@ -0,0 +1,207 @@
...
+   :etag etag
+   :last_modified (.getMillis last-modified)})
+
+;; Retrieves all files that have been uploaded via Storage APIs
+(defn list-files! [app-id subdirectory]
+  (let [_ (storage-beta/assert-storage-enabled! app-id)
+        prefix (if (string/blank? subdirectory)
+                 app-id
+                 (str app-id "/" subdirectory))
+        objects-resp (s3-util/list-app-objects prefix)
```
dwwoelfel, line 102: This has an implicit limit of 1000, right?

client/packages/admin/src/index.ts
```diff
@@ -556,6 +558,142 @@ class Auth {
...
+  };
+
+  /**
+   * Deletes multiple files by their path names (e.g. "photos/demo.png", "essays/demo.txt").
+   *
+   * @see https://instantdb.com/docs/storage
+   * @example
+   *   const ok = await db.storage.bulkDelete(["images/1.png", "images/2.png", "images/3.png"]);
+   */
+  bulkDelete = async (pathnames: string[]): Promise<boolean> => {
```
stopachka: nit: I would name it `deleteBulk`, mainly due to the intellisense

client/packages/admin/src/index.ts
```diff
@@ -556,6 +558,142 @@ class Auth {
   }
 }
 
+type UploadMetadata = { contentType?: string } & Record<string, any>;
```
stopachka, line 561: heck yeah!

client/packages/admin/src/index.ts
```diff
@@ -556,6 +558,142 @@ class Auth {
...
+type StorageFile = {
+  key: string;
+  name: string;
+  size: number;
+  etag: string;
+  last_modified: number;
```
stopachka, line 567: nit: should we make this `lastModified`?

client/sandbox/admin-sdk-express/src/index.ts
```diff
@@ -130,10 +131,50 @@ async function testDeleteUser() {
...
+}
+
 // testCreateToken();
 // testQuery();
 // testTransact();
 // testScoped();
 // testSignOut();
 // testFetchUser();
 // testDeleteUser();
+// testAdminStorage("src/demo.jpeg", "admin/demo.jpeg", "image/jpeg");
```
stopachka, line 175: nice!

stopachka (commented):
Cloose!

server/src/instant/util/storage.clj
reichert621: sounds good!

server/src/instant/util/storage.clj
reichert621: yeahhh planning on handling pagination in a followup!

client/packages/admin/src/index.ts
reichert621: makes sense! grammatically "delete bulk" sounds a little weird, but maybe it's just me... how do we feel about `deleteBulk` vs `deleteMany`?

server/src/instant/util/storage.clj
reichert621: nice catch, good thing to keep an eye out for!

server/src/instant/util/storage.clj
reichert621: (realized this was a bit silly anyway, since what I was really testing was the `assert-storage-permission!` function... removed and just added tests for `assert-storage-permission!` below instead)

client/packages/admin/src/index.ts
reichert621: this comes directly from the server, and it seems like the pattern used in the rest of the codebase is to keep server API responses in snakecase 🤷 happy to change, but we should probably all align on what the convention should be

stopachka: SGTM! Yeah we have not been consistent. Let's keep snake_case. Eventually we can inc a version, and write some middleware that converts based on the client asking.

client/packages/admin/src/index.ts
```diff
@@ -556,6 +558,142 @@ class Auth {
...
+      `${this.config.apiURI}/admin/storage/files?filename=${encodeURIComponent(pathname)}`,
+      {
+        method: "DELETE",
+        headers: authorizedHeaders(this.config),
+      },
+    );
+
+    // Just return true if the request was successful,
+    // since the s3 response might be confusing
+    return !!data;
```
stopachka: nit: is it ever possible for `data` _not_ to be true? 

If not, I would void instead. 

Reasoning: By returning `true`, the client may think: 'do I need to handle the false case?'

stopachka (approved):
Woohoo!

client/packages/admin/src/index.ts
stopachka: Oo -- I like `deleteMany`!

client/packages/admin/src/index.ts
reichert621: agreed

PR #97, opened 2024-08-30
Add support for Clerk

dwwoelfel:
Adds support for using Clerk as your auth provider.

We utilize the same `signInWithIdToken` machinery we built for logging in with the native google web button. 

Here's the process:

1. You go to the Auth tab on the Instant dashboard and add your clerk publishable key
2. We look up the domain from the publishable key (it's base64-encoded right in the key) and use the key to construct an openid connect discovery url at `${domain}/.well-known/openid-configuration`.
3. We create a new `app_oauth_clients` entry in the database with that discovery url.
4. On the user's website, the user signs in to Clerk with their clerk app, then calls Clerk's `getToken` to get an id token. They call Instant's `db.auth.signInWithIdToken` to create a long-lived session with instant.
5. When they call `signInWithIdToken`, we use our existing oauth machinery to validate the token with the `app_oauth_client` we created and sign them in.

The Instant app owner doesn't have to write any backend code--all of it works directly from the client.

The UI is behind a feature flag so that we don't expose the frontend before the backend is deployed.

I made some changes to support oauth providers that only validate auth tokens:
1. client_id, client_secret, token_endpoint, and authorization_endpoint are now optional on the `app_oauth_clients` table
  2. We might want to think about only making the discovery endpoint required and dropping the token and authorization endpoints. It would simplify the db a bit.
3. client_name is now required to be unique across `app_oauth_clients`, instead of across clients for a single provider. That should have always been the case, it was just an oversight by me when I initially created the table.

<details>
<summary>Screenshots of the flow</summary>

<table>
<tr>
<td>
image
</td>
</tr>

<tr>
<td>
image

</td>
</tr>

<tr>
<td>
image
</td>
</tr>

</table>
</details>


client/packages/core/src/index.ts
```diff
@@ -390,7 +390,7 @@ class Auth {
   signInWithIdToken = (params: {
     idToken: string;
     clientName: string;
-    nonce: string | undefined | null;
+    nonce?: string | undefined | null;
```
dwwoelfel, line 393: Made `nonce` optional. I thought `undefined` meant you could leave it out, but I guess not.

client/packages/core/src/Reactor.js
```diff
@@ -1107,7 +1107,7 @@ export default class Reactor {
    * @param {Object} params
    * @param {string} params.clientName - The name of the client requesting authorization.
    * @param {string} params.idToken - The id_token from the external service
-   * @param {string | null | undefined} params.nonce - The nonce used when requesting the id_token from the external service
+   * @param {string | null | undefined} [params.nonce] - The nonce used when requesting the id_token from the external service
```
dwwoelfel, line 1116: The `[]` makes the parameter optional

client/www/components/dash/AppAuth.tsx
```diff
@@ -720,11 +735,544 @@ function GoogleClients({
...
+
+    if (!idToken) {
+      // No jwt, can't sign in to instant
+      return;
+    }
+
+    // Create a long-lived session with Instant for your clerk user
+    // It will look up the user by email or create a new user with
+    // the email address in the session token.
+    db.auth.signInWithIdToken({
```
stopachka, line 861: Thought: 

With this `useEffect`, afaik `signInWithIdToken` would end up running on each page load, even if we already authed before. 

I worry this would cause some thrash in Reactor, because we would always send an extra ['user changed' update](https://github.com/instantdb/instant/blob/main/client/packages/core/src/Reactor.js#L1120C10-L1120C27)
 
One option to I could think of, is to show a demo where we do the login, as part of Clerk's redirect. Here's an example I wrote for some users: 

https://github.com/stopachka/clerk_auth_example/blob/main/frontend/src/App.tsx#L31-L35

Another could be to make `signInWithIdToken` smarter, though I am not 100% sure how we could achieve it. 

What do you think @dwwoelfel ?

server/src/instant/util/exception.clj
```diff
@@ -127,6 +127,15 @@
                              :original-input param}}))]
     coerced))
 
+(defn get-optional-param! [obj ks coercer]
```
stopachka, line 130: nice!

stopachka (approved):
🫡🫡🫡🫡 Fantastic work Daniel. From client to dashboard to backend 🔥

client/www/components/dash/AppAuth.tsx
dwwoelfel: Resolved with https://github.com/instantdb/instant/pull/97/commits/e9d98ad7a4ed0b9190a3084a0327fed84df9d995

We'll reuse the user's existing token if it's still valid and belongs to the same user. Then in `changeCurrentUser`, we'll exit early if the current user is the same as the new user.

PR #101, opened 2024-08-30
Use inferred types to set default type for add row

dwwoelfel:
Makes a couple of changes to add/edit row:

1. If we don't have a value for the field, looks up the type from the attr's inferred types.
2. Adds an `id` field for the `Add row` dialog and prepopulates it with an id. There's a little `refresh` button you can click to get a new one or you can paste in your own. Will refuse to save with a handy error message if the uuid is invalid.
  a. I set tabIndexes so that the first editable row would be the first thing selected instead of the id row

image


stopachka (approved):
👨‍🍳👨‍🍳👨‍🍳 awesome!

PR #102, opened 2024-08-30
Run clojure tests on every push

dwwoelfel:
Now that we're open source, the repo gets unlimited action minutes.

Runs the clojure tests on every push so you don't have to open a PR to see what the build thinks of your changes.


stopachka (approved):
Heck yeah! Thank you GH

PR #108, opened 2024-09-01
[quickfix] Set instant-config-app-id for dev

stopachka:
Afaik we use the same instant-config app id in both dev and prod. I went ahead and updated the config dev.edn 

@dwwoelfel @nezaj


nezaj (approved):
Nice!

PR #111, opened 2024-09-03
Add missing await to authn example

dnathe4th:
The final step in the example of the vanilla authn logic was missing an `await` inside the async function verifying the login code.


dwwoelfel (approved):
Thanks for fixing!

PR #115, opened 2024-09-03
Increase honeycomb sample rates

nezaj:
We reduced events from ~17M → ~12M https://github.com/instantdb/instant/pull/109. The problem is handle-receive and handle-events [generate a ton of subtraces](https://ui.honeycomb.io/instantdb/environments/prod/result/bGk9tRFm9tk?hideCompare) for querying/book-keeping. So let's just wholesale increase the sample rate for all event types

### DIFF
--- a/server/src/instant/reactive/session.clj
+++ b/server/src/instant/reactive/session.clj
@@ -300,8 +300,7 @@
 (defn event-sample-rate [{:keys [op]}]
   (cond
     (= op :set-presence) 0.01
-    (#{:client-broadcast :join-room} op) 0.1
-    :else 1))
+    :else 0.1))
 
 (defn handle-event [store-conn eph-store-atom session event]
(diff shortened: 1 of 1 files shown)


stopachka (approved):
We riffed on this later. Honestly, I don't think the cost of shipping this is going to be so high. I don't know _when_ we'll get to adding a BatchProcessor. Afaik, at least there is an option to do this _within_ process, so we don't have to use refinery [1]. 

Let's just try to run with 0.1% trace sampling, and see where we get. 

One thing that would be nice, if we could timebox and ship: 
- I like @dwwoelfel idea, of _only_ sampling to honeycomb, but keeping the logs to logdna 
  - I am not 100% sure how to do it, but if we could do that, it would be amazing

nezaj:
Closing in favor of https://github.com/instantdb/instant/pull/131

PR #119, opened 2024-09-03
Adds instructions to the LLM on where to find our docs

dwwoelfel:
Helps cursor find all of the docs links so that it can build an index for custom docs https://docs.cursor.com/context/@-symbols/@-docs#add-custom-docs

Adds a hidden div with all of the docs links to every docs page. From my testing, it needs the links to be absolute paths.


stopachka (approved):
Wow! Cool idea!

PR #120, opened 2024-09-03
Add loading to a couple of buttons

### LINKED ISSUE #106: idempotency issue: When creating new application, with slighty lag i unknowingly clicked twice, which created 2 apps.
image

dwwoelfel:
Adds `loading` in a couple of places so that you can't double-click the button and can see that something is happening. Fixes #106

### DIFF
--- a/client/www/components/dash/Onboarding.tsx
+++ b/client/www/components/dash/Onboarding.tsx
@@ -228,7 +228,11 @@ function CreateFirstAppScreen(props: {
               props.onAppNameChange(t);
             }}
           />
-          <Button type="submit" disabled={props.appName.trim().length === 0}>
+          <Button
+            type="submit"
+            loading={props.isLoading}
+            disabled={props.appName.trim().length === 0}
+          >
             Let's build!
           </Button>
           {props.error ? (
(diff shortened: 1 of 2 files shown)


client/www/components/dash/Onboarding.tsx
```diff
@@ -228,7 +228,11 @@ function CreateFirstAppScreen(props: {
               props.onAppNameChange(t);
             }}
           />
-          <Button type="submit" disabled={props.appName.trim().length === 0}>
+          <Button
+            type="submit"
+            loading={props.isLoading}
+            disabled={props.appName.trim().length === 0}
```
stopachka, line 234: nit: perhaps instead of `loading`, what if we added this to the `disabled` check? 

Reasoning: afaik if we do this, it will have a very quick 'flash' in the UI

stopachka (approved):
Nice!

client/www/components/dash/Onboarding.tsx
dwwoelfel: You'll get a flash in either case because the disabled state also dims the button. The only difference that I can see is that the `loading` should also show a spinner in your mouse (but doesn't seem to do so in Chrome b/c we set the `disabled` prop on the button when loading.

stopachka: Ah cool! I had an assumption that the text would change or something. Carry on 🫡

PR #122, opened 2024-09-03
Encode more cel values to JSON

dwwoelfel:
Fixes encoding null and cel expression values to JSON so that the sandbox doesn't break on rules that return null.


stopachka (approved):
TIL encoder! Great

PR #124, opened 2024-09-03
Coerce NULL_VALUE to null in cel check

dwwoelfel:
If you do `user.isAdmin` in your rule, that will return a NullObject from cel and we'll evaluate that to truthy.

This change coerces NULL_VALUE to nil before we return from `eval-program!`


stopachka (approved):
Awesome!

PR #131, opened 2024-09-04
Add refinery for honeycomb sampling

dwwoelfel:
Adds refinery to do honeycomb tail sampling for us.

Our configuration is set up to send all events that have an `error` property, and throttle all other events so that we send a max of about 30/second. That should keep us below our Honeycomb plan limits.

The refinery container sits next to the server container and we redirect all of our honeycomb API calls to the refinery container. Then the refinery container makes a decision on what to send based on the rules we defined in `server/refinery/rules.yaml`. More docs on those [here](https://docs.honeycomb.io/manage-data-volume/sample/honeycomb-refinery/sampling-methods/) and [here](https://github.com/honeycombio/refinery/blob/main/rules_complete.yaml).

We use a `EMAThroughputSampler` to set a target throughput for non-errors. It uses the `name` field to group spans for throttling, so that it doesn't completely drop certain spans. If we want to throttle certain spans more or with more granularity, we can add additional rules to the yaml file.

### DIFF
--- a/.github/workflows/clojure.yml
+++ b/.github/workflows/clojure.yml
@@ -149,7 +149,8 @@ jobs:
   publish:
     runs-on: ubuntu-latest
     needs: [ clj-check ]
-    if: github.ref == 'refs/heads/main'
+    # XXX: refinery BACK TO main
+    if: github.ref == 'refs/heads/refinery'
     permissions:
       id-token: write
       contents: read
@@ -200,7 +201,7 @@ jobs:
       uses: "./.github/actions/elastic-beanstalk"
       with:
         working-directory: "server"
-        files: '["docker-compose.yml", ".platform/hooks/prebuild/logdna.sh"]'
+        files: '["docker-compose.yml", ".platform/hooks/prebuild/logdna.sh", "refinery/config.yaml", "refinery/rules.yaml"]'
         aws-region: "us-east-1"
         bucket: "elasticbeanstalk-us-east-1-597134865416"
         application-name: "instant-docker-prod"
--- a/.github/workflows/clojure.yml
+++ b/.github/workflows/clojure.yml
@@ -146,9 +146,8 @@ jobs:
       working-directory: server
       run: clojure -M:lint "--config" ".clj-kondo/ci-config.edn"
 
-  publish:
+  publish-image:
     runs-on: ubuntu-latest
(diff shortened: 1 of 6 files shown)


.github/workflows/clojure.yml
```diff
@@ -146,10 +146,11 @@ jobs:
       working-directory: server
       run: clojure -M:lint "--config" ".clj-kondo/ci-config.edn"
 
-  publish:
+  publish-image:
```
dwwoelfel, line 149: Small change to make the deploys faster, we build the docker image while running the tests, but only create the eb application version if the tests pass.

server/docker-compose.yml
```diff
@@ -22,3 +24,12 @@ services:
       - /var/log/eb-engine.log:/var/log/eb-engine.log:ro
       - /var/log/messages:/var/log/messages:ro
       - /var/log/eb-docker:/var/log/eb-docker:ro
+  refinery:
+    image: honeycombio/refinery:2.7.1
+    volumes:
+      - ./refinery/config.yaml:/etc/refinery/refinery.yaml
+      - ./refinery/rules.yaml:/etc/refinery/rules.yaml
+    environment:
+      - REFINERY_HONEYCOMB_API_KEY=${REFINERY_HONEYCOMB_API_KEY}
```
dwwoelfel, line 33: I set the `REFINERY_HONEYCOMB_API_KEY` env var in the elastic beanstalk config.

server/src/instant/util/tracer.clj
```diff
@@ -166,21 +167,21 @@
            (end-span! *span*))))))
 
 ;; (XXX)
-;; Given a `sample-rate`, we will randomly skip spans at that rate 
+;; Given a `sample-rate`, we will randomly skip spans at that rate
```
nezaj, line 170: Nit. Should we delete this comment now?

dwwoelfel: I think we should leave everything the same for a bit while we get comfortable with refinery, then come back and move any sampling configuration to the rules.yaml.

nezaj: Sounds good!

stopachka:
Woohoo!!

nezaj (approved):

PR #142, opened 2024-09-05
Strong Init fixes

markyfyi:
1. Fix a bug in the InstaQL types where we were failing to propagate RCI config to subqueries.  Not to self: be more scrupulous around generic type param defaults.
1. Loosen constraints on the generic param for `i.json`.  This removes the "this is a plain old object" guarantee, but allows developers to better integrate foreign types and interfaces.  One thing to note - we should sanitize objects before sending them down-the-wire to avoid the "failed to serialize circular structure" error.
1. Improve `db.tx` support for `.optional()` attrs by allowing for `null` as a value in a addition to the specified type.
1. Enforce "strict mutations" for `db.tx`.  Specifying unknown attributes will raise a type error.


stopachka (approved):
Love it!!

PR #144, opened 2024-09-05
Strong Init utility types, plus infra for entity override types

markyfyi:


stopachka (approved):
dam!

client/packages/core/src/index.ts
```diff
@@ -28,6 +30,7 @@ import type {
 } from "./presence";
 import * as i from "./schema";
 import { createDevtool } from "./devtool";
+import { DB } from "./types";
```
stopachka: I would call this IDatabase

client/packages/core/src/index.ts
```diff
@@ -601,4 +606,6 @@ export {
   PresenceSlice,
   PresenceResponse,
   InstaQLQueryParams,
+  InstantQuery,
```
stopachka, line 609: nit: would call InstaQLQuery, to match the others

stopachka (approved):
LFG!

client/packages/core/src/index.ts
markyfyi: I named it this way because it _doesn't_ match the others.  `InstaQLQuery*` are all internal helpers that aren't user-friendly, whereas `InstantQuery` and `InstantQueryResult` are meant for external consumption.  Look at the generic signatures, they're super different.  Open to changing the naming scheme, but I want to keep helper types and internal types distinct.

PR #145, opened 2024-09-05
[Intern] Faster top

nezaj:
The top page is getting slower and slower because of pulling too much historical data. Feels like the most interesting data is in the recent week.

To allow for more flexibility, you can use a query param to alter the number of days. E.g. adding `?n=10` to the url will fetch 10 days worth of data. The default is 7.

### DIFF
--- a/server/src/instant/util/int.clj
+++ b/server/src/instant/util/int.clj
@@ -1,0 +1,11 @@
+(ns instant.util.int)
+
+(defn parse-int [s default]
+  (try
+    (Integer/parseInt s)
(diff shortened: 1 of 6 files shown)


stopachka (approved):
Nice!

server/src/instant/util/int.clj
```diff
@@ -0,0 +1,11 @@
+(ns instant.util.int)
```
stopachka: nit would call this utility `number`, as I think we'll likely want other utlity functions that use longs etc

PR #146, opened 2024-09-05
Delete unused import

gunayanil:
Deleted `Linking` import which is not used anywhere in the component.


stopachka (approved):
Thank you @gunayanil !

PR #147, opened 2024-09-06
Lookup by a ref attr

dwwoelfel:
If the user has a unique ref attr with cardinality one, this allows them to use `lookup` to update an entity in the namespace using `forward-name.id` as the lookup key.

Here's an example:

If the `user_prefs` namespace has a unique link to `users`, then the developer can update `user_prefs` with:

```js
tx.user_prefs[lookup("users.id", userId)].update({new_pref: 'new-value})
```

Updated the backend to support this with the admin api.

I attempted to solve the problem where you can't create a new entity with a lookup set the lookup value at the same time, but could not find any easy solutions, so I just gave it a better error message. Now when you do:

```js
db.transact([
    tx.users[lookup('handle', 'h')].update({handle: 'h'})
]);
```

It will throw a validation error with message: `"Updates with lookups can only update the lookup attribute if an entity with the unique attribute value already exists."`

### DIFF
--- a/.github/workflows/js.yml
+++ b/.github/workflows/js.yml
@@ -3,7 +3,6 @@ name: JS CI
 
 on:
   push:
-    branches: ["main"]
   pull_request:
     types: [opened, synchronize]
 
--- a/server/src/instant/admin/model.clj
+++ b/server/src/instant/admin/model.clj
@@ -64,7 +64,7 @@
       (ex/throw-validation-err!
        :lookup
        lookup
-       [{:message (str ident-name " is not a valid link lookup.")}]))
+       [{:message (str ident-name " is not a valid lookup attribute.")}]))
     fwdName))
 
 (defn extract-lookup [attrs etype eid]
--- a/server/src/instant/admin/model.clj
+++ b/server/src/instant/admin/model.clj
@@ -1,19 +1,19 @@
 (ns instant.admin.model
-  "Context: 
-   
-   Internally, we process transactions as `tx-steps`. The model for that is in 
-   `instant.db.transaction`, and looks like: 
-   
-   [[:add-triple eid ...] 
+  "Context:
+
+   Internally, we process transactions as `tx-steps`. The model for that is in
+   `instant.db.transaction`, and looks like:
+
+   [[:add-triple eid ...]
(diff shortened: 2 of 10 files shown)


.github/workflows/js.yml
```diff
@@ -3,7 +3,6 @@ name: JS CI
 
 on:
   push:
-    branches: ["main"]
```
dwwoelfel, line 6: Just making the js tests run on every commit so you don't have to open a PR to see them.

server/src/instant/db/model/triple.clj
```diff
@@ -7,8 +7,10 @@
    [clojure.spec.alpha :as s]
    [instant.util.spec :as uspec]
    [instant.util.json :refer [->json]]
+   [instant.util.string :refer [multiline->single-line]]
    [instant.util.tracer :as tracer]
-   [instant.data.constants :refer [empty-app-id]])
+   [instant.data.constants :refer [empty-app-id]]
+   [instant.util.exception :as ex])
```
dwwoelfel, line 13: This file is much easier to review with whitespace off https://github.com/instantdb/instant/pull/147/files?w=1

stopachka:
Woohoo!

.github/workflows/js.yml
stopachka: Love it!

server/src/instant/admin/model.clj
```diff
@@ -156,60 +177,116 @@
...
+   :add-ops (conj add-ops [:add-attr attr])})
+
+(defn add-attrs-for-obj [acc op]
+  (let [[action etype _eid obj] op
+        labels (conj (keys obj) "id")]
+    (reduce (fn [{:keys [attrs] :as acc} label]
+              (let [fwd-attr (attr-model/seek-by-fwd-ident-name [etype label] attrs)
+                    rev-attr (when (contains? ref-actions action)
+                               (attr-model/seek-by-rev-ident-name [etype label] attrs))]
+                (cond (and (contains? update-actions action)
```
stopachka, line 227: niice

stopachka (approved):
Heeck yeah! Let's 🚢🚢🚢 it @dwwoelfel

PR #151, opened 2024-09-06
Fix the refinery sampler rules

dwwoelfel:
Had the wrong environment, it should have been "prod" instead of "production".

Tested in production and it's sampling now.


stopachka (approved):
Niice!

PR #152, opened 2024-09-06
Use types from core in admin

markyfyi:
Context: https://stackoverflow.com/questions/78109569/type-expansion-causes-an-infinite-loop-in-package-but-not-in-project

Something got rid of the crazy type expansion described above, probably the bump to TypeScript 5.x last month.  This change dedupes our typedefs.

### DIFF
--- a/client/scripts/publish_packages.clj
+++ b/client/scripts/publish_packages.clj
@@ -62,7 +62,7 @@
 (defn -main [& _args]
   (let [tag (first _args)
         version (str/trim (slurp "version.md"))]
-    (when-not tag (set-package-versions! version))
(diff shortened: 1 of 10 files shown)


stopachka (approved):
Beautiful!!!

client/scripts/publish_packages.clj
```diff
@@ -62,7 +62,7 @@
 (defn -main [& _args]
   (let [tag (first _args)
         version (str/trim (slurp "version.md"))]
-    (when-not tag (set-package-versions! version))
+    (set-package-versions! version)
```
markyfyi, line 65: We actually need this for dependent packages

PR #155, opened 2024-09-06
remove more loading states from `useAuth` and `useQuery`

stopachka:
**Before**

https://github.com/user-attachments/assets/7968d979-f605-41f2-b606-d5e3af513d5b

**After** 

https://github.com/user-attachments/assets/6824528b-3d68-4ac9-b260-30ee72159354

## Problem

When a component _first_ mounted, `useAuth` and `useQuery` always returned `isLoading: true`. 

We did this before because of SSR. If `useAuth` or `useQuery` was used on the server, we would _have_ to return `isLoading: true`. If on the client we then used our local cache and had `isLoading: false`, it would cause a hydration error. 

I wrote a stackoverflow about this: https://stackoverflow.com/questions/78889459/is-there-a-way-to-detect-when-render-is-part-of-hydration

## Solution

[`useSyncExternalStore`](https://react.dev/reference/react/useSyncExternalStore) is our weapon. It lets us provide a state for hydration, and a state for mounting. 

```javascript
const state = useSyncExternalStore(
  subscribe, // this subscribes to our state
  /*getSnapshot*/ () => resultCacheRef.current, // this gives the current value to the client
  /*getServerSnapshot*/ () => defaultAuthState, // this gives the value **during hydration**
);
```

- **During hydration** `getServerSnapshot` runs during hydration. This lets us make sure that if `useQuery` was called on the server, we return the `isLoading: true` result
- **After hydration** `getSnapshot` runs after hydration. This means that all further mounts, or further client-side navigation, will return _whatever was cached_! 

### Update to useAuth

I updated `useAuth` to use `useSyncExternalStore`. Crucially, this meant we would need some kind of 'synchronous' value that could set for the initial snapshot. 

**Hack:** Reactor._currentUserCached

I added a `_currentUserCached` value to Reactor. This lets us synchronously access the value of the current user, if we had already loaded it. The cleaner solution would be to refactor `CurrentUser`, into an observable of sorts. This observable could handle subscribing, and give a 'sync' view to our data. I think it's something we should explore, but it would be a complicated lift, so I started caching calls with calls to `getCurrentUser` and `updateUser`

### Update to useQuery

We already had `useSyncExternalStore` in useQuery, but we needed to make sure that the 'first' value in the ref was whatever we had in the store. Made a quick `getPreviousResult` function, to achieve this. Eventually, it would be _really_ cool if we _only_ updated a query, when the underlying store, or the list of mutations changed. 

## Bonus: Added app router pages to the nextjs-react index page 

We currently listed all pages-router 'play' files, but not the app-router play files. I refactored the code a bit to also show the app-router play files 

@markyfyi @dwwoelfel @nezaj @reichert621


client/packages/react/src/useQuery.ts
```diff
@@ -37,10 +47,9 @@ export function useQuery<
   // uses `Object.is` to compare the previous and next state.
   // If we don't use a ref, the state will always be considered different, so
   // the component will always re-render.
-  const resultCacheRef =
-    useRef<LifecycleSubscriptionState<Q, Schema, WithCardinalityInference>>(
-      defaultState,
-    );
+  const resultCacheRef = useRef<
+    LifecycleSubscriptionState<Q, Schema, WithCardinalityInference>
```
stopachka, line 52: Hmm...I am seeing some caching issues when I play with some data in Explorer. I will look a bit more deeply

client/packages/core/src/Reactor.js
```diff
@@ -79,6 +79,7 @@ export default class Reactor {
 
   _presence = {};
   _broadcastSubs = {};
+  _currentUserCached = { isLoading: true, error: undefined, user: undefined };
```
stopachka, line 82: This lets us synchronously access the value of the current user, if we had already loaded it. 

The cleaner solution would be to refactor CurrentUser, into an observable of sorts. This observable could handle subscribing, and give a 'sync' view to our data. 

I think it's something we should explore, but it would be a complicated lift, so I started caching calls with calls to getCurrentUser and updateUser

client/packages/react/src/useQuery.ts
```diff
@@ -33,17 +43,17 @@ export function useQuery<
...
+  // to compare the previous and next state.
   // If we don't use a ref, the state will always be considered different, so
   // the component will always re-render.
-  const resultCacheRef =
-    useRef<LifecycleSubscriptionState<Q, Schema, WithCardinalityInference>>(
-      defaultState,
-    );
+  const resultCacheRef = useRef<
+    LifecycleSubscriptionState<Q, Schema, WithCardinalityInference>
+  >(stateForResult(_core._reactor.getPreviousResult(query)));
```
stopachka, line 53: Technically, this means that we _could_ skip the first 'callback' for subscribeQuery. Reason: since we already cached the previous result here, `subscribeQuery` would in-effect return the same cached result in the first callback response.

But, it's not 100% true. Something _could_ have happened between the useEffect call and this cached call. It would be amazing if we could refactor Reactor, so we could have more stable values for query results. This way, we would know exactly _when_ it was okay to skip an update. i.e if `store` hasn't changed, and `transactions queue` has not changed.

client/packages/core/src/Reactor.js
```diff
@@ -1002,9 +1014,14 @@ export default class Reactor {
...
     this.authCbs.push(cb);
+    let alreadySentImmediate = false;
+    if (!this._currentUserCached.isLoading) {
+      cb(this._currentUserCached);
+      alreadySentImmediate = true;
+    }
     let unsubbed = false;
     this.getCurrentUser().then((resp) => {
-      if (unsubbed) return;
+      if (unsubbed || alreadySentImmediate) return;
```
dwwoelfel: Should this check if the `_currentUserCached` we sent is the same as `resp`?

dwwoelfel (approved):
One comment, but looks good to me!

client/packages/core/src/Reactor.js
stopachka: done! Used `areObjectsDeepEqual`, as the call to `getCurrentUser` itself will re-write `_currentUserCached`

PR #156, opened 2024-09-06
[Docs] Add client side sign out

nezaj:
A user asked how to sign out from the client in the discord. We have this method but I realize we didn't explicitly talk about it in our docs (we only talk about it from the server)


stopachka (approved):
Nice!

vinciarts:
It's me haha

PR #157, opened 2024-09-06
String Init: Use schema instead of attrs when resolving cardinality

markyfyi:
To maintain type-safety, honor the specified schema and not the synced attrs when determining whether to return one/many results.


client/packages/core/__tests__/src/instaqlInference.test.js
```diff
@@ -68,6 +83,20 @@ test("many-to-many with inference", () => {
 });
 
 test("one-to-one with inference", () => {
+  const schema = i.graph(
+    "",
```
stopachka, line 87: nit: would say 'app-id' or something

stopachka (approved):
LGTM!

PR #161, opened 2024-09-06
Update lookups and links in rewrite mutations

dwwoelfel:
Updates rewrite mutations to rewrite lookups and fix links.

This fixes a couple of bugs:

1. lookups weren't getting updated, so if you used the lookup before init finished, it would fail at the server with "attr is not unique"
2. creating links in the opposite order (e.g. users.links(books) when the forward attr is already books.users) would cause the attribute to be added twice, causing a ident is not unique failure at the server

This PR also changes the behavior of `db.transact` a bit. Before, we would throw immediately if `instaml.transform` threw an error. Now we wait until the promise is created and deliver the error to the promise. This should be nicer for callers of `db.transact` because now `db.transact().then(x => console.log('ok')).catch(e => console.error(e))` won't throw.


client/packages/core/src/Reactor.js
```diff
@@ -518,23 +519,30 @@ export default class Reactor {
...
+          const newTxStep = instaml.rewriteStep(mapping, txStep);
+
           retTxSteps.push(newTxStep);
           return [mapping, retTxSteps];
         },
         [mapping, []],
       );
+      if (chunks) {
+        try {
+          return [newMapping, instaml.transform(attrs, chunks)];
```
dwwoelfel: Found a problem with doing transform to rewrite the mutations--it will recreate the`add-attr` every time you call it, so you won't keep consistent attr ids for the new attrs. I think I'll go back to the existing method and find a different way to solve the links issue.

dwwoelfel: Resolved. Now we're just augmenting the existing rewriter.

client/packages/core/src/InMemoryStorage.js
```diff
@@ -0,0 +1,14 @@
+export default class IndexedDBStorage {
```
stopachka: nit: would name it `InMemoryStorage`

stopachka (approved):
Love the test!

client/packages/core/src/InMemoryStorage.js
dwwoelfel: Whoops! Thanks for catching.

PR #166, opened 2024-09-08
Don't assume periods are refs in lookups if the attr exists

dwwoelfel:
Allows `lookup` to succeed if the attr with a dot in it exists instead of assuming it's a ref lookup.


stopachka (approved):
🫡🫡🫡🫡 love it!

PR #170, opened 2024-09-09
Reuse created_at for a triple while waiting for server

dwwoelfel:
Reuses the created_at for an optimistic triple if we already have the triple in the store.

This fixes a bug where editing an entity would shoot it to the top/bottom of the list while we waited for the server to respond.


stopachka (approved):
:mr_clean: 

love it!

PR #178, opened 2024-09-09
Strong Init Docs

markyfyi:


client/www/pages/docs/strong-init.md
```diff
@@ -0,0 +1,60 @@
+---
+title: Strong Init (experimental)
+---
+
+We're piloting strictly-typed variants of our client libraries that leverage your `instant.schema.ts` definition. We're calling it "Strong Init". Strong Init enhances InstaQL and InstaML methods with deep typesafety and enhances your editor experience with rich intellisence.
```
stopachka:
```suggestion
What if you could use the types you defined in `instant.schema.ts`, inside `init`? We have a new version of `init` out that lets you do this. It's in beta, but if you use it, `useQuery` and `transact` will automatically get typesafety and intellisense.

Here's how it works
```

Main thoughts: 
- I would try to use simpler words that you would use in conversation with a developer. i.e 'leverage', 'variant', 'client libraries', 'enhances', 'rich', feel a bit out of place
- I would avoid the word 'simply', as it could get the developer who is confused thinking 'if it's so simple, why am I confused'

client/www/pages/docs/strong-init.md
```diff
@@ -0,0 +1,60 @@
...
+
+If you want to migrate, simply swap out `init` with `init_experimental` and provide it your app's graph schema object.
```
stopachka: If you want to try it out, swap `init` with `init_experimental`, and pass in your 'schema' object. Here's how that looks:

client/www/pages/docs/strong-init.md
```diff
@@ -0,0 +1,60 @@
...
+
+```ts
+import { init_experimental } from '@instantdb/react';
+import schema from '~/instant.schema.ts';
```
stopachka:
```suggestion
// 1. Import your schema file
// Don't have a schema? check out https://www.instantdb.com/docs/cli to get started
import schema from '../instant.schema.ts';
```

client/www/pages/docs/strong-init.md
```diff
@@ -0,0 +1,60 @@
...
+
+const db = init_experimental({
+  appId: '__APP_ID__',
+  schema,
```
stopachka, line 20:
```suggestion
  // 2. Use it inside `init_experimental`
  schema,
```

client/www/pages/docs/strong-init.md
```diff
@@ -0,0 +1,60 @@
...
+});
+
+function App() {
+  return <Main />;
+}
+```
+
+## Strong InstaML
```
stopachka: I see the example for InstaML, but not for useQuery. 

Maybe we integrate this in the code example above, so there's an example with `useQuery` that is typed, as well as db.tx

client/www/pages/docs/strong-init.md
```diff
@@ -0,0 +1,60 @@
...
+
+In order to leverage your schema in InstaML, just replace all `tx` references with `db.tx`.
```
stopachka:
```suggestion
To make sure your transactions are typed, here's what you need to do: 

If you used Instant before, you would use the global `tx` object. Instead, use `db.tx`: 
```

client/www/pages/docs/strong-init.md
```diff
@@ -0,0 +1,60 @@
...
+
+```ts
+// ❌ instead of global `tx`
+tx.todos[id()].update({ done: true });
+
+// ✅ use bound `db.tx`
```
stopachka:
```suggestion
// ✅ use `db.tx`
```

client/www/pages/docs/strong-init.md
```diff
@@ -0,0 +1,60 @@
...
+db.tx.todos[id()].update({ done: true }); // note the `db`
+```
+
+## Breaking changes
+
+Switching to Strong Init will automatically turn on a feature we're calling RCI - Runtime Cardinality Inference.
+
+Since the new useQuery is schema-aware, we know when to return a single item instead of an array. 🎉 Bear in mind that you'll need to update your call sites for "has one" relationships.
+
+If you want to enable Strong Init without this feature, you can opt out by adding a `cardinalityInference` flag set to `false` in your `init_experimental` call.
```
stopachka:
```suggestion
## Changes to useQuery

Once you switch to `init_experimental`, your query results get more powerful too. 

Previously, all responses in a `useQuery` returned arrays. Now, we can use your schema to decide. If you have a 'has one' relationship, we can return _just_ one item directly. 

```
const { data } = useQuery({users: {author: {}}); 
const firstUser = data.users[0]
// before 
const author = user.author[0]
// after
const author = user.author // no more array! 🎉
```

🎉 Bear in mind that you'll need to update your call sites for "has one" relationships.

If you want to enable Strong Init without this feature, you can opt out by adding a `cardinalityInference` flag set to `false` in your `init_experimental` call.
```

client/www/pages/docs/strong-init.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Rooms support
```
stopachka, line 68: ### Typing rooms

client/www/pages/docs/strong-init.md
```diff
@@ -0,0 +1,60 @@
...
+
+Rooms are still expressed in pure TypeScript. You can enrich your schema with your rooms shape by calling `schema.withRoomSchema<R>`.
```
stopachka:
```suggestion
Rooms are still expressed in pure TypeScript. You can type rooms with `schema.withRoomSchema<R>`. Here's how it looks:
```

stopachka (commented):
Lesgoo!

client/www/pages/docs/strong-init.md
```diff
@@ -0,0 +1,91 @@
...
+```
+
+## Changes to `tx`
+
+If you used Instant before, you would use the global `tx` object. Instead, use the schema-aware `db.tx`.
+
+```ts
+// ❌ instead of global `tx`
+tx.todos[id()].update({ done: true });
+
```
nezaj: Nit. Let's remove this, I remember when I tried to do a similar think w/ the merge docs (saying don't do this, do this instead) users found it confusing in the docs

```suggestion
```

client/www/pages/docs/strong-init.md
```diff
@@ -0,0 +1,91 @@
...
+Since the new `useQuery` is schema-aware, we know when to return a single item instead of an array. 🎉 Bear in mind that if you're migrating from `init`, you'll need to update all of your call sites that reference these "has-one" relationships.
+
+```ts
+const { data } = useQuery({ users: { author: {} });
+
+const firstUser = data.users[0]
+
+// ❌ before
+const author = user.author[0]
```
nezaj: Similar to above, would remove the `before` and `after`

```suggestion
```

client/www/data/docsNavigation.js
```diff
@@ -41,6 +41,7 @@ module.exports = [
       { title: 'Instant CLI', href: '/docs/cli' },
       { title: 'Storage (beta)', href: '/docs/storage' },
       { title: 'Schema-as-code', href: '/docs/schema' },
+      { title: 'Strong Init (experimental)', href: '/docs/strong-init' },
```
nezaj: Let's change the order so that the beta/experimental things are next to each other
```suggestion
      { title: 'Schema-as-code', href: '/docs/schema' },
      { title: 'Storage (beta)', href: '/docs/storage' },
      { title: 'Strong Init (experimental)', href: '/docs/strong-init' },
```

nezaj (approved):
Great write up! Made a few nits

PR #179, opened 2024-09-10
[superadmin] support for transferring apps 

stopachka:
**1)** Added An endpoint to initiate transferring an app. This will send an email like so: 
CleanShot 2024-09-09 at 19 54 42@2x

**2)** A user can then login, and claim ownership of the app: 
CleanShot 2024-09-09 at 19 57 46@2x

**3)** The superadmin also has an API to _revoke_ access

Notes: 
1. Right now, only the superadmin api lets you create a 'creator' transfer. Eventually we could enhance our normal invitation flow to also allow this. 

@nezaj @dwwoelfel @markyfyi

### DIFF
--- a/server/src/instant/dash/routes.clj
+++ b/server/src/instant/dash/routes.clj
@@ -446,14 +446,17 @@
 
 (defn ephemeral-claim-post [req]
   (let [app-id (ex/get-param! req [:params :app_id] uuid-util/coerce)
-        admin-token (ex/get-param! req [:body :token] uuid-util/coerce)
+        token (ex/get-param! req [:body :token] uuid-util/coerce)
         {app-creator-id :creator_id} (app-model/get-by-id! {:id app-id})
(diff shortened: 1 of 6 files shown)


server/src/instant/dash/routes.clj
```diff
@@ -751,13 +754,21 @@
   (let [{user-email :email user-id :id} (req->auth-user! req)
         invite-id (ex/get-param! req [:body :invite-id] uuid-util/coerce)
         {:keys [invitee_role status app_id invitee_email]} (instant-app-member-invites-model/get-by-id! {:id invite-id})]
+    (tool/def-locals!)
```
dwwoelfel: debug code

dwwoelfel (approved):
One comment, but looks good to me!

server/src/instant/dash/routes.clj
stopachka: 🫡

PR #181, opened 2024-09-10
Invalidation fixes

dwwoelfel:
Fixes some issues with the invalidator:

1. Updates attr change topics to put the attr id in a set so that it will match against ordinary topics.
2. Prevents the inferred types from updating the attr row on every update. If not for the first issue, we would have been doing a lot more unnecessary invalidating. Now it only creates a wal log if the inferred types change.
 
   i. This could be even better--if you set `replica identity full` on the table we should get a full diff (https://github.com/eulerto/wal2json/issues/152), then we could only re-render on attr changes we care about.
4. Coerces `where` inputs to uuids when they're a ref, e.g. `{users: {$: {where: {bookshelves: "bookshelf-uuid"}}}`

   i. This was causing the topic not to match because the string didn't equal the uuid.

### DIFF
--- a/server/src/instant/db/instaql.clj
+++ b/server/src/instant/db/instaql.clj
@@ -600,11 +600,14 @@
 (defn guarded-where-query [ctx {:keys [etype level] :as form}]
   (try
     (where-query ctx form)
-    (catch clojure.lang.ExceptionInfo _e
-      (list true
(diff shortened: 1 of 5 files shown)


stopachka (approved):
Monster PR @dwwoelfel !!

PR #182, opened 2024-09-10
[schema] move schema ops into a shared namespace

stopachka:
Refactoring our schema functions into a single namespace. This way we can use them in both `dash-routes` and `superadmin`

@nezaj @dwwoelfel @markyfyi

### DIFF
--- a/server/src/instant/admin/routes.clj
+++ b/server/src/instant/admin/routes.clj
@@ -435,17 +435,6 @@
                         :headers {"app-id" (str counters-app-id)
                                   "authorization" (str "Bearer " admin-token)}}))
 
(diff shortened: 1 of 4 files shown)


server/src/instant/admin/routes.clj
```diff
@@ -435,17 +435,6 @@
                         :headers {"app-id" (str counters-app-id)
                                   "authorization" (str "Bearer " admin-token)}}))
 
-(defn schema-experimental-get [req]
```
stopachka, line 438: From a loong while ago. It was never used -- the new schema stuff obviates this endpoint : )

dwwoelfel (approved):
Looks good to me!

PR #183, opened 2024-09-10
Notify subs when transaction times out

dwwoelfel:
Fixes a bug where the user would see stale data if a mutation times out. We weren't notifying the queries that they needed to re-evaluate without the mutation after the timeout.

Refactors `finishTransaction` a bit so that you just pass the status instead of passing the status and whether it was an error. We infer whether it was an error from the status. 

Moves error handling for a failed mutation to its own method so that we can use it whether we failed for a timeout or failed for an error.

### DIFF
--- a/client/packages/core/src/Reactor.js
+++ b/client/packages/core/src/Reactor.js
@@ -156,9 +156,16 @@ export default class Reactor {
     );
   }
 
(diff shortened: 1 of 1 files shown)


stopachka (approved):
🫡🫡🫡yess

PR #185, opened 2024-09-10
Throttle persisting items

dwwoelfel:
Adds a 100ms throttle to persisting to storage. 

Should improve cpu-usage/latency if you get a flurry of `add-query-ok` messages.


G2Jose:
Will this make things feel more laggy?

dwwoelfel:
> Will this make things feel more laggy?

No, this won't affect the render flow.

The data is persisted to storage in the background, and this just makes us do it less often. The only potential downside is that it extends the window where data might not be persisted to storage by 100ms. But now that I think of it, I could add a beforeunload hook to flush any pending writes to storage.

stopachka (approved):
Wohoo! I have a good feeling this is going to help a lot for Kein / Kosmik too 🫡

PR #190, opened 2024-09-10
fix rooms

markyfyi:
- **Fix rooms**
- **fix fix**

### DIFF
--- a/client/packages/core/src/Reactor.js
+++ b/client/packages/core/src/Reactor.js
@@ -77,7 +77,9 @@ export default class Reactor {
   /** @type BroadcastChannel | undefined */
   _broadcastChannel;
 
+  _rooms = {};
   _presence = {};
+  _broadcastQueue = [];
   _broadcastSubs = {};
   _currentUserCached = { isLoading: true, error: undefined, user: undefined };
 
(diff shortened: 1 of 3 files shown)


client/packages/core/src/Reactor.js
```diff
@@ -1295,8 +1332,8 @@ export default class Reactor {
 
     const handler = { ...opts, roomId, cb, prev: null };
 
+    this._rooms[roomId].isLoading = true;
```
stopachka: Would we 'always' want to set this? i.e what if the user has already joined the room before?

client/packages/core/src/Reactor.js
```diff
@@ -1217,7 +1231,14 @@ export default class Reactor {
   // Rooms
 
   joinRoom(roomId) {
-    this._trySendAuthed(uuid(), { op: "join-room", "room-id": roomId });
+    this._rooms[roomId] = {
+      isLoading: true,
```
stopachka: Same as above, I wonder how this will function if there are two places in one page that call `joinRoom`. 

Perhaps we should de-dupe?

client/packages/core/src/Reactor.js
```diff
@@ -360,11 +362,15 @@ export default class Reactor {
...
-        const loadingRoom = this._presence[loadingRoomId];
-        if (loadingRoom) {
-          loadingRoom.isLoading = false;
+        const joinedRoom = this._rooms[loadingRoomId];
+
+        if (!joinedRoom) {
+          break;
         }
+
+        joinedRoom.isLoading = false;
```
stopachka: Makes me also realize: what happens if we go offline and back online? 

Now each 'joinedRoom` should be set back to 'isLoading', as we will be sending new join events.

client/packages/core/src/Reactor.js
markyfyi: You're right.  This shouldn't be set here at all!

stopachka (approved):
Let's go!!

PR #191, opened 2024-09-10
[vercel] try new lockfile

stopachka:
image

The builds were breaking on vercel. Looking at logs, it seems like vercel doesn't respect our lockfile anymore.


stopachka:
no dice

PR #197, opened 2024-09-12
chore: restrict npm uploads to dist folder

wobsoriano:
This PR ensures we only upload the `dist` folder to `npm` when publishing. Right now all packages uploads all files/folder to `npm`:

Screenshot 2024-09-11 at 10 37 17 PM


client/packages/admin/package.json
```diff
@@ -6,6 +6,7 @@
   "module": "dist/module/index.js",
   "types": "dist/module/index.d.ts",
   "unpkg": "dist/index.js",
+  "files": ["dist"],
```
stopachka: Thank you for the PR @wobsoriano 

Would you mind adding an npmignore file instead for `.turbo`? 

Reason: We want to include most of the other files. I think it's a bit of a nicer experience -- it won't show up in people's bundles, but it can be nice for folks to explore the code if they ever need to dig into the node_modules folder. (I sometimes find myself doing this for other projects) 

cc @markyfyi if you have thoughts

wobsoriano: @stopachka gotchu, updated!

stopachka (approved):
Thank you @wobsoriano !

PR #198, opened 2024-09-12
Add Yourkit to the readme

dwwoelfel:
Adds YourKit to the readme. 

It's a nice, easy-to-use java profiler that can profile both memory and cpu in a running JVM.

If we add their logo and link to their site, they'll give us a free license.


stopachka (approved):
Nice!

PR #202, opened 2024-09-12
Add missing todos namespace

gunayanil:
I guess `todos` namespace was forgotten in db.tx.


stopachka (approved):
Thank you @gunayanil !

PR #203, opened 2024-09-12
[www][sandbox] Log query and transaction errors in the output pane

markyfyi:
Screenshot 2024-09-12 at 2 06 08 PM


stopachka (approved):
Woohoo!

PR #204, opened 2024-09-12
Use a single db connection for transact

dwwoelfel:
_Easier to review with whitespace off https://github.com/instantdb/instant/pull/204/files?w=1_

We can get into a situation in `permissioned-transaction/transact!` where all of the db connections are in use, but we're waiting on a connection to fetch data for permissions before we can close our connection, leading to a deadlock. Eventually the connection will time out and things will proceed, but the db will be locked up in the meantime.

The solution is to use the connection we opened for all queries in the transaction, but we do a lot of parallel querying (`O(count(eids))`) to get permissions information and a single connection might hurt throughput. This PR adds pre-fetching so that we can only need to run 3 queries to get permissions (though we could get that down to 2 (or maybe even 1 if we're optimistic) if we really tried).

There are two kinds of prefetching we do: 
1. we prefetch triples for all eids from the tx-steps so that we can build the entity maps.
2. we prefetch relations for `data.ref` in the permissions rules.

To make 2. work, we walk the cel asts looking for `data.ref` calls and extract the paths.

There may be situations where we somehow don't prefetch the data we need. In that case, we'll just fall back to the old behavior and make another call to the database. In order to catch those situations, there are a couple of new macros in `instant.util.io`:

1. `io/tag-io` is used to tag a function as using io. I added `io/tag-io` to all of the sql helpers, since that's the majority of our io.
2. `io/warn-io` takes a keyword `context` arg and a body and will warn with a honeycomb trace if anybody tries to do any io in the body. We can look for `warn-io` in the logs to find places where we're not prefetching properly and improve our prefetching.

Since we have it, I also prefetch `data.ref` relations in instaql/permissioned-query. Testing with the `stickers` example saw a slight speedup, but nothing dramatic.

### DIFF
--- a/server/src/instant/db/permissioned_transaction.clj
+++ b/server/src/instant/db/permissioned_transaction.clj
@@ -12,13 +12,14 @@
    [next.jdbc :as next-jdbc]
    [instant.jdbc.sql :as sql]
    [instant.util.async :as ua]
-   [instant.util.exception :as ex]))
+   [instant.util.exception :as ex]
+   [instant.util.io :as io]))
 
 (defn extract-etype [{:keys [attrs]} attr-id]
   (attr-model/fwd-etype (attr-model/seek-by-id attr-id attrs)))
 
-;; -------------- 
-;; Check Commands 
+;; --------------
+;; Check Commands
 
 (defn tx-change-type [[action]]
   (cond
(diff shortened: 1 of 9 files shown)


server/src/instant/db/permissioned_transaction.clj
```diff
@@ -158,31 +168,31 @@
...
+     stopa-eid [[:add-triple stopa-eid :users/name \"Stopa\"]
+                [:add-triple stopa-eid :users/age 30]]
    }
 
    With this, we can generate a grouped `check` command for each `eid`."
   [ctx tx-steps]
   (let [eid->tx-steps (group-by second tx-steps)
         eid-actions (mapcat ->eid-actions eid->tx-steps)]
     (->> eid-actions
-         (ua/vfuture-pmap (partial object-check ctx)))))
```
dwwoelfel, line 185: No longer using `vfuture-pmap` because we shouldn't be doing any db fetches except for the preloads below. All of the work should be cpu-based.

stopachka: Beautiful!

server/src/instant/db/permissioned_transaction.clj
```diff
@@ -222,93 +232,162 @@
...
+                ;; update-delete data-ref dependencies in one go.
+                preloaded-triples (preload-triples ctx tx-steps)
+                ctx (assoc ctx :preloaded-triples preloaded-triples)
+                check-commands (io/warn-io :check-commands
+                                 (get-check-commands ctx tx-steps))
+
+                {create-checks true update-delete-checks false}
+                (group-by create-check? check-commands)
+
+                preloaded-update-delete-refs (preload-refs ctx update-delete-checks)
```
stopachka, line 376: 🫡🫡🫡

stopachka (approved):
ooh man! So darn cool

Issue #54, opened 2024-08-26
Setting an object key to undefined doesn't work locally


jmandel:
Here's an example of code that works _only when it's able to round-trip through the instantdb backend_. In other words,  my React component updates appropriately when the network is online, but fails to work locally when network is offline:

```
    await db.transact([
      tx.room[room.id].merge({
        players: {
          [playerName]: undefined
        }
      })
    ]);

```

I can work around this with the following (which works offline as well as online):

```
    await db.transact([
      tx.room[room.id].update({
        ...room,
        players: Object.fromEntries(
          Object.entries(room.players || {}).filter(([name]) => name !== playerName)
        ),
      })
    ]);

```

... but this shouldn't be required, according to https://www.instantdb.com/docs/instaml#:~:text=Sometimes%20you%20may%20want%20to%20remove%20keys, and the inconsistent behavior (offline vs online) is hard to reason about.

nezaj:
Thanks for reporting this! We're going to look into this soon!

jmandel:
Thanks for the response and the quick fix! You all are doing great work.

PR #205, opened 2024-09-12
[core][merge] Fix client deep-merge null logic

markyfyi:
Fixes https://github.com/instantdb/instant/issues/54.  I added a util to the `patch` example to replicate the case outlined.

Our logic in `immutableDeepMerge` was plain wrong, and we weren't deleted properties that were "nulled" out.

### DIFF
--- a/client/sandbox/react-nextjs/pages/play/patch.tsx
+++ b/client/sandbox/react-nextjs/pages/play/patch.tsx
@@ -35,6 +35,9 @@ export default function Patch() {
       <button className="border border-black" onClick={mergeWithNull}>
         merge `nestedData` with `null`
       </button>
+      <button className="border border-black" onClick={mergeWithDeepUndef}>
+        merge `nestedData` with `{`{new_key:undefined}`}`
+      </button>
       <button className="border border-black" onClick={deleteAttr}>
         delete `nestedData` attribute
       </button>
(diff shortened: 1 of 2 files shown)


stopachka (approved):
Cool!

stopachka (approved):
Cool!

client/sandbox/react-nextjs/pages/play/patch.tsx
```diff
@@ -65,6 +68,14 @@ export default function Patch() {
     ]);
   }
 
+  function mergeWithDeepUndef() {
+    db.transact([
+      tx.blocks[singletonIdX].merge({
+        nestedData: { new_key: undefined },
```
stopachka, line 74: Will this still work if the user goes offline and then back online? I thought `new_key` would disappear, since we save this to IDB

markyfyi: I don't follow.  This is the behavior that was broken

markyfyi: Oh, I see!  My original hypothesis was wrong, in fact, we were transforming undefined into null before stashing it into indexeddb

stopachka: Oh awesome! Where do we do it?

markyfyi: Right when it hits `Reactor`, in `pushTx`

PR #207, opened 2024-09-12
[phantom-presence] handle stray websockets

stopachka:
This PR: 
1. On the client, we make sure to close the previous websocket connection, when we are about to start a new one 
2. On the server, we started a ping-pong mechanism, where we detect idle connections and proactively kill them 

### Context

Kosmik reported a curious bug: If you went offline and went back online, you would get a phantom presence object. What happened?

### Bugs

The fundamental bug was this: tcp connections stayed 'active', _even when we went offline_. 

We assumed that the underlying websocket would close when we went offline. But this is not the case. Google Chrome does not close the socket automatically when you go offline. [1] 

What happens to idle TCP connections really depends on the machine [2]. Sometimes it can take minutes to close the connection.

### Solutions

I did two things in this PR: 

**1) Client: Close previous connections when starting new ones**

In Reactor, when we go back online, we run _startSocket. We used to assume that previous `_ws` would have been closed. Now no more: when we start a new socket, we'll make sure to close the previous one. 

**2) Server: Start ping/pong** 

Having the client close stray connections is a good step. But it doesn't quite cover it. For example, if the user goes offline, and _never_ goes back online, then what? 

The server could keep the connection around for minutes. Here's what we do now: the server sends `ping` messages every 5 seconds. If we don't receive some message within 10 seconds, we close the connection. 

[1] [Here's an interesting chrome bug report](https://code.google.com/p/chromium/issues/detail?id=76358)
[2] [Interesting microsoft article hiredman shared with us](https://learn.microsoft.com/en-us/troubleshoot/windows-client/networking/tcp-ip-port-exhaustion-troubleshooting#:~:text=After%20a%20graceful%20closure%20or%20an%20abrupt%20closure%20of%20a%20session%2C%20after%20a%20period%20of%204%20minutes%20(default)%2C%20the%20port%20used%20by%20the%20process%20or%20application%20would%20be%20released%20back%20to%20the%20available%20pool.%20During%20this%204%20minutes%2C%20the%20TCP%20connection%20state%20will%20be%20TIME_WAIT%20state.)

I kicked off a conversation about this on the [clojurians slack](https://clojurians.slack.com/archives/C03S1KBA2/p1726170248831079)

### DIFF
--- a/server/src/instant/lib/ring/websocket.clj
+++ b/server/src/instant/lib/ring/websocket.clj
@@ -10,7 +10,8 @@
   (:refer-clojure :exclude [send])
   (:require [ring.adapter.undertow.headers :refer [set-headers]]
             [instant.util.json :refer [->json]]
-            [instant.util.tracer :as tracer])
+            [instant.util.tracer :as tracer]
+            [instant.util.delay :as delay])
   (:import
    [io.undertow.server HttpServerExchange]
    [io.undertow.websockets
@@ -21,7 +22,7 @@
     BufferedBinaryMessage
     BufferedTextMessage
     CloseMessage
-    WebSocketChannel
+    StreamSourceFrameChannel WebSocketChannel
     WebSockets
     WebSocketCallback]
    [io.undertow.websockets.spi WebSocketHttpExchange]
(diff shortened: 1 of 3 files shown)


server/src/instant/lib/ring/websocket.clj
```diff
@@ -21,43 +22,71 @@
...
                      :data    (.getData message)}))
       (onFullBinaryMessage [^WebSocketChannel channel ^BufferedBinaryMessage message]
+        (.set atomic-last-received-at (System/currentTimeMillis))
         (let [pooled (.getData message)]
           (try
             (let [payload (.getResource pooled)]
               (on-message {:channel (channel-wrapper channel)
                            :data    (Util/toArray payload)}))
             (finally (.free pooled)))))
+      (onPong [^WebSocketChannel channel ^StreamSourceFrameChannel channel]
```
stopachka, line 61: You may be wondering: 

How does the client know to send a `pong`, we did not change Reactor. 

Most clients now can handle websocket pings by default. So since the server sends `ping`, chrome will automatically respond with a pong. 

(Tested this with chrome and expo iOS)

server/src/instant/reactive/session.clj
```diff
@@ -548,7 +533,6 @@
                                :http-req http-req
                                :ws-conn ws-conn
                                :receive-q receive-q
-                               :ping-job (start-ping-job store-conn id)
                                :pending-handlers pending-handlers}]
                    (on-open store-conn socket)))
       :on-message (fn [{:keys [data]}]
```
dwwoelfel, line 538: Should we update the `atomic-last-received-at` in `on-message` also?

dwwoelfel (approved):
Nice work!

server/src/instant/reactive/session.clj
stopachka: I pushed this down to the [caller of on-message](https://github.com/instantdb/instant/pull/207/files#diff-bef8276a9e702296afb121d63ec060dac0cca22a08b6e20994e347e62f810bf8R51-R52), so this already happens. Took the inspiration of updating on every message from Precursor : )

PR #209, opened 2024-09-13
Update useQuery code snippet

gunayanil:
**I am still not 100% sure** but I tried to make a few corrections in the `useQuery` code snippet.
- Add missing unclosed curly bracket `}` in the `useQuery`.
- Add "before" code and use `firstUser` instead of undefined `user` variable. 
- For "after" code, use `data.users` instead of undefined `user` variable. (I am still not sure if `data.users` will really return a single object instead of an array)

### DIFF
--- a/client/www/pages/docs/strong-init.md
+++ b/client/www/pages/docs/strong-init.md
@@ -62,7 +62,7 @@ const firstUser = data.users[0];
 const author = firstUser.author[0];
 
 // after
-const author = data.users.author; // no more array! 🎉
+const author = firstUser.author; // no more array! 🎉
 ```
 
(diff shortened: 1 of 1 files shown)


client/www/pages/docs/strong-init.md
```diff
@@ -55,12 +55,14 @@ Previously, all responses in a `useQuery` returned arrays. Now, we can use your
...
+const firstUser = data.users[0];
 
-const firstUser = data.users[0]
+// before
+const author = firstUser.author[0];
 
-// 🎉 after
-const author = user.author // no more array! 🎉
+// after
+const author = data.users.author; // no more array! 🎉
```
stopachka:
```suggestion
// after
const author = firstUser.author; // no more array! 🎉
```

stopachka (approved):
Thank you @gunayanil !

PR #213, opened 2024-09-13
Fix muts offline bug

markyfyi:
Fixes a bug where we were resolving query data before folding in pending mutations.  This was due to us resolving queries after loading cached query responses, but not waiting for cached mutations.

Context: https://discord.com/channels/1031957483243188235/1148284450992574535/1280807824334782484


stopachka (approved):
Lesgoo!

PR #215, opened 2024-09-13
[www][dash] save last viewed app

markyfyi:
Saves the last viewed app in LocalStorage so that we can show it when a user returns to the dash.

### DIFF
--- a/client/www/pages/dash/index.tsx
+++ b/client/www/pages/dash/index.tsx
@@ -327,7 +327,6 @@ function Dashboard() {
     if (!firstApp) return;
 
     const lastAppId = getLocal('dash_app_id');
-    console.log({ lastAppId });
 
     const defaultAppId = lastAppId ?? firstApp.id;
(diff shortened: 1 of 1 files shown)


client/www/pages/dash/index.tsx
```diff
@@ -320,19 +320,26 @@ function Dashboard() {
...
+    if (appId && isAppIdValid) return;
+
     const firstApp = apps?.[0];
     if (!firstApp) return;
 
-    const isAppIdValid = Boolean(apps.find((a) => a.id === appId));
-    if (appId && isAppIdValid) return;
+    const lastAppId = getLocal('dash_app_id');
+
+    const defaultAppId = lastAppId ?? firstApp.id;
```
stopachka, line 334: Would it be possible here that `lastAppId` is no longer valid? 

i.e someone deletes that app. 

We may want to do a check for that

stopachka (approved):
Lesgoo!

PR #220, opened 2024-09-16
Remove old CLI logi client-side flow

markyfyi:
Removes v0 of CLI auth where we tried hitting an ephemeral localhost server.


stopachka (commented):
Woohoo!

PR #222, opened 2024-09-17
[fix] suppress IOException errors on sendPingBlocking

stopachka:
We get some IOExceptions, when trying to send a ping to a socket that becomes closed. 

Afaik this can happen due to a race condition: the socket closes while we are running a sendPingBlocking 

[Example trace 1](https://ui.honeycomb.io/instantdb/environments/prod/datasets/instant-server/result/2FA3CtCUCka/trace?source=query&trace_id=1d3131aa8fee0d361069181df4ad87ef)
[Example trace 2](https://ui.honeycomb.io/instantdb/environments/prod/datasets/instant-server/result/2FA3CtCUCka/trace/jbnpdnkjv4A?fields[]=s_name&fields[]=s_serviceName&source=query&span=080bc468fb0979b3)

I say we suppress these errors, as they are expected. Open to other ideas though. 

@dwwoelfel @markyfyi @nezaj

### DIFF
--- a/server/src/instant/lib/ring/websocket.clj
+++ b/server/src/instant/lib/ring/websocket.clj
@@ -79,8 +79,10 @@
       (if (> ms-since-last-message idle-timeout-ms)
         (tracer/with-span! {:name "socket/close-inactive"}
           (IoUtils/safeClose channel))
-        (WebSockets/sendPingBlocking (ByteBuffer/allocate 0)
-                                     channel)))
+        (try (WebSockets/sendPingBlocking
+              (ByteBuffer/allocate 0)
+              channel)
+             (catch java.io.IOException _))))
     (catch Exception e
       (tracer/record-exception-span! e {:name "socket/ping-err"
                                         :escaping? false}))))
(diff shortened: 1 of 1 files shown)


nezaj (approved):
This is scoped to just `ping` events so fine by me!

server/src/instant/lib/ring/websocket.clj
```diff
@@ -79,8 +79,10 @@
       (if (> ms-since-last-message idle-timeout-ms)
         (tracer/with-span! {:name "socket/close-inactive"}
           (IoUtils/safeClose channel))
-        (WebSockets/sendPingBlocking (ByteBuffer/allocate 0)
-                                     channel)))
+        (try (WebSockets/sendPingBlocking
+              (ByteBuffer/allocate 0)
+              channel)
+             (catch java.io.IOException _))))
```
dwwoelfel: Can we just catch the closedchannelexception instead of all io exceptions?

stopachka: good idea! Will do this

stopachka: cc @dwwoelfel 

It looks like this can throw either a `ClosedChannelException`, _or_ a `IOException` with the message ['Channel is closed'](https://github.com/undertow-io/undertow/blob/main/core/src/main/java/io/undertow/websockets/core/WebSocketChannel.java#L353-L355).

I went ahead and updated the code to specifically catch: ClosedChannelException, and IOException with this message.

PR #223, opened 2024-09-17
Fix Typo in Clerk Auth Docs

shelwinsunga:
The package name is `@clerk/nextjs` and not `@clerk/next`


dwwoelfel:
Thanks for fixing!

PR #224, opened 2024-09-17
Env app IDs

markyfyi:
Change how we do app IDs in CLI/schema.

- Completely decouple schema.ts/`i.graph` from the concept of app IDs.  They're exclusively handled at a higher level in the CLI.
- Since we can no longer read an app ID from the schema, I modified the CLI to look for `process.env.INSTANT_APP_ID`.
- Update docs, examples, and error messages accordingly.
- Also add preliminary support for reading named IDs in `instant.config.ts`.  Users with multiple named apps can specify one in the CLI, e.g. `instant-cli pull dev`, where `dev` is a key in a map of names to IDs. **Note, I'm not documenting/announcing this feature until I get feedback from our core testers.**

Once we land this, I'll announce the breaking changes in Discord.

How to test:

cd into `sandbox/cli-nodejs` and try these out

1. Run a command without a `.env`.  It should give you a helpful error.
1. Run a command with `INSTANT_APP_ID` in `.env`.  It should work.
1. Run a command with a malformed `INSTANT_APP_ID` in `.env`.  It should give a helpful error.
1. Run a command with an app ID as an arg.  It should work.
1. Run a command with a malformed app ID as an arg.  It should give a helpful error.
1. BONUS: Run a command with an `instant.config.ts` file like the example below.  It should work.

```ts
// example config file
export default {
  apps: {
    coolApp: {
      id: "ID_HERE_PLS",
    },
  },
};

```

### DIFF
--- a/client/packages/cli/index.js
+++ b/client/packages/cli/index.js
@@ -15,16 +15,16 @@ import openInBrowser from "open";
 
 // config
 
-const dev = Boolean(process.env.DEV);
-const verbose = Boolean(process.env.VERBOSE);
+const dev = Boolean(process.env.INSTANT_CLI_DEV);
+const verbose = Boolean(process.env.INSTANT_CLI_VERBOSE);
 
 const instantDashOrigin = dev
   ? "http://localhost:3000"
   : "https://instantdb.com";
 
-const instantBackendOrigin = dev
-  ? "http://localhost:8888"
-  : "https://api.instantdb.com";
+const instantBackendOrigin =
+  process.env.INSTANT_CLI_API_URI ||
+  (dev ? "http://localhost:8888" : "https://api.instantdb.com");
 
 // cli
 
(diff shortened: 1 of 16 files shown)


client/packages/cli/index.js
```diff
@@ -855,11 +909,16 @@ import { i } from "${instantModuleName ?? "@instantdb/core"}";
...
 ${indentLines(JSON.stringify(linksEntriesCode, null, "  "), 1)}
 );
 
 export default graph;
 `;
 }
+
+const noAppIdErrorMessage = `
+No app ID found.
+Add \`INSTANT_APP_ID=<ID>\` to your .env file.
```
stopachka, line 933: nit: I would link to the docs, or explain that they can use NEXT_ VITE_ etc

stopachka (approved):
Woohoo! Nice

PR #226, opened 2024-09-17
CLI env auth

markyfyi:
Enhancing CLI to support CI/automation.

1. Adds a `--print` flag to `login` so that a user can grab their auth token for later consumption.
2. Updates CLI actions to check for `INSTANT_CLI_AUTH_TOKEN`, similar to how we handle `INSTANT_APP_ID`.  This is useful for customers in build automation envs like Vercel or GHA.


stopachka (approved):
Cool!

PR #230, opened 2024-09-18
Expose peerId in presence user and peers objects

markyfyi:
https://discord.com/channels/1031957483243188235/1281153699179135027

```
Right now we only get peer ids for other users, which forces us to maintain our own user.peerId  to be able to properly do multiplayer.
This is because the same user can be connected with more than one client, so we can't rely on the user.id  for that matter.
Exposing the current user's peerId would allow to uniquely identify users no matter what
```

### DIFF
--- a/client/packages/core/src/presence.ts
+++ b/client/packages/core/src/presence.ts
@@ -18,10 +18,17 @@ export type PresenceOpts<PresenceShape, Keys extends keyof PresenceShape> = {
   keys?: Keys[];
 };
 
+type PresencePeer<PresenceShape, Keys extends keyof PresenceShape> = Pick<
+  PresenceShape,
+  Keys
+> & {
+  peerId: string;
+};
+
 export type PresenceSlice<PresenceShape, Keys extends keyof PresenceShape> = {
-  user?: Pick<PresenceShape, Keys> & { peerId: string };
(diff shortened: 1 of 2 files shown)


client/packages/core/src/presence.ts
```diff
@@ -19,9 +19,9 @@ export type PresenceOpts<PresenceShape, Keys extends keyof PresenceShape> = {
 };
 
 export type PresenceSlice<PresenceShape, Keys extends keyof PresenceShape> = {
-  user?: Pick<PresenceShape, Keys>;
+  user?: Pick<PresenceShape, Keys> & { peerId: string };
```
stopachka: nit: perhaps we create a new type for `Pick<PresenceShape, Keys> & { peerId: string };` -- `PresenceEntry` or something?

stopachka (approved):
SGTM!

client/packages/core/src/presence.ts
```diff
@@ -42,6 +42,7 @@ export function buildPresenceSlice<
     peers: Record<string, PresenceShape>;
   },
   opts: PresenceOpts<PresenceShape, Keys>,
+  userPeerId: string,
 ) {
```
stopachka: nit: Perhaps we should have an output type here?

PR #231, opened 2024-09-18
[clean] remove babashka-pod files

stopachka:
Realized we had some stray babashka-pod files. Removed them and added pattern to gitignore

@nezaj @dwwoelfel @markyfyi


dwwoelfel (approved):
Thanks!

PR #233, opened 2024-09-18
rough perf sandbox

stopachka:
We want to massively improve perf for queries and transactions. Some users provided great repros. I added this repro in our sandbox, so we can play with it on latest instant and local backend. 

```
git co perf-research 
pnpm i 
pnpm run dev 
# load  http://localhost:5002 
```

I don't think this will stay as a separate sandbox for the long-term; Likely we should merge this into our nextjs examples page, but for now just keeping it separate to iterate.

@nezaj @dwwoelfel @markyfyi


dwwoelfel (approved):
Love it!

PR #237, opened 2024-09-18
Select attrs for the namespace

dwwoelfel:
When we fetch the attributes for an entity in instaql queries, we fetch all attributes for the namespace. Some users reuse ids across namespaces, so we end up polluting their entities with attributes from other namespaces.

This PR updates instaql and permissions checks to fetch only the attributes that are defined in the namespace. We're also more careful about which attributes we delete when deleting an entity.

Also creates a new `Attrs` type in `attr-model` that looks and acts like a list, but has some extra methods to efficiently look up attrs by id, etype, fwd-name, and rev-name. The first time you call one of those methods, it will index the attrs, then it will use the indexes on any future calls. It lets us get rid of the manual indexing that we were doing.


server/src/instant/db/model/attr.clj
```diff
@@ -341,41 +340,118 @@
...
+(defprotocol AttrsExtension
+  (seekById [this id])
+  (seekByFwdIdentName [this fwd-ident])
+  (seekByRevIdentName [this revIdent])
+  (attrIdsForEtype [this etype]))
+
+;; Creates a wrapper over attrs. Makes them act like a regular list, but
+;; we can also index them on demand so that our access patterns will be
+;; efficient.
+(deftype Attrs [elements cache]
```
stopachka, line 374: biyutiful!

stopachka (approved):
Let's do thiis!

PR #238, opened 2024-09-18
[perf][1] cache dataForResult

stopachka:
We don't cache `dataForResult` right now. This means that when we render in dev, we get: 

1. 1 call for dataForResult from notifyOne
2. 2 calls for `dataForResult`, when running `useQuery`

This is a very light cache for dataForResult; there's still going to be false positives [1]. _But_ this will remove the 2 unnecessary in dev, and 1 in prod. Here's how it looks from a flame graph: 

**Before**
CleanShot 2024-09-18 at 16 56 43@2x

**After** 
CleanShot 2024-09-18 at 16 57 20@2x

Lots moar coming! 

@nezaj @dwwoelfel @markyfyi 

[1] Eventually, we should really look into turning a bunch of our objects into observables. It would be _so_ cool if we could quickly derive whether a `dataForResult` is stale or not, based on whether it's dependents have changed.

### DIFF
--- a/client/packages/core/src/Reactor.js
+++ b/client/packages/core/src/Reactor.js
@@ -90,6 +90,7 @@ export default class Reactor {
   _broadcastSubs = {};
   _currentUserCached = { isLoading: true, error: undefined, user: undefined };
   _beforeUnloadCbs = [];
+  _dataForResultCache = {};
 
   constructor(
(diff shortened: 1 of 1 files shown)


dwwoelfel (approved):
LGTM!

PR #241, opened 2024-09-19
Improve dataForResult cache

stopachka:
In our previous perf improvement, we lost one advantage: 

Consider the following scenario: 
1. We load everything from storage
2. Afterwards, we get a `useQuery`

In this scenario, there would be no data for `getPreviousResult`, since 1. would not have done a `notifyOne`. 

I circumvented this problem by making the cache a bit smarter. Now, we run the `dataForQuery` in _both_ `notifyOne` _and_ `getPreviousResult`. If those calls run on the same 'version' of querySubs and pendingMutations, they return from the cache.

**Before**
image

**After** 
(No change)
CleanShot 2024-09-19 at 11 08 22@2x

@nezaj @dwwoelfel @markyfyi

### DIFF
--- a/client/packages/core/tsconfig.json
+++ b/client/packages/core/tsconfig.json
@@ -9,6 +9,7 @@
     "outDir": "dist",
     "rootDir": "src",
     "sourceMap": true,
+    "inlineSources": true,
     "target": "es2015",
     "esModuleInterop": true,
     "moduleResolution": "node",
--- a/client/packages/core/src/Reactor.js
+++ b/client/packages/core/src/Reactor.js
@@ -90,7 +90,7 @@ export default class Reactor {
   _broadcastSubs = {};
   _currentUserCached = { isLoading: true, error: undefined, user: undefined };
   _beforeUnloadCbs = [];
-  _dataForResultCache = {};
+  _dataForQueryCache = {};
 
   constructor(
     config,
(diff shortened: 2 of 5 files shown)


client/packages/core/tsconfig.json
```diff
@@ -9,6 +9,7 @@
     "outDir": "dist",
     "rootDir": "src",
     "sourceMap": true,
+    "inlineSources": true,
```
stopachka, line 12: It _seems_ like this is okay: 

https://chatgpt.com/share/66ec6a75-9ca0-8003-ae2a-916d472ba29a

dwwoelfel, line 12: Will inlineSources increase the size of the bundle we publish to npm?

dwwoelfel:
It look good to me, but any idea why the preview deploy is failing with `TypeError: Cannot read properties of undefined (reading 'version')` while building examples?

stopachka:
> It look good to me, but any idea why the preview deploy is failing with TypeError: Cannot read properties of undefined (reading 'version') while building examples?

It was a real bug! When run on server, `this.querySubs` came in as undefined. Fixed 🫡

client/packages/core/tsconfig.json
stopachka: Afaik it will not. Here's what I did: 

1. Asked chatgpt: https://chatgpt.com/share/66ec6a75-9ca0-8003-ae2a-916d472ba29a 
2. Looked at the `dist` folder. It looks like all the js files just have a single comment at the bottom, pointing to a .map file. Afaik I don't think those .map files will ship on a production bundle

PR #242, opened 2024-09-19
Add validation to check for duplicate attrs

markyfyi:
Adds some logic to our schema change planner to bail when a user defines a link twice.

```
~/instant/client/sandbox/cli-nodejs λ INSTANT_CLI_DEV=1 px instant-cli push -y 
Planning...
Failed to update schema.
Validation failed for schema
schema: Duplicate entry found for attribute: tags->posts. Check your schema file for duplicate link definitions.
schema: Duplicate entry found for attribute: posts->tags. Check your schema file for duplicate link definitions.
```

Also fixes a bug where we weren't bailing `push perms` when `push schema` failed.

### DIFF
--- a/server/src/instant/model/schema.clj
+++ b/server/src/instant/model/schema.clj
@@ -189,5 +189,3 @@
              :rules (rule-model/get-by-app-id aurora/conn-pool
                                               {:app-id app-id})}]
     (permissioned-tx/transact! ctx steps)))
-
-*e
--- a/server/src/instant/model/schema.clj
+++ b/server/src/instant/model/schema.clj
@@ -12,7 +12,7 @@
   (into {} (map (fn [[k v]] [k (f [k v])]) m)))
 
 (defn attr-ident-names [attr]
-  (keep identity [(attr-model/fwd-ident-name attr) (attr-model/rev-ident-name attr)]))
+  (keep seq [(attr-model/fwd-ident-name attr) (attr-model/rev-ident-name attr)]))
 
 (defn schemas->ops [current-schema new-schema]
   (let [{new-blobs :blobs new-refs :refs} new-schema
@@ -151,7 +151,6 @@
 
 (defn assert-unique-idents! [current-attrs steps]
   (let [current-ident-names (->> current-attrs
-                                 (filter (comp #{:ref} :value-type))
                                  (mapcat attr-ident-names)
                                  (map vec))
         ident-names (->>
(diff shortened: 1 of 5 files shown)


server/src/instant/model/schema.clj
```diff
@@ -158,3 +189,5 @@
              :rules (rule-model/get-by-app-id aurora/conn-pool
                                               {:app-id app-id})}]
     (permissioned-tx/transact! ctx steps)))
+
+*e
```
stopachka: nit: remove

server/src/instant/model/schema.clj
```diff
@@ -4,13 +4,22 @@
...
+            [instant.util.exception :as ex])
   (:import (java.util UUID)))
 
 (defn map-map [f m]
   (into {} (map (fn [[k v]] [k (f [k v])]) m)))
 
-(defn schemas->ops [current-schema new-schema]
+(defn link-dir-keys [{ri :reverse-identity
+                      fi :forward-identity}]
+  (concat [[(-> fi second)
```
stopachka: I would write: 

```
(defn attr-ident-names [attr] 
     (keep identity [(attr-model/fwd-ident-name attr) (attr-model/rev-ident-name attr)))
```

server/src/instant/model/schema.clj
```diff
@@ -77,10 +86,24 @@
...
+              steps
+              (mapcat (fn [[op data]]
+                        (when (= op :add-attr)
+                          (link-dir-keys data))))
+              (frequencies)
+              (filter (fn [[_ v]] (> v 1))))]
+    (when (seq dups)
+      (ex/assert-valid! :schema
+                        :steps
+                        (map (fn [[[ns attr]]]
```
stopachka: nit: would use the names `etype label`

stopachka (approved):
Cool!

server/src/instant/model/schema.clj
markyfyi: `keep identity` neat

server/src/instant/model/schema.clj
```diff
@@ -135,20 +141,63 @@
...
+  (str "Duplicate entry found for attribute: "
+       etype
+       "->"
+       label
+       ". "
+       "Check your schema file for duplicate link definitions."))
+
+(defn assert-unique-idents! [current-attrs steps]
+  (let [current-ident-names (->> current-attrs
+                                 (filter (comp #{:ref} :value-type))
```
stopachka: q: is it necessary to scope only to `ref` attrs?

stopachka (approved):
LGTM!

server/src/instant/model/schema.clj
markyfyi: Good catch.  Duplicate blobs should be filtered out by `schemas->ops`, but we can make `assert-unique-idents!` more robust by removing this filter.

server/src/instant/model/schema.clj
markyfyi: `fwd-ident-name` returns an empty list in the null case, so it was `keep seq`

nezaj:
@markyfyi is this ready to merge?

PR #243, opened 2024-09-19
Use mutative instead of immer

stopachka:
Immer autofreezes objects. When the object is really large, this gets more and more costly. I switched us to mutative, which does not rely on auto-freeze. [They have a comparison about immer vs mutative here](https://mutative.js.org/docs/extra-topics/comparison-with-immer). 

I made a bunch of transactions in our perf tester. 

**Before** 
[many_tx_immer.json](https://github.com/user-attachments/files/17065226/many_tx_immer.json)

A single transaction ontop of the bench query on immer is about 8ms. As you add more pending transactions, it grows to about 40ms.
image

**After** 
[many_tx_mutative.json](https://github.com/user-attachments/files/17065227/many_tx_mutative.json)

Mutative is about 3ms on a single transaction. As I add more pending transactions, the size _seems to stay_ at 3ms! 

CleanShot 2024-09-19 at 12 10 26@2x

@dwwoelfel @markyfyi @nezaj

### DIFF
--- a/client/packages/core/package.json
+++ b/client/packages/core/package.json
@@ -33,7 +33,7 @@
     "vitest": "^1.6.0"
   },
   "dependencies": {
-    "immer": "9.0.14",
+    "mutative": "^1.0.10",
     "object-hash": "^3.0.0",
     "uuid": "^9.0.0"
   }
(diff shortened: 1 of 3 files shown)


dwwoelfel (approved):
LGTM!

PR #244, opened 2024-09-19
[perf][store] improve from and to json

stopachka:
**Our `to` and `from` json parsing is now 10x faster!**

For large query results, `toJSONWithMaps` and `fromJSONWithMaps` was taking a long time. 

I had two hunches
- I think when we add a 'reviver', the codepath inside v8 likely goes through a less optimized path. 
- We also added more data than was strictly needed to parse. For example, we don't need the indexes -- we could derive them. 

## Before
CleanShot 2024-09-19 at 12 39 50@2x
CleanShot 2024-09-19 at 12 40 23@2x

## Attempt 1: Using `*JSONwithStores` | ~3.5X improvement
This still used a reviver, but 
- a) removed the indexes
- b) only checked for the 'store' type, rather than all Maps 

CleanShot 2024-09-19 at 12 41 04@2x
CleanShot 2024-09-19 at 12 41 27@2x

## Attempt 2: Hand-tuned | ~10x improvement
This _did not_ use a reviver. Instead: 
- a) we parsed the json
- b) then we walked it. 

CleanShot 2024-09-19 at 12 41 04@2x
CleanShot 2024-09-19 at 12 41 04@2x


dwwoelfel (approved):
So fast!

PR #246, opened 2024-09-19
[perf] improve local query perf 2.7x

stopachka:
**This PR makes queries ~2.7x faster** 

## Main Change: datalog query _only_ for ids

Say we had a query like `{ posts: {} }`. This would produce a datalog query like:

```clojure
[[?id :posts/id] 
 [?id ?attr ?v]]
```

datalog would end up taking up to 25ms to finish this query. Likely this is because our join algo needs more tuning. Instead, I changed it so: 

1. datalog _only_ queries for eids. 
2. Once we have eids, we 'generate' the object manually

**Before**
CleanShot 2024-09-19 at 16 24 01@2x

**After**
CleanShot 2024-09-19 at 16 23 02@2x

**Note:** In both instances `query` is a bit contaminated by a react update. Afaik react starts rendering when we run `cb()`, and for some reason the profiler counts it as part of `query`. I am not sure why. For this demo I just made React return null, so we saw the results closer to what `query` is actually doing 

## Total impact so far: 

https://paper.dropbox.com/doc/Perf-Impact--CXK5lkdqoXPXTiH7WurEs1kyAg-DEYYXdPeU5ZrP1CnCqsNO

[5200_before.json](https://github.com/user-attachments/files/17068121/5200_before.json)

[5200_after.json](https://github.com/user-attachments/files/17068126/5200_after.json)

@nezaj @dwwoelfel @markyfyi

### DIFF
--- a/client/packages/core/__tests__/src/data/zeneca/attrs.json
+++ b/client/packages/core/__tests__/src/data/zeneca/attrs.json
@@ -18,7 +18,7 @@
   },
   {
     "id": "1fcc2b72-02be-405b-918c-8e441f75dcab",
-    "value-type": "ref",
+    "value-type": "blob",
     "cardinality": "one",
     "forward-identity": [
       "a09d83c4-c587-41fd-b660-bd02a2269dfa",
@@ -35,7 +35,7 @@
   },
   {
     "id": "6eebf15a-ed3c-4442-8869-a44a7c85a1be",
-    "value-type": "ref",
+    "value-type": "blob",
     "cardinality": "one",
     "forward-identity": ["4ceb593d-8953-4441-81c4-21cc521ef6dd", "books", "id"],
     "unique?": true,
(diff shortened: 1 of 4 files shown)


client/packages/core/__tests__/src/data/zeneca/attrs.json
```diff
@@ -18,7 +18,7 @@
   },
   {
     "id": "1fcc2b72-02be-405b-918c-8e441f75dcab",
-    "value-type": "ref",
+    "value-type": "blob",
```
stopachka, line 21: I changed these to be refs, as all our current apps keep ids as blobs

client/packages/core/src/instaql.js
```diff
@@ -333,43 +317,29 @@ function isBefore(startCursor, direction, [e, a, _v, t]) {
...
-  for (const e of toRemove) {
-    delete res[e];
+  let objects = {}
+  const startCursor = pageInfo?.["start-cursor"];
+  const blobAttrs = s.blobAttrs(store, etype);
+  for (const [id, time] of idVecs) {
+    if (
+      startCursor &&
+      aid === startCursor[1] &&
+      isBefore(startCursor, direction, [id, aid, id, time])
```
dwwoelfel, line 333: We'll have to rewrite this a bit if we add sorting on user-defined attributes.

dwwoelfel (approved):
LGTM!

client/packages/core/src/instaql.js
stopachka: 🫡

PR #254, opened 2024-09-22
[Dashboard] Set boolean default to false

nezaj:
[Two users](https://discord.com/channels/1031957483243188235/1148284450992574535/1286330010818576486) noted that when you use the schema editor we set booleans to true by default. We should actually set it to false by default so users don't accidentally turn it on when editing the row.


stopachka:
Lesgoo!

PR #255, opened 2024-09-22
Update and

nezaj:
[Cam](https://discord.com/channels/1031957483243188235/1148284450992574535/1286828614951632936) asked a great question about `AND` vs `IN.` Clarifying our docs to explain the difference and to also make it clear that `IN` is a shorthand for `OR` on one key


dwwoelfel (approved):
LGTM!

Issue #250, opened 2024-09-20
`npx instant-cli login` fails


kenrick95:
Causes by #235 

Apparently it is because it uses `import.meta.dirname` which is available only in Node >=21.2.0 && ^20.11.0

Unluckily, I was using Node 20.10.0. I could upgrade my Node to solve my problem, however a nicer error message would be nice saying on which Node versions are required to run instant-cli.

At the moment the error is quite cryptic:

```
node:internal/validators:162
    throw new ERR_INVALID_ARG_TYPE(name, 'string', value);
          ^

TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received undefined
    at validateString (node:internal/validators:162:11)
    at join (node:path:433:7)
    at file:///path/to/my/project/node_modules/.pnpm/instant-cli@0.14.0/node_modules/instant-cli/index.js:20:16
    at ModuleJob.run (node:internal/modules/esm/module_job:218:25)
    at async ModuleLoader.import (node:internal/modules/esm/loader:329:24)
    at async loadESM (node:internal/process/esm_loader:34:7)
    at async handleMainPromise (node:internal/modules/run_main:113:12) {
  code: 'ERR_INVALID_ARG_TYPE'
}

Node.js v20.10.0
```

nezaj:
@kenrick95 thanks for letting us know! This is fixed #256. `npx instant-cli ...` should now work!

kenrick95:
Thank you~

PR #256, opened 2024-09-22
[instant-cli] Update version command

nezaj:
Provides a fix for #250.

I also thought it would be nice to let `-v` work for version. (previously you would have to use uppercase V, which was not consistent with our other options)


stopachka:
Niice!

PR #263, opened 2024-09-23
Fix errors in generated d.ts files

markyfyi:
Somehow, we're getting type errors in the `.d.ts` files generated in our builds, but not in the source `.ts` files. 😱  I'm still looking into how/why, but shipping a fix in the meantime.


stopachka (approved):
Thank you @markyfyi !

PR #264, opened 2024-09-23
Fix errors in generated d.ts files in the rest of our packages

markyfyi:
I just checked all of our `.d.ts` files and found a handful of similar type errors.

cc @stopachka


stopachka (approved):
woohoo!

PR #266, opened 2024-09-23
Production core types

markyfyi:
When debugging a repo with TS's `skipLibCheck = false`, I was getting type errors about `object-hash` et al missing type defs.

Also, noticed we had types for `lodash`, but no `lodash`!

### DIFF
--- a/client/packages/core/package.json
+++ b/client/packages/core/package.json
@@ -21,10 +21,7 @@
   "devDependencies": {
     "@babel/core": "^7.17.9",
     "@babel/preset-env": "^7.16.11",
-    "@types/lodash": "^4.14.182",
     "@types/node": "^18.11.18",
-    "@types/object-hash": "^3.0.3",
-    "@types/uuid": "^8.3.4",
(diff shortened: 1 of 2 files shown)


client/packages/core/package.json
```diff
@@ -33,6 +30,8 @@
     "vitest": "^1.6.0"
   },
   "dependencies": {
+    "@types/object-hash": "^3.0.3",
+    "@types/uuid": "^8.3.4",
```
dwwoelfel: I stopped exporting the uuid class in https://github.com/instantdb/instant/pull/48/files, so we might not need `@types/uuid` in `dependencies`

dwwoelfel (approved):
One comment, but LGTM! 

I guess this explains why someone was having problems with missing uuid types.

client/packages/core/package.json
markyfyi: Nice, I think you're right!  We're not importing `uuid` in any of the `.d.ts` files.

PR #269, opened 2024-09-24
[Dash] Make sandbox transact return more similar

nezaj:
[[One of our users]](https://discord.com/channels/1031957483243188235/1287593736930463856/1287926659915452476) was using the sandbox to debug and was confused by the return of `transact.` In the sandbox we extract out the `tx-id` but in actual instaml we don't.

This change makes the sandbox `transact` return match more closely to the real `transact`


dwwoelfel (approved):
Looks good to me!

PR #271, opened 2024-09-24
[Admin] Add test for ignoring id

nezaj:
Realized I should also include a test for #270 to help get notified of a future regression


stopachka (approved):
Niice!

PR #277, opened 2024-09-24
Surface advanced features in docs

markyfyi:
I notice users in Discord still asking about types and schemas.  This patch makes an effort to surface awareness of these features in our onboarding docs.

### DIFF
--- a/client/www/pages/docs/index.md
+++ b/client/www/pages/docs/index.md
@@ -8,7 +8,7 @@ Instant is the Modern Firebase. With Instant you can easily build realtime and c
 
 Curious about what it's all about? Try a {% blank-link href="https://instantdb.com/tutorial" label="demo" /%}. Have questions? {% blank-link href="https://discord.com/invite/VU53p7uQcE" label="Join us on discord!" /%}
 
-And if you're ready, follow the quick start below, to build a live app in less than 5 minutes! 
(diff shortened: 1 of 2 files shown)


client/www/pages/docs/index.md
```diff
@@ -257,3 +257,11 @@ export default App
 Go to `localhost:3000` and follow the final instruction to load the app!
 
 Huzzah 🎉 You've got your first Instant web app running! Check out the [**Explore**](/docs/init) section to learn more about how to use Instant :)
+
+{% callout type="note" %}
+
+**Level up your app!**
```
nezaj: I don't think we should include this yet in our Getting Started. This is a more advanced concept, I think it would be better to recommend folks to go deeper and check out our `init` docs

client/www/pages/docs/init.md
```diff
@@ -41,3 +41,11 @@ const db = init<MyAppSchema>({ appId: APP_ID });
 ```
 
 You'll now be able to use `InstaQL` and `InstalML` throughout your app!
+
+{% callout type="note" %}
+
+**Level up your app!**
```
nezaj:
```suggestion
**Schema as code and type safety**
```

nezaj (commented):
Left some comments!

PR #280, opened 2024-09-25
Users table

dwwoelfel:
Adds a `$users` namespace that provides a read-only view into all of the users from the `app_users` table.

It comes with default rules that only allow the authenticated user to view their entry in the `$users` namespace.

### How it works

When you enable from the `Admin` page, we add `$users.id` and `$users.email` attributes and we add rules.

For queries, we keep track of which namespaces are referenced in the `instaql` query. If we see the `$users` namespace, we prepend a `triples` CTE to the postgres query that includes all of the rows from the `triples` table, plus all of the rows from the `app_users` table. The `app_users` rows are coerced into the triples format, using the attr ids we added. That way queries further down will see the data in `app_users` as ordinary triples and all of our query machinery just works.

There is some extra overhead from merging in the `$users` triples, so we only do it when the `$users` namespace needs to be queried.

For transactions, we reject any writes to the `$users` namespace. You can still add reverse links to the $users namespace, e.g. you could add `posts.created_by` -> `$users.id`. We'll verify that you have view access to the user before we let you link it.

For reactivity, we watch for inserts to the `app_users` table and convert them into inserts into the `triples` table before they get passed to the invalidator. We need the mapping from `$users.id` and `$users.email` to their attr ids, so we also watch for changes to `attrs` and `idents` to keep that mapping up to date.

### Various changes

- Users can no longer create a namespace that starts with `$`. Those are now explicitly reserved for system tables.
- You'll need to have `view` access on any linked values in order to link them.
- There's a new `merge!` function in `rules` that will merge the new rules with the existing rules.

Before merging, I'll add a temporary feature flag on the frontend so that we can have time to deploy the backend.


server/resources/migrations/31_attrs_replica_full.up.sql
```diff
@@ -0,0 +1 @@
+alter table attrs replica identity full;
```
stopachka, line 1: awwh yeah. Would be interesting to later try this with triples too! Maybe when we have better observability over invalidation

server/src/instant/db/datalog.clj
```diff
@@ -1556,10 +1566,12 @@
...
 (defn send-query-single
   "Sends a single query, returns the join rows."
-  [conn app-id named-patterns]
+  [ctx conn app-id named-patterns]
   (tracer/with-span! {:name "datalog/send-query-single"}
     (let [{:keys [query pattern-metas]} (match-query :match-0- app-id named-patterns)
-          sql-query (hsql/format query)
+          sql-query (hsql/format
+                     (update query :with (fn [ctes]
+                                           (maybe-add-users-shim ctx app-id ctes))))
```
stopachka, line 1574: impressive how little needed to change in this ns!

server/src/instant/db/permissioned_transaction.clj
```diff
@@ -172,47 +227,48 @@
...
+                                                          value
+                                                          [{:message "Expected link value to be a uuid."
+                                                            :hint {:tx-step tx-step}}])))]))]
+              (cond-> acc
+                true (update {:eid eid
+                              :etype etype
+                              :action (tx-object-action-type tx-step)}
+                             (fnil conj [])
+                             patched-step)
+                rev-etype (update {:eid rev-eid
```
stopachka, line 267: (thinking through) 

I am feeling a bit ~iffy about this additional check. My main worry is that the `view` rule is being overloaded here. It isn't always true that if you can `view` an object, you can link to it. 

One alternative thought I had was that perhaps it should be the `update` rule, but I realize this would mean $users wouldn't quite work. 

I think the 'most general' solution would have to look like an 'after_save' hook or something.  

But, given all the constraints, I think adding this extra view rule is strictly more helpful than the present state. 

So en fin, let's roll with it 🫡

server/src/instant/db/transaction.clj
```diff
@@ -121,12 +122,23 @@
 (comment
   (batch (gen/generate (s/gen ::tx-steps))))
 
-(defn transact-without-tx-conn! [conn app-id tx-steps]
+(defn prevent-$users-updates [op attrs]
+  (doseq [attr attrs
+          :let [ns (-> attr :forward-identity (nth 1))]]
```
stopachka: nit: would call this `etype`. Perhaps even use `attr-model/fwd-etype`. (Main reason: `ns` is already defined in clojure and makes the syntax colors look a bit off)

server/src/instant/model/rule.clj
```diff
@@ -17,6 +18,17 @@
...
+(defn merge!
+  ([params] (merge! aurora/conn-pool params))
+  ([conn {:keys [app-id code]}]
+   (sql/execute-one!
+    conn
+    (hsql/format {:insert-into :rules
+                  :values [{:app-id app-id
+                            :code [:cast (->json code) :jsonb]}]
+                  :on-conflict :app-id
+                  :do-update-set {:code [:|| :rules.code :excluded.code]}}))))
```
stopachka, line 30: daam! :||

server/src/instant/reactive/invalidator.clj
```diff
@@ -332,25 +432,27 @@
...
-      {:wal-chan wal-chan
-       ;; Nothing will ever be put on this chan,
-       ;; it will be closed when the wal-chan is closed
-       ;; so that the consumer can know to stop waiting for
-       ;; its puts to complete
-       :close-signal-chan (a/chan)
-       :worker-chan worker-chan
-       :byop-chan byop-chan})))
+  (let [users-shims (tracer/with-span! {:name "invalidator/init-users-shims"}
+                      (atom (attr-model/get-all-users-shims aurora/conn-pool)))]
```
stopachka: Is there a race condition, where in the split second before wal-listener starts, some updates could miss? (i.m.o am fine to keep that race-condition around!)

dwwoelfel: There is a race condition. I wasn't going to worry about it because it seems unlikely that we'd hit it and I thought it would be a lot of work to fix, but turns out its pretty easy to fix.

stopachka: Monster! 🫡

PR #281, opened 2024-09-25
Extract base tsconfig from client packages

markyfyi:
Utilizes [`extends`](https://www.typescriptlang.org/tsconfig/#extends) to share common config across our core client packages.  I wanted to do `www` and `sandbox` too but they were all pretty different.

### DIFF
--- a/client/packages/react/tsconfig.json
+++ b/client/packages/react/tsconfig.json
@@ -1,21 +1,6 @@
 {
-  "include": ["src"],
-  "exclude": ["node_modules", "dist"],
   "compilerOptions": {
     "jsx": "react-jsx",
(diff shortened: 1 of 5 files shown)


client/packages/react/tsconfig.json
```diff
@@ -1,21 +1,10 @@
 {
+  "extends": "../../tsconfig.base.json",
```
suhasdeshpande, line 2: This extend is complaining that there is no `tsconfig.base.json`. Is it possible that base json is not part of final artifact when this extends in packages requires it to be? 

Screenshot 2024-10-31 at 08 20 41
Screenshot 2024-10-31 at 08 21 01

PR #282, opened 2024-09-26
Show a landing page if the oauth flow redirects to external app

dwwoelfel:
If our oauth flow redirects to an external app, the redirect will leave the redirecting window dangling. In the case of Google, the last page they redirect from is a loading page, so it looks like the google page is loading forever.

With this change, we'll show a landing page that automatically opens the external redirect url. It also has a button in case the user clicks "Cancel".

The page looks like this:
image

Confirmed that it doesn't modify the oauth flow for expo.


nezaj (approved):
Nice!!

PR #285, opened 2024-09-26
Query Once

markyfyi:
Adds `queryOnce` to our suite of client SDKs.  `queryOnce` is basically `useQuery`, but returns a promise instead of registering a subscription.

- `queryOnce` is a hack on top of `subscribeQuery`/`add-query`.
- Along the way, I implemented "true" un-subscription, i.e. sending the `remove-query` op to the server when we no longer need to listen to a query.  I updated the current `subscribeQuery` flow with this, too.  Credit to @stopachka for the idea.

I added an example (`/play/query-once`) to demonstrate various cases:
- `queryOnce` in `useEffect` on init
- `queryOnce` on an existing live subscription (the `add-query-exists` case)
- `queryOnce` with a new query

I'll add docs/docstrings once we're happy with the implementation!

### DIFF
--- a/client/packages/core/src/Reactor.js
+++ b/client/packages/core/src/Reactor.js
@@ -881,13 +881,15 @@ export default class Reactor {
     const prevData = this._dataForQueryCache[hash]?.data;
     const data = this.dataForQuery(hash)?.value;
     if (!data) return;
-    if (areObjectsDeepEqual(data, prevData)) return;
+    const haveChanged = !areObjectsDeepEqual(data, prevData);
     rs.forEach((r) => {
       if (r.once) {
         this._unsubQuery(r.q, hash, r.cb);
       }
 
-      r.cb(data);
+      if (haveChanged || r.once) {
+        r.cb(data);
+      }
     });
   };
 
--- a/client/packages/core/src/Reactor.js
+++ b/client/packages/core/src/Reactor.js
@@ -648,10 +648,7 @@ export default class Reactor {
       (r) => r.dfd !== dfd,
     );
 
-    if (!this.queryOnceDfds[hash].length) {
-      delete this.queryOnceDfds[hash];
-      this._trySendAuthed(uuid(), { op: "remove-query", q });
-    }
+    this._cleanupQuery(q, hash);
   }
 
   _unsubQuery(q, hash, cb) {
@@ -659,10 +656,19 @@ export default class Reactor {
 
     this.queryCbs[hash] = this.queryCbs[hash].filter((r) => r.cb !== cb);
 
-    if (!this.queryCbs[hash].length) {
-      delete this.queryCbs[hash];
-      this._trySendAuthed(uuid(), { op: "remove-query", q });
-    }
+    this._cleanupQuery(q, hash);
+  }
+
+  _cleanupQuery(q, hash) {
+    const hasListeners =
+      this.queryCbs[hash].length && this.queryOnceDfds[hash].length;
+
+    if (hasListeners) return;
+
+    delete this.queryCbs[hash];
+    delete this.queryOnceDfds[hash];
+
+    this._trySendAuthed(uuid(), { op: "remove-query", q });
   }
 
   // When we `pushTx`, it's possible that we don't yet have `this.attrs`
--- a/client/packages/core/src/Reactor.js
+++ b/client/packages/core/src/Reactor.js
@@ -386,6 +386,9 @@ export default class Reactor {
           this._tryJoinRoom(roomId);
         }
         break;
+      case "add-query-exists":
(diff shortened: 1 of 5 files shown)


nezaj:
@markyfyi I'm not quite sure if this works as expected. I put up an updated example to show how someone may use `queryOnce` for things like checking unique-ness on the client, but it doesn't seem to work as expected.

stopachka (commented):
Awesome! 

I think one piece information that is missing, which we may need to add to make all this work is the following: 

Which `results` in `querySubs` is actually 'from' the server on this ws? 

i.e 

It's possible to get a result back from IndexedDB. But what we need to make sure of, is that the `add-query` for this query has run, and returned `add-query-ok`. Once we have that, we know that we have the latest result, and we are subscribed.

markyfyi:
@nezaj @stopachka PTALA

After @stopachka's comment, I took another pass at attempting to read from the cache.  I discovered `dataForQuery`, and if IIUC, it does everything we need to make sure we're working with fresh data.

My only concern now is race conditions, i.e. is it possible that the cache might be stale?  For instance, right after `init-ok`.  Or does `pendingMutationsVersion` take care of that?  Might be worth a 5mins deep dive.

client/packages/core/src/Reactor.js
```diff
@@ -729,7 +855,7 @@ export default class Reactor {
       querySubVersion === cached.querySubVersion &&
       pendingMutationsVersion === cached.pendingMutationsVersion
     ) {
-      return cached.data;
+      return { fresh: true, value: cached.data };
```
stopachka: How do you know at this point that the query is fresh?

stopachka (commented):
I think we are getting closer, but it's not 100% clear to me how `fresh` works. I also think we may want to remove the `once` call, and manually do the unsubscribe. Afaik the checks for `once` are complicating the logic here.

client/packages/core/src/Reactor.js
```diff
@@ -583,25 +593,81 @@ export default class Reactor {
...
+    if (!this._isOnline) {
+      throw new Error(
+        "Offline: Cannot execute query because the device is offline.",
+      );
+    }
+
+    const result = await new Promise((resolve, reject) => {
+      let done = false;
+
+      const unsub = this._subscribeQuery(
```
stopachka: is there a way we can remove the `once` check? 

Maybe we can follow a different codepath from `_subscrubeQuery` or something. 

I would really love to avoid it if possible

stopachka (commented):
Soo close!

client/packages/core/src/index.ts
```diff
@@ -385,6 +385,32 @@ class InstantCore<
     delete globalInstantCoreStore[this._reactor.config.appId];
     this._reactor.shutdown();
   }
+
+  /**
+   * Use this for one-off queries.
+   * Returens local data if available, otherwise fetches from the server.
```
orenaksakal:
```suggestion
   * Returns local data if available, otherwise fetches from the server.
```

client/packages/core/src/Reactor.js
```diff
@@ -583,25 +601,70 @@ export default class Reactor {
...
+  }
+
+  _completeQueryOnce(q, hash, dfd) {
+    if (!this.queryOnceDfds[hash]) return;
+
+    this.queryOnceDfds[hash] = this.queryOnceDfds[hash].filter(
+      (r) => r.dfd !== dfd,
+    );
+
+    if (!this.queryOnceDfds[hash].length) {
```
stopachka: small nit: would move this logic into another helper fn.

stopachka (approved):
Woohoo!!

client/packages/core/src/Reactor.js
```diff
@@ -24,6 +24,8 @@ const STATUS = {
   ERRORED: "errored",
 };
 
+const QUERY_ONCE_TIMEOUT = 3000;
```
nezaj: Nit. Would make this `5000` to match `transact` timeout

client/packages/react/src/InstantReact.ts
```diff
@@ -463,4 +465,30 @@ export abstract class InstantReact<
     );
     return state;
   };
+
+  /**
+   * Use this for one-off queries.
+   * Returens local data if available, otherwise fetches from the server.
```
kenrick95:
```suggestion
   * Returns local data if available, otherwise fetches from the server.
```

same as the suggestion above

client/packages/core/src/Reactor.js
markyfyi: Yeah, this was the issue I mentioned last evening.  We need to check both `queryOnceDfds` and `queryCbs` before sending a `remove-query`

PR #286, opened 2024-09-27
docs(schema-as-code): fixed syntax error in default export of i.graph

KunalSin9h:
Either do 

```ts
export default i.graph();

# or

const g = i.graph():
export default g;
```

image


stopachka (approved):
Thank you @KunalSin9h !

PR #288, opened 2024-09-27
[Dash] Add clear app to admin

nezaj:
[One of our users](https://discord.com/channels/1031957483243188235/1148286343638687786/1288277649805938729) asked if we could make it easy to clear apps. This is especially useful if someone is prototyping on a paid app, since it's not easy for them to re-create an app from scratch.

One thing that was strange was that the explorer would be stale after clearing the data. I get around this by doing a reload, but I'd be open to a better solution, just wasn't sure how.

It might also be nice to offer some more checkboxes / configuration for what things you want to clear, but I figured this is decent solution for now and if folks need something more customizable they will let us know.

I also held off on adding this to the cli-helper, just wanted to make sure this actually meets his needs for now before increasing the surface area of other packages

https://github.com/user-attachments/assets/b2982671-b975-4035-8ccf-e6d7321a06b5

### DIFF
--- a/client/www/pages/dash/index.tsx
+++ b/client/www/pages/dash/index.tsx
@@ -1023,7 +1023,9 @@ function Admin({
 }) {
   const token = useContext(TokenContext);
   const [deleteAppOk, updateDeleteAppOk] = useState(false);
+  const [clearAppOk, updateClearAppOk] = useState(false);
   const [editMember, setEditMember] = useState<InstantMember | null>();
+  const clearDialog = useDialog();
   const deleteDialog = useDialog();
   const inviteDialog = useDialog();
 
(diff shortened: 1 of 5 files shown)


client/www/pages/dash/index.tsx
```diff
@@ -1288,21 +1290,56 @@ function Admin({
...
+                  await jsonFetch(`${config.apiURI}/dash/apps/${app.id}/clear`, {
+                    method: 'POST',
+                    headers: {
+                      authorization: `Bearer ${token}`,
+                      'content-type': 'application/json',
+                    },
+                  });
+
+                  clearDialog.onClose();
+                  window.location.reload();
```
dwwoelfel: Do we need the reload? If you want to make sure the rules change, you can do:

```
await dashResponse.mutate();
```

Also might be nice to add a success toast here:

```
successToast('App cleared!');
```

dwwoelfel (approved):
One comment, but LGTM!

client/www/pages/dash/index.tsx
nezaj: Thanks @dwwoelfel ! `dashResponse.mutate()` was what I was looking for!

server/src/instant/model/app.clj
```diff
@@ -273,6 +275,17 @@
                             WHERE a.id = ?::uuid"
                            new-creator-id id])))
 
+(defn clear-by-id!
+  "Deletes attrs, rules, and triples for the specified app_id"
+  ([params] (clear-by-id! aurora/conn-pool params))
+  ([conn {:keys [id]}]
+   (next-jdbc/with-transaction [tx-conn conn]
+     (attr-model/delete-by-app-id! tx-conn id)
+     (rule-model/delete-by-app-id! tx-conn {:app-id id}))))
```
dwwoelfel: To get this to invalidate queries, you probably need to add:

```
(transaction-model/create! conn {:app-id id})
```

inside the transaction.

PR #289, opened 2024-09-27
[Dev tool] Add clear app to devtool

nezaj:
[Cam](https://discord.com/channels/1031957483243188235/1288277649805938729/1289306430880350240) mentioned that having clear app in the devtool would be the most useful place. So this PR stubs an Admin section in the devtool

Not all the Admin actions in the dashboard makes sense here, for example, you probably don't want to delete the app in the devtool.

It would probably be better to have the Admin page more configurable so you could select which pieces you want. That feels like a bigger lift right now, and if more admin features are requested in the dev tool, I think we can take on a more elegant solution then.

**Authorized View**

image

**Insufficent Role View**

image


dwwoelfel (approved):
LGTM!

PR #290, opened 2024-09-27
Only evaluate span-opts once

dwwoelfel:
Each time we do `~span-opts`, it will evaluate the form. This change ensures we only evaluate the form once per call to `with-span!`.


nezaj (approved):
Nice!

PR #291, opened 2024-09-28
Add test that the invalidator is working

dwwoelfel:
Adds an integration test to make sure that the invalidator picks up changes to the triples table.

Inspired by this small bug that I luckily caught before deploying the users table pr: https://github.com/instantdb/instant/pull/280/commits/a27e2497c6c2345053910d9f3b37d331056a0d1a. It would have broken invalidation.

Makes a couple of small changes to the invalidator:

1. You can start and stop it without setting global state. There's a `start-global` and `stop-global` we now use in `instant.core` that keeps the previous behavior.
2. The `kick-wal` helper writes to the `config` table, which is more effective than `pg_notify`.

### DIFF
--- a/.github/workflows/clojure.yml
+++ b/.github/workflows/clojure.yml
@@ -48,11 +48,7 @@ jobs:
           POSTGRES_USER: test_user
           POSTGRES_PASSWORD: test_password
           POSTGRES_DB: test_db
-        options: >-
-          -c wal_level=logical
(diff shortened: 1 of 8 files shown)


dwwoelfel:
Not quite ready for review--still trying to figure out how to get postgres in CI to support logical replication

stopachka (approved):
Great idea @dwwoelfel !

PR #292, opened 2024-09-29
Fix view checks, make them warn-only for now

dwwoelfel:
Fix for view checks that were causing problems for some users.

The problem happened when an object was created and linked in the same transaction. We would try to run the `view` check on an empty object, which would fail.

Now we'll only run the `view` check if the object already exists, counting on the `create` check to reject the transaction if the user isn't allowed to create it.

To prevent more problems for users, we just log the results of the `view` check for now. Once we have some confidence that everything looks ok, we can turn it on.

### DIFF
--- a/server/README.md
+++ b/server/README.md
@@ -60,4 +60,4 @@ If you want to make any changes to your configuration, update the `resources/con
 
 # Questions?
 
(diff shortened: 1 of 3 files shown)


stopachka (approved):
Niice!

PR #296, opened 2024-09-30
Improved error messaging for conflicting link defs in schema.ts

markyfyi:
The schema planner doesn't account for a sneaky invariant: what if the user has a link def, but in the _opposite direction_ of one in the database?  As in, the forward and reverse identities are swapped.  Right now, we catch this, but throw a misleading "check your schema file for dupes" error.  In this change, I add some code to check for this case, and throw a (hopefully) more informative error.

To repro - go to `sandbox/cli-nodejs/instant.schema.ts` and swap the `forward` and `reverse` props of one of the links.

### DIFF
--- a/server/src/instant/model/schema.clj
+++ b/server/src/instant/model/schema.clj
@@ -8,8 +8,6 @@
             [instant.util.exception :as ex])
   (:import (java.util UUID)))
 
-
-
 (defn map-map [f m]
   (into {} (map (fn [[k v]] [k (f [k v])]) m)))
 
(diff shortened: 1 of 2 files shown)


nezaj (approved):
LGTM w/ one nit.

I like the informative error, Stripe goes pretty with some of their error messages and I think it's a really nice user experience

server/src/instant/model/schema.clj
```diff
@@ -8,6 +8,8 @@
             [instant.util.exception :as ex])
   (:import (java.util UUID)))
 
+
```
nezaj: Nit. extra white-space

server/src/instant/model/schema.clj
```diff
@@ -149,21 +151,39 @@
...
-                               (when (= op :add-attr)
-                                 (attr-ident-names data)))))
-        dups (->> (concat current-ident-names ident-names)
+  (let [current-fwd-ident-names (->> current-attrs
+                                     (map attr-model/fwd-ident-name)
+                                     (map vec))
+        current-all-ident-names (->> current-attrs
+                                     (mapcat attr-ident-names)
+                                     (map vec))
+        steps-ident-names (->>
```
nezaj: Nit. Might be nice to call this `new-ident-names`

stopachka (approved):
LGTM!

PR #297, opened 2024-09-30
Instant Entity helper type

markyfyi:
Users wanted a way to quickly pluck a single entity's type from their schema, so I made `InstantEntity`. 💥

```ts
type Checkin = InstantEntity<
  DB,
  "checkins"
>;
```

But it's easy to imagine many cases where you want the entity with some links, so `InstantEntity` accepts an optional 3rd param specifying a "subslice" of InstaQL links relative to the entity. 🎉 ✨ 

```ts
type Checkin = InstantEntity<
  DB,
  "checkins",
  {
    habit: {
      category: {};
    };
  }
>;
```

### DIFF
--- a/client/packages/admin/src/index.ts
+++ b/client/packages/admin/src/index.ts
@@ -20,6 +20,7 @@ import {
   type InstantQueryResult,
   type InstantSchema,
   type InstantObject,
+  type InstantEntity,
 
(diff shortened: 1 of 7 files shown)


stopachka (approved):
LGTM!

PR #298, opened 2024-09-30
[Bump] 0.14.8

nezaj:
Looks like my `node_modules` was borked which caused some type issues. What was strange was that `instant-cli` got published but the other packages did not. It would be nicer if either everything got published or nothing did. Something to look into.

I got `pnpm run build` to work by re-creating `node_modules`

Bumping up everything to `0.14.8` so all our package numbers stay consistent


stopachka (approved):
LGTM!

PR #299, opened 2024-10-01
instant db from schema hepler type

markyfyi:
Exports a helper for users who require a DB type prior to initialization.  For example, when using a DI framework that defers initialization.

Context: https://discord.com/channels/1031957483243188235/1289204418515832903


stopachka (approved):
SGTM

PR #301, opened 2024-10-01
Enable view checks behind a feature flag

dwwoelfel:
Enables view checks in permissioned-transaction behind a feature flag.

Everything is looking good in the logs so far, if something goes wrong we can disable it for individual apps while we investigate.


stopachka (approved):
lesgoo!

PR #306, opened 2024-10-02
Admin "gen magic code" method

markyfyi:
Users who owned their auth email stack were asking for a magic code helper to integrate Instant into their emails.

### DIFF
--- a/server/src/instant/admin/routes.clj
+++ b/server/src/instant/admin/routes.clj
@@ -8,7 +8,9 @@
             [instant.util.uuid :as uuid-util]
             [instant.db.instaql :as iq]
             [instant.util.email :as email]
+            [instant.model.instant-user :as instant-user-model]
             [instant.model.app-user :as app-user-model]
+            [instant.model.instant-user-magic-code :as  instant-user-magic-code-model]
             [instant.model.app-user-refresh-token :as app-user-refresh-token-model]
             [instant.db.datalog :as d]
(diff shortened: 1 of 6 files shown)


server/src/instant/model/instant_user.clj
```diff
@@ -96,6 +96,10 @@
    (sql/execute-one! conn
                      ["DELETE FROM instant_users WHERE email = ?" email])))
 
+(defn get-or-create-by-email! [{:keys [email]}]
+  (or  (get-by-email {:email email})
```
stopachka: nit: space

server/src/instant/admin/routes.clj
```diff
@@ -263,6 +265,18 @@
   (let [{user-id :id app-id :app_id} (req->app-user! req)]
     (response/ok {:deleted (app-user-model/delete-by-id! {:id user-id :app-id app-id})})))
 
+(defn magic-code-post [req]
+  (let [email (ex/get-param! req [:body :email] email/coerce)
+        {user-id :id} (instant-user-model/get-or-create-by-email! {:email email})
```
stopachka: Afaik this is for the wrong model. This should be the `app-user-model` no?

stopachka (commented):
Change to app-user-model and i.m.o G2G!

server/src/instant/admin/routes.clj
markyfyi: :o

PR #311, opened 2024-10-02
Allow deleting refs from the `users` schema form

dwwoelfel:
Allow users to delete backward links from the users edit namespace form.


nezaj (approved):
Nice!

PR #313, opened 2024-10-02
Fix over-eager queryOnce case

markyfyi:
@stopachka pointed out that `queryOnce` calls that fired before `init-ok` were timing out.  This is because we silently failed sending them over WebSocket (due to the connection not being in an authenticated state) and never tried again once a connection was established.  This change enhances `_flushPendingMessages` to send events for every pending `queryOnce` like we do with subs.

### DIFF
--- a/client/packages/core/src/Reactor.js
+++ b/client/packages/core/src/Reactor.js
@@ -24,7 +24,7 @@ const STATUS = {
   ERRORED: "errored",
 };
 
-const QUERY_ONCE_TIMEOUT = 5000;
+const QUERY_ONCE_TIMEOUT = 10_000;
 
 const WS_OPEN_STATUS = 1;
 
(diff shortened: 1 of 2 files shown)


client/packages/core/src/Reactor.js
```diff
@@ -969,6 +969,13 @@ export default class Reactor {
     safeSubs.forEach(({ eventId, q }) => {
       this._trySendAuthed(eventId, { op: "add-query", q });
     });
+
+    Object.values(this.queryOnceDfds)
+      .flat()
+      .forEach(({ eventId, q }) => {
+        this._trySendAuthed(eventId, { op: "add-query", q });
```
stopachka, line 976: Afaik there is now one weird invariant, where we could send an add-query twice. i.e `q` exists both in queryOnceDfds _and_ safeSubs. 

Is there a way we can collect both `subs` and `queryOnceDfds`, and return a single result of `qs`?

--- 

(Alternatively we can leave this as is, but it may lead to add-query-ok returning twice, or something like that)

stopachka (approved):
Lesgoo!

PR #315, opened 2024-10-03
[Docs] Add more context to query once

nezaj:
I liked that we included `pageInfo` but I thought it would be nicer to be more explicit that you would use `pageInfo` in a paginated case. The way the docs were written prior to this change may have given the impression that `pageInfo` is necessary for `queryOnce`


stopachka (approved):
Nice!

PR #316, opened 2024-10-03
fix: breaking typos in tutorials

orenaksakal:
Just noticed tutorial's delete cases are broken because of typo `tx.message` instead of `tx.messages`

reproduction video and details: https://discord.com/channels/1031957483243188235/1148284450992574535/1291236006921637930


stopachka (approved):
Good catch. Thank you @orenaksakal

Issue #287, opened 2024-09-27
Fetch Failed: too much in CLI


KunalSin9h:
Whenever i use cli, there is 70% chance it give "Fetch Failed error". My internet is relatively good for a CLI to work!.

![image](https://github.com/user-attachments/assets/d9148898-a547-4a64-b9b7-e6a1ea31d5c6)

KunalSin9h:
full error

```ocaml
Planning...
node:internal/deps/undici/undici:13178
      Error.captureStackTrace(err);
            ^

TypeError: fetch failed
    at node:internal/deps/undici/undici:13178:13
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async fetchJson (file:///usr/local/lib/node_modules/instant-cli/index.js:531:15)
    at async pushSchema (file:///usr/local/lib/node_modules/instant-cli/index.js:378:19) {
  [cause]: AggregateError [ETIMEDOUT]: 
      at internalConnectMultiple (node:net:1118:18)
      at internalConnectMultiple (node:net:1186:5)
      at Timeout.internalConnectMultipleTimeout (node:net:1712:5)
      at listOnTimeout (node:internal/timers:583:11)
      at process.processTimers (node:internal/timers:519:7) {
    code: 'ETIMEDOUT',
    [errors]: [
      Error: connect ETIMEDOUT 172.67.149.9:443
          at createConnectionError (node:net:1648:14)
          at Timeout.internalConnectMultipleTimeout (node:net:1707:38)
          at listOnTimeout (node:internal/timers:583:11)
          at process.processTimers (node:internal/timers:519:7) {
        errno: -110,
        code: 'ETIMEDOUT',
        syscall: 'connect',
        address: '172.67.149.9',
        port: 443
      },
      Error: connect ENETUNREACH 2606:4700:3035::ac43:9509:443 - Local (:::0)
          at internalConnectMultiple (node:net:1182:16)
          at Timeout.internalConnectMultipleTimeout (node:net:1712:5)
          at listOnTimeout (node:internal/timers:583:11)
          at process.processTimers (node:internal/timers:519:7) {
        errno: -101,
        code: 'ENETUNREACH',
        syscall: 'connect',
        address: '2606:4700:3035::ac43:9509',
        port: 443
      },
      Error: connect ETIMEDOUT 104.21.29.133:443
          at createConnectionError (node:net:1648:14)
          at Timeout.internalConnectMultipleTimeout (node:net:1707:38)
[...]

Issue #138, opened 2024-09-05
Update failing with timeout


nikshepsvn:
![image](https://github.com/user-attachments/assets/bf85cfd6-3e9d-4f60-bca7-a35bb0e024f3)
as seen here, I can't seem to update my data because of this, is it because the object is too big? What is going on and how can I get around this?

Himasnhu-AT:
Try the solution specified in the above PR by increasing the timeout duration.

Here's a sample code from the PR to implement the timeout:

```js
Copy code
const controller = new AbortController();
const id = setTimeout(() => controller.abort(), timeout); // Set the timeout here

try {
  const res = await fetch(`${instantBackendOrigin}${path}`, {
    method: method ?? "GET",
    headers: {
      ...(withAuth ? { Authorization: `Bearer ${authToken}` } : {}),
      "Content-Type": "application/json",
    },
    body: body ? JSON.stringify(body) : undefined,
    signal: controller.signal, // Passing the signal to abort the fetch if timeout is reached
  });

  clearTimeout(id); // Clear the timeout to free up resources
} catch (error) {
  // Handle errors, including the timeout abort error
}
```

PR #318, opened 2024-10-03
Refactor fetchJson to handle custom timeout

Himasnhu-AT:
Found this, apparently the issue isn't closed https://github.com/nodejs/undici/issues/1373

In Chromium browser, the default timeout is somewhere around 300 second
whereas in node, it is around 30 sec. 

possible solution: update the timeouts to prevent this issue

Fixes ISSUES: #287 

> [!IMPORTANT]
> Similar approach to be taken for browser, to resolve the issue: #138
> Possible solution: Have npm package, which lets user define their own timeout, or maybe pass args to set timeout in backend

### DIFF
--- a/client/packages/cli/index.js
+++ b/client/packages/cli/index.js
@@ -3,7 +3,7 @@
 import { readFileSync } from "fs";
 import { mkdir, writeFile, readFile, stat } from "fs/promises";
 import { dirname, join } from "path";
-import { fileURLToPath } from 'url';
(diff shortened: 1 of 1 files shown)


nezaj:
Thanks for this PR! Great ideas. We decided instead of having an additional parameter, we'll increase the cap to 5 minutes to match Chrome. https://github.com/instantdb/instant/pull/345

PR #320, opened 2024-10-03
Schema validations rework

markyfyi:
Taking a second pass at schema validation.  I wanted something more explicit and readable.  Covers a handful of new cases, and fixes a false-positive for backward links.


server/src/instant/model/schema.clj
```diff
@@ -141,50 +141,171 @@
...
-                      errors)))
+  (let [current-link-attrs
+        (filter (comp #{:ref} :value-type) current-attrs)
+
+        current-blobs
+        (filter (comp #{:blob} :value-type) current-attrs)
+
+        current-blob-idents
+        (->> current-blobs
+             (map #(attr-model/fwd-ident-name %))
```
stopachka: nit: you can write (map attr-model/fwd-ident-name)

server/src/instant/model/schema.clj
```diff
@@ -141,50 +141,171 @@
...
+             (into #{}))
+
+        current-links-mapping-fwd
+        (->> current-link-attrs
+             (map #(vector
```
stopachka: If you want to get a bit clojure fancy, I _think_ you can write this as: 

(map (juxt attr-model/fwd-ident-name attr-model/rev-ident-name

server/src/instant/model/schema.clj
```diff
@@ -141,50 +141,171 @@
...
+                    (attr-model/rev-ident-name %)
+                    (attr-model/fwd-ident-name %)))
+             (into {}))
+
+        errors
+        (->>
+         steps
+         (filter (comp #{:add-attr} first))
+         (map second)
+         (map #(let [fwd-name
```
stopachka, line 188: nit: would extract this into a helper function. 

Reason: later on when debugging, it can be very helpful to add a tool/def-locals here, but right now it will be hard because of the %.

stopachka (approved):
Beautiful!

server/src/instant/model/schema.clj
markyfyi: Should I do a helper, or just change the `#` shorthand to `fn`?

stopachka: I would probably do a helper, as it's pretty big

PR #322, opened 2024-10-04
Auth users ref

dwwoelfel:
Adds support for `auth.ref` for querying off of the `$users` namespace in permissions rules.

This will allow for more flexibility in permission rules and make it easier to implement role-based permissions.

If you had a `capabilities` namespace with an `action` attr that linked to `$users`, you would protect the `tasks` ns with something like:

```json
"tasks": {
  "allow": {
    "view": "'view_tasks' in auth.ref('$user.capabilties.action')"
  }
}
```

Also creates a new `CelMap` `deftype` so that we can store metadata like `ctx` and `etype` outside of the data in the map. `->cel-map` takes two arguments, the metadata and the object, then automatically cycles the object through ->json and <-json.

### DIFF
--- a/.github/workflows/clojure.yml
+++ b/.github/workflows/clojure.yml
@@ -17,7 +17,7 @@ jobs:
     - name: Get changed files
       id: changed-files
       run: |
-        if ${{ github.event_name == 'pull_request' }}; then
+        if ${{ github.event_name == 'pull_request' || github.event.before == '0000000000000000000000000000000000000000' }}; then
             echo "changed_files=$(git diff --name-only -r HEAD^1 HEAD | xargs)" >> $GITHUB_OUTPUT
         else
             echo "changed_files=$(git diff --name-only ${{ github.event.before }} ${{ github.event.after }} | xargs)" >> $GITHUB_OUTPUT
--- a/server/src/instant/db/cel.clj
+++ b/server/src/instant/db/cel.clj
@@ -339,9 +339,16 @@
 ;; but this will do for now.
 (defn collect-ref-uses
   "Returns a set of `path-str` used in `data.ref` calls in the given cel ast,
-   grouped by the object, e.g. {\"data\": #{\"a.b\"}, \"auth\" #{\"c.d\"}}."
+   grouped by the object, e.g. {\"data\": #{\"a.b\"}, \"auth\" #{\"c.d\"}}.
+   Automatically strips `$user` from auth path-str"
   [^CelAbstractSyntaxTree ast]
-  (expr->ref-uses (.getExpr ast)))
+  (-> (expr->ref-uses (.getExpr ast))
+      (update "auth" (fn [x]
+                       (set (map (fn [path-str]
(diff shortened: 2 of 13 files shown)


server/src/instant/db/cel.clj
```diff
@@ -73,9 +76,9 @@
 (defn- find-val-path [query]
   (first
    (for [[i pattern] (map-indexed vector query)
-         :let [pos (.indexOf pattern value-sym)]
-         :when (not= pos -1)]
-        ;; `dec` because these patterns include the `index`, but join-rows don't.
+         :let [pos (ucoll/index-of value-sym pattern)]
```
dwwoelfel, line 79: Gets rid of a reflection warning

.github/workflows/clojure.yml
```diff
@@ -17,7 +17,7 @@ jobs:
     - name: Get changed files
       id: changed-files
       run: |
-        if ${{ github.event_name == 'pull_request' }}; then
+        if ${{ github.event_name == 'pull_request' || github.event.before == '0000000000000000000000000000000000000000' }}; then
```
dwwoelfel, line 20: Now we'll run the tests on the first push to a new branch.

server/src/instant/db/cel.clj
```diff
@@ -132,33 +135,79 @@
 ;; ----
 ;; Cel
 
+(defprotocol CelMapExtension
+  (getMeta [this]))
+
+(deftype CelMap [metadata m]
```
stopachka, line 141: biyutiful!

server/src/instant/db/cel.clj
```diff
@@ -132,33 +135,79 @@
...
+
+  (entrySet [_]
+    (set (seq (or m {}))))
+
+  CelMapExtension
+  (getMeta [_]
+    metadata))
+
+(defn ->cel-map [metadata m]
+  (CelMap. metadata (<-json (->json m))))
```
stopachka, line 172: Nice that the <-json ->json is in one place now. Eventually we can get it rid of it and make the `get` in CelMap smarter. 

Feeling our cel code getting beetter!

server/src/instant/db/cel.clj
```diff
@@ -366,11 +399,57 @@
                  patterns
                  results))))
 
+(defn make-validator ^CelAstValidator [attrs]
+  (reify CelAstValidator
```
stopachka, line 403: so cool!

stopachka (approved):
Beautiful work Daniel

.github/workflows/clojure.yml
nezaj: By run the tests on the first push to a new branch -- does this mean tests won't re-run if we push up a new commit?

dwwoelfel: It will still run on pushes that aren't the first push. This just fixes a bug where you either had to open a PR or do another push to the branch to get the clojure tests to run.

PR #323, opened 2024-10-04
Handle simultaneous add-attr

dwwoelfel:
Handles simultaneous add-attr transactions that create the same attributes.

If we have multiple database transactions that are adding the same attrs, one of the transactions can fail with "Record not unique ident".

With this change, we'll do an upsert on the `id` and only throw if one of the fields on the ident or attr is different.

### DIFF
--- a/server/src/instant/db/model/attr.clj
+++ b/server/src/instant/db/model/attr.clj
@@ -292,9 +292,9 @@
                                                    [:raise_exception_message
                                                     [:||
                                                      "The attribute with id "
-                                                     [:cast [:to_json :EXCLUDED] :text]
+                                                     [:cast :EXCLUDED.id :text]
                                                      " conflicts with an existing attribute with id "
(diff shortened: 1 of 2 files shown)


server/src/instant/db/model/attr.clj
```diff
@@ -230,6 +230,25 @@
...
+                                                     [:=
+                                                      (qualify-col :idents col)
+                                                      (qualify-col :EXCLUDED col)])
+                                                   ident-table-cols))
+                                       :EXCLUDED.etype
+                                       ;; raise_exception_message is typed to return
+                                       ;; a boolean, so we cast it to text so that it
+                                       ;; can throw its exception
+                                       :else [:cast
+                                              [:raise_exception_message
```
stopachka, line 249: TIL `:||`. 

_Makes martial artist bow, in respect to your SQL skills_

stopachka (approved):
Heeck yeah!

PR #324, opened 2024-10-04
Keep track of how many connections we have

dwwoelfel:
Sends metrics with gauges about how many connections and requests are open on the server.

### DIFF
--- a/server/src/instant/core.clj
+++ b/server/src/instant/core.clj
@@ -38,6 +38,7 @@
    [ring.middleware.multipart-params :refer [wrap-multipart-params]]
    [ring.middleware.params :refer [wrap-params]])
   (:import
+   (io.undertow Undertow UndertowOptions Undertow$Builder Undertow$ListenerInfo)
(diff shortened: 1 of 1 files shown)


server/src/instant/core.clj
```diff
@@ -97,13 +98,36 @@
...
+  (def server ^Undertow (undertow-adapter/run-undertow
+                         (handler)
+                         {:host "0.0.0.0"
+                          :port (config/get-server-port)
+                          :configurator (fn [^Undertow$Builder builder]
+                                          (.setServerOption builder UndertowOptions/ENABLE_STATISTICS true ))}))
+  (def stop-gauge (gauges/add-gauge-metrics-fn
+                   (fn []
+                     (let [^Undertow server server
+                           ^Undertow$ListenerInfo listener (some-> server
```
stopachka, line 110: oo heck yeah!

stopachka (approved):
Niice!

PR #329, opened 2024-10-08
[small] allow promo codes for some emails

stopachka:
We can use this to help pro customers enable some other apps. 

@nezaj @dwwoelfel


nezaj (approved):
Nice!

PR #330, opened 2024-10-08
[CLI] Add .env file if it doesn't exist

nezaj:
I notice when starting a new project it may be easy to `npx instant-cli pull <APP_ID>`, edit some schema, and then be surprised that `npx instant-cli push` doesn't work, This is because you also need to create a `.env` file for `push` to know the app id, which I kind of expected the cli would handle for me if I already pulled earlier.

Figured we can just prompt the user if they want to create a `.env` file


stopachka (approved):
Lesgoo!

PR #333, opened 2024-10-09
Hazelcast for ephemeral state

dwwoelfel:
This stores our ephemeral state in Hazelcast, which will allow us to run multiple instances. It will also make the rooms transition smoother on deploys because we won't have a split state during the deploy window.

### Latency
I tested latency by temporarily adding `publishedAt: new Date()` when we set presence in the cursors example and calculating the difference between now and publishedAt in the sibling iframes.

The latency for both was ~85ms, surprisingly there was no difference between hazelcast and non-hazelcast.

### One map vs multiple maps
I went with one map per room rather than a large map with all of the rooms because it is a lot easier to manage making updates to the map. However, a single large map would support eviction policies that would make it easier to prevent a stale room from living in memory forever. I’ll re-evaluate if we end up having a lot of trouble with rooms not getting cleaned up.

### Potential issues
With Hazelcast, the ephemeral state will persist across deploys unless all of the machines are down. If we lose a machine and it doesn’t clean up its sessions, that data will be stuck forever. To head off this problem, I switched session-ids to use `squuids`, which contain a timestamp. This will give us a way to clean out old data, since we know that a squuid older than the oldest instance must be stale. There’s a function that we can use to manually clean out stale data, but we can automate it if it becomes an issue.

### Rollout
For rollout, I created a new hazelcast config flag. It has a circuit breaker we can pull to shut off hazelcast completely if we run into unexpected issues. You can also turn hazelcast on/off for an individual app. In order to make the transition seamless, we update both hazelcast and the state atom in the background. If hazelcast is on, then it will use a map listener to distribute updates. If it's off, then it will use the existing watcher.

When this deploys, we'll start updating hazelcast in the background on a single machine. If everything looks normal, then I'll turn on Hazelcast for the home page. If that looks good, I'll turn it on for a few apps, then turn it on by default.

After we're comfortable with it, we can increase the number of instances.

I haven't had any problems with it in staging so far, so I'm not expecting too many issues in production, but this will be the first time we'll see how it performs under our production load.

### Other changes

- You can specify a port in the url on www and the config will pick that up if dev mode is enabled—useful for testing hazelcast with multiple instant-server instances locally. It’s what the new `/play/hazelcast` sandbox uses.
- We run a transform on the results from the flags query each time the query result is updated. That lets us more efficiently check flag values.
- session ids use `squuids` so that we can track creation time and clean up stale data

### Testing locally
Start a new dev server with `PORT=8889 NRPEL_PORT=6006 make dev`. Visit [http://localhost:4000/play/hazelcast](http://localhost:4000/play/hazelcast). In your config app, set `default-value` to `true` for the `hazelcast` flag and you should see all panes update when you move your mouse over one of them.


server/docker-compose.yml
```diff
@@ -4,15 +4,12 @@ services:
...
+      - BEANSTALK_PORT=80
       - PRODUCTION=true
       - HONEYCOMB_ENDPOINT=http://refinery:8082
     env_file:
       - .env
-    ports:
-      - "80:5000"
-      - "6005:6005"
-      - "10001:10001"
+    network_mode: "host"
```
dwwoelfel, line 12: With host mode we don't need to specify a port mapping because it's no longer on a separate network.

client/sandbox/react-nextjs/pages/play/hazelcast.tsx
```diff
@@ -0,0 +1,75 @@
...
+function nextId() {
+  return ++__id;
+}
+
+function Page() {
+  const [pages, setPages] = useState([
+    { port: 8888, id: nextId() },
+    { port: 8888, id: nextId() },
+    { port: 8888, id: nextId() },
+    { port: 8889, id: nextId() },
```
stopachka, line 15: awwh yeah!!

server/docker-compose.yml
dwwoelfel: We need this because Hazelcast gets confused and tries to send messages to itself without it.

server/src/instant/core.clj
```diff
@@ -125,9 +125,9 @@
                            :value (.getMaxProcessingTime stats)}]))))))
 
 (defn stop []
-  (when (bound? (resolve 'server))
+  (when (bound? #'server)
```
dwwoelfel, line 128: This was just broken before. Tested and this works.

server/src/instant/reactive/ephemeral.clj
```diff
@@ -153,13 +407,31 @@
 
 (declare ephemeral-store-atom)
 
+(defn start-hz []
+  (let [result (deref (future @hz) (* 60 1000) :timeout)]
```
dwwoelfel, line 411: Once hazelcast is the default, we'll have to remove this timeout.

server/src/instant/reactive/ephemeral.clj
```diff
@@ -65,6 +121,158 @@
...
+                        (a/<!! complete-ch)))
+                    (recur (a/<!! ch)))))
+
+    (let [listener-id (.addEntryListener m listener true)]
+      (fn []
+        (.removeEntryListener m listener-id)
+        (a/close! ch)))))
+
+(defn register-room-map [^IMap m app-id room-id sess-id]
+  (let [map-name (.getName m)]
```
stopachka, line 205: would check about diff string vs

server/src/instant/util/aws.clj
```diff
@@ -0,0 +1,48 @@
+(ns instant.util.aws
+  (:require [clj-http.client :as clj-http]
+            [amazonica.aws.ec2 :as ec2]))
+
+(def environment-tag-name "elasticbeanstalk:environment-name")
+
+(defn get-tag [tag-name]
+  (let [token (-> (clj-http/put
+                   "http://169.254.169.254/latest/api/token"
```
stopachka, line 14: curious q: where does this ip come from?

stopachka (approved):
Biyutiful!

server/src/instant/util/aws.clj
dwwoelfel: This is the ip that aws (and the rest of the cloud services) uses for their metadata API: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html

I'll add a comment.

PR #334, opened 2024-10-09
Fix roomId

dwwoelfel:
I think this was just a copy/paste typo because `roomId` isn't defined at this part of the file. I think the reason we didn't see an error is because of the weird way switches bleed into eachother in js.


stopachka (approved):
thank you @dwwoelfel !

PR #335, opened 2024-10-09
Make cursors work with touch on mobile

dwwoelfel:
Adds ontouch handlers to `<Cursors>` so that they work on mobile.

When people look at the examples page on their phones, the cursors examples don't do anything.

Also tells the browser to load the iframes lazily, which should help shorten the delay for page interactivity. Lighthouse gave the old page a 68 in performance and the new a 75, so there's some evidence it helps.

Try it on your phone: https://instant-www-js-touch-jsv.vercel.app/examples?app=61daa7fa-fc4c-40a6-a9e4-6f1569a9f252#3-cursors


stopachka (approved):
SGTM!

PR #336, opened 2024-10-09
[Docs] Add users docs

nezaj:
Adds some docs on the `$users` namespace!

[Preview link](https://instant-www-js-add-users-docs-jsv.vercel.app/docs/users)


client/www/pages/docs/users.md
```diff
@@ -0,0 +1,224 @@
...
+  const profileId = lookup('email', currentUser.email);
+  db.transact([
+    tx.profiles[profileId].update({ userId: currentUser.id, nickname: newNick })
+      // Link the profile to the user with the `user` label we defined in the schema
+      .link({ user: currentUser.id }),
+  ]);
+}
+```
+
+If attr creation on the client [is enabled](/docs/permissions#attrs) is enabled,
```
dwwoelfel:
```suggestion
If attr creation on the client [is enabled](/docs/permissions#attrs),
```

dwwoelfel (approved):
Nice work on the docs!

It might be a good idea to mention that the `$users` namespace is in beta and that you'll be able to modify the namespace once we move the `app_users` table into the `triples` table.

nezaj:
@dwwoelfel I worry we have too much `beta` stuff on our website and I would love for us to start cleaning that up.

I think in this case we can just update the docs once the `$users` namespace is modifiable

PR #337, opened 2024-10-09
[delete-entity] fix automatic deletes on the value side

stopachka:
Cam reported a bug, where he was seeing 'stray' entities in instaql results. 

The root cause was the following:

If you have a `boards` namespace, and a `nodes` namespace. If the link was: `[boardId :boards/nodes nodeId]`. 

When a `node` was deleted, the link was not cleaned up. 

Looking into it, I realized there was a stray conditional. 

Added a test, and removed the stray conditional.

Once landed, will write a quick script to delete stray entities 🫡

@dwwoelfel @nezaj

### DIFF
--- a/server/src/instant/data/resolvers.clj
+++ b/server/src/instant/data/resolvers.clj
@@ -38,6 +38,19 @@
        (string/join "-")
        (str "eid-")))
 
+(defn make-attr-resolver
+  [{:keys [conn-pool] :as _db} app-id]
+  (let [attrs (attr-model/get-by-app-id conn-pool app-id)
+        aid->friendly-name (->> attrs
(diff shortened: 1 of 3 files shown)


server/src/instant/data/resolvers.clj
```diff
@@ -38,6 +38,19 @@
        (string/join "-")
        (str "eid-")))
 
+(defn make-attr-resolver
```
stopachka, line 41: This isn't used in the PR, but thought it would be nice. 

I found myself wanting to 'transform' the node results for an app, but didn't have the 'mapping' for ids. So, this just does the mapping for attrs

dwwoelfel (approved):
Nice bug hunting!

PR #338, opened 2024-10-10
Optimistic attrs for lookups in the link value position

dwwoelfel:
This handles adding optimistic attrs when a lookup is used in the link position.

If you use a lookup in a link before the attrs are loaded:

```
tx.users[userId].link({posts: lookup('number', 10)})
```

Then the request will fail with an attribute is not unique error.


stopachka (approved):
Niice!

PR #340, opened 2024-10-10
Remove view check flag

dwwoelfel:
There was an error when we tried to report the status of the view check in the trace. We were running it in admin-mode when it was disabled, so it was returning a map. When I flipped the flag to run it for real, it returned a boolean instead.

I think we can push forward with the view check, as it seems to be working as designed.


stopachka (approved):
Lesgoo!

PR #343, opened 2024-10-10
Send wal alerts directly to discord

dwwoelfel:
Sends wal alerts directly to the errors discord channel and `@` instanteam when we send it.

Adds an `agent` to send the messages in the background.

We now have the ability to do mentions in discord messages, so this also adds an `@instanteam` mention for new paying users and churns.


stopachka (approved):
Lesgoo!

nezaj (approved):
Nice!

PR #345, opened 2024-10-11
[CLI] Increase default timeout for CLI

nezaj:
One of our users put up a PR to increase the timeout of the CLI https://github.com/instantdb/instant/pull/318

I figured instead of making this an extra option, we can just cap it to 5 minutes, which is what Chrome uses https://github.com/nodejs/undici/issues/1373

Tested this with a local backend to ensure timeouts are respected!


stopachka (approved):
sgtm!

PR #349, opened 2024-10-11
Lower refinery goal throughput

dwwoelfel:
We're just over the threshold at the current 30/second rate, lowering it to 25/second may nudge us under.


nezaj (approved):
Nice!

PR #350, opened 2024-10-11
Show attr checks in permissions

dwwoelfel:
Shows attr checks in the sandbox:

image

### DIFF
--- a/server/src/instant/db/permissioned_transaction.clj
+++ b/server/src/instant/db/permissioned_transaction.clj
@@ -333,6 +333,8 @@
     {:scope :attr
      :etype :attrs
      :action :create
+     :program program
+     :data {:updated attr}
      :check (fn [_ctx]
               (if-not program
                 true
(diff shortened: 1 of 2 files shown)


server/src/instant/db/permissioned_transaction.clj
```diff
@@ -333,6 +333,8 @@
     {:scope :attr
      :etype :attrs
      :action :create
+     :program program
```
nezaj, line 336: Just curious, why does `:program` and `:data` need be included?

nezaj (approved):
LGTM! Just one question out of curiosity

server/src/instant/db/permissioned_transaction.clj
dwwoelfel: We use `data` to display the value:
image

I don't think we need program, but it's more consistent with what the other checks return.

PR #351, opened 2024-10-11
Wal2json v2

dwwoelfel:
Updates the wal listener to use format_version 2 in wal2json. It produces one JSON object per tuple instead of one per transaction.

This should put less strain on the db connection when there is a very large update.

Before when there was a transaction, we would have something that looked like:

```js
{change: [
  {table: "triples",...},
  {table: "triples",...},
  {table: "transactions", ...}]}
```

Now we'll get a stream of changes for one transaction:
```js
{action: "B"}
{action: "U", table: "triples", ...}
{action: "U", table: "triples", ...}
{action: "U", table: "triples", ...}
{action: "C"}
```

Our `produce` function is now stateful and will collect all of the changes between the "B" and the "C" to collect the set of changes for the invalidator.

### DIFF
--- a/server/src/instant/jdbc/wal.clj
+++ b/server/src/instant/jdbc/wal.clj
@@ -171,15 +171,29 @@
    a PGConnection with some specific settings. Use `get-pg-replication-conn`
    to get the right kind of PGConnection."
   ^PGReplicationStream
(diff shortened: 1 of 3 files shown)


stopachka (approved):
🚢

PR #352, opened 2024-10-11
Health page for the wal

dwwoelfel:
Adds a page that will return a 500 if the wal unexpectedly restarts and isn't able to recover.

Each time the wal is interrupted, we'll set a flag in config for the machine to indicate that the wal is unhealthy. If we're able to restart, then we'll remove the flag. But if we restart a few times, then we'll stop removing the flag and leave it as unhealthy. In that case, someone will have to manually reset the config flag.

A new page at `/health/system` will return a 500 whenever the config flag is tripped.


stopachka (approved):
Awesome!

PR #353, opened 2024-10-12
Track the time it takes a websocket to respond to our ping

dwwoelfel:
I'm curious if long ping times will help us identify sessions that are unhealthy and if we could use this as a signal that the websocket connection should be closed so that it can reconnect.

We set the last ping time every time we ping and then we set the ping latency when we get a pong. This isn't perfect because we don't have a way to correlate a ping with a pong, but it should still be roughly accurate.

We send the ping-latency along with the other event metrics in `receive-worker/handle-event`.


stopachka (approved):
Heeck yeah!

PR #354, opened 2024-10-13
Use one map for all rooms in Hazelcast

dwwoelfel:
Uses a single map in hazelcast for all of the rooms.

In the previous PR https://github.com/instantdb/instant/pull/333, I mentioned that I wasn't sure if one big map or a map per room was best. Now that we've had it in production for a bit, I think one big map will be better. It turns out that creating the listeners on the map is expensive and can cause the virtual thread to park. If we do one big map, then we only have to create listeners once for the big map instead of every time a new room is joined.

The downside is that our listener fires for every change to any room, even if our instance doesn't have any sessions subscribed to the room. To reduce the impact, we don't ask for values in our listener, just the keys, so less data is transferred.

To make the updates efficient, we use merge functions that hazelcast can serialize and send to the instance that is holding the map.  For example, to remove a session, we call merge with our new `RemoveSessionMerge` class with the session id. Our `remove-session-serializer` sends just the session-id to the instance that holds that key and applies the change. Then the listener will distribute that change to all of the connected sessions.

### DIFF
--- a/server/deps.edn
+++ b/server/deps.edn
@@ -23,6 +23,7 @@
         incanter/incanter {:mvn/version "1.9.3"
                            :exclusions [bouncycastle/bctsp-jdk14
                                         postgresql/postgresql]}
+        com.taoensso/nippy {:mvn/version "3.4.2"}
 
         nrepl/nrepl {:mvn/version "0.9.0"}
         jarohen/chime {:mvn/version "0.3.3"}
--- a/server/src/instant/jdbc/wal.clj
+++ b/server/src/instant/jdbc/wal.clj
@@ -308,7 +308,7 @@
         " The wal handler threw an exception. Check if it restart automatically."
         " If it didn't, redeploy the server.\n\nIf you're quick enough you can "
         "peek at the transaction that caused the error:\n\n"
-        (format "```\nselect data from pg_logical_slot_peek_changes('%s', null, null);```"
+        (format "```\nselect data from pg_logical_slot_peek_changes('%s', null, null, 'format-version', '2', 'include-lsn', 'true');```"
                 slot-name))))
 
 (defn start-worker
--- a/server/src/instant/reactive/ephemeral.clj
+++ b/server/src/instant/reactive/ephemeral.clj
@@ -259,9 +259,9 @@
       ;; The last session left the room, so we should close out the go loop.
       (a/close! chan-before))
 
-    (.merge! (hz-util/->RemoveSessionMergeV1 sess-id)
-             (get-hz-rooms-map)
-             room-key)))
+    (hz-util/merge! (hz-util/->RemoveSessionMergeV1 sess-id)
+                    (get-hz-rooms-map)
+                    room-key)))
 
 (defn clean-old-sessions []
   (let [oldest-timestamp (aws-util/oldest-instance-timestamp)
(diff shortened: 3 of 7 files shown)


server/deps.edn
```diff
@@ -23,6 +23,7 @@
         incanter/incanter {:mvn/version "1.9.3"
                            :exclusions [bouncycastle/bctsp-jdk14
                                         postgresql/postgresql]}
+        com.taoensso/nippy {:mvn/version "3.4.2"}
```
dwwoelfel, line 26: We use nippy to serialize the `data` field, since it can be anything and nippy is good at serializing any clojure data structure efficiently: https://github.com/taoensso/nippy

server/src/instant/jdbc/wal.clj
```diff
@@ -308,7 +308,7 @@
         " The wal handler threw an exception. Check if it restart automatically."
         " If it didn't, redeploy the server.\n\nIf you're quick enough you can "
         "peek at the transaction that caused the error:\n\n"
-        (format "```\nselect data from pg_logical_slot_peek_changes('%s', null, null);```"
+        (format "```\nselect data from pg_logical_slot_peek_changes('%s', null, null, 'format-version', '2', 'include-lsn', 'true');```"
```
dwwoelfel, line 311: Not related, but wasn't really worth its own PR.

server/src/instant/reactive/ephemeral.clj
```diff
@@ -162,116 +198,88 @@
...
-      (fn []
-        (.removeEntryListener m listener-id)
-        (a/close! ch)))))
-
-(defn register-room-map [^IMap m app-id room-id sess-id]
-  (let [map-name (.getName m)]
-    ;; Use locking to ensure that we only create one listener
-    ;; for the app and that we don't get a race condition when
-    ;; we destroy the map after the last person leaves.
-    (locking m
```
dwwoelfel, line 209: No more `locking`!

server/src/instant/util/hazelcast.clj
```diff
@@ -0,0 +1,168 @@
...
+           (java.util UUID)
+           (java.util.function BiFunction)))
+
+;; Be careful when you update the records and serializers in this
+;; namespace. Hazelcast shares them across the fleet, so they must be
+;; updated in a backwards compatible way. The old versions have to
+;; work while the new and old versions are simultaneously deployed.
+
+;; To make breaking changes, follow these steps:
+;; 1. Create a new version of the record, e.g. JoinRoomMergeV2
```
stopachka, line 18: nice deets!

server/src/instant/reactive/ephemeral.clj
```diff
@@ -162,116 +198,88 @@
...
+                                            (get-in [:rooms room-key :session-ids])
+                                            (disj sess-id))]
+                        (cond-> m
+                          true (disj-in [:sessions sess-id] room-id)
+                          (empty? session-ids) (dissoc-in [:rooms room-key])
+                          (seq session-ids) (assoc-in [:rooms room-key :session-ids]
+                                                      session-ids)))))
+        chan-before (get-in old-val [:rooms room-key :chan])
+        chan-after (get-in new-val [:rooms room-key :chan])]
+    (when (and chan-before (not= chan-before chan-after))
```
stopachka, line 258: really nice use of atoms to get this kind of concurrency control! 🤩

stopachka (approved):
Awesome work @dwwoelfel!

PR #356, opened 2024-10-14
[query-once] finish line

stopachka:
Made a series of different changes, to help us get queryOnce over the finish line. Some were quality-of-life fixes, some small bugs, and one bigger change. 

I could have separated this as separate PRs, but thought they were reasonable to have in one holistic PR. Here are the changes:

### Quality of Life Improvements

**1) Remove wrapping around our `log` util**

We had a nifty `log` util, which we used sparingly in Reactor. One problem with it was that when you used it, you would not get the correct line numbers. 

This was because we wrapped `console.log`, so all the invocations would report `'log.js'`. 

Instead of _wrapping_ console.log, I now expose it directly. This maintains the correct line numbers. 

I also updated our flag from `loggingEnabled` to `__instantLogging`. This is mainly because `loggingEnabled` was too general.

**2) Add an `id` to the websocket, and add more logs in the Reactor**

Once we had a better `log` function, I also started to add more logs in the reactor. I added an `id` field to our websocket too. This is purely for helping us see the flow of actions in logs. We can still use the WebSocket for reference equality

### Small Bugfixes

**3) Increase `queryOnce` timeout to 30s**

I noticed our queryOnce timeout was 10 seconds. I think this is a bit too short, given that the normal `fetch` timeout on Chrome is 300 seconds. Afaik for firefox it's 90 seconds. I went a bit more conservative, at 30 seconds. 

**4) Make sure we report an error if `queryOnce` was called during SSR**

Before, `queryOnce` would throw a random error if you called it from the backend (It would say `this.querySubs is not defined`). This is because Reactor is not initialized on the backend. Instead I updated the code to give a more helpful error, and say to use the admin API instead.

## Main bugfix: Make sure we have only one socket active

Previously, it was possible to get into states where we had _multiple_ sockets up at the same time. 

One way this could have happened: 

1. socket 1 is in a 'connecting' state 
2. we go offline and back online
3. socket 2 starts. We try to 'close' the previous socket; [but we only checked for the 'connected' state ](https://github.com/instantdb/instant/blob/main/client/packages/core/src/Reactor.js#L1064)

In general, we wrote code like `ws.close(); ws.open()`, assuming these would happen synchronously. If they did not, we would get into weird states. 

I refactored the code to make sure that we only have 1 socket active. **To do this, I made it so our websocket callbacks _only_ work on the current socket.**

I also made an optimization: If we are trying to start a socket, and we already have one in a 'connecting' state, we re-use it.

@dwwoelfel @nezaj

### DIFF
--- a/client/packages/core/src/Reactor.js
+++ b/client/packages/core/src/Reactor.js
@@ -43,7 +43,7 @@ let _wsId = 0;
 function createWebSocket(uri) {
   const ws = new WebSocket(uri);
   // @ts-ignore
-  ws.id = _wsId++;
+  ws._id = _wsId++;
   return ws;
 }
 
@@ -1027,7 +1027,7 @@ export default class Reactor {
     if (this._ws !== targetWs) {
       log.info(
         "[socket][open]",
-        targetWs.id,
+        targetWs._id,
         "skip; this is no longer the current ws",
       );
       return;
@@ -1055,7 +1055,7 @@ export default class Reactor {
     if (this._ws !== targetWs) {
       log.info(
         "[socket][message]",
-        targetWs.id,
+        targetWs._id,
         m,
         "skip; this is no longer the current ws",
       );
@@ -1069,12 +1069,12 @@ export default class Reactor {
     if (this._ws !== targetWs) {
       log.info(
         "[socket][error]",
-        targetWs.id,
+        targetWs._id,
         "skip; this is no longer the current ws",
       );
       return;
     }
-    log.error("[socket][error]", targetWs.id, e);
+    log.error("[socket][error]", targetWs._id, e);
   };
 
   _wsOnClose = (e) => {
@@ -1082,7 +1082,7 @@ export default class Reactor {
     if (this._ws !== targetWs) {
       log.info(
         "[socket][close]",
-        targetWs.id,
+        targetWs._id,
         "skip; this is no longer the current ws",
       );
       return;
@@ -1097,14 +1097,14 @@ export default class Reactor {
     if (this._isShutdown) {
       log.info(
         "[socket][close]",
-        targetWs.id,
+        targetWs._id,
         "Reactor has been shut down and will not reconnect",
       );
       return;
     }
     log.info(
       "[socket][close]",
-      targetWs.id,
+      targetWs._id,
       "schedule reconnect, ms =",
       this._reconnectTimeoutMs,
     );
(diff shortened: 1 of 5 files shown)


client/packages/core/src/Reactor.js
```diff
@@ -24,8 +24,9 @@ const STATUS = {
   ERRORED: "errored",
 };
 
-const QUERY_ONCE_TIMEOUT = 10_000;
+const QUERY_ONCE_TIMEOUT = 30_000;
```
dwwoelfel, line 27: Won't we time out on the backend after 5 seconds anyway?

client/packages/core/src/Reactor.js
```diff
@@ -38,6 +39,14 @@ const OAUTH_REDIRECT_PARAM = "_instant_oauth_redirect";
 
 const currentUserKey = `currentUser`;
 
+let _wsId = 0;
+function createWebSocket(uri) {
+  const ws = new WebSocket(uri);
+  // @ts-ignore
+  ws.id = _wsId++;
```
dwwoelfel: Might want to prefix this with an underscore in case `Websocket` gets a native id field in the future.

client/packages/core/src/utils/log.ts
```diff
@@ -1,24 +1,15 @@
...
-    debug: makeLogger("debug"),
-    error: makeLogger("error"),
-  };
-})();
+let isEnabled = false;
+if (
+  typeof window !== "undefined" &&
+  typeof window.localStorage !== "undefined"
+) {
+  isEnabled = !!window.localStorage.getItem("__instantLogging");
```
dwwoelfel: Could we enable it by default if `devBackend` is true?

client/packages/core/src/Reactor.js
stopachka: Yes! My thinking is that this timeout will help in cases like: `the user kicks off queryOnce, goes offline for 25 seconds, etc` 

I do think we may want to increase backend timeout too; I am just not 100% sure how our receive-workers will take it; I worry we'll end up saturating the pool. (Perhaps we can look into making an infinite pool of virtual threads, but that can make debugging more difficult)

client/packages/core/src/utils/log.ts
stopachka: oo I like that!

stopachka: done!

PR #358, opened 2024-10-14
[sandbox] set up a repro for the tx order issue

stopachka:
This is a good repro, where sending two transactions from the client returns a kind of 'out-of-order' response. Kosmik team gave us a good repro for this; moving it into the sandbox, to play with it a bit more with our dev infra. 

@dwwoelfel @nezaj


dwwoelfel (approved):
LGTM!

PR #359, opened 2024-10-14
Fix optimistic attrs for lookups on admin model

dwwoelfel:
There was a typo and it was looking at the fwd-ident when it meant to look at the rev-ident.


nezaj (approved):
Nice!

PR #361, opened 2024-10-15
Use a grouped-queue in session

stopachka:
Reference: https://paper.dropbox.com/doc/Receive-Q-Causal-Order-RAfjwje81o6jWqDT6jLie

### Problem
Since we removed the transaction lock, there’s a race condition: if you send two transactions txA, txB, txB can finish before txA. 

```
    // Client: 
    txA { name = "Alice" }
    txB { name = "Bob" } 
    // Expected result: 
    {name: "Bob"} 
    // Got: 
    {name = "Alice" }
```

### Solution: grouped-queue

Created a `grouped-queue` abstraction, so we can maintain causal order for events, and used it inside session. 

`grouped-queue` takes a `group-fn`. Each enqueued item goes through a `group-fn`. We make sure that only _one_ item can run for a particular group at a time. We also distribute work into the queue more fairly. 

Right now we group the following events:
- (`transact`) by session-id 
- (`set-presence` `join-room` `leave-room` `broadcast`) by session-id
- (`add-query` `remove-query`) by session-id and query

### Future work

Now that we have this abstraction, we can use it in more places: for example, we can use it in invalidator to separate work per app

We can also make smarter batching decisions: 
  - we can treat N refreshes as 1 refresh command. 
  - we can batch `transacts` and `set-presence` together, etc

@dwwoelfel @nezaj

### DIFF
--- a/server/src/instant/grouped_queue.clj
+++ b/server/src/instant/grouped_queue.clj
@@ -37,17 +37,16 @@
         (process-fn item))
       :group-key
       (let [group-key arg
-            [prev curr] (swap-vals! group-key->subqueue update group-key pop)
-            prev-subqueue (get prev group-key)
-            item (first prev-subqueue)
+            item (first (get @group-key->subqueue group-key))
+            _ (process-fn item)
+            curr (swap! group-key->subqueue update group-key pop)
             curr-subqueue (get curr group-key)]
-        (process-fn item)
         (when (seq curr-subqueue)
           (.put main-queue [:group-key group-key]))))))
 
 (comment
-  (def gq (create :k))
-  (enqueue! gq {:k :a})
+  (def gq (create identity))
+  (enqueue! gq :a)
   (enqueue! gq {:k :a})
   (enqueue! gq {:k :b})
   (enqueue! gq {:not-grouped :c})
--- a/server/src/instant/grouped_queue.clj
+++ b/server/src/instant/grouped_queue.clj
@@ -1,6 +1,6 @@
 (ns instant.grouped-queue
   (:import
-   (java.util.concurrent LinkedBlockingQueue)
+   (java.util.concurrent LinkedBlockingQueue TimeUnit)
    (clojure.lang PersistentQueue)))
 
 (def empty-q PersistentQueue/EMPTY)
@@ -28,26 +28,29 @@
         (when first-enqueue?
           (.put main-queue [:group-key group-key]))))))
 
-(defn process! [{:keys [main-queue group-key->subqueue]
-                 :as _grouped-q} process-fn]
-  (let [[t arg] (.take main-queue)]
-    (case t
-      :item
-      (let [item arg]
-        (process-fn item))
-      :group-key
+(defn process-polling! [{:keys [main-queue group-key->subqueue]
+                         :as _grouped-q} process-fn]
+  (let [[t arg :as entry] (.poll main-queue 1000 TimeUnit/MILLISECONDS)]
+    (cond
+      (nil? entry) nil
+
+      (= t :item)
+      (process-fn arg)
+      (= t :group-key)
       (let [group-key arg
-            item (first (get @group-key->subqueue group-key))
-            _ (process-fn item)
-            curr (swap! group-key->subqueue update group-key pop)
-            curr-subqueue (get curr group-key)]
-        (when (seq curr-subqueue)
-          (.put main-queue [:group-key group-key]))))))
+            item (first (get @group-key->subqueue group-key))]
+        (try
+          (process-fn item)
+          (finally
+            (let [curr (swap! group-key->subqueue update group-key pop)
+                  curr-subqueue (get curr group-key)]
+              (when (seq curr-subqueue)
+                (.put main-queue [:group-key group-key])))))))))
 
 (comment
-  (def gq (create identity))
-  (enqueue! gq :a)
+  (def gq (create :k))
+  (enqueue! gq {:k :a})
   (enqueue! gq {:k :a})
   (enqueue! gq {:k :b})
   (enqueue! gq {:not-grouped :c})
-  (process! gq println))
+  (process-polling! gq println))
--- a/server/src/instant/grouped_queue.clj
+++ b/server/src/instant/grouped_queue.clj
@@ -16,7 +16,7 @@
 (defn size [{:keys [size] :as _grouped-q}]
   (.get size))
 
-(defn add! [{:keys [group-fn main-queue group-key->subqueue size]
+(defn put! [{:keys [group-fn main-queue group-key->subqueue size]
              :as _grouped-q} item]
   (let [group-key (group-fn item)]
     (if (nil? group-key)
@@ -74,9 +74,9 @@
 
 (comment
   (def gq (create {:group-fn :k}))
-  (add! gq {:k :a})
-  (add! gq {:k :a})
-  (add! gq {:k :b})
-  (add! gq {:not-grouped :c})
+  (put! gq {:k :a})
+  (put! gq {:k :a})
+  (put! gq {:k :b})
+  (put! gq {:not-grouped :c})
   (peek gq)
   (process-polling! gq println))
--- a/server/src/instant/grouped_queue.clj
+++ b/server/src/instant/grouped_queue.clj
@@ -1,23 +1,31 @@
 (ns instant.grouped-queue
   (:import
    (java.util.concurrent LinkedBlockingQueue TimeUnit)
+   (java.util.concurrent.atomic AtomicInteger)
    (clojure.lang PersistentQueue)))
 
 (def empty-q PersistentQueue/EMPTY)
 
 (defn create [{:keys [group-fn]}]
-  {:group-fn group-fn
+  {:size (AtomicInteger. 0)
+   :group-fn group-fn
    :group-key->subqueue (atom {})
    :main-queue (LinkedBlockingQueue.)})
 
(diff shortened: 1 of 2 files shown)


server/src/instant/grouped_queue.clj
```diff
@@ -0,0 +1,53 @@
...
+  (let [[t arg] (.take main-queue)]
+    (case t
+      :item
+      (let [item arg]
+        (process-fn item))
+      :group-key
+      (let [group-key arg
+            item (first (get @group-key->subqueue group-key))
+            _ (process-fn item)
+            curr (swap! group-key->subqueue update group-key pop)
```
stopachka: Note: `swap!` happens _after_ `process-fn`, to avoid a race condition. 

I originally wrote: 

```clojure
      (let [group-key arg
            [prev curr] (swap-vals! group-key->subqueue update group-key pop)
            prev-subqueue (get prev group-key)
            item (first prev-subqueue)
            curr-subqueue (get curr group-key)]
        (process-fn item)
        (when (seq curr-subqueue)
          (.put main-queue [:group-key group-key]))))))
```

But I realized this could lead to a race condition, where two threads were handling items for the same `group-key`: 

1. Imagine a group size = 1, 2 workers
2. Worker1 runs process!, does swap-vals!, and starts process-fn
3. While that is happening, we receive an enqueue! 
4. Now, enqueue would see that we went from no elements to one element, and add to main-queue
5. Worker2 would pick it up, and voila, 2 items are running in parallel for the same group-key

server/src/instant/grouped_queue.clj
```diff
@@ -0,0 +1,53 @@
...
+   :main-queue (LinkedBlockingQueue.)})
+
+(defn enqueue! [{:keys [group-fn main-queue group-key->subqueue]
+                 :as _grouped-q} item]
+  (let [group-key (group-fn item)]
+    (if (nil? group-key)
+      ;; This item is not to be grouped.
+      (.put main-queue [:item item])
+      ;; This item will be grouped on `group-key` 
+      (let [[prev] (swap-vals! group-key->subqueue
```
stopachka: I am using 1 atom, rather than an atom per queue, for the following reason: 

If we wanted to keep `group-key->subqueue` as introspectable state, we would have to keep it in something like an atom. But then, if each subqueue also was an atom, we would need to coordinate changes between the two atoms.

My thinking is since the contention here is strictly for adding and removing elements, it's okay to have everything in one atom. 

Open to other ideas too though!

server/src/instant/grouped_queue.clj
```diff
@@ -0,0 +1,56 @@
...
+
+      (= t :item)
+      (process-fn arg)
+      (= t :group-key)
+      (let [group-key arg
+            item (first (get @group-key->subqueue group-key))]
+        (try
+          (process-fn item)
+          (finally
+            (let [curr (swap! group-key->subqueue update group-key pop)
```
stopachka: Note: `swap!` happens _after_ `process-fn`, to avoid a race condition. 

I originally wrote: 

```clojure
      (let [group-key arg
            [prev curr] (swap-vals! group-key->subqueue update group-key pop)
            prev-subqueue (get prev group-key)
            item (first prev-subqueue)
            curr-subqueue (get curr group-key)]
        (process-fn item)
        (when (seq curr-subqueue)
          (.put main-queue [:group-key group-key]))))))
```

But I realized this could lead to a race condition, where two threads were handling items for the same `group-key`: 

1. Imagine a group size = 1, 2 workers
2. Worker1 runs process!, does swap-vals!, and starts process-fn
3. While that is happening, we receive an enqueue! 
4. Now, enqueue would see that we went from no elements to one element, and add to main-queue
5. Worker2 would pick it up, and voila, 2 items are running in parallel for the same group-key

server/src/instant/grouped_queue.clj
```diff
@@ -0,0 +1,56 @@
...
+  {:group-fn group-fn
+   :group-key->subqueue (atom {})
+   :main-queue (LinkedBlockingQueue.)})
+
+(defn enqueue! [{:keys [group-fn main-queue group-key->subqueue]
+                 :as _grouped-q} item]
+  (let [group-key (group-fn item)]
+    (if (nil? group-key)
+      ;; This item is not to be grouped.
+      (.put main-queue [:item item])
```
stopachka: I am thinking: 

Perhaps we should have a separate `ungrouped-queue` inside the state.

The `main-queue` would only keep a 'signal', and become more of a `dispatch-queue`. 

The reason I am thinking of doing this: 

Right now, it would be hard to determine a true `queue-size`. If we had items stored in a separate queue, than we could derive `queue-size` by adding up all the respective child queues.

wdyt @dwwoelfel ?

dwwoelfel: Could you just keep a counter that you increment when you add an item to the queue and decrement when you start processing an item?

stopachka: Honestly good point; was trying to be 'functional' about it; but the counter feels right. Much less indirection, and only a few places where we have to deal with it.

Will update

server/src/instant/grouped_queue.clj
```diff
@@ -0,0 +1,82 @@
...
+             :as _grouped-q} item]
+  (let [group-key (group-fn item)]
+    (if (nil? group-key)
+      ;; This item is not to be grouped.
+      (do (.incrementAndGet size)
+          (.put dispatch-queue [:item item]))
+
+      ;; This item will be grouped on `group-key` 
+      (let [_ (.incrementAndGet size)
+            [prev] (swap-vals! group-key->subqueue
```
dwwoelfel: One downside of a single queue atom is that there will be a lot of contention on the atom because every event will have to touch the same atom twice.

server/src/instant/grouped_queue.clj
```diff
@@ -0,0 +1,82 @@
...
+                               update
+                               group-key
+                               (fn [subqueue]
+                                 (conj (or subqueue empty-q)
+                                       item)))
```
dwwoelfel:
```suggestion
                               (fnil conj empty-q)
                               item)
```

server/src/instant/grouped_queue.clj
```diff
@@ -0,0 +1,82 @@
...
+      (nil? entry) nil
+      (= t :item) arg
+      (= t :group-key) (first (get @group-key->subqueue arg)))))
+
+(defn process-polling!
+  ([gq process-fn] (process-polling! gq process-fn {:poll-ms 1000}))
+  ([{:keys [dispatch-queue group-key->subqueue size] :as _grouped-q}
+    process-fn
+    {:keys [poll-ms]}]
+   (let [[t arg :as entry] (.poll dispatch-queue poll-ms TimeUnit/MILLISECONDS)]
```
dwwoelfel, line 52: Why do we call `poll` here instead of `take`? Is it because we want to be able to handle shutdown?

server/src/instant/reactive/session.clj
```diff
@@ -445,62 +444,49 @@
...
+                            :attributes {:worker-n worker-n
+                                         :session-id session-id}})
+
+      :else
+      (straight-jacket-handle-receive
+       store-conn
+       eph-store-atom
+       (assoc (into {} session)
+              :worker-n worker-n)
+       (assoc event
```
dwwoelfel, line 456: Is there no way to get the `receive-q-delay-ms` here?

server/src/instant/grouped_queue.clj
dwwoelfel: You might want to add a `locking` around it--I read somewhere that it played more nicely with virtual threads.

server/src/instant/grouped_queue.clj
stopachka: Yes! I wanted a way to be able to handle the shutdown. Open for other ways though

server/src/instant/grouped_queue.clj
stopachka: c.f https://ericnormand.me/guide/clojure-virtual-threads#:~:text=Don%27t%20access%20synchronized%20blocks%20or%20methods

I thought `locking` wasn't good for virtual threads

stopachka: For posterity: 

https://mikemybytes.com/2024/02/28/curiosities-of-java-virtual-threads-pinning-with-synchronized/

server/src/instant/grouped_queue.clj
```diff
@@ -0,0 +1,83 @@
...
+
+       (= t :item)
+       (do
+         (process-fn arg)
+         (.decrementAndGet size)
+         true)
+
+       (= t :group-key)
+       (let [group-key arg
+             item (first (get @group-key->subqueue group-key))]
```
dwwoelfel, line 64: Just an idea, not sure I would implement it, but you might have a place in the queue object for the item/items that you're operating on. That way you could take it off the queue and it might make it easier to combine pending items, since you'd know you can combine anything in the queue.

It might look like:

```clojure
queue = {pending: PersistentQueue[item]
         working: [item]}
 ```
 
 And then the get would look something like:
 
 ```clojure
(swap! 
  big-queue
  key 
  (fn [] 
    (-> queue 
      (update :pending (.pop)
      (assoc :working [(first (:pending queue]))])
 ```
 
 And the complete would just clear the `:working` array.

dwwoelfel (approved):
LGTM!

server/src/instant/grouped_queue.clj
stopachka: Oo. I like this idea; it adds more introspectability too. 

I can try this in a follow-on PR (will do shortly). Maybe we can already start batching!

PR #367, opened 2024-10-16
[bugfix] clear dataForQuery cache when changing users

stopachka:
In our authhello example, I noticed a time where a query gets stuck in a loading state. 

Without refreshing the page: 

1. Log in 
2. Log out
3. Log back in 

This happened because:
- In `notifyOne`, we exit early whenever we see a result that has already been in our cache
- But logging in and out, we never cleared our cache

I am not 100% sure this is the root cause. There may be other situations where our cache backfires and prevents us from notifying a cb. This is a clearly broken case, so first fixing it.

@dwwoelfel @nezaj

### DIFF
--- a/client/packages/core/src/Reactor.js
+++ b/client/packages/core/src/Reactor.js
@@ -1413,6 +1413,7 @@ export default class Reactor {
   updateUser(newUser) {
     const newV = { error: undefined, user: newUser };
     this._currentUserCached = { isLoading: false, ...newV };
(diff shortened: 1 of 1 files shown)


dwwoelfel (approved):
LGTM!

clarkg:
@stopachka did this potentially cause the issue I mentioned here also? 

https://discord.com/channels/1031957483243188235/1148284450992574535/1294297398763458641

PR #369, opened 2024-10-16
Add app-id in the oauth state param

dwwoelfel:
We put the app-id in the state param. When we move the `app_oauth_redirects` in the `triples` table, we'll need to know which app to fetch the redirect from. This lets us figure out which app to use.

Two other small changes:
1. Replaces `(UUID/randomUUID)` with `(random-uuid)` in the route ns
2. Changes our cookie-handling logic a bit to add a prefix to the cookie value. On localhost, clerk was overwriting the `__session` key. In order to know which cookie is ours, we now add a `instantdb_` prefix to the value.

### DIFF
--- a/server/src/instant/runtime/routes.clj
+++ b/server/src/instant/runtime/routes.clj
@@ -431,16 +431,18 @@
                         state
                         (return-error "Missing state param in OAuth redirect."))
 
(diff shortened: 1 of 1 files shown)


nezaj (approved):
Nice!

PR #370, opened 2024-10-16
[Client] Add button for Hide/Show App ID on Copyable component

unllamas:
Hello!

I started using Instantdb today, and I find it great! The only problem I found is that when I use it, my App ID is always visible, so if I want to share a screen or screens, it is impossible.

Because of this, I decided to make a small modification to the client so that it always appears hidden and can be shown if necessary.

Captura de pantalla 2024-10-16 a la(s) 4 38 31 p  m

### DIFF
--- a/client/www/components/ui.tsx
+++ b/client/www/components/ui.tsx
@@ -20,6 +20,8 @@ import {
   CheckCircleIcon,
   ClipboardCopyIcon,
   XIcon,
+  EyeIcon,
+  EyeOffIcon,
 } from '@heroicons/react/solid';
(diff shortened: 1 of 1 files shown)


unllamas:
@dwwoelfel Thanks for the quick reply!

I think what you say is correct.
Below I leave you a preview:

https://github.com/user-attachments/assets/5c870aaa-1120-4684-b455-c4dbe9aca5fd

dwwoelfel:
Thanks for making those changes! I made some minor edits and will merge it in now.

PR #371, opened 2024-10-16
[in-progress] supporting batching in grouped-queue

stopachka:
This is an in-progress PR: 

- Adds an abstraction called an `inflight-queue`. This queue can keep track of a `working` and `pending` state 
  - (Credit to @dwwoelfel [for this idea ](https://github.com/instantdb/instant/pull/361#discussion_r1803428858))

### DIFF
--- a/server/src/instant/grouped_queue.clj
+++ b/server/src/instant/grouped_queue.clj
@@ -5,7 +5,38 @@
    (java.util.concurrent.atomic AtomicInteger)
    (clojure.lang PersistentQueue)))
 
-(def empty-q PersistentQueue/EMPTY)
+;; ----------- 
+;; inflight-queue
+
+(def persisted-q-empty PersistentQueue/EMPTY)
+
+(def inflight-queue-empty
+  {:pending persisted-q-empty
+   :working []})
+
+(defn inflight-queue-put [inflight-queue item]
+  (update inflight-queue :pending conj item))
+
+(defn inflight-queue-empty? [{:keys [pending working] :as _inflight-queue}]
+  (and (empty? pending) (empty? working)))
+
+(defn inflight-queue-pop [{:keys [pending working]}]
+  {:pending (pop pending)
+   :working (if-let [item (first pending)]
+              (conj working item)
+              working)})
+
+(defn inflight-queue-workset [{:keys [working]}]
+  working)
+
+(defn inflight-queue-workset-clear [inflight-queue]
+  (assoc inflight-queue :working []))
+
+(defn inflight-queue-peek-pending [{:keys [pending] :as _inflight-queue}]
+  (first pending))
+
+;; ----------- 
+;; grouped-queue
 
 (defn create [{:keys [group-fn]}]
   {:size (AtomicInteger. 0)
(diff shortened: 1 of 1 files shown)


server/src/instant/grouped_queue.clj
```diff
@@ -61,15 +92,20 @@
 
        (= t :group-key)
        (let [group-key arg
-             item (first (get @group-key->subqueue group-key))]
+             marked (locking group-key->subqueue
+                      (swap! group-key->subqueue update group-key inflight-queue-pop))
```
stopachka: My thinking is, `process-polling` can take a fn like `reserve-fn`, which is takes an `group-key` and an `inflight-queue`, and must return a new `inflight-queue`. 

This way, the caller can decide how much to drain the queue. 

We can than pass the reuslt on to `process-fn`

server/src/instant/grouped_queue.clj
```diff
@@ -5,7 +5,38 @@
    (java.util.concurrent.atomic AtomicInteger)
    (clojure.lang PersistentQueue)))
 
-(def empty-q PersistentQueue/EMPTY)
+;; ----------- 
+;; inflight-queue
+
+(def persisted-q-empty PersistentQueue/EMPTY)
+
+(def inflight-queue-empty
```
stopachka, line 21: After we get this right, can using deftype, so peek, pop, count, et al 'just work' (Clojure ftw)

server/src/instant/grouped_queue.clj
```diff
@@ -5,7 +5,38 @@
...
+  working)
+
+(defn inflight-queue-workset-clear [inflight-queue]
+  (assoc inflight-queue :working []))
+
+(defn inflight-queue-peek-pending [{:keys [pending] :as _inflight-queue}]
+  (first pending))
+
+;; ----------- 
+;; grouped-queue
```
stopachka, line 49: I realize I could use a deftype to create a `GroupedQueue` as well; I worry about having an atom inside a deftype, but heck, why not

server/src/instant/grouped_queue.clj
```diff
@@ -5,7 +5,42 @@
...
+(defn inflight-queue-workset [{:keys [working]}]
+  working)
+
+(defn inflight-queue-workset-clear [inflight-queue]
+  (assoc inflight-queue :working []))
+
+(defn inflight-queue-peek-pending [{:keys [pending] :as _inflight-queue}]
+  (first pending))
+
+(defn inflight-queue-drain [{:keys [pending working]}]
```
dwwoelfel: Might be a good idea to have a `:max-items` field that will limit the number of items that can be moved from pending into working.

server/src/instant/reactive/session.clj
```diff
@@ -479,7 +484,13 @@
                                :attributes {:worker-n n}})
          (do (grouped-queue/process-polling!
               receive-q
-              (fn [item] (straight-jacket-process-receive-q-item store-conn eph-store-atom n item)))
+              {:reserve-fn receive-worker-reserve-fn
+               :process-fn (fn [_ [{{:keys [op]} :item :as entry} :as batch]]
+                             (tracer/with-span! {:name "receive-worker/process-receive-q-item"
+                                                 :attributes {:work-n n
+                                                              :op op
```
dwwoelfel, line 491: Might be nice to log the group-key here too.

dwwoelfel (approved):
Beautiful!

server/src/instant/grouped_queue.clj
stopachka: Good idea! 

I went ahead and refactored like so: 

`(inflight-queue-reserve n iq)`

Using this to take `1` and to take `max-items` from pending

`(inflight-queue-reseve-all iq)` 

This takes all from pending. I could have had a signifier like `:all ` in inflight-queue-reserve, but thought separate functions would lead to a better interface

server/src/instant/reactive/session.clj
stopachka: Done!

PR #373, opened 2024-10-17
Add an lru cache for an app's attrs

dwwoelfel:
Adds a cache for attrs so that we don't have to fetch them on every operation.

They are automatically invalidated when we see the attrs get updated in the wal and we also manually invalidate them when we make changes to the attrs.


stopachka (approved):
Niice!

PR #374, opened 2024-10-17
Add cli config for the instant-config app

dwwoelfel:
Adds the cli to make changes to the instant-config app.

This will make it easier to add new config flags in development and keep them in sync.


stopachka (approved):
Woohoo! Instant on Instant ftw 🔥

PR #375, opened 2024-10-17
Better error for invalid lookup

dwwoelfel:
Returns a nicer error when we can't parse the lookup value.

We were throwing 500s when someone encoded `undefined` in the lookup value. Now we'll throw a validation error.


stopachka (approved):
Niice!

PR #376, opened 2024-10-17
App users in triples

dwwoelfel:
Adds a migration from app_users and related tables into the triples table, so that all of the data for an app is stored with that app.

We created a new "System catalog" app that holds all of the attrs for the user and oauth tables. This allows us to make changes to the attrs later without having to migrate every single app. The special app has id `a1111111-1111-1111-1111-111111111ca7`, and the user that owns the app has id `e1111111-1111-1111-1111-111111111ca7`. There are database triggers that will prevent anyone from accidentally deleting the app or its user. 

There are also triggers that prevent deleting attrs/idents on the system catalog app, but you can set a special postgres setting to override:

```
set instant.allow_system_catalog_app_attr_delete to true;
set instant.allow_system_catalog_app_ident_delete to true;
```

Except for `$users.id` and `$users.email`, all of the system attributes are hidden from the user's view. We may want to expose them later, but right now you can't do anything with them and it would just be confusing to see a ton of new tables when you view your app.

There are default rules for all of the namespaces that prevent view/create/update/delete, except for $users, which has a view rule of `auth.id == data.id`. We also prevent any writes to the system tables from admins, but we'll want to relax that in a follow-up PR for `$users` at least.

Migration plan:
1. Migrate some apps manually and make sure everything is good
2. Set the default `users_in_triples=true` when we create an app
3. Migrate the rest of the apps
4. Wait a week to make sure everything is good
5. Remove all of the old code and old tables

Incidental changes:
1. `tx/transact!` now returns the results of each of the grouped steps. This lets us write a `delete!` helper that gets the thing that was deleted.
2. The perms check routes for the sandbox now return the code that the rule executed. I did that so that we could show the default rule that was applied if they view/modify a system table.

TODO before deploy:
- [x] Push up the flags schema changes
- [ ] Run migrations


stopachka:
🥘👨‍🍳

server/resources/migrations/34_hardcoded_objects.up.sql
```diff
@@ -0,0 +1,105 @@
...
+  execute function prevent_delete_system_catalog_app();
+
+create or replace function prevent_delete_system_catalog_attr()
+returns trigger as $$
+declare
+  disable_trigger boolean := false;
+begin
+
+  begin
+    disable_trigger := current_setting('instant.allow_system_catalog_app_attr_delete')::boolean;
```
stopachka, line 56: So cool 🤩

server/src/instant/db/model/entity.clj
```diff
@@ -55,13 +56,17 @@
...
-    m))
+(defn triples->map [{:keys [attrs include-server-created-at?] :as _ctx} triples]
+  (reduce (fn [acc [_e a v t]]
+            (let [label (attr-model/fwd-label (attr-model/seek-by-id a attrs))]
+              (cond-> acc
+                true (assoc label v)
+
+                (and (= label "id")
+                     include-server-created-at?)
+                (assoc "$serverCreatedAt" (Date. t)))))
```
stopachka, line 67: nice!

server/src/instant/flags_impl.clj
```diff
@@ -87,3 +90,53 @@
...
+    (let [attrs (attr-model/get-by-app-id config-app-id)
+          eid (random-uuid)
+          id-attr-id (resolve-attr-id attrs
+                                      :app-users-to-triples-migration/id)
+          app-id-attr-id (resolve-attr-id attrs
+                                          :app-users-to-triples-migration/appId)
+          machine-attr-id (resolve-attr-id
+                           attrs
+                           :app-users-to-triples-migration/processId)]
+      (tx/transact! aurora/conn-pool
```
stopachka, line 111: Preetty cool!

server/src/instant/system_catalog.clj
```diff
@@ -0,0 +1,249 @@
...
+                 (when (not= chars '(\1 \1 \1 \1 \1))
+                   (let [bitstring (apply str chars)]
+                     (bitstring->char bitstring)))))
+         (apply str))))
+
+(defn decode-system-uuid [uuid]
+  (let [system-str (decode-long->string (.getMostSignificantBits uuid))
+        [etype-shortcode label-shortcode] (string/split (decode-long->string (.getLeastSignificantBits uuid))
+                                                        #"\/")]
+    {:catalog (subs system-str 0 (- (count system-str) 2))
```
stopachka, line 110: Cool idea!

stopachka (approved):
Beautiful work @dwwoelfel !!!

server/src/instant/db/model/attr.clj
```diff
@@ -128,7 +128,7 @@
 ;; -------
 ;; caching
 
-(defonce attr-cache (cache/lru-cache-factory {} :threshold 256))
+(def attr-cache (cache/lru-cache-factory {} :threshold 256))
```
stopachka, line 131: Q: was this intended? (Assuming you may have changed this for repl)

dwwoelfel, line 131: Turns out to be nicer to make this `def` instead of `defonce` because if you change how the attrs are represented, then you automatically clear the cache when you reload the namespace.

server/src/instant/db/transaction.clj
```diff
@@ -140,6 +140,37 @@
...
+   for `a`), so that the delete query can be statically defined in
+   triple-model/delete-multi!"
+  ([conn attrs app-id results]
+   (enforce-on-deletes conn attrs app-id results (:delete-entity results)))
+  ([conn attrs app-id results last-delete-entities]
+   (if-let [deleted-triples (seq last-delete-entities)]
+     (let [deletes (keep (fn [{:keys [triples/attr_id triples/entity_id]}]
+                           (let [attr (attr-model/seek-by-id attr_id attrs)]
+                             (when (and (= :ref (:value-type attr))
+                                        (= :cascade (:on-delete attr)))
```
dwwoelfel, line 161: Only `:cascade` for now, but we could also support `:restrict`, which would throw an error if you had more items to delete.

Our default is technically `:setnull`, so we could make that the default on the attr.

server/src/instant/db/model/attr.clj
stopachka: oh, makes sense!

PR #377, opened 2024-10-17
[bugfix] remove top-level process-receive-q-entry

stopachka:
We started getting some false alerts when we added this top-level log. I also realized we don't sample it. 

Looking at the previous version of [`session.clj`](https://github.com/instantdb/instant/blob/cbf4653a23158433f1292343e267c36d19ad83af/server/src/instant/reactive/session.clj), I realized our highest level log was handle-receive. So, I updated the code to be similar to that. 

Moving forward i'll look into removing our custom logic for sampling in clojure, as now have refinery. With that, I'll also try to flatten the number of 'handle-x' functions we have in session.clj 

@dwwoelfel @nezaj


dwwoelfel (approved):
LGTM!

PR #378, opened 2024-10-17
move our sampling to refinery

stopachka:
Previously, we had an in-house, 'poor-man's' sampler. This led to some child events being lost. Now that we have refinery, we don't need to sample at the event level. Instead, we can use refinery rules and sample at the trace level. 

Moved our current use of sampling into refinery.

@dwwoelfel @nezaj


dwwoelfel (approved):
LGTM!

PR #379, opened 2024-10-18
Remove sampling inside clojure

stopachka:
[Now that all our sampling is happening in refinery](https://github.com/instantdb/instant/pull/378), we can remove our 'poor man's sampler! 

@dwwoelfel @nezaj


dwwoelfel (approved):
LGTM!

PR #383, opened 2024-10-20
Return res (which would include res.user) from signInWithCustomToken

clarkg:
pls 🙏🏾

This would be thus also be returned in `db.auth.signInWithToken` which would be very convenient and helpful 

I noticed that `user` from `db.useAuth` doesn't sync immediately after calling `db.auth.signInWithToken`, causing issues with our React Native navigation that depends on `user` from `db.useAuth`

in any case it seems like an oversight not to return `res`

### DIFF
--- a/client/packages/core/src/Reactor.js
+++ b/client/packages/core/src/Reactor.js
@@ -1452,6 +1452,7 @@ export default class Reactor {
       refreshToken: authToken,
     });
     this.changeCurrentUser(res.user);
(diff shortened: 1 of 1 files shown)


clarkg:
idk why I can't assign reviewers but pls and ty in advance @stopachka @nezaj @dwwoelfel

stopachka (approved):
Landing to unblock! 🫡

PR #386, opened 2024-10-21
Clear the token on the dashboard if we get unauthorized

dwwoelfel:
Logs the user out if the token is bad instead of showing "Record not found: instant-user".

I don't think we hit this much in production, but if we did, the user would have to figure out that they needed to clear localStorage to reset the auth state. 

Useful when switching between databases in development.


stopachka (approved):
Kaizen 🫡

PR #387, opened 2024-10-21
[CEL] Add strings extensions

nezaj:
A user pointed out that we should include a trailing slash for our storage permission rule. This PR updates the docs and also adds the string extensions for CEL

Note: The string extension is not needed for this example, but I thought it might be useful to add this in since it's part of our current version of CEL. Separately I notice we're currently on `0.5.2` of cel-java but the latest is [0.8.0](https://github.com/google/cel-java/tags) which includes some improvements / additional extensions. May be good to upgrade/test in a separate PR


dwwoelfel (approved):
LGTM!

PR #390, opened 2024-10-24
[Admin SDK] Encode getUser for admin sdk

nezaj:
One of our users reported `getUser` borks when there's a `+` in email. This is due to us not encoding the query string [[SO]](https://stackoverflow.com/a/6855723)


dwwoelfel (approved):
LGTM!

clarkg:
@nezaj i think we encountered this too.  subscribing so i know when this is merged and released, thank you!

PR #393, opened 2024-10-28
Make users-in-triples the default for an app

dwwoelfel:
Most apps have been migrated, this will make it the default for new apps


nezaj (approved):
Nice!

PR #394, opened 2024-10-28
Update docs for users table, make it easier to edit view rule

dwwoelfel:
image

image

Also makes it easier to edit the view rule for users. If you don't provide a `create`, `update`, or `delete` rule, we won't throw an error. We'll only throw an error if you set it to something that isn't `"false"`.


nezaj (approved):
Nice!

PR #395, opened 2024-10-28
[Core] Remove object-hash

nezaj:
`object-hash` has a dependency on `crypto` which has caused folks issues in environments like [[Cloudflare](https://discord.com/channels/1031957483243188235/1280983629270487080/1290051152737472542)]

We only use `object-hash` for creating a unique value for our queries but we can use something else.

I looked into a few hashing alternatives and asked claude to come up with something. Based on the results this seemed like the fastest algo. You can see benchmarks here https://github.com/nezaj/hash-benchmarks

### DIFF
--- a/client/packages/core/package.json
+++ b/client/packages/core/package.json
@@ -31,9 +31,7 @@
     "vitest": "^1.6.0"
   },
   "dependencies": {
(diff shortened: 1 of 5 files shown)


stopachka (approved):
SGTM!

PR #399, opened 2024-10-29
Hide the admin token

dwwoelfel:
Uses the new `hide` functionality for copyable inputs to hide the admin token by default:

image


nezaj (approved):
Nice!

PR #400, opened 2024-10-29
[Docs] Fix  $users

nezaj:
One of users got confused because the cli tool would complain that we were trying to modify `$users` table. Re-pulling the schema fixed it for him, but he noticed that the default `$users` schema did not match what was in the docs.


dwwoelfel (approved):
Nice!

PR #401, opened 2024-10-30
[client] Make sure we ignore id key in instaml

stopachka:
On the backend, we already ignored the `id` inside a tx-step arg: 

https://github.com/instantdb/instant/blob/main/server/src/instant/admin/model.clj#L180-L181

We did not do the same on the frontend. This meant that if a user did something like: 

```javascript
tx.posts[idA].update({id: idB})
```

This would lead erroneous results. 

Fixing at least from the client, so it's not possible to send data like this

@dwwoelfel @nezaj

### DIFF
--- a/client/packages/core/__tests__/src/instaml.test.js
+++ b/client/packages/core/__tests__/src/instaml.test.js
@@ -26,6 +26,22 @@ test("simple update transform", () => {
   }
 });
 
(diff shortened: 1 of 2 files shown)


dwwoelfel (approved):
LGTM!

PR #403, opened 2024-10-30
[sandbox] add a `display-code` variable to `program`

stopachka:
daformat noticed that the 'rules' pain in sandbox shows the 'semi-compiled' view of rules:

```cel
cel.bind(...)
```

I added a special `display-code` parameter, which 

- a) shows the version of the code before we add `binds`
- b) shows the pretty-printed code for system tables

@dwwoelfel @nezaj

### DIFF
--- a/server/src/instant/admin/routes.clj
+++ b/server/src/instant/admin/routes.clj
@@ -188,7 +188,7 @@
                                    (dissoc :check)
                                    (update :program
                                            select-keys
-                                           [:etype :action :code])))
+                                           [:etype :action :code :display-code])))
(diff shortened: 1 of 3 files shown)


server/src/instant/admin/routes.clj
```diff
@@ -183,11 +183,12 @@
                         :committed? (:committed? result)
                         :check-results
                         (map (fn [r]
+                               (tool/def-locals!)
```
dwwoelfel: def locals

dwwoelfel (approved):
LGTM!

PR #404, opened 2024-10-30
[Metrics] Filter out migration date for analytics

nezaj:
The migration skewed our metrics, a simple fix is to just ignore transactions for that day, this may filter out some legitmate apps/devs that should be active, but not in a meaningful way


dwwoelfel (approved):
LGTM!

PR #405, opened 2024-10-30
[admin] order admin responses

stopachka:
A user pointed out that responses in admin were not coming in serverCreatedAt order. 

This makes it happen. 

Note: 

Right now we assume that we only order by serverCreatedAt. Eventually, this will either become unnecessary due to our instaql CTE update. But in case it's not, we'll need to update the logic when we allow for ordering by other keys. 

@dwwoelfel @nezaj


dwwoelfel (approved):
Very nice!

PR #406, opened 2024-10-30
[www] Engineered in San Francisco

nezaj:
Minor, but had it as a todo to check if we mention our NY address anywhere on the website. We don't list the address, but figured would update the footer


stopachka (approved):
LGTM!

PR #408, opened 2024-10-30
[admin] add a GET /schema

stopachka:
Nowlow mentioned he'd like an admin schema endpoint. I thought it would be nice to set one up! 

This exposes the 'schema' layout we have for CLI. I won't add it in the official API just yet, but would be good to open it up and get feedback. 

@dwwoelfel @nezaj

### DIFF
--- a/server/src/instant/admin/routes.clj
+++ b/server/src/instant/admin/routes.clj
@@ -21,7 +21,8 @@
    [instant.util.string :as string-util]
    [instant.util.uuid :as uuid-util]
    [ring.util.http-response :as response]
-   [instant.model.schema :as schema-model])
+   [instant.model.schema :as schema-model]
+   [clojure.string :as string])
   (:import
    (java.util UUID)))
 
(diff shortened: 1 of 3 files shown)


server/src/instant/admin/routes.clj
```diff
@@ -407,6 +409,16 @@
...
 
+;; Experimental. If we change this let the Kosmik folks know
+(defn schema-get [req]
+  (let [{app-id :app_id} (req->admin-token! req)
+        current-attrs (attr-model/get-by-app-id app-id)
+        current-schema (schema-model/attrs->schema current-attrs)]
+    (response/ok {:schema (update current-schema
+                                  :refs
+                                  update-keys
+                                  (partial string/join "-"))})))
```
stopachka, line 420: I do this, as the key on `refs` is a vec. I don't think we use this vec in any place, but instead of potentially introducing a bug, thought for now I'd update it here, so it becomes:

{
  "schema": {
    "refs": {
      "cats-people-people-cats": {
        "forward-identity": [

dwwoelfel:
You might want to do attr-model/remove-hidden on the attrs so they don't get all of the non-$user system attrs, like $magicCode.

dwwoelfel:
> You might want to do attr-model/remove-hidden on the attrs so they don't get all of the non-$user system attrs, like $magicCode.

Oh nvm, looks like schema-model/attrs->schema does that already

dwwoelfel (approved):
lgtm!

PR #409, opened 2024-10-31
Add `$not` and `$isNull` operators 

dwwoelfel:
Adds a `$not` and `$isNull` operators for where queries.

$not looks like this:

```js
{books: {$: {where: {title: {:$not "The Things We Make"}}}}}
```

$isNull takes a boolean, true to return rows where it is null and false to return rows where it isn't:
```js
{books: {$: {where: {title: {:$isNull true}}}}}
```

At the datalog layer, the operators are kind of crammed in to the `value` position. For `$not`, the pattern looks like:

```
[?e aid {:$not "val"}]
```

For `$isNull`, the pattern looks like:

```
[?e id-eid {:$isNull {:attr-id aid, :nil? true}}]
```

It might be nicer if we had predicate functions, but I had a difficult time figuring out how to map that to our postgres queries.

The `$not` CTE is pretty simple, it just looks like:

```sql
    AND (
      attr_id = 'aid'::uuid
    )
    AND (value <> CAST('"a"' AS JSONB))
```

When you use a `$not`, we also add a `$isNull: true`, since our `$not` will only match on triples that have something set for the attribute. We still want to return rows where the value is "undefined".

The `$isNull` is a little more complicated. It checks for `null` as a value for the attr and for entities where the attr is missing:

```sql
    AND (
      attr_id = 'id-attr-id'::uuid
    )
    AND (
      entity_id NOT IN (
        SELECT
          t.entity_id
        FROM
          triples AS t
        WHERE
          (t.entity_id = entity_id)
          AND (
            t.attr_id = 'aid'::uuid
          )
          AND (t.value <> CAST('null' AS JSONB))
      )
    )
```

Also changes the `in` operator to `$in`, but keeps the old one around for backwards compatibility.


client/sandbox/react-nextjs/pages/play/notandnil.tsx
```diff
@@ -0,0 +1,263 @@
...
+
+async function provisionEphemeralApp() {
+  const r = await fetch(`${config.apiURI}/dash/apps/ephemeral`, {
+    method: "POST",
+    headers: {
+      "Content-Type": "application/json",
+    },
+    body: JSON.stringify({
+      title: "Pagination example",
+      // Uncomment and start a new app to test rules
```
stopachka, line 219: cool!!

client/www/pages/docs/instaql.md
```diff
@@ -662,7 +766,7 @@ In these cases, you can use `queryOnce` instead of `useQuery`. `queryOnce` retur
 Unlike `useQuery`, `queryOnce` will throw an error if the user is offline. This is because `queryOnce` is intended for use cases where you need the most up-to-date data.
 
 ```javascript
-const query = { todos: {}, };
+const query = { todos: {} };
```
stopachka, line 769: _mrclean_

server/src/instant/db/datalog.clj
```diff
@@ -502,6 +536,24 @@
     :a (in-or-eq :attr-id v)
     :v (in-or-eq :value (map value->jsonb v))))
 
+(defn- function-clauses [named-pattern]
```
stopachka, line 559: man impressed with how localized this change can be inside datalog.clj

stopachka (approved):
Beautiful!

dwwoelfel:
Found an issue with how this handles references. It's putting the nil and not checks at the wrong levels on the CTEs.

Still a little bit of work to do before shipping.

PR #410, opened 2024-10-31
Do background hz updates in a single thread

dwwoelfel:
It turns out running the hz background sync in multiple threads was a bad idea because the events are handled out of order and then we set presence after they've left the room.

This puts the background updates in a single queue and handles them from one background worker. I'll be watching `instant.ephemeral.hz-ops-q.size` to make sure the queue isn't getting overwhelmed (if it does there's a feature flag that we can flip).


stopachka (approved):
Heck yeah!

PR #411, opened 2024-10-31
Handle hazelcast shutdown on our own

dwwoelfel:
Hazelcast is shutting down on its own when it sees the sigterm signal, but that happens before we're finished shutting down the server. It leaves a trail of errors as the sessions close and can't update hazelcast. 

After this change, we'll wait to gracefully shutdown hazelcast until after the undertow server shuts down.


nezaj (approved):
Nice!

PR #412, opened 2024-10-31
Add newsletter

nezaj:
Adds tooling for sending out broadcast emails!

This PR contains three things

1. Preview tool for emails located at `intern/emails`
2. Adds content for the october newsletter
3. Adds a server scripts `newsletter.clj` with instructions on how to send out newsletters

image

### DIFF
--- a/server/src/instant/scripts/newsletter.clj
+++ b/server/src/instant/scripts/newsletter.clj
@@ -43,12 +43,8 @@
 (defn read-file
   "Reads a file relative to a base path"
   [base-path file-path]
-  (try
-    (let [full-path (io/file base-path file-path)]
-      (slurp full-path))
-    (catch Exception _
-      (println "Error reading file:" file-path)
-      nil)))
+  (let [full-path (io/file base-path file-path)]
+    (slurp full-path)))
 
 (comment
   (read-file base-path (html-path "oct2024"))
--- a/server/src/instant/scripts/newsletter.clj
+++ b/server/src/instant/scripts/newsletter.clj
@@ -127,7 +127,7 @@
 
 (comment
   ;; 1. Set your params
-  (def params {:title "Instant News - Oct 2024" :slug "oct2024"})
+  (def params {:title "Instant News - Oct 2024" :slug "UPDATE ME"})
 
   ;; 2. Send test email to yourself, verify looks good
   (let [htmlBody (read-file base-path (html-path (:slug params)))
--- a/server/src/instant/scripts/newsletter.clj
+++ b/server/src/instant/scripts/newsletter.clj
@@ -1,0 +1,148 @@
+(ns instant.scripts.newsletter
+  "Send newsletters to our users
+
+  Usage: First run through these steps to prepare the email content
+    1. Create email content in Dropbox. Get sign off on the copy.
+    2. Export content as markdown and save to `www/intern/_emails/[slug].md`
+    3. Preview the styled markdown at `localhost:3000/intern/emails/[slug]`
+    4. Once preview looks good, generate html and text versions of the email
+       from `client/Makefile` via `make build-email`
+       Note: we want both html and text to maximize delivery
+    5. Refresh the page at `localhost:3000/intern/emails/[slug]`
+       and verify text version looks good via the text view button
+
+    And now go through the comment block at the bottom of this file to send
+    out the newsletter. Huzzah!
+  "
+
+  (:require
+   [instant.config :as config]
+   [clj-http.client :as clj-http]
+   [clojure.java.io :as io]
+   [clojure.string :as str]
+   [instant.util.json :refer [->json]]
+   [instant.util.tracer :as tracer]
+   [clojure.tools.logging :as log]))
+
+(def base-path (io/file (.getParent (io/file *file*)) "../../../../"))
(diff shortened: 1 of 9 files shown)


server/src/instant/scripts/newsletter.clj
```diff
@@ -0,0 +1,156 @@
...
+  (reset! once false)
+  (reset! errors []))
+
+(defn read-file
+  "Reads a file relative to a base path"
+  [base-path file-path]
+  (try
+    (let [full-path (io/file base-path file-path)]
+      (slurp full-path))
+    (catch Exception _
```
stopachka: nit: are you sure you want to add a try around this? Likely if the file doesn't exist, it'd be nice for the code execution to break out anyways

server/src/instant/scripts/newsletter.clj
```diff
@@ -0,0 +1,156 @@
...
+                     :message message}]
+    (swap! errors conj error-entry)))
+
+(defn handle-email-error! [e to]
+  (if (inactive-recipient? e)
+    (add-error! to "INACTIVE_RECIPIENT" "This email address has been marked inactive.")
+    (add-error! to "ERROR" (.getMessage e)))
+  nil)
+
+(defn send! [{:keys [to title textBody htmlBody
```
stopachka, line 70: nit: Perhaps we could reuse `postmark/send!`, but call it with the alias stuff from here?

server/src/instant/scripts/newsletter.clj
```diff
@@ -0,0 +1,156 @@
...
+            :body (->json body)})
+          (catch Exception e
+            (handle-email-error! e to)))))))
+
+(defn send-newsletter! [{:keys [title slug live?]}]
+  (let [emails (str/split-lines (read-file base-path emails-path))]
+    (if-not live?
+      (doseq [email emails]
+        (log/info "Simulating sending newsletter to" email))
+      (if @once
```
stopachka: nit: I would consider naming the atom something like `sent-already?` or something

server/src/instant/scripts/newsletter.clj
```diff
@@ -0,0 +1,156 @@
...
+            (let [htmlBody (read-file base-path (html-path slug))
+                  textBody (read-file base-path (txt-path slug))]
+              (send! {:to email
+                      :title title
+                      :htmlBody htmlBody
+                      :textBody textBody}))))))))
+
+(comment
+  ;; 1. Set your params
+  (def params {:title "Instant News - Oct 2024" :slug "oct2024"})
```
stopachka: nit: perhaps we don't provide the slug here, and force the user to write it. This way, they would not accidentally eval the forms and end up sending an old version

server/src/instant/scripts/newsletter.clj
nezaj: Agreed! Pushed up an update!

server/src/instant/scripts/newsletter.clj
```diff
@@ -0,0 +1,152 @@
...
+
+  ;; 2. Send test email to yourself, verify looks good
+  (let [htmlBody (read-file base-path (html-path (:slug params)))
+        textBody (read-file base-path (txt-path (:slug params)))]
+    (send! (assoc params
+                  :to "joeaverbukh@gmail.com"
+                  :htmlBody htmlBody
+                  :textBody textBody)))
+
+  ;; 3. Add emails to server/resources/emails/emails.txt
```
stopachka, line 140: Cool! How does once generate these? Maybe a lil script would be nice

stopachka (approved):
Love the completeness here! @nezaj

server/src/instant/scripts/newsletter.clj
nezaj: @stopachka One difference with `postmark/send!` is that it errors out on a failure, in our case we want to keep going and just collect errors.

server/src/instant/scripts/newsletter.clj
nezaj: Updated!

server/src/instant/scripts/newsletter.clj
nezaj: Since we're warming up right now I choose not to automate this part for this month. I dumped the list of users from the top page (so the most active are first) to a google sheet and am manually sending 100 daily emails this week, 200 daily the next week, etc.

Starting next month we can fetch this via a query. Will push an update then

server/src/instant/scripts/newsletter.clj
nezaj: I like it!

server/src/instant/scripts/newsletter.clj
stopachka: Love it!

server/src/instant/scripts/newsletter.clj
stopachka: i.m.o you could still use the normal postmark/send!, but add a try catch around it: 

```clojure
  (try 
     ... 
     (catch ... 
        (add-error!
```

This can be cleaned up later though, so if there's some kludges to doing it this way we can punt it

nezaj: Discussed in person! We'll leave it as is!

PR #413, opened 2024-10-31
[client] make sure that auth is finished after awaiting db.auth

stopachka:
If you await `db.auth.signIn*`, you would expect that afterwards the system would be logged in. But Clark noticed this was not the case. A basic repro: 

```javascript
function App() {
  const { isLoading, user } = db.useAuth();
  const [calledSignIn, setCalledSignIn] = useState(false);
  const [result, setResult] = useState<any>(null);
  useEffect(() => {
    if (calledSignIn && !result) {
      setResult({ calledSignIn, user }); // Problem: `user` won't exist, when `calledSignIn` is already true
    }
  }, [user, calledSignIn]);
  return (
    <button
      onClick={async () => {
        const ret = await getToken();
        await db.auth.signInWithToken(ret.token);
        setCalledSignIn(true);
      }}
    >
      Sign in
    </button>
  );
}
```

This is because we did not await on `this.changeCurrentUser`. Adding an await fixed the issue 

@dwwoelfel @nezaj

### DIFF
--- a/client/packages/core/src/Reactor.js
+++ b/client/packages/core/src/Reactor.js
@@ -1441,7 +1441,7 @@ export default class Reactor {
       email,
       code,
     });
-    this.changeCurrentUser(res.user);
+    await this.changeCurrentUser(res.user);
     return res;
   }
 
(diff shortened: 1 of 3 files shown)


nezaj (approved):
Nice!

PR #415, opened 2024-10-31
[test] perf for large queries

stopachka:
A user gave a good repro to test large nested queries, and how tx affects the store. 

Moved the demo inside here, so we can iterate on it. 

@dwwoelfel @nezaj


dwwoelfel:
LGTM!

PR #418, opened 2024-11-01
[perf][client] improve client-side nested queries by 12x

stopachka:
While looking into the large query perf optimizations, I realized that most of the time spent on those queries were looking up attrs: 

CleanShot 2024-11-01 at 11 21 43@2x

**I added an `attrs` index, for fwd ident, rev ident, blobs, and primary. This reduced one of the bigger queries from 240ms to about 20ms!** 

**Before**
CleanShot 2024-11-01 at 11 13 54@2x

**After** 
CleanShot 2024-11-01 at 11 15 12@2x

### DIFF
--- a/client/packages/core/src/store.js
+++ b/client/packages/core/src/store.js
@@ -47,7 +47,7 @@ function setInMap(m, path, value) {
   setInMap(nextM, tail, value);
 }
 
-function createIndexMap(attrs, triples) {
+function createTripleIndexes(attrs, triples) {
   const eav = new Map();
   const aev = new Map();
   const vae = new Map();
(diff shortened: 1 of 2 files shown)


client/packages/core/src/store.js
```diff
@@ -106,14 +114,19 @@ export function fromJSON(storeJSON) {
...
+
 export function createStore(
   attrs,
   triples,
   enableCardinalityInference,
   linkIndex,
 ) {
-  const store = createIndexMap(attrs, triples);
+  const store = createTripleIndexes(attrs, triples);
   store.attrs = attrs;
```
stopachka, line 128: At some point, it would be nicer to have an `Attr` abstraction like we do in clojure. 

I skipped it in this PR, as we don't have the really nice protocol abstraction, so we'd have to change callsites

dwwoelfel (approved):
Wow!

PR #420, opened 2024-11-01
Remove session from store before clearing ephemeral state

dwwoelfel:
If a session disconnects just as they join the room, we can end up with data for the session in the ephemeral state that will never go away.

This change removes the session from the store before we remove the ephemeral state.


stopachka (approved):
Heck yeah!

PR #421, opened 2024-11-03
Set up refinery to use redis for peer-to-peer communication

dwwoelfel:
Sets up refinery to use redis for peer communication.

This should help us keep under the honeycomb rate limit as we add additional machines.

It should also help when we deploy new instances, since each new instance will see how many events were delivered instead of starting from scratch each time.

Uses a serverless valkey (aws's redis fork) elasticache. We'll have to keep an eye on usage to see what it costs, but I wouldn't expect it to go over the free tier.

Also makes a couple of changes to our `docker-compose.yml` to try to improve graceful shutdown. I wasn't seeing the `shutdown` logs I would expect in honeycomb or in the cloudwatch logs. It turns out that we are gracefully shutting down, but we're not seeing the logs because refinery and the cloudwatch logger get shut down first.

Instead of doing `docker compose down`, elasticbeanstalk calls `docker stop` on each running container. That means that the dependency tree we set up in the docker-compose.yml isn't respected and the refinery container stops before our app container.

We could fix this moving refinery into its own separate cluster. Not sure how we'd fix the cloudwatch logs not uploading.

### DIFF
--- a/server/scripts/prod_ssh.sh
+++ b/server/scripts/prod_ssh.sh
@@ -1,9 +1,19 @@
-instance_id=$(
-  aws ec2 describe-instances \
-    --filter "Name=tag:elasticbeanstalk:environment-name,Values=Instant-docker-prod-env" \
-    --query "Reservations[].Instances[?State.Name == 'running'].InstanceId[]" \
-    --output text
-)
+while [[ "$#" -gt 0 ]]; do
+  case $1 in
+    --instance-id) instance_id="$2"; shift ;;
+    *) echo "Unknown parameter passed: $1"; exit 1 ;;
+  esac
+  shift
+done
+
+if [ -z "$instance_id" ]; then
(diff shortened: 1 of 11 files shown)


server/scripts/prod_ssh.sh
```diff
@@ -1,12 +1,22 @@
-instance_id=$(
-  aws ec2 describe-instances \
-    --filter "Name=tag:elasticbeanstalk:environment-name,Values=Instant-docker-prod-env" \
-    --query "Reservations[].Instances[?State.Name == 'running'].InstanceId[]" \
-    --output text
-)
+while [[ "$#" -gt 0 ]]; do
+  case $1 in
+    --instance-id) instance_id="$2"; shift ;;
```
stopachka, line 3: nice!

stopachka (approved):
Niice!

PR #423, opened 2024-11-04
Fix for multiple $not with references

dwwoelfel:
Two issues that are fixed:

1. We were comparing `:value` in the or conditions when we should have been comparing `:value-blob`.

2. We wern't handling the `symbol-map` datastructure properly. With ands and ors nested in just the right way, we would stop traversing the tree early.


stopachka (approved):

stopachka:
🫡 🫡 🫡

PR #424, opened 2024-11-04
Custom serializer for hazelcast room keys

dwwoelfel:
Seeing a weird issue in prod where sessions aren't being removed because the key isn't recognized as being in the map. I haven't been able to replicate it locally, but it's possible that there is some inconsistency in how the map key is serialized. 

This PR writes our own serializer/deserializer for the room-map keys. Hopefully that will make hazelcast recognize the key.

### DIFF
--- a/server/src/instant/reactive/ephemeral.clj
+++ b/server/src/instant/reactive/ephemeral.clj
@@ -57,8 +57,9 @@
         network-config (.getNetworkConfig config)
         join-config (.getJoin network-config)
         tcp-ip-config (.getTcpIpConfig join-config)
-        aws-config (.getAwsConfig join-config)]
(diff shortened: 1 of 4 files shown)


server/src/instant/reactive/ephemeral.clj
```diff
@@ -78,7 +78,7 @@
                            hz-util/serializer-configs)
 
     (let [hz (Hazelcast/getOrCreateHazelcastInstance config)
-          hz-rooms-map (.getMap hz "rooms")
+          hz-rooms-map (.getMap hz "rooms-v2")
```
dwwoelfel, line 85: Changed the room-map name since the keys won't be compatible.

stopachka (approved):
LGTM!

PR #430, opened 2024-11-05
[docs] show npx instead of npm i for isntant-cli

stopachka:
In our docs page, we explictly wrote: `npm i -D instant-cli`. This will end up pinning a specific version of the cli to this app. 

However, we don't really want this to be the default experience. The default experience should be to use npx instant-cli. 

The downside of this is that we will have to worry more about backwards compatibility, but it is a better experience for the user. 

@dwwoelfel @nezaj


nezaj (approved):
Sounds good!

PR #432, opened 2024-11-05
[Devtool] Add ids and adjust escape

nezaj:
What it says on the tin!


dwwoelfel (approved):
LGTM!

Would we ever have multiple of these elements? If so, might be better to add a class name instead of an id.

nezaj:
I don't think we would have multiple of these but no harm in making them a class instead!

PR #435, opened 2024-11-06
[clean] remove stray sandbox clients

stopachka:
We no longer use `perf_1` and `tx_order` repros. Removing them to keep the sandbox clean

@dwwoelfel @nezaj


dwwoelfel (approved):
Farewell old friends!

PR #440, opened 2024-11-07
[DOCS] correction in rooms presence example

uinstinct:
The [topics](https://www.instantdb.com/docs/presence-and-topics#topics) in presence and topics section seems to have non-functional code.

This PR fixes the same. It also introduces some more emojis and rendering them for a wider understanding of the topics feature example. (I kept it minimal for simple demos)

### Before

Upon just copy pasting the topics code, we get errors and a non-functional code.

image

### After

We get a functional code with the corrected example!

https://github.com/user-attachments/assets/790a5ced-1bf6-4694-b4bc-8d69bd0eb517

### DIFF
--- a/client/www/pages/docs/presence-and-topics.md
+++ b/client/www/pages/docs/presence-and-topics.md
@@ -183,14 +183,35 @@ const db = init<Schema, RoomSchema>({ appId: APP_ID });
 const room = db.room('video', 'room-id-123');
 
 function Component() {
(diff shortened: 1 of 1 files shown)


nezaj:
Thanks for pointing this out! Agreed we should have a fuller example! Opted to use our live reactions example for consistency. https://github.com/instantdb/instant/pull/449

PR #442, opened 2024-11-07
[experiment][strong-init-sandbox] export a `Graph` interface in instant.schema.ts

stopachka:
**Context:** 

We used a lot of inference to generate the `Schema` type. When you infer types though, they tend to auto-expand in intellisense, so you get these kind of inscrutable types like so: 

**Before** 

CleanShot 2024-11-07 at 15 27 15@2x

CleanShot 2024-11-07 at 15 27 15@2x

This is a [known issue](https://github.com/microsoft/TypeScript/issues/45954) in Typescript. 

However, there's a "trick" we can do. Typescript doesn't expand interfaces. So if we 'cast' our graph to an interface, the types look a lot better: 

**After** 

CleanShot 2024-11-07 at 15 29 27@2x

CleanShot 2024-11-07 at 15 29 44@2x

**Related Stackoverflow: https://stackoverflow.com/questions/79164542/how-can-i-mask-a-generic-type-for-intellisense**
--- 

Right now, I'm just doing this manually in our strong-init playground. If we end up liking this, I'll make this a part of the instant.schema.ts generation. For now, there's still more parts where the types look off, so I'm going to keep pushing and playing. 

@dwwoelfel @nezaj

### DIFF
--- a/client/sandbox/strong-init-vite/instant.schema.ts
+++ b/client/sandbox/strong-init-vite/instant.schema.ts
@@ -1,6 +1,6 @@
 import { i } from "@instantdb/react";
 
-const graph = i.graph(
+const _graph = i.graph(
   {
     messages: i.entity({
       content: i.string(),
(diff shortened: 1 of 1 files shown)


nezaj (approved):
Neato!

PR #443, opened 2024-11-08
[in-progress] strong-init upgrades

stopachka:
In-progress PR: am going to use this as a master to change a bunch of things.

--- 

### Change 1: Split Paths
Previously, _both_ `init` and `init_experimental` reused a bunch of the same subclasses:  classes like `InstantReact`, `InstantReactWeb`, `InstantCore`, and so on. 

This approach is good for code sharing, but it also complects the two features:

1. We have to be extra careful about changes we make to experimental types, because it _could_ affect the current `init` types.
2. Type changes on strong init will affect normal init: For example, `CardinalityInference` has seeped through all our classes, even though the normal `init` does not have to deal with this at all. 
3. Strong init is affected by normal init: For example, we need to handle the case where `RoomSchema` comes in a different position for arguments. 

I went ahead and split the codepaths: `init` creates `InstantReact`, `InstantCore`, etc, while `init_experimental` creates `InstantReactExperimental`, etc 

This will let us be a lot more bold with changes to strong_init, knowing we won't affect the current flow. 

Once we land this, I am going to fast follow to: 

a. Remove strong-init bits in the normal `init` flow 
b. Remove the normal-init bits in the `strong-init` flow 

### Change 2: Remove strong-init parts from normal-init 

- Removes `WithCardinalityInference` from normal init flows
- Removes InstantGraph checks in normal init flows

### Change 3: cardinalityInference by default for strong-init

- Updated strong-init so we do cardinalityInference by default

### Change 4: Schema <- RoomSchema for strong init classes

- We used to extract RoomSchema, but this was only due to backwards compatibility with normal init classes. Updated our strong init classes to just care about Schema

### DIFF
--- a/client/packages/admin/src/index.ts
+++ b/client/packages/admin/src/index.ts
@@ -159,14 +159,12 @@ function init<Schema extends {} = {}>(config: Config) {
 
 function init_experimental<
   Schema extends InstantGraph<any, any, any>,
-  WithCardinalityInference extends boolean = true,
 >(
   config: Config & {
     schema: Schema;
-    cardinalityInference?: WithCardinalityInference;
   },
 ) {
-  return new InstantAdminExperimental<Schema, WithCardinalityInference>(config);
+  return new InstantAdminExperimental<Schema>(config);
 }
 
 /**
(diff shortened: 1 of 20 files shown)


nezaj (approved):
Go forth and prosper!

stopachka:
(Taking a different approach...will push up a different PR!)

PR #446, opened 2024-11-08
[DOCS] Cursors prop name correction

uinstinct:
We are using an incorrect prop name `currentUserColor` in docs whereas the [correct prop name is `userCursorColor`](https://github.com/instantdb/instant/blob/ff52fb3024a7c1c3e3e37498ecd23280b7fa8834/client/packages/react/src/Cursors.tsx/#L20).

Also, adding `className="h-full w-full"` helps us understand styles/classes need to be applied to the `Cursors` element to make it visible.


nezaj (approved):
Nice catch!

PR #448, opened 2024-11-08
Adds opt-in enforcement of data types

dwwoelfel:
**Easier to review with whitespace hidden**: https://github.com/instantdb/instant/pull/448/files?w=1

Adds the ability to opt-in to type enforcement for an attribute.

Supported types are: `string`, `number`, `boolean`, and `date`.

`date` will accept either `number` as ms since the epoch, or `string` if it's anything that postgres can cast to a date (`JSON.stringify(new Date())` works).

We create an indexing job when you add a type to an attribute. The job updates all triples in the background, 1000 at a time. The jobs themselves are stored in the `indexing_job` table in the database and the job runner breaks the job into smaller sequential tasks. Each time it executes a task for the job, it will check the job out from the db, then release the job after it finishes the task. We keep the tasks small so that we're resilient to instances starting and stopping.

This is what it looks like when you set the type from the Explorer:

https://github.com/user-attachments/assets/d6137f4e-74b9-4c04-b34a-46915735db21

If you click on the row, it takes you to the full row in the explorer.

This is what it looks like when you add a type from the cli:

https://github.com/user-attachments/assets/d87a4f40-37d7-43dd-8905-476db31fd782

This is what it looks like when there is invalid data:

https://github.com/user-attachments/assets/03b295fa-ec9f-4c77-8540-67e50a7c16f8

If you're using a terminal that supports it, you can Cmd+click on the `id` to take you to the explorer page with that entity selected.

Minor changes:
1. Changes how we store `where` clauses in the explorer. Instead of storing a namespace and id, it stores a tuple with field + value. That lets us link directly to a row for an individual entity. We link to the entity when we find invalid data for the checking process.
2. Fixes a bunch of reflection warnings
3. Adds a new `gauged-chan` which will publish gauges with the size and capacity of the channel.
4. Adds `process-id` to all spans
5. Makes the bootstrap migrations resilient against adding new fields to the database

Before deploying:
- [ ] Run migration

After deploying:
- [ ] Run `alter table triples validate constraint valid_value_data_type;` (prevents locking table during up migration)
- [ ] Run migration on system tables


client/www/components/dash/explorer/EditRowDialog.tsx
```diff
@@ -45,6 +45,12 @@ function tryJsonParse(value: any) {
 }
 
 function getAppropriateFieldType(attr: SchemaAttr, value: any): FieldType {
+  if (!value && attr.checkedDataType) {
```
stopachka: Not required for this PR, just realized, these checks could be rewritten as something like: 

if (value) {
   ...
}
if (attr.checkedDataType) { 
  ...
} 
if (attr.inferredTyped?.length) { 
 ...

client/www/pages/docs/schema.md
```diff
@@ -40,10 +40,18 @@ The first parameter to `i.graph` is a dictionary of entities, where the key repr
 
 Entity definitions accept a map of attribute definitions, where the key represents the attribute name and the value contains configuration for the attribute.
 
-First we specify the expected type of the attribute: `i.string()`, `i.number()`, `i.boolean()`, `i.json()` and `i.any()`.
+First we specify the expected type of the attribute: `i.string()`, `i.number()`, `i.boolean()`, `i.date()`, `i.json()`, or `i.any()`.
+
+{% callout %}
+
+`i.date()` accepts dates as either a numeric timestamp (in milliseconds), or an ISO 8601 string.
```
stopachka: nit: maybe we should give examples of this, as i.m.o most folks don't know what ISO 8601 is. Perhaps if they know it's like `date.toString()`, they'll get it

server/resources/migrations/36_checked_data_type.up.sql
```diff
@@ -0,0 +1,200 @@
...
+          when 'string' then
+            return is_jsonb_valid_datestring(value);
+          else
+            return false;
+        end case;
+      else
+        return data_type is null;
+    end case;
+  end;
+$$ language plpgsql immutable;
```
stopachka, line 145: I salute your plpgsql foo sir 🫡

server/src/instant/db/indexing_jobs.clj
```diff
@@ -0,0 +1,678 @@
...
+
+(defn update-work-estimate!
+  ([job next-stage]
+   (update-work-estimate! aurora/conn-pool next-stage job))
+  ([conn next-stage job]
+   (let [estimate (case (:job_type job)
+                    ("check-data-type" "remove-data-type" "index" "unique")
+                    (-> (sql/select-one
+                         conn
+                         (hsql/format {:select :%count.*
```
stopachka, line 252: TIL :%count.* :O

server/src/instant/db/indexing_jobs.clj
```diff
@@ -0,0 +1,678 @@
...
+  (loop []
+    (when-let [job-id (a/<!! chan)]
+      (try
+        (tracer/with-span! {:name "indexing-jobs/grab-job"
+                            :attributes {:job-id job-id}}
+          (if-let [job (grab-job! job-id)]
+            (process-job chan job)
+            (tracer/add-data! {:attributes {:job-not-grabbed true}})))
+        (catch Throwable t
+          (discord/send-error-async! (format "%s unexpected job error job-id=%s msg=%s"
```
stopachka, line 556: awh yeah!

server/src/instant/db/indexing_jobs.clj
```diff
@@ -0,0 +1,678 @@
...
+        @process))))
+
+(defn restart []
+  (stop)
+  (start))
+
+;; ----------------------
+;; Migrate system catalog
+
+(defn migrate-system-catalog []
```
stopachka, line 647: awwh yeah!

server/src/instant/util/async.clj
```diff
@@ -118,3 +126,53 @@
...
+         buf (if (number? buf-or-n) (a/buffer buf-or-n) buf-or-n)
+         chan (a/chan xform ex-handler)]
+     (gauges/add-gauge-metrics-fn
+      (fn [{:keys [cleanup]}]
+        (if (a-impl/closed? chan)
+          (cleanup)
+          [{:path (format "%s.%s.count"
+                          (namespace chan-name)
+                          (name chan-name))
+            :value (count buf)}
```
stopachka, line 169: TIL I had no idea you could count the buffer, to get the items that are still in it

server/src/instant/util/exception.clj
```diff
@@ -237,11 +242,40 @@
...
+               :value
+               :value-md5
+               :ea
+               :eav
+               :av
+               :ave
+               :vae
+               :created-at
+               :checked-data-type]
+              (map (fn [[start end]]
```
stopachka, line 270: awh yeah

server/test/instant/db/indexing_jobs_test.clj
```diff
@@ -0,0 +1,164 @@
+(ns instant.db.indexing-jobs-test
```
stopachka, line 1: 🫡🫡🫡

stopachka (approved):
Masterful work @dwwoelfel

client/www/pages/docs/schema.md
dwwoelfel: Updated!

client/www/components/dash/explorer/EditRowDialog.tsx
dwwoelfel: Updated!

Also, I think this fixes a potential bug where `value` is `0`, `false`, or some other falsey thing.

PR #449, opened 2024-11-08
[Docs] Add full topics example

nezaj:
Following up from https://github.com/instantdb/instant/pull/440, putting together a fuller topics example with real code that folks can run

https://github.com/user-attachments/assets/1af211f4-35e4-4c5e-bb66-de162eaa4428


dwwoelfel (approved):
LGTM!

PR #450, opened 2024-11-09
start init_experimental_v2

stopachka:
We are making bunch of upgrades to [strong_init](https://paper.dropbox.com/doc/Strong-Init-CLI-to-The-Finish-Line--CadDwG0cJfhSU5rzmONBBacwAg-qpwvypWaKiPxVO0xiYGEc). 

Here's what's new in this PR: 

## split paths to `init_experimental_v2`

Before, our codepaths for `init` and `init_experimental` were complected. This made it hard to evolve `init_experimental`, without breaking existing users. 

It also made it so we _had_ to support type inference for both the 'old' case (passing in a `Schema` type), and the new case (passing in a `Graph` object). 

Now, I've split the codepaths into an `init_experimental_v2` function. This lets us evolve this codepath without having to worry about breaking existing callers. I can also write these functions without having to deal with backwards compatibility to `init`. 

## `rooms` are now part of schema

I upgrade our `schema`, function, so instead of: 

```typescript
const graph = i.graph(
  entities,
  links,
).withRoomSchema<Rooms> 
```

You can now write: 

```typescript 
const schema = i.schema({ 
   entities, 
   links, 
   rooms
})
```

This lets us a) do runtime checks for ephemeral data too down the road, and b) keep our system consistent (it's weird that we had to define 'entities' using runtime functions, while rooms where a type you had to cludge in)

## `cardinalityInference` is now the default. 

Previously, you couldn't _just_ rely on the `Schema` type, because `db` could _also_ be set up with a `cardinalityInference` flag. This caused a few issues, where you couldn't actually figure out a query result from a schema. Instead, you needed to know the schema _and_ the value of the 'cardinalityInference' value. 

This is why we used to have the `helperTypes` namespace. For init_experimental_v2, I made it so cardinalityInference is just on by default. If you like at how we do `helperTypes` now, we can just use our underlying types directly.

## Better intellisense

I jigged up how we do types, so the intellisense starts to look a lot better. For example:  

**Before** 

CleanShot 2024-11-12 at 17 25 50@2x

CleanShot 2024-11-12 at 17 26 11@2x

CleanShot 2024-11-12 at 17 26 35@2x

**After** 
CleanShot 2024-11-12 at 17 26 50@2x

CleanShot 2024-11-12 at 17 27 12@2x

CleanShot 2024-11-12 at 17 27 23@2x

## Still to do: 

I need to run through a bunch of runtime tests, to make sure react, react-native, admin, core, are working on our examples.

## Follow-on PRs

We can land this as is, but before we launch, we have a few more things to do: 
1. I want to add a `$once` modifier to queries
2. I want to add more typescript tests, perhaps inside the packages 
3. I need to update our cli, so we generate an `i.schema` from pull, instead of i.graph 
4. I need to make it so `schema` is optional. Right now strong_init _requires_ a schema.

@dwwoelfel @nezaj


client/packages/core/src/schemaTypes.ts
```diff
@@ -55,11 +45,37 @@ export class DataAttrDef<ValueType, IsRequired extends boolean> {
...
+    ? isRequired extends true
+      ? ValueType
+      : ValueType | undefined
+    : never;
+
+export class LinkAttrDef<
+  Cardinality extends CardinalityKind,
+  EntityName extends string,
+> {
+  constructor(
```
dwwoelfel, line 59: It's kind of odd that the order of the arguments to the constructor doesn't match the order of the arguments to the type.

client/packages/react/src/useQuery.ts
```diff
@@ -89,3 +92,63 @@ export function useQuery<
...
+          ...result,
+        };
+
+        cb();
+      });
+
+      return unsubscribe;
+    },
+    // Build a new subscribe function if the query changes
+    [queryHash],
```
dwwoelfel, line 143: Don't forget to add `_core` as a dependency to the hook like in https://github.com/instantdb/instant/pull/447

client/sandbox/strong-init-vite/instant.schema.v2.ts
```diff
@@ -0,0 +1,50 @@
...
+  entities: {
+    messages: i.entity({
+      content: i.string(),
+    }),
+    $users: i.entity({
+      email: i.string().unique().indexed(),
+    }),
+  },
+  links: {
+    messageCreator: {
```
dwwoelfel, line 13: Probably out of scope, but I've always found it a little weird that you have to give your links a name.

client/sandbox/strong-init-vite/src/App.tsx
```diff
@@ -1,15 +1,23 @@
...
+const room = db.room('chat');
+
 function App() {
-  const { isLoading, data, error } = db.useQuery({ messages: {} });
+  const publishEmoji = room.usePublishTopic("emoji");
+  const presence = room.usePresence({ keys: ["name"] });
+  const user = presence.user!;
+  const peers = presence.peers!;
+  const publishPresence = presence.publishPresence;
+  ((..._args) => {})(publishEmoji, user, peers, publishPresence);
```
dwwoelfel: What does this line do?

dwwoelfel (commented):
Seems reasonable to me, but a lot of they typescript stuff is over my head.

We'll have to be extra careful to update both the `experimental` file and the regular file when making changes as long as we have two files. 

Couple of general questions:

1. How long do you expect we'll have both versions in the codebase?

2. Any thoughts about the migration path for users?

dwwoelfel:
> I want to add a $once modifier to queries

What would the `$once` modifier do?

client/packages/react/src/useQuery.ts
stopachka: Oh, nice catch!

client/sandbox/strong-init-vite/instant.schema.v2.ts
stopachka: Yeah, agreed! Will noodle on something better perhaps in follow-ons

client/sandbox/strong-init-vite/src/App.tsx
stopachka: Ah, forgot to add a note: it's just to make typescript quiet down about unused variable names. 

Will add a comment

stopachka:
> What would the $once modifier do?

My thinking is it would be analogous to a 'select one'. So the user can explicitly say: 'don't give me back an array, I want just one item'

stopachka:
> How long do you expect we'll have both versions in the codebase?

> Any thoughts about the migration path for users?

Here's the plan discussed! 

1. (Before init_experimental_v2 is launched, the codebase has it as `init_experimental_do_not_use`) 
2. Week 1: fold `init_experimental_do_not_use` _into_ `init_experimental`
     1. We make i.graph, return an i.schema-like thing, which has a `withRoomSchema` function
     2. This will be marked deprecated
3. Week 3: init_experimental becomes init
    1. Anyone who migrates from `init`, _should_ work 
    2. However, if they over-road init with at type, like `init<Schema>(...)`, they will need to convert to an i.schema. However, `instant-cli` pull should help.

client/packages/core/src/schemaTypes.ts
stopachka: gp! Re-ordered constructor

client/packages/react/src/useQuery.ts
stopachka: (Going to do this in a separate PR, as both useQueries have this, and I want to do some more sanity checking)

dwwoelfel (approved):
LGTM!

PR #451, opened 2024-11-09
Send library version info in `init`, cli requests, and admin requests

dwwoelfel:
Stores a version.js file in all of the packages that we populate as part of `make publish` with the library version.

Then we send the version info along with `init`, in the headers for cli requests, and in the headers for admin requests.

Useful for debugging.

I was planning to use it for determining whether to automatically create the checked data types when you do cli push, but now I think maybe it would be better to do just add a flag. Either way, it will be nice to have version info in the honeycomb logs.


client/packages/react-native/src/version.js
```diff
@@ -0,0 +1,3 @@
+const version = "v0.14.27-dev";
```
stopachka, line 2: nit: should we add a comment here saying it's auto-generated?

stopachka (approved):
Yaas!!

client/packages/react-native/src/version.js
dwwoelfel: Good idea!

PR #452, opened 2024-11-09
Exclude set-presence from the cloudwatch logs in prod

dwwoelfel:
Should reduce our log expenses and maybe make things a little faster.


nezaj:
Nice!

nezaj (approved):

PR #453, opened 2024-11-09
Add back in sampling

nezaj:
Looks like our Cloudwatch bill is exceeding our EC2 bill. Daniel and I noticed most of the logs are from things like `set-presence` which probably don't add a lot of value. Let's go back to filtering those out?


nezaj:
Closing in favor of https://github.com/instantdb/instant/pull/452

PR #454, opened 2024-11-11
Don't throw 500s on invalid JWT tokens

dwwoelfel:
Use `ex/throw-oauth-err!` in `verify-jwt` so that we return a 400 with a good error message instead of a 500.


stopachka (approved):
byutiful!

PR #457, opened 2024-11-11
Send refresh-presence through receive-q

dwwoelfel:
This should prevent a slow session from tying up the queue.

When we do `set-presence`, we'll put one `refresh-presence` in the receive-q for each session instead of immediately doing a broadcast. This makes `set-presence` finish much faster. If one of the sessions subscribed to the room is slow, then they won't slow down the set-presence broadcast for everyone in the room. 

We also drop `refresh-presence` events if there are multiple in the queue for a session and room, further preventing a slow session from clogging things up. Also acts as a relief valve for sessions that are getting overloaded.

cc @stopachka @nezaj

### DIFF
--- a/server/src/instant/reactive/session.clj
+++ b/server/src/instant/reactive/session.clj
@@ -31,7 +31,6 @@
    [instant.util.exception :as ex]
    [instant.util.uuid :as uuid-util]
    [instant.grouped-queue :as grouped-queue]
-   [instant.gauges :as gauges]
    [instant.reactive.receive-queue :as receive-queue :refer [receive-q]])
   (:import
    (java.util.concurrent CancellationException)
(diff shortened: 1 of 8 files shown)


server/src/instant/reactive/session.clj
```diff
@@ -586,29 +609,25 @@
...
 (defn group-fn [{:keys [item] :as _input}]
   (let [{:keys [session-id op room-id]} item]
-    (condp contains? op
-      #{:transact}
+    (case op
+      :transact
       [:transact session-id]
 
-      #{:join-room :leave-room :set-presence :client-broadcast}
+      (:join-room :leave-room :set-presence :client-broadcast)
```
stopachka, line 618: TIL :O

stopachka (approved):
Brilliant!

PR #458, opened 2024-11-12
Simplify hazelcast now that we're using receive-q for refresh-presence

dwwoelfel:
Follow up to https://github.com/instantdb/instant/pull/457.

Now that we're putting `refresh-presence` messages on the receive-q instead of broadcasting them, we can do that work directly in the map listener. It lets us remove a bunch of the core.async complexity from the hazelcast code.


stopachka (approved):
Ahh biyutiful!

PR #461, opened 2024-11-12
Quick followup for checked types PR

dwwoelfel:
1. Fixes a bug where we notify every time we check for stuck jobs instead of checking if there are stuck jobs
2. Backdates the migration to indicate the constraint is checked.


stopachka (approved):
LGTM!

stopachka (approved):
lesgoo!

PR #463, opened 2024-11-12
Remove the code for the transition to system catalog

dwwoelfel:
Removes all of the $users table shim code and all of the legacy code to transition from users in the `app_users` table to users in the triples table.

TODO after merging
 - [ ] Run migration after backend is fully deployed


stopachka (approved):
_mrclean_

PR #464, opened 2024-11-12
[strong-init] add a few typescript tests in strong-init sandbox, and fix bugs

stopachka:
I added two new files: 

- typescript_tests_normal.tsx
- typescript_tests_experimental.tsx

These go through a full flow of rooms, queries, transactions with db's in `core, react, and react-native`. 

While doing this, I realized we had a typescript error in our current implementation, for core and React. We were inferring `RoomSchema` in the wrong argument position. 

Fixed, so that rooms come with typehints again in core and react-native.

@dwwoelfel @nezaj

### DIFF
--- a/client/packages/core/src/index.ts
+++ b/client/packages/core/src/index.ts
@@ -143,12 +143,12 @@ function init_experimental<
   NetworkListener?: any,
 ): InstantCore<
   Schema,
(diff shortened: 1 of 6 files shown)


dwwoelfel (approved):
kind of cool that you can just write example code to test the typescript types

PR #465, opened 2024-11-13
[strong-init] schema helpers: fix up docs and update our strong-init sandbox

stopachka:
- Our schema helper example was really tough to read. I made a small improvement, so it at least looks a bit nicer. There's a lot more to do with the docs (i.e instead of `myTable`, using actual examples, and fixing this up in general, but for now just starting off simple with the visual improvement)
- I also started to use these examples in our strong-init sandboxes 

**Docs: before**
CleanShot 2024-11-12 at 16 27 26@2x

**Docs: after** 

CleanShot 2024-11-12 at 16 27 44@2x

@dwwoelfel @nezaj


dwwoelfel (approved):
Nice and clean!

PR #466, opened 2024-11-13
[sm][types] sm type fix for jobs

nezaj:
Was reviewing #448 and noticed a small typo for types. Not a big deal but figured would push up


dwwoelfel (approved):
Thanks!

PR #467, opened 2024-11-13
Set a default query timeout of 30 seconds

dwwoelfel:
Sets a default query timeout of 30 seconds. This should only affect admin queries because client queries coming through the websocket already have a 5 second timeout.

If the requests takes longer than 30 seconds, we'll return a 429 that looks like:

```js
> await adminDb.query({posts: {}}).catch(e => console.log(e))
{
  status: 429,
  body: {
    type: 'timeout',
    message: 'The query took too long to complete.',
    hint: null
  }
}

```


stopachka (approved):
Yass!

PR #468, opened 2024-11-13
[Docs] Pagination and order limited to top-level namespace

nezaj:
Adding a note so folks aren't surprised when `limit/offset/order/etc` don't work for nested queries


dwwoelfel (approved):
LGTM!

PR #469, opened 2024-11-14
Make broadcast work with multiple instances

dwwoelfel:
This updates broadcast to work across multiple instances.

Some of the room's sessions may be on other physical machines when we broadcast a message to everybody in the room. We use a hazelcast topic to notify those other sessions that they need to broadcast the message.

Two changes to make this work:
1. Similar to https://github.com/instantdb/instant/pull/457, we send `:server-broadcast` events through the receive-q instead of doing a single `rs/try-broadcast-event!`
2. In `:client-broadcast` we first notify any sessions that are on the same machine, then we send a message through a hazelcast topic with the list of remote session ids and the data to broadcast. The listener on the topic will put a `:server-broadcast` event on the `receive-q` for the sessions that it has locally.

There is a new [http://localhost:4000/play/hazelcast-broadcast](http://localhost:4000/play/hazelcast-broadcast) playground to make it easy to test.

### DIFF
--- a/client/sandbox/react-nextjs/pages/play/hazelcast-broadcast.tsx
+++ b/client/sandbox/react-nextjs/pages/play/hazelcast-broadcast.tsx
@@ -1,0 +1,74 @@
+import React, { useState } from "react";
+import Head from "next/head";
+import config from "../../config";
+
+let __id = 0;
+function nextId() {
+  return ++__id;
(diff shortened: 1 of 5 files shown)


nezaj (approved):
Woohoo! Love the examples!

PR #472, opened 2024-11-14
Index process

dwwoelfel:
Uses the background indexing jobs for adding/removing index and uniqueness.

Here's what setting and removing `indexed` from the explorer looks like

https://github.com/user-attachments/assets/1926295e-20d3-4842-9455-2ac6c36b5808

Uniqueness is similar, with helpful errors if the attribute has duplicate values. Clicking on the "multiple entities with value X" link will go to the explorer with all entities with that value present.

https://github.com/user-attachments/assets/fbe82ed9-42ec-44a2-b8ca-ef370d287782

If you have data that's too big to index, you'll also get an error:

https://github.com/user-attachments/assets/49b78e56-c51b-451b-954e-b8ad1c9eedf3

Updating from the cli is similar

https://github.com/user-attachments/assets/a261a193-2d33-479c-a8ee-f8bc0a6c413b

TODO before deploying:
- [ ] Run migration

### DIFF
--- a/client/www/components/dash/explorer/EditRowDialog.tsx
+++ b/client/www/components/dash/explorer/EditRowDialog.tsx
@@ -284,6 +284,7 @@ export function EditRowDialog({
                     tabIndex={tabIndex}
                     value={value}
                     options={[
+                      { value: '', label: '-' },
                       { value: 'false', label: 'false' },
                       { value: 'true', label: 'true' },
                     ]}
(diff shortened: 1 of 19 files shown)


client/www/components/dash/explorer/EditRowDialog.tsx
```diff
@@ -284,6 +284,7 @@ export function EditRowDialog({
                     tabIndex={tabIndex}
                     value={value}
                     options={[
+                      { value: '', label: '-' },
```
dwwoelfel, line 287: This lets us distinguish between `false` and `null`. Without this change you can't set a value to `false` from the edit row dialog.

nezaj: Nice!

nezaj (approved):
LGTM! Really cool to see how to add more types of jobs!

PR #473, opened 2024-11-14
[strong-init-v2] make schema optional

stopachka:
Previously, strong-init _required_ that a `schema` was passed in. This now makes `schema` optional. 

There's two benefits to this. First, users will still be able to hack as soon as they install instant, without having to define a schema first. Second, it makes upgrading from normal init easier: if they don't define a schema, the behavior is the same as old init.

There's some cool typescript trickery going on, but the big insight was to default our `Schema` type parameter, to an 'UnknownSchema`, which had the correct settings for links. More on this in the [stackoverflow discussion](https://stackoverflow.com/questions/79187112/what-is-the-idiomatic-way-to-make-a-generic-type-optional)

@dwwoelfel @nezaj

### DIFF
--- a/client/packages/admin/src/index.ts
+++ b/client/packages/admin/src/index.ts
@@ -39,7 +39,7 @@ import {
   type ResolveAttrs,
   type ValueTypes,
   type DoNotUseInstantSchema,
-  DoNotUseUnknownSchema,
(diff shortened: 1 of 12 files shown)


dwwoelfel (approved):
LGTM!

PR #476, opened 2024-11-14
Add an lru cache for an app's rules

dwwoelfel:
Similar to https://github.com/instantdb/instant/pull/373, caches rules so that we don't have to fetch them repeatedly.

They are automatically invalidated when we see the rules get updated in the wal and we also manually invalidate them when we make changes to the rules.


stopachka (approved):
Nice!

PR #481, opened 2024-11-15
Faster init through better caching

dwwoelfel:
Improves the caching of attrs and adds caching to `app/get-by-id` and `instant_users/get-by-app-id`.

When all is cached, reduces the number of queries we run for `:init` from 2 to 0 when the user isn't logged in, and 4 to 1 when we need to look up the user by their auth token.

We were fetching attrs a lot more often than we needed to because the `query-op` gets passed a conn-pool that it uses when it calls `attrs/get-by-app-id`. That was causing us to bypass the cache. With this PR we'll stop bypassing the cache if the passed in conn is `aurora/conn-pool`.

We fetch the app creator by the app id, so the cache needs to be evicted when either the app's creator changes or the instant_user changes. This PR introduces a novel lru cache that allows you to evict based on data in the cached item. The main cache key is the app_id, but the cache registers the user_id as an alternative eviction key, allowing you to evict by either the app_id or the user_id.

### DIFF
--- a/server/src/instant/util/cache.clj
+++ b/server/src/instant/util/cache.clj
@@ -71,14 +71,13 @@
                          limit))
 
   MultiEvictLRUCacheImpl
-  (evict-impl [this key]
+  (evict-impl [_ key]
     (MultiEvictLRUCache. (dissoc cache key)
                          (dissoc lru key)
                          (disj-in mapping
                                   [(mapping-fn (get cache key))]
                                   key)
                          mapping-fn
-                         ;; XXX: Only tick once??
                          (inc tick)
                          limit))
 
--- a/server/src/instant/util/cache.clj
+++ b/server/src/instant/util/cache.clj
@@ -7,14 +7,13 @@
   [base start-at]
   (into (priority-map/priority-map) (for [[k _] base] [k start-at])))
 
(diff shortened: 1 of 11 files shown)


server/src/instant/util/cache.clj
```diff
@@ -0,0 +1,106 @@
...
+                                  key)
+                         mapping-fn
+                         (inc tick)
+                         limit))
+
+  Object
+  (toString [_]
+    (str cache \, \space lru \, \space mapping \, \space tick \, \space limit)))
+
+(defn multi-evict-lru-cache-factory
```
stopachka, line 87: 🫡🫡🫡

server/src/instant/util/cache.clj
```diff
@@ -0,0 +1,106 @@
...
+                             (assoc lru item tick+)
+                             lru)
+                           mapping
+                           mapping-fn
+                           tick+
+                           limit)))
+  (miss [_ item result]
+    (let [tick+ (inc tick)]
+      (if (>= (count lru) limit)
+        (let [k (if (contains? lru item)
```
nezaj, line 38: @dwwoelfel curious, for the miss case why do we check whether the item is in the cache? Could we just make` k (first (peek lru))`?

PR #482, opened 2024-11-15
Run receive worker in a virtual thread

dwwoelfel:
I'm worried about how many threads we're creating to run the receive workers.

This PR runs the workers in virtual threads.

We already perform the `handle-receive` work in a virtual thread, so this should be fine. The new stuff that happens in a virtual thread will just be the `.poll` on the LinkedBlockingQueue.


stopachka (approved):
Love it!

PR #483, opened 2024-11-15
[do-not-review] instant-cli v2

stopachka:


stopachka:
landed these changes separately

PR #484, opened 2024-11-15
Initialize presence on join room

dwwoelfel:
When we call `useSyncPresence` in the typing indicator example, we're dropping the `publishPresence` command because `this._presence` is empty. Now we set it so that we'll call the initial setPresence after joining the room.

Before the grouped queue improvements, this might have raced on the server and caused a "you haven't joined the room yet" error. But now that we handle events in order, it should be fine.

Before:

Screenshot 2024-11-15 at 10 11 35 AM

After:

Screenshot 2024-11-15 at 10 09 33 AM

### DIFF
--- a/client/packages/core/src/Reactor.js
+++ b/client/packages/core/src/Reactor.js
@@ -1535,6 +1535,8 @@ export default class Reactor {
       };
     }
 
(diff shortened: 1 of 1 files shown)


nezaj (approved):
Nice!

PR #485, opened 2024-11-15
[Dash] Show correct billing for collaborators

nezaj:
We enabled showing the billing tab to collaborators and above but needed to update a few more places

* Show the billing tab on the left
* Fetch subscription based on just app id, no need for user info
* Show friendly error message for portal link as well


dwwoelfel (approved):
LGTM!

PR #486, opened 2024-11-15
[init] merge into init_experimental

stopachka:
This PR: 

Promotes`do_not_use_init_experimental` PR into the latest `init_experimental`. 

The main changes:
- `schema` is now optional in `init_experimental`
  - This makes upgrading from `init` to `init_experimental` much easier
- `cardinalityInference` is no longer an option
  - If `schema` is provided, cardinality inference is on. If it is not, cardinality inference is off
- Better intellisense
  - We are more careful about how we write our types, so the intellisense looks better

**Will strong-init users experience a breaking change?**

`i.graph` returns an InstantSchemaDef now, so they _will not_ experience breaking changes from a different `graph` type. 

However, if some users wrote `init_experimental({ schema, cardinalityInference: false })`, we will not respect the cardinality inference. In this case, typescript will let them know though, and they will need to fix up callsites. This is unlikely to happen, as afaik most users of strong init _use_ cardinality inference. 

**Follow-on PRs**

- I will update instant-cli, so we write an example `i.schema`, instead of `i.graph`. 
- Once I do that, I will make i.graph deprecated
- And with that, I'll launch strong_init updates!

@dwwoelfel @nezaj

### DIFF
--- a/client/packages/admin/src/index.ts
+++ b/client/packages/admin/src/index.ts
@@ -16,14 +16,15 @@ import {
 
   // query types
   type QueryResponse,
-  type DoNotUseQueryResponse,
-  type InstaQLQueryParams,
+  type InstaQLResponse,
+  type InstaQLParams,
   type InstantQuery,
   type InstantQueryResult,
   type InstantSchema,
   type InstantSchemaDatabase,
   type InstantObject,
   type InstantEntity,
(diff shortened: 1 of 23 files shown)


dwwoelfel (approved):
LGTM!

PR #487, opened 2024-11-15
defsql

dwwoelfel:
Creates a new `defsql` macro to centralize all of the boilerplate we add for each sql fetching function.

Before:
```clj
(defn select
  [conn query]
  (tracer/with-span! {:name "sql/select"
                      :attributes (span-attrs conn query)}
    (with-translating-psql-exceptions
      (io/tag-io
        (sql/query conn query {:builder-fn rs/as-unqualified-maps
                               :timeout *query-timeout-seconds*})))))
```

After:
```clj
(defsql select sql/query {:builder-fn rs/as-unqualified-maps})
```

[Direct link to the changes to sql.clj](https://github.com/instantdb/instant/compare/defsql?expand=1#diff-4c3638c4ecdec4d03b7ae656618dabf60e715e115daa42b50a9b27c2cf2fe928)

Also adds an optional prefix argument to all of the `sql/*` functions that takes a tag to put in the span. Example:

```clojure
(sql/select ::get-apps conn ["select * from apps"])
```

```log
[98c3] 1ms [sql/select] query_tag=:apps/get-apps
```

This will help us determine which queries are slow and which queries are running.

I tagged a bunch of the existing queries, but not all of them.

### DIFF
--- a/server/.clj-kondo/hooks/defsql.clj
+++ b/server/.clj-kondo/hooks/defsql.clj
@@ -1,0 +1,22 @@
+(ns hooks.defsql
+  (:require [clj-kondo.hooks-api :as api]))
+
+
+(defn defsql [{:keys [node]}]
+  (let [[name query-fn] (rest (:children node))
+        conn-token (api/token-node 'conn)
+        query-token (api/token-node 'query)
+        tag-token (api/token-node '_tag)
+        new-node (api/list-node
+                  (list
+                   (api/token-node 'defn)
+                   name
+                   (api/list-node
+                    (list
+                     (api/vector-node [conn-token query-token])
+                     (api/list-node (list query-fn conn-token query-token))))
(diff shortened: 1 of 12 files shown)


server/.clj-kondo/hooks/defsql.clj
```diff
@@ -0,0 +1,22 @@
+(ns hooks.defsql
+  (:require [clj-kondo.hooks-api :as api]))
+
+
+(defn defsql [{:keys [node]}]
```
nezaj, line 5: Neat! Curious, what made you want to add this in as opposed to some of our other macros?

dwwoelfel: There's a lot of boilerplate and I'm adding more with the graceful query shutdown.

PR #489, opened 2024-11-16
[strong-init] support shapes for backwards compatibility

stopachka:
## Context

In normal init, folks could previously provide types as shapes, like so: 

```typescript
const db = init<Schema, Rooms>(...)
```

With `init_experimental`, this will no longer work. The current solutions are either to go untyped: 

```typescript
const db = init_experimental({ appId } );
```

Or, to upgrade to the schema object: 

```typescript
ocnst db = init({ appId, schema })
```

### Problem 

However, it can be cumbersome to create this schema object. When we upgrade users to strong init, it would suck to force them into a breaking change. 

### Solution 

I added a helper type, called `BackwardsCompatibleSchema`. This takes in the old `Schema` and `Rooms` shape, and creates an `InstantSchemaDef` type, which behaves _just_ as the Rooms and schema types did before. 

@dwwoelfel @nezaj

### DIFF
--- a/client/packages/admin/src/index.ts
+++ b/client/packages/admin/src/index.ts
@@ -24,6 +24,7 @@ import {
   type InstantSchemaDatabase,
   type InstantObject,
   type InstantEntity,
+  type BackwardsCompatibleSchema,
   type IInstantDatabase,
 
   // schema types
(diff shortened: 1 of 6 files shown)


dwwoelfel (approved):
LGTM!

PR #490, opened 2024-11-16
Graceful shutdown of queries on timeout

dwwoelfel:
This fixes a problem where the postgres connection gets killed when we call `future-cancel` on handle-receive timeouts. 

If the database is overloaded and it causes a bunch of handle-receives to time out, we can get into a feedback loop where canceling the query causes the connection to get killed, putting additional wear on the database when we recreate the connection.

With this PR, we keep track of the in-progress queries and cancel them before we call `future-cancel` on the handler.

There's a dynamic var `*in-progress-stmts*` that you can bind to an atom with a set in it. If bound, the `defsql` handler will put a `Cancelable` in the set while the query is active. You can call `sql/cancel-in-progress` to cancel all of the in-flight queries. We do just that in `session/handle-receive` when the handler times out. Then we can call `future-cancel` without killing our connection.

### DIFF
--- a/server/src/instant/jdbc/sql.clj
+++ b/server/src/instant/jdbc/sql.clj
@@ -1,28 +1,23 @@
 (ns instant.jdbc.sql
   (:require
-   [next.jdbc :as next-jdbc]
-   [next.jdbc.sql :as sql]
-   [next.jdbc.result-set :as rs]
-   [next.jdbc.prepare :as p]
-   [next.jdbc.connection :as connection]
    [clojure.string :as string]
+   ;; load all pg-ops for hsql
+   [honey.sql.pg-ops]
+   [instant.util.exception :as ex]
    [instant.util.io :as io]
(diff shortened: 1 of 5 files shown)


server/src/instant/jdbc/sql.clj
```diff
@@ -152,11 +173,36 @@
...
+                    (~query-fn ps# nil opts#))
+                  (finally
+                    ;; Don't close the connection if a java.sql.Connection was
+                    ;; passed in, or we'll end transactions before they're done.
+                    (when create-connection?#
+                      (.close c#))))))
             (catch PSQLException e#
               (throw (ex/translate-and-throw-psql-exception! e#)))))))))
 
+(defn next-do-execute!
```
dwwoelfel: Created an issue on the `next-jdbc` repo: https://github.com/seancorfield/next-jdbc/issues/287

nezaj (approved):
Nice!

PR #491, opened 2024-11-16
Use cluster size in refinery sampler

dwwoelfel:
It seems like the default is false https://docs.honeycomb.io/manage-data-volume/sample/honeycomb-refinery/sampling-methods/#useclustersize

This might help us keep under the rate limit.

### DIFF
--- a/server/test/instant/db/indexing_jobs_test.clj
+++ b/server/test/instant/db/indexing_jobs_test.clj
@@ -242,7 +242,7 @@
                                            :index? false
                                            :value-type :blob
                                            :cardinality :one}]])
-              _ (dotimes [x 20]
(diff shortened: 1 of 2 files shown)


server/test/instant/db/indexing_jobs_test.clj
```diff
@@ -242,7 +242,7 @@
                                            :index? false
                                            :value-type :blob
                                            :cardinality :one}]])
-              _ (dotimes [x 20]
+              _ (dotimes [x 10]
```
dwwoelfel, line 245: Just makes the test less prone to timing out.

stopachka (approved):
SGTM!

PR #497, opened 2024-11-19
Nested future-cancel

dwwoelfel:
Followup to https://github.com/instantdb/instant/pull/490

When we call `future-cancel` on events that timeout in `handle-receive`, we can see child spans that continue long after the parent has been canceled and completes.

My hypothesis is that the child futures we create aren't being canceled when we cancel the parent future and that's what is causing the child spans to continue working. This leads to the system getting overwhelmed because we just keep piling up canceled work in the background.

A small test that demonstrates the problem:

```clojure
(let [signal (atom nil)
      go-ahead (promise)
      v (future @(future @go-ahead (reset! signal :whoops!)))]
  (Thread/sleep 50)
  (future-cancel v)
  (Thread/sleep 50)
  (deliver go-ahead true)
  (Thread/sleep 50)
  (is (thrown? java.util.concurrent.CancellationException @v))
  (is (= :whoops! @signal))))
```

This PR adds tracking of the `vfuture`s created as direct descendents in a `vfuture` and will cancel any child `vfuture` before canceling itself when `future-cancel` is called on it.

One limitation is that we lose track of the child futures if we don't deref them in the parent (e.g. `(vfuture (vfuture @go-ahead (reset! signal :whoops!)))` won't cancel the inner future), but that seems fine to me. If you wanted it to be canceled, you should have kept track of it.

We don't keep track of children in `vfut-bg` since it will spawn a lot of child futures and we don't normally cancel them.

I was seeing thread pinning locally when we called `.close` on the connection in `sql/cancel-in-progress`, so I also updated that to only close the connection if we didn't open the connection in the body of `defsql`. The connection will get closed in the `finally` clause--that happens in the same thread, which seems to prevent the pinning issue.

### DIFF
--- a/server/src/instant/jdbc/sql.clj
+++ b/server/src/instant/jdbc/sql.clj
@@ -138,23 +138,32 @@
 
 (def ^:dynamic *in-progress-stmts* nil)
 
-(defn cancel-in-progress [stmts]
-  (doseq [stmt stmts]
-    (.cancel stmt)))
-
 (defprotocol Cancelable
   (cancel [this]))
+
+(defn cancel-in-progress [stmts]
+  (doseq [stmt stmts]
+    (cancel stmt)))
 
 (defn register-in-progress
   "Registers the statement in the in-progress set (if we're tracking it)
   and returns a closeable that will remove the statement from the set at
   the end of the query."
-  [conn stmt]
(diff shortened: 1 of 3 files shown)


stopachka (approved):
Beautiful!

PR #498, opened 2024-11-19
Put errors back into the receive queue

dwwoelfel:
This should prevent a session that is causing a lot of errors from backing up the receive queue.

When we get an error in handle-receive, instead of sending the error over the websocket immediately, we'll put the error in the receive queue. We use the grouped-queue to handle those one at a time so that any session having trouble receiving messages won't block the entire queue.

Other things that would be nice here:
  1. Batching messages to the client
  2. Some way to disconnect sessions that are showing lots of errors.

### DIFF
--- a/server/src/instant/reactive/session.clj
+++ b/server/src/instant/reactive/session.clj
@@ -223,6 +223,25 @@
                      :tx-id tx-id
                      :client-event-id client-event-id})))
 
+;; -----
+;; error
+
+(defn handle-error! [store-conn sess-id {:keys [status
+                                                client-event-id
+                                                original-event
+                                                type
+                                                message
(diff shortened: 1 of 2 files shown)


server/src/instant/reactive/session.clj
```diff
@@ -563,24 +589,27 @@
 
     (grouped-queue/inflight-queue-reserve 1 inflight-q)))
 
+(defn start-receive-worker [store-conn eph-store-atom receive-q stop-signal id]
```
dwwoelfel, line 592: Extracted this into a function so that it would be easier to use inside of the session tests.

nezaj (approved):
Nice!

PR #499, opened 2024-11-19
[client][strong-init] use schema to help our attr generation

stopachka:
Previously, if a user creates an object that didn't exist before, we would make our best guess and create the object for them. 

This was useful, but would sometimes create surprising results if you used non-standard links. For example, for `posts.owner`, our guess would assume that `owner` was the namespace. 

Now that we pass in a schema in strong init, we can use the schema to help user create missing attrs. 

For missing object attrs:
1. We use the `schema` to decide if they need to be indexed or unique 

For missing ref attrs: 
1. We use the schema to set the forward identity, reverse identity, cardinality, indexed, unique 

If a missing attr is _not_ in the schema:

This is likely a mistake, and we throw a validation error

@dwwoelfel @nezaj

### DIFF
--- a/client/sandbox/react-nextjs/pages/play/missing-attrs.tsx
+++ b/client/sandbox/react-nextjs/pages/play/missing-attrs.tsx
@@ -1,0 +1,197 @@
+import { useEffect, useState } from "react";
+import config from "../../config";
+import { init, init_experimental, tx, id, i } from "@instantdb/react";
+import { useRouter } from "next/router";
+
+const schema = i.graph(
+  {
+    comments: i.entity({
+      slug: i.string().unique().indexed(),
+    }),
+    $users: i.entity({
+      email: i.string().unique().indexed(),
+    }),
+  },
+  {
+    commentAuthors: {
+      forward: {
+        on: "comments",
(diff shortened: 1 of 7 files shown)


client/sandbox/react-nextjs/pages/play/missing-attrs.tsx
```diff
@@ -0,0 +1,197 @@
...
+    });
+    return unsub;
+  });
+
+  return (
+    <div>
+      <div>
+        <button
+          className="bg-black text-white m-2 p-2"
+          onClick={() =>
```
dwwoelfel, line 50: Could you add a button that tries to add an attribute that doesn't exist in the schema? Curious to see what the error looks like in the sandbox.

dwwoelfel (approved):
Love this!

client/sandbox/react-nextjs/pages/play/missing-attrs.tsx
stopachka: CleanShot 2024-11-19 at 15 00 03@2x

Adding! Looks like we have a try catch

PR #500, opened 2024-11-19
[admin-sdk][strong-init] throw on missing attrs 

stopachka:
Previously, we would do our best guess to generate attrs, when a transact came in. 

However, in the case of strong init, it doesn't really make sense to do that. The user already has an explicit schema. 

In this case, we throw an error, letting the user know that the there was an unknown attribute

@dwwoelfel @nezaj

### DIFF
--- a/server/src/instant/admin/model.clj
+++ b/server/src/instant/admin/model.clj
@@ -330,8 +330,20 @@
       (create-lookup-attrs ops)
       (create-attrs-from-objs ops)))
 
-(defn transform [attrs steps]
+(defn transform [{:keys [attrs throw-on-missing-attrs?] :as _ctx} steps]
   (let [{attrs :attrs add-attr-tx-steps :add-ops} (create-missing-attrs attrs steps)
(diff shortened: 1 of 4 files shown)


server/src/instant/admin/model.clj
```diff
@@ -330,8 +330,20 @@
...
+            (let [ident-names (->> add-attr-tx-steps
+                                   (map (comp
+                                         #(string/join "." %1)
+                                         attr-model/ident-name
+                                         :forward-identity
+                                         second)))]
+              (ex/throw-validation-err!
+               :steps
+               steps
+               [{:message "Attributes are missing in you schema"
```
dwwoelfel:
```suggestion
               [{:message "Attributes are missing in your schema"
```

server/src/instant/admin/routes.clj
```diff
@@ -139,7 +139,11 @@
 ;; Transact
 
 (defn transact-post [req]
+  (tool/def-locals!)
```
```suggestion
```

dwwoelfel (approved):
Couple of comments, but LGTM!

Issue #496, opened 2024-11-19
bug with error message using cli


uriva:
```
file:///home/uri/.npm/_npx/2a2cb36dc5f602ce/node_modules/instant-cli/index.js:1098
    console.error(noAppIdErrorMessage);
                  ^

ReferenceError: Cannot access 'noAppIdErrorMessage' before initialization
    at getAppIdWithErrorLogging (file:///home/uri/.npm/_npx/2a2cb36dc5f602ce/node_modules/instant-cli/index.js:1098:19)
    at pushSchema (file:///home/uri/.npm/_npx/2a2cb36dc5f602ce/node_modules/instant-cli/index.js:584:23)
    at Command.<anonymous> (file:///home/uri/.npm/_npx/2a2cb36dc5f602ce/node_modules/instant-cli/index.js:72:5)
    at Command.listener [as _actionHandler] (/home/uri/.npm/_npx/2a2cb36dc5f602ce/node_modules/commander/lib/command.js:542:17)
    at /home/uri/.npm/_npx/2a2cb36dc5f602ce/node_modules/commander/lib/command.js:1502:14
    at Command._chainOrCall (/home/uri/.npm/_npx/2a2cb36dc5f602ce/node_modules/commander/lib/command.js:1386:12)
    at Command._parseCommand (/home/uri/.npm/_npx/2a2cb36dc5f602ce/node_modules/commander/lib/command.js:1501:27)
    at /home/uri/.npm/_npx/2a2cb36dc5f602ce/node_modules/commander/lib/command.js:1265:27
    at Command._chainOrCall (/home/uri/.npm/_npx/2a2cb36dc5f602ce/node_modules/commander/lib/command.js:1386:12)
    at Command._dispatchSubcommand (/home/uri/.npm/_npx/2a2cb36dc5f602ce/node_modules/commander/lib/command.js:1261:25)

Node.js v22.7.0
```

stopachka:
Fixed in 0.15.4 with #502 ! Thank you @uriva

PR #502, opened 2024-11-20
[cli] Hoist up `noAppIdErrorMessage`

stopachka:
See: https://github.com/instantdb/instant/issues/496

when instant couldn't find the app id, we would not actually show the error message. Instead, we'd throw a js error: 

```bash
file:///Users/stopa/projects/instant/client/packages/cli/index.js:1106
      noAppIdErrorMessage
      ^

ReferenceError: Cannot access 'noAppIdErrorMessage' before initialization
    at getAppIdWithErrorLogging (file:///Users/stopa/projects/instant/client/packages/cli/index.js:1106:7)
    at pushSchema (file:///Users/stopa/projects/instant/client/packages/cli/index.js:584:23)
    at Command.<anonymous> (file:///Users/stopa/projects/instant/client/packages/cli/index.js:72:5)
```

Hoisting it up so we actually show the error 

@dwwoelfel @nezaj

### DIFF
--- a/client/packages/cli/index.js
+++ b/client/packages/cli/index.js
@@ -23,6 +23,13 @@ const verbose = Boolean(process.env.INSTANT_CLI_VERBOSE);
 
 // consts
 
+const noAppIdErrorMessage = `
+No app ID found.
+Add \`INSTANT_APP_ID=<ID>\` to your .env file.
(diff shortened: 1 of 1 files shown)


dwwoelfel (approved):
Good fix!

PR #505, opened 2024-11-20
[cli] support global options

stopachka:
### Context

`commander` has some rough support for global options, but they work in a weird way. Once you enable global options, this is how they work: 

**Calling from root: global options show up as local options**

If you write `npx instant-cli -h`, global options will show up as `options`: 

CleanShot 2024-11-20 at 13 18 50@2x

This is a bit confusing, but happens because in the context of the root command, these options are _not_ global. 

**Calling from a subcommand, all _but_ `-h` show up as global options**: 

If you wrote `npx instant-cli login -h`, you would see: 

CleanShot 2024-11-20 at 13 20 52@2x

Here, it _almost_ works as we'd like. Except, the `help` command shows up under `options`, rather than `global options`. 

### Change

It was a bit annoying that in `npx instant-cli -h`, the "options" section was prominently the first thing. These were all global flags, and likely not what the user was most interested in: the user would be most interested in the commands they can run.

So, I enabled Global Options. But to do this, I started to hack the library a bit, so: 

a. The global options showed up at the bottom 
b. The global options showed a consistent list, no matter what command we called.

To do this, 
1. I added a `globalOption`, `getLocalAndGlobalOptions` helper, which returns a consistent list for us 
2. Overrode `formatHelp`, so we could use this function to load the options. 

### After 

CleanShot 2024-11-20 at 13 26 12@2x

CleanShot 2024-11-20 at 13 26 30@2x

@dwwoelfel @nezaj

### DIFF
--- a/client/packages/cli/index.js
+++ b/client/packages/cli/index.js
@@ -166,6 +166,11 @@ function globalOption(flags, description, argParser) {
     opt.argParser(argParser);
   }
   // @ts-ignore
+  // __global does not exist on `Option`, 
+  // but we use it in `getLocalAndGlobalOptions`, to produce 
+  // our own custom list of local and global options.
+  // For more info, see the original PR:
+  // https://github.com/instantdb/instant/pull/505
   opt.__global = true;
   return opt;
 }
--- a/client/packages/cli/index.js
+++ b/client/packages/cli/index.js
@@ -6,7 +6,7 @@ import { join } from "path";
 import { randomUUID } from "crypto";
 import dotenv from "dotenv";
 import chalk from "chalk";
-import { program } from "commander";
+import { program, Option } from "commander";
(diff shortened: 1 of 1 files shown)


client/packages/cli/index.js
```diff
@@ -57,14 +58,129 @@ program.addHelpText("after", helpFooterChalk);
...
+  showGlobalOptions: true,
+  formatHelp,
+});
+
+function globalOption(flags, description, argParser) {
+  const opt = new Option(flags, description);
+  if (argParser) {
+    opt.argParser(argParser);
+  }
+  // @ts-ignore
```
dwwoelfel, line 168: Might be nice to note that we need the ts-ignore because we're using a custom field of our own to signify global options.

stopachka: _makes salute_ will do!

PR #507, opened 2024-11-20
[cli] single push / pull command

stopachka:
Previously, We had 6 total commands for push and pull: 

- `push-schema`: push _just_ the schema
- `push-perms`: push _just_ the perms
- `push`: push _both schema and perms
- `pull-schema`: pull _just_ the schema
- `pull-perms`: pull _just_ the perms
- `pull`: pull _both_  perms and schema

## Change

In this PR, we merge them into two commands: 

- `push schema | perms | all, defaults to all`
- `pull schema | perms | all, defaults to all`

I also moved the `-app` override to be an option, rather than an argument. 

CleanShot 2024-11-20 at 15 28 53@2x

## What about backwards compatibility?

**Callers of push-perms, pull-schema, et al**

These commands will continue working, but with a deprecation warning message: 

CleanShot 2024-11-20 at 15 27 51@2x

**Callers of push [app-id]**

If a caller included an app-id-or-name argument which _would_ have resolved to an app id before, we will continue supporting it, but print a deprecation message: 

CleanShot 2024-11-20 at 15 29 56@2x

@dwwoelfel @nezaj

### DIFF
--- a/client/packages/cli/index.js
+++ b/client/packages/cli/index.js
@@ -38,6 +38,35 @@ const instantBackendOrigin =
   process.env.INSTANT_CLI_API_URI ||
   (dev ? "http://localhost:8888" : "https://api.instantdb.com");
 
+const PUSH_PULL_OPTIONS = new Set(["schema", "perms", "all"]);
+async function resolveBagAndAppWithErrorLogging(cmdName, arg, opts) {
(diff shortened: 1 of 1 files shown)


dwwoelfel (approved):
Appreciate the deprecation warnings!

PR #510, opened 2024-11-21
Move refinery to its own environment

dwwoelfel:
Moves refinery to its own elasticbeanstalk environment so that we get better rate limiting.

I thought making the refineries communicate via redis would cause them to share rate-limiting info, but we're increasing the number of events as we increase the number of instances.

It uses an internal load balancer so that refinery isn't accessible outside of our aws account.

If you want to make changes to refinery (e.g. bump the version), update the config in `server/refinery` and it will automatically publish a new version to eb when merging into main. Then you can deploy the change from the elasticbeanstalk UI.

Also disables logdna since we weren't using it and it's just eating up cpu+bandwidth for no reason.


server/.ebextensions/resources.config
```diff
@@ -0,0 +1,5 @@
+Resources:
+  AWSEBAutoScalingGroup:
+    Type: AWS::AutoScaling::AutoScalingGroup
+    Properties:
+      PlacementGroup: "Instant-docker-prod-placement-group"
```
dwwoelfel, line 5: I figured out how to make elasticbeanstalk put our instances in a placement group!

nezaj (approved):
Nice!

server/.ebextensions/resources.config
stopachka: Daam!

server/docker-compose.yml
```diff
@@ -1,43 +1,14 @@
...
   web:
     image: IMAGE_REPLACE_ME
-    depends_on:
-      - refinery
     environment:
       - BEANSTALK_PORT=80
       - PRODUCTION=true
-      - HONEYCOMB_ENDPOINT=http://localhost:8082
+      # Internal load balancer
+      - HONEYCOMB_ENDPOINT=http://refinery.us-east-1.elasticbeanstalk.com:8082
```
stopachka, line 8: oh cool! Curious q: how does the domain work? i.e did we nab refinery.us-east-1 ? Or can other people have this endpoint within their networks?

dwwoelfel: It's all ours--the elastic beanstalk cnames are all in one global namespace (well, a namespace per region) and you keep it as long as you don't terminate the environment.

stopachka: Heeck yeah!!

PR #512, opened 2024-11-21
[InstaQL] Add $like operator

nezaj:
As part of the storage upgrade, I found myself on a side quest to implement `$like`

https://github.com/user-attachments/assets/bf2362a6-d7d3-462e-bde9-eb665014fb93

This PR does the following:

* Adds support for `$like` in instaql on the backend
* Updates our topic logic so `$like` queries can be properly refreshed
* Adds `$like` to instaql on the frontend
* Adds a sandbox app to play with it / verify behavior
* Adds some docs and tests!

image

### DIFF
--- a/server/src/instant/db/datalog.clj
+++ b/server/src/instant/db/datalog.clj
@@ -332,13 +332,14 @@
       #{(-> v second :$isNull :attr-id)}
       '_]]
 
-    ;; Handle $like patterns by using wildcard for v
+    ;; There's probably a more clever way to invalidate against like
+    ;; values but this is a simple way for now
     (and (= :function (first v))
          (contains? (second v) :$like))
     [[idx
       (component->topic-component symbol-values :e e)
       (component->topic-component symbol-values :a a)
-      '_]] ;; Wildcard for v
+      '_]]
 
     :else
     [[idx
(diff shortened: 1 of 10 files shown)


server/src/instant/db/datalog.clj
```diff
@@ -541,6 +554,7 @@
     :function (let [[func val] (first v-value)]
                 (case func
                   :$not [[:not= :value (value->jsonb val)]]
+                  :$like [[:like [:->> :value :0] val]]
```
dwwoelfel, line 558: Can you run an explain on the query that this generates? My expectation is that it's going to have to do a sequential scan over all the triples with `attr_id = :attr-id`. If you wait for my comparison operators PR, I can make it use the string index.

dwwoelfel (commented):
We probably need to prevent using the `$like` operator on anything that isn't checked as a string and indexed to keep the load on the database down.

server/src/instant/db/datalog.clj
```diff
@@ -328,6 +331,17 @@
       '_
       #{(-> v second :$isNull :attr-id)}
       '_]]
+
+    ;; There's probably a more clever way to invalidate against like
+    ;; values but this is a simple way for now
+    (and (= :function (first v))
+         (contains? (second v) :$like))
+    [[idx
```
nezaj, line 339: I wasn't sure here if using `:ea` is better?

PR #514, opened 2024-11-21
New elastic beanstalk environment

dwwoelfel:
The old elasticbeanstalk environment is stuck in an invalid state, so this updates us to use a new one.

The new environment is already running and everything looks good.

Deployment strategy:

1. Reduce number of instances in old env to 1 and wait for connections to drain
2. Update cloudflare to use the new environment
3. Reduce number of instances in old env to 0 and wait for connections to drain
4. Terminate old environment

There will be a split brain on session state while sessions transition to the new environment, but no worse than it was when we were on a single instance.

### DIFF
--- a/server/Makefile
+++ b/server/Makefile
@@ -97,12 +97,12 @@ psql-dump:
 	pg_dump $${DATABASE_URL} -f dump.sql
 
 tail-web:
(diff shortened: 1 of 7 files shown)


stopachka (approved):
Sounds good!

PR #516, opened 2024-11-21
[small] make missing instant module an error rather than a warning

stopachka:
When doing pull, if we couldn't find Instant's module in the package.json, we would warn and continue on. 

However, if we didn't have the module name, we would likely create an erroneous schema file. We default to `@instantdb/core`, so for all react and react-native users, they would get a broken file. 

In an ideal world we would prompt them here and install instant for them, but for now just stopping with an error instead: 

image

@dwwoelfel @nezaj

### DIFF
--- a/client/packages/cli/index.js
+++ b/client/packages/cli/index.js
@@ -58,7 +58,7 @@ async function resolveBagAndAppWithErrorLogging(cmdName, arg, opts) {
   // check, once we're confident that users no longer
   // provide app ID as their first argument
   if (arg && !PUSH_PULL_OPTIONS.has(arg)) {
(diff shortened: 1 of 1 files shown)


client/packages/cli/index.js
```diff
@@ -241,11 +241,9 @@ function globalOption(flags, description, argParser) {
   return opt;
 }
 
-function deprecationWarning(oldCmd, newCmd) {
-  console.log(
-    chalk.yellow(
-      "[warning]: " + "`instant-cli " + oldCmd + "` is deprecated.",
-    ) +
+function warnDeprecation(oldCmd, newCmd) {
```
stopachka, line 244: as a small kaizen updated this to use a `warn` helper

dwwoelfel (approved):
LGTM!

PR #518, opened 2024-11-21
[RFC] Attempt to also print out query in handle-receive-timeout exception

clarkg:
timeout exception is not helpful currently:

```
 (NOBRIDGE) ERROR  Operation timed out: handle-receive {"client-event-id": null, "op": "error", "original-event": {"op": "refresh", "session-id": "673f84f9-ef42-4572-bccc-d7e85b3aa2e7", "total-delay-ms": 65, "ws-ping-latency-ms": 0}, "status": 500, "type": "operation-timed-out"}
 (NOBRIDGE) ERROR  This error comes with some debugging information. Here it is:
 {"timeout-ms": 5000}
 ```

with regards to the actual Clojure change, I need that FB macro meme "I have no idea what I'm doing". and I would suspect attempting to print `in-progress-stmts` would not actually show anything useful. But not sure how to access the original query passed in from the client

### DIFF
--- a/server/src/instant/reactive/session.clj
+++ b/server/src/instant/reactive/session.clj
@@ -476,7 +476,7 @@
                                     ;; If false, then canceling the queries let
                                     ;; the future complete before we could cancel it
                                     :future-cancel-result cancel-res}}))
-              (ex/throw-operation-timeout! :handle-receive handle-receive-timeout-ms)))
+              (ex/throw-operation-timeout! :handle-receive handle-receive-timeout-ms :in-progress-stmts in-progress-stmts)))
 
           (catch CancellationException _e
             ;; We must have cancelled this in the on-close, so don't try to do any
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: server/src/instant/reactive/session.clj [hunk 1]
```diff
@@ -476,7 +476,7 @@
                                     ;; If false, then canceling the queries let
                                     ;; the future complete before we could cancel it
                                     :future-cancel-result cancel-res}}))
-              (ex/throw-operation-timeout! :handle-receive handle-receive-timeout-ms)))
+              (ex/throw-operation-timeout! :handle-receive handle-receive-timeout-ms :in-progress-stmts in-progress-stmts)))
```
dwwoelfel, line 479: The `in-progress-stmts` is just a list closures that let use cancel database queries for an event that times out.

Created a PR here that should do what you want: https://github.com/instantdb/instant/pull/523

clarkg: @dwwoelfel perfect ty I'll close this!

PR #519, opened 2024-11-21
[cli] if no instant module found in package.json, prompt to install one

stopachka:
**Context** 
When we generate files for the user, we need some instant module to exist. This is because in schema.ts for example, we need to generate an import statement like: 

```typescript
import { i } from '@instantdb/react'
```

Previously: 
1. We only detected instantdb/react or instantdb/core. 
2. We would fail if they weren't there

This PR makes the CLI more helpful, by: 
1. Detecting react, core, react-native, and admin
2. If none are detected, prodding the user to choose one and installing it for them.

## When a module is detected:
CleanShot 2024-11-21 at 12 59 52@2x

## If no module is detected:
CleanShot 2024-11-21 at 13 01 43@2x
CleanShot 2024-11-21 at 13 01 12@2x

@dwwoelfel @nezaj

### DIFF
--- a/client/packages/cli/src/util/packageManager.js
+++ b/client/packages/cli/src/util/packageManager.js
@@ -1,3 +1,7 @@
+// Note: 
+// Extracted the main logic for `detectPackageManager` from:
+// https://github.com/vercel/vercel/blob/eb7fe8a9266563cfeaf275cd77cd9fad3f17c92b/packages/build-utils/src/fs/run-user-scripts.ts
+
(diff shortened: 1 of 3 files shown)


client/packages/cli/src/util/packageManager.js
```diff
@@ -0,0 +1,94 @@
...
+}
+
+
+export {
+  detectPackageManager,
+  getInstallCommand,
+}
+
+
+
```
dwwoelfel:
```suggestion
```

PR #522, opened 2024-11-22
Fix instaql with reverse references

dwwoelfel:
Fixes Cam's `$isNull` bug.

We weren't handling reverse references correctly. Before this PR you can't use the linked field name in the where without the `.id`:

```clojure
{bookshelves: {$: {where: {users: myUserId}}}}
```
(`bookshelves.users` is the reverse link to `users.bookshelves`).

You could leave off the `.id` in the forward direction:

```clojure
{users: {$: {where: {bookshelves: bookshelfId}}}}
```

Now both directions are supported.

Also fixes how we check `$isNull` for reverse references, copying the same strategy we use on the server.

We actually patched this same issue on the server as part of the `isNull` PR https://github.com/instantdb/instant/pull/409, but didn't think to check if it was a problem on the client.

### DIFF
--- a/client/packages/core/__tests__/src/instaql.test.js
+++ b/client/packages/core/__tests__/src/instaql.test.js
@@ -464,6 +464,47 @@ test("multiple connections", () => {
   ]);
 });
 
+test("query reverse references work with and without id", () => {
+  const stopa = query(
+    { store },
+    {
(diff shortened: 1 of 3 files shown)


stopachka (approved):
Niice!

PR #523, opened 2024-11-22
Add some debug info on refresh timeout/failure

dwwoelfel:
Adds some debug info about the queries we were executing when `:refresh` times out.

Passes a `debug-info` atom into handle-receive, then populates that atom with the queries we're going to run in `:refresh`. If we time out, then we'll put those queries in error we send to the client.

Here's what's new in the error event:
```diff
{
    "op": "error",
    "status": 500,
    "client-event-id": null,
    "original-event": {
        "op": "refresh",
        "session-id": "67401257-49b0-4f14-9fe3-2b838d83b339",
        "total-delay-ms": 0,
        "ws-ping-latency-ms": 0.463833,
+       "processed-tx-id": 1156752,
+       "instaql-queries": [
+           {
+               "colors": {}
+           }
+       ]
    },
    "type": "operation-timed-out",
    "message": "Operation timed out: handle-receive",
    "hint": {
        "timeout-ms": 5000
    }
}
```

Not sure if `original-event` is the best place to put that information.

### DIFF
--- a/server/src/instant/reactive/session.clj
+++ b/server/src/instant/reactive/session.clj
@@ -165,7 +165,7 @@
      :instaql-result instaql-result
      :result-changed? result-changed?}))
 
-(defn- handle-refresh! [store-conn sess-id _event]
+(defn- handle-refresh! [store-conn sess-id _event debug-info]
   (let [auth (get-auth! store-conn sess-id)
         app-id (-> auth :app :id)
         current-user (-> auth :user)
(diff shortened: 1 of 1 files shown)


clarkg:
thank you @dwwoelfel ! 🙏

server/src/instant/reactive/session.clj
```diff
@@ -490,8 +500,12 @@
               (cond
                 instant-ex (handle-instant-exception session
                                                      original-event
-                                                     instant-ex)
-                :else (handle-uncaught-err session original-event root-err))))
+                                                     instant-ex
```
clarkg, line 503: do you not need to pass it in to `throw-operation-timeout` above also?

clarkg: oh sorry ignore me, `handle-instant-exception` handles `throw-operation-timeout`

clarkg (approved):

PR #525, opened 2024-11-22
[strong-init][experimental] Expand

stopachka:
When we returned a query result type, it would look something like:

CleanShot 2024-11-22 at 10 02 25@2x

This is not _too_ helpful, as it doesn't actually show what the result would look like. 

I added a `Expand` type, which tries to convert this to it's most concrete form: 

CleanShot 2024-11-22 at 10 02 43@2x

This is a bit experimental, as I am not 100% sure how this will interact with other types in the real world. 

However, I thought, let's ship it and get some feedback. 

@dwwoelfel @nezaj

### DIFF
--- a/client/packages/admin/src/index.ts
+++ b/client/packages/admin/src/index.ts
@@ -42,6 +42,8 @@ import {
   type ValueTypes,
   type InstantSchemaDef,
   type InstantUnknownSchema,
+  type InstaQLEntity,
+  type InstaQLResult,
 } from "@instantdb/core";
 
 import version from "./version";
@@ -926,4 +928,6 @@ export {
   type ValueTypes,
   type InstantSchemaDef,
   type InstantUnknownSchema,
+  type InstaQLEntity,
+  type InstaQLResult,
 };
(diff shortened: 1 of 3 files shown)


client/packages/core/src/queryTypes.ts
```diff
@@ -10,6 +10,12 @@ import type {
   ResolveEntityAttrs,
 } from "./schemaTypes";
 
+type Expand<T> = T extends object
```
stopachka, line 13: I originally had a more fully-featured version of this, which handled non-objects too. See: 

https://github.com/sindresorhus/type-fest/blob/main/source/simplify-deep.d.ts

However, I think it's actually better to _just_ handle objects.[ I chose to use the 4 liner from jcatz](https://stackoverflow.com/a/57683652/1292663)

My main reason:

This maps our types. It _should_ be a no-op for TS, but the more complicated we make it, the more likely that it in fact is not a no-op.

client/packages/admin/src/index.ts
```diff
@@ -926,4 +928,6 @@ export {
   type ValueTypes,
   type InstantSchemaDef,
   type InstantUnknownSchema,
+  type InstaQLEntity,
+  type InstaQLResult,
```
stopachka, line 932: Unrelated to this PR, but I noticed we forgot to export these in admin. Thought it wasn't worth an extra PR

dwwoelfel (approved):
LGTM!

PR #526, opened 2024-11-22
Distinguish between etypes on permission checks

dwwoelfel:
Fixes a bug where we could apply the permission check for a different namespace if two objects in the different namespaces shared an eid.

Now we group the permission checks by etype and eid, ensuring that we run the correct permission check for each entity.


nezaj (approved):
Nice!

PR #528, opened 2024-11-22
[cli] Handle existing apps in `instant-cli init`

stopachka:
`instant-cli init` used to _only_ handle the case where a user wanted to create a brand new app. 

But most of the time, users already created an app in the dashboard. For this case, they would have to know that the right command would be to call `instant-cli pull` with an app id param. 

Instead, I updated the `instant-cli init` flow, so we ask the user if they want to create a new app, or import an existing one. 

**Future work**
- In `instant-cli pull`, if no app id is provided, we should just prompt the user for an app id, like here
- In `instant-cli init`, we should try to detect the app id. If we do, we should ask them if they want to import that app

**Here's how things look:**

### First, ask: create a new app, or import an existing one?
image

### If creating a new app, go through this flow:

image

### Or, prompt for an app id 

CleanShot 2024-11-22 at 11 30 01@2x

@dwwoelfel @nezaj

### DIFF
--- a/client/packages/cli/index.js
+++ b/client/packages/cli/index.js
@@ -342,7 +342,9 @@ program
     warnDeprecation("pull-schema", "pull schema");
     const appId = await getAppIdWithErrorLogging(appIdOrName);
     if (!appId) return;
-    pullSchema(appId);
+    const pkgAndAuthInfo = await resolvePackageAndAuthInfoWithErrorLogging();
+    if (!pkgAndAuthInfo) return;
+    pullSchema(appId, pkgAndAuthInfo);
   });
 
 // Note: Nov 20, 2024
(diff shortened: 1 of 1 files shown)


client/packages/cli/index.js
```diff
@@ -553,13 +546,51 @@ async function init() {
...
+  console.log(
+    "Great! Grab your app id from: " +
+      chalk.blueBright.underline("https://instantdb.com/dash"),
+  );
+  const _id = await input({
+    message: "What's the app ID?",
+    required: true,
+  }).catch(() => null);
+  const id = _id?.trim();
+  if (!id) {
```
nezaj, line 559: Might be worth validating the input is a uuid? I wonder if someone may have a copy/paste error and would be nice if we can tell them can't find an app id?

(Maybe the error from `pull` hints this? I think right now the error message is terse?)

stopachka: Good point! I actually have a follow on PR which removes the manual prompt here entirely: 

https://github.com/instantdb/instant/pull/532

I also check now when app id is imported from env, whether it's a UUID or not, so the user should get a good error message 🫡

nezaj (approved):
Nice!

PR #529, opened 2024-11-22
[Docs] Add ref lookup

nezaj:
Wanted to pop this in real quick. Adds docs on how to use `lookup` for traversing links

### DIFF
--- a/client/www/pages/docs/instaml.md
+++ b/client/www/pages/docs/instaml.md
@@ -250,8 +250,9 @@ If your entity has a unique attribute, you can use `lookup` in place of the id t
 ```javascript
 import { lookup } from '@instantdb/core';
 
+// Update the company name from Facebook to Meta without knowing the id
 db.transact([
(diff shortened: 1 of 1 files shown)


client/www/pages/docs/instaml.md
```diff
@@ -264,13 +265,34 @@ It can be used with `update`, `delete`, `merge`, `link`, and `unlink`.
...
+
+```javascript
+import { lookup } from '@instantdb/core';
+
+// Update a user's profile information without knowing the user's id
+// This works only if 1) the email is unique and 2) profile is a one-to-one
+// link to the user
+const email = "alyssa_p_hacker@instantdb.com";
+db.transact([
+  tx.profile[lookup('$users.email', email)].update({ favoriteLisp: 'Clojure' }),
```
dwwoelfel, line 292: Does this work? I thought we only supported `$users.id`.

nezaj: Ah I realize I misunderstood how lookup with refs work. We only support doing linked lookups by ids right now.

PR #531, opened 2024-11-22
[in-progress] Comparison operators on indexed+checked attrs

dwwoelfel:
Adds `$gt`, `$lt`, `$gte`, and `$lte` comparison operators for indexed attrs with `date` or `number` checked types.

**How it works**
We add new partial indexes on the `triples` table that extracts the typed data from the json. This is the number index for example:

```sql
create index if not exists triples_number_type_idx on triples (
    app_id,
    attr_id,
    triples_extract_number_value(value),
    entity_id
  )
  where ave and checked_data_type = 'number';
```

Our comparison query uses the index we created:

```sql
select * from triples
where app_id = :app_id
  and ave
  and attr_id = :attr_id
  and checked_data_type = 'number'
  and triples_extract_boolean_value(value) > 10;
```

```sql
->  Index Scan using triples_number_type_idx on triples  (cost=0.27..8.30 rows=1 width=143) (actual time=0.027..0.226 rows=221 loops=1)
Index Cond: ((app_id = :app-id)
              AND (attr_id = :attr-id) 
              AND (triples_extract_number_value(value) > '10'::double precision))
```

Still TODO:
- [ ] Clean up error messages a bit
- [ ] Fix topics
- [ ] Tests
- [ ] Client
- [ ] Create a PR into Joe's $like PR

TODOS before deploy:
- [ ] Manually add all of the indexes with `create index concurrently` so that we don't lock up the db when running the migration
- [ ] Run the migration

### DIFF
--- a/client/sandbox/react-nextjs/pages/play/operators.tsx
+++ b/client/sandbox/react-nextjs/pages/play/operators.tsx
@@ -1,0 +1,200 @@
+import { useEffect, useState } from "react";
+import config from "../../config";
+import { init_experimental, tx, id, i } from "@instantdb/react";
+import { useRouter } from "next/router";
+
+const schema = i.graph(
+  {
+    comments: i.entity({
+      slug: i.string().unique().indexed(),
+      someString: i.string().indexed(),
+      date: i.date().indexed(),
+      order: i.number().indexed(),
+      bool: i.boolean().indexed(),
+    }),
+    $users: i.entity({
+      email: i.string().unique().indexed(),
+    }),
+  },
+  {
+    commentAuthors: {
+      forward: {
+        on: "comments",
+        has: "one",
+        label: "author",
+      },
+      reverse: {
+        on: "$users",
+        has: "many",
+        label: "authoredComments",
+      },
+    },
+  },
+);
+
+function randInt(max: number) {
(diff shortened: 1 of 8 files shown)


client/packages/core/src/datalog.js
```diff
@@ -17,39 +17,43 @@ function matchExact(patternPart, triplePart, context) {
...
-    return context;
-  }
-  return null;
-}
-
 function matcherForPatternPart(patternPart) {
   switch (typeof patternPart) {
     case "string":
       return patternPart.startsWith("?") ? matchVariable : matchExact;
-    case "object":
```
dwwoelfel, line 44: We already do all the filtering in the store, so this saves us an extra comparison on the data.

server/test/instant/db/instaql_test.clj
```diff
@@ -57,7 +58,7 @@
      (when (seq aggregates)
        {:aggregate aggregates}))))
 
-(defn- is-pretty-eq?
+(defmacro is-pretty-eq?
```
dwwoelfel, line 61: Making this a macro helps jump-to-definition jump to the place `is-pretty-eq?` is used instead of jumping here.

stopachka (approved):
...This is beautiful!

PR #532, opened 2024-11-22
[cli] prompt for an app selector rather than pasting an id

stopachka:
Previously, we prompted for an app id. That's kind of a sadder experience though. Instead, we now fetch apps and let the user select: 

image

### DIFF
--- a/client/packages/cli/index.js
+++ b/client/packages/cli/index.js
@@ -554,18 +554,31 @@ async function createApp({ pkgDir, instantModuleName }) {
   }
 }
 
-async function promptForAppId() {
-  console.log(
-    "Great! Grab your app id from: " +
-      chalk.blueBright.underline("https://instantdb.com/dash"),
-  );
-  const _id = await input({
-    message: "What's the app ID?",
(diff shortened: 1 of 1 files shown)


client/packages/cli/index.js
```diff
@@ -554,18 +554,31 @@ async function createApp({ pkgDir, instantModuleName }) {
...
+  if (!apps.length) {
+    const ok = await "You don't have any apps. Want to create a new one?";
+    if (!ok) return;
+    await createApp(pkgAndAuthInfo);
+    return;
+  }
+  const choice = await select({
+    message: "Which app would you like to import?",
+    choices: res.data.apps.map((app) => {
+      return { name: app.title, value: app.id };
```
dwwoelfel: Maybe we ought to do ``` `${app.title} (${app.id})` ```?

dwwoelfel (approved):
LGTM!

client/packages/cli/index.js
stopachka: I like it!

PR #534, opened 2024-11-25
[cli] add a `-a` flag to init

stopachka:
This PR: 
- Adds a `--app` flag to init, so you can provide an app id upfront
- Updates how we detect app ids, so if we experience an error in some input, we exit early rather than continue with other methods
  - Previously, if an `env` app id was formatted incorrectly, we would print an error and then continue, prompting the user to create an app. But instead, i.m.o it's better to exit at that point 
- Generalizes the logic for detecting or creating an app into one function 

### Where this is headed

Now that `init` has a `--app` flag, there's a real similarity to `pull`. If you think about it, `init` is like `pull`, but it lets you create an app. 

Which begs the question: why shouldn't `pull` let you create an app, if it doesn't exist? 

I think it should. This first PR generalizes the logic for 'detecting' or 'creating' an app. In a follow-up PR, I am going to use these in `pull`

### DIFF
--- a/client/packages/cli/index.js
+++ b/client/packages/cli/index.js
@@ -286,7 +286,14 @@ program
   .option("-p --print", "Prints the auth token into the console.")
   .action(login);
 
-program.command("init").description("Create a new app").action(init);
+program
+  .command("init")
+  .description("Set up a new project.")
+  .option(
+    "-a --app <app-id>",
+    "If you have an existing app ID, we can pull schema and perms from there.",
+  )
+  .action(init);
 
 // Note: Nov 20, 2024
 // We can eventually delete this,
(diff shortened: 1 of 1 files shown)


client/packages/cli/index.js
```diff
@@ -604,18 +672,57 @@ async function init() {
...
+  appId,
+  appTitle,
+) {
+  const schema = await readLocalSchemaFile();
+  const { perms } = await readLocalPermsFile();
+
+  console.log(chalk.green(`Successfully created your Instant app "${appId}"`));
+  console.log(`Please add your app ID to your .env config:`);
+  console.log(chalk.magenta(`INSTANT_APP_ID=${appId}`));
+  console.log(chalk.underline(appDashUrl(appId)));
```
dwwoelfel, line 695: Would be nice to use terminalLink here so supported terminals would get a clickable link, e.g. https://github.com/instantdb/instant/blob/main/client/packages/cli/index.js#L832

dwwoelfel (approved):
LGTM!

client/packages/cli/index.js
stopachka: Oo this is great! I will include this in a follow-on PR, where I update every link with `terminalLink`

PR #535, opened 2024-11-25
[cli] detect _or_ create apps in push/pull commands

stopachka:
Previously: 
- If we couldn't detect an app id, we would exit and print an error.

Now:
- We use `detectOrCreateApp`, so in case we can't detect an app, we prompt the user to provide one, or create one for them.

Made these changes for `push,push-schema,push-perms,pull,pull-schema,pull-perms` 

Aside: 
- This whole `ok: true`, `withErrorLoggin` business is def a schlep. Noting for a future refactor, to use exceptions a la instant.exception on the backend  

@dwwoelfel @nezaj

### DIFF
--- a/client/packages/cli/index.js
+++ b/client/packages/cli/index.js
@@ -309,11 +309,7 @@ program
   )
   .action(async (appIdOrName, opts) => {
     warnDeprecation("push-schema", "push schema");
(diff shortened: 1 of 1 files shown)


nezaj (approved):
Nice!

PR #537, opened 2024-11-25
Fix data.ref with checked types

dwwoelfel:
Most of the changes here are to make testing against apps with checked types easier.

The change that fixes the bug is in `attr_pat` where we don't try to validate the type when we're passed in a placeholder symbol (which cel's data.ref does).


nezaj (approved):
Nice!

PR #538, opened 2024-11-25
[cli] Make sure `instant-cli` login responds to `deny`

stopachka:
Previously, there were a few bugs with our `instant-cli login` flow: 
1. Once we entered the waitForAuthToken phase, we couldn't ctrl + c out of the program
   1. This was because we had a try catch over the fetch json, which swallowed everything
1. If a user clicked 'Deny' on the dashboard, the system would not exit
  1.  This was because there was no difference between a) a missing login request, a claimed one, or an expired one

I updated our backend, so we return more descriptive errors when we are waiting for authentication. If we are simply waiting for user input, we recurse. But, if there's any other error, we now exit. 

I also removed the high-level try catch. 

@dwwoelfel @nezaj

### DIFF
--- a/client/packages/cli/index.js
+++ b/client/packages/cli/index.js
@@ -444,7 +444,7 @@ async function login(options) {
   if (!registerRes.ok) return;
 
   const { secret, ticket } = registerRes.data;
-
+  console.log("Let's log you in!");
   const ok = await promptOk(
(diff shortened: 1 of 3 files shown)


dwwoelfel (approved):
LGTM!

PR #539, opened 2024-11-25
[Storage] Add storage user

nezaj:
Still working on the storage PR but felt wanted to ship this separately to users earlier. 

Created a an IAM user specifically for S3 privileges and supply it's credentials for generating signed urls

### DIFF
--- a/server/src/instant/storage/s3.clj
+++ b/server/src/instant/storage/s3.clj
@@ -52,28 +52,33 @@
   (list-app-objects default-bucket app-id)
   (list-app-objects app-id))
 
+(defn generate-presigned-url
+  ([opts]
+   (let [access-key (config/s3-storage-access-key)
+         secret-key (config/s3-storage-secret-key)]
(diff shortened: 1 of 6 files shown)


server/src/instant/storage/s3.clj
```diff
@@ -54,19 +55,25 @@
 (defn signed-upload-url
   ([object-key] (signed-upload-url default-bucket object-key))
   ([bucket-name object-key]
-   (s3/generate-presigned-url {:method :put
-                               :bucket-name bucket-name
-                               :key object-key})))
+   (s3/generate-presigned-url
```
dwwoelfel: Could we make this optional so that people who get this working on OSS won't have to set up the access keys if they don't want to?

dwwoelfel (approved):
One comment, but LGTM!

server/src/instant/storage/s3.clj
nezaj: Done!

PR #541, opened 2024-11-25
[cli] use prettier to autoformat our exported files

stopachka:
Previously, the files we generated would be a bit off with formatting:
- Keys would be written out as strings, even rather than as symbols
- Spacing was off

Added prettier to the CLI. This makes it so we print a formatted file

@dwwoelfel @nezaj


dwwoelfel (approved):
❤️ ❤️ ❤️

PR #542, opened 2024-11-26
gzip honeycomb traces

dwwoelfel:
Compresses honeycomb traces with gzip. Should help us to transfer a lot less data between our instances. 

Tested locally talking both to the honeycomb API and a local refinery instance.

### DIFF
--- a/server/src/instant/util/tracer.clj
+++ b/server/src/instant/util/tracer.clj
@@ -12,7 +12,7 @@
    (io.opentelemetry.context Context)
    (io.opentelemetry.exporter.otlp.trace OtlpGrpcSpanExporter)
    (io.opentelemetry.sdk.resources Resource)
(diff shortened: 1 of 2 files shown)


stopachka (approved):
awwh yeah!

server/src/instant/util/tracer.clj
```diff
@@ -71,13 +88,13 @@
         default-attributes (cond-> @last-calculated-metrics
                              true (assoc "host.name" (config/get-hostname)
                                          "process-id" @config/process-id)
-                             thread (assoc SemanticAttributes/THREAD_NAME
+                             thread (assoc "thread.name"
                                            (.getName thread)
-                                           SemanticAttributes/THREAD_ID
+                                           "thread.id"
```
dwwoelfel, line 93: Removed SemanticAttributes because I couldn't get it to install with the latest versions of opentelemetry.

PR #543, opened 2024-11-26
Log how much data we're sending over websockets per app

dwwoelfel:
Adds a trace to log how much data we're sending over the websockets. This way we can get a better idea of how much different apps are contributing to network usage.

Ignores the span in prod for the console log so that we don't spike our cloudwatch costs.


stopachka (approved):
SGTM!

PR #544, opened 2024-11-26
[cli] merge all `pull` file creation into one path

stopachka:
**Context** 

Previously, we had a separate flow, where if you used `init` and chose to create an app, we would make some empty files for you. 

However, those files had a few issues: 
1. The default state was filled with data that wasn't actually in the production app. This felt a bit confusing
2. There wasn't _much_ explanation. For example, the `rules` file just had some example rules, but nothing more

We also had unique behavior when pulling an app: 
1. If there were no perms in an app, we would not create a perms file 

**This PR** 

Now, whether you 'create' an app or not, we will go through the same `pull` flow. 

In this new pull flow, we detect: 
1. If the schema does not have any entities defined, we add a comment explaining what they are 
2. If the schema does not have links defined, we add a comment explaining what they are 
3. If the perms are empty, we still make a perms file, but add a little description 

**Coming up**

I still have more to do, to make these files look better (i.schema instead of graph for schema, intellisense for perms, comparing previous files). But, we'll get to it soon 🔥

--- 

CleanShot 2024-11-25 at 17 03 50@2x

CleanShot 2024-11-25 at 17 04 07@2x

CleanShot 2024-11-25 at 17 04 21@2x

@dwwoelfel @nezaj

### DIFF
--- a/client/packages/cli/index.js
+++ b/client/packages/cli/index.js
@@ -401,7 +401,7 @@ async function push(bag, appId, opts) {
 
 function printDotEnvInfo(source, appId) {
   if (source === "imported" || source === "created") {
-    console.log(`\nPicked app ${chalk.green(appId)}! \n`);
+    console.log(`\nPicked app ${chalk.green(appId)}!\n`);
     console.log(
       `To use this app automatically from now on, update your ${chalk.green("`.env`")} file:`,
     );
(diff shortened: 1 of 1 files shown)


client/packages/cli/index.js
```diff
@@ -73,7 +63,7 @@ const PUSH_PULL_OPTIONS = new Set(["schema", "perms", "all"]);
 function convertArgToBagWithErrorLogging(arg) {
   if (!arg) {
     return { ok: true, bag: "all" };
-  } else if (PUSH_PULL_OPTIONS.has(arg.trim().lowercase())) {
+  } else if (PUSH_PULL_OPTIONS.has(arg.trim().toLowerCase())) {
```
stopachka: oi, messed this up in a previous PR, but it's not shipped to prod

client/packages/cli/index.js
```diff
@@ -564,29 +567,32 @@ async function promptImportAppOrCreateApp() {
     if (!ok) return { ok: false };
     return await promptCreateApp();
   }
+
+  apps.sort((a, b) => +new Date(b.created_at) - +new Date(a.created_at));
```
stopachka, line 571: Small change, so including here: 

Sorting apps by created at desc, so your newly created apps show first

PR #546, opened 2024-11-26
Make like use index

dwwoelfel:
Makes like use the new string index. Refuses to run `$like` if the attr isn't an indexed string.

There are 2 PRs to make it easier to see the changes. `like-merge-main` (#547) merges `main` into #512--it looks like a lot of changes, but it's just everything we've done since the PR was opened. Then this PR is based off of `like-merge-main`. 

It's probably easiest to look at the the changes in this PR, then change the base branch to `add-like` and merge it into the `add-like` branch. Then you can merge the `add-like` branch into main.


nezaj (approved):
Wooohooo!

PR #551, opened 2024-11-26
Move drop-refresh-spam to flag

dwwoelfel:
Moves [drop-refresh-spam](https://github.com/jsventures/instant/pull/1138) to a flag so we can easily turn it on and off.

### DIFF
--- a/server/flags-config/instant.perms.ts
+++ b/server/flags-config/instant.perms.ts
@@ -1,79 +1,91 @@
-export default {
-  "attrs": {
-    "allow": {
-      "create": "false"
-    }
+// Docs: https://www.instantdb.com/docs/permissions
+
+const rules = {
+  attrs: {
+    allow: {
+      create: "false",
+    },
   },
   "storage-whitelist": {
-    "allow": {
-      "view": "false",
-      "create": "false",
-      "delete": "false",
-      "update": "false"
-    }
+    allow: {
+      view: "false",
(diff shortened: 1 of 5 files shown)


server/flags-config/instant.perms.ts
```diff
@@ -1,79 +1,91 @@
...
+    allow: {
+      view: "false",
+      create: "false",
+      delete: "false",
+      update: "false",
+    },
+  },
+};
+
+export default rules;
```
stopachka, line 91: awwh yeahh : )

server/flags-config/instant.schema.ts
```diff
@@ -1,50 +1,50 @@
-// instant-config
-// https://instantdb.com/dash?s=main&t=home&app=24a4d71b-7bb2-4630-9aee-01146af26239
+// http://localhost:3000/dash?s=main&t=home&app=24a4d71b-7bb2-4630-9aee-01146af26239
```
stopachka, line 1: (Looking here, I am going to just nix this url in the generation; I don't think it's too helpful)

stopachka (approved):
niice!

server/flags-config/instant.schema.ts
dwwoelfel: It seems kind of useful to me! Just a little annoying for us since we use dev and prod.

PR #552, opened 2024-11-26
export InstantAdminDatabase

stopachka:
User reported that this type was missing. It's not longer _strictly_ necessary to export it, as all other types just take the Schema, but it's good to have. 

https://discord.com/channels/1031957483243188235/1280983629270487080/1310893036401856523

@dwwoelfel @nezaj


nezaj (approved):
🚢

PR #553, opened 2024-11-26
[Docs] Update  docs with checked type

nezaj:
What it says on the tin!

### DIFF
--- a/client/www/pages/docs/instaql.md
+++ b/client/www/pages/docs/instaql.md
@@ -870,9 +870,10 @@ console.log(data)
 
 ### $like
 
-The `where` clause supports `$like` queries that will return entities that match
-a case-insensitive substring of the provided value for the field. You can use
-`$like` to do queries like `contains`, `startsWith`, and `endsWith`.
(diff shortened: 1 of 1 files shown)


client/www/pages/docs/instaql.md
```diff
@@ -870,16 +870,19 @@ console.log(data)
 
 ### $like
 
-The `where` clause supports `$like` queries that will return entities that match
-a case-insensitive substring of the provided value for the field. You can use
-`$like` to do queries like `contains`, `startsWith`, and `endsWith`.
+The `where` clause supports `$like` on fields that are indexed with a checked `string` type.
 
-`{ $like: "%promoted!" }` looks for values that end with "promoted!".
+`$like` queries will return entities that match a case-insensitive substring of the provided value for the field. Here's how you can do queries like `startsWith`, `endsWith` and `includes`.
```
dwwoelfel: We're using `like` on the postgres side, which should be case sensitive.

nezaj: Good callout! I meant to say case-sensitive but mixed them up

dwwoelfel (approved):
LGTM!

PR #556, opened 2024-11-26
[cli] add a plan step to push perms

stopachka:
When we pushed perms, previously we would say: "This will overwrite your production perms. Ok?" 

But that's a bit scary, as you can't really tell _what_ has changed. 

In this PR, we compare production perms with what we are pushing, and show the user: 

image

@dwwoelfel @nezaj


client/packages/cli/index.js
```diff
@@ -1070,10 +1071,27 @@ async function pushPerms(appId) {
...
+  if (!diffedStr.length) {
+    console.log("There are no changes to apply.");
+    return; 
+  }
+  
+  console.log("The following changes will be applied to your perms:");
+  console.log(diffedStr);
+  
+  const okPush = await promptOk("OK to proceed?");
+  if (!okPush) return;
```
stopachka, line 1094: There is a race condition in both push schema and perms, where if someone _else_ pushed something in the meantime, we may potentially override it. 

I think we can punt this race condition for now, but perhaps we can do in the future, is to pass in some version signifier, so in case it's not the same, we error out

client/packages/cli/index.js
```diff
@@ -1070,10 +1071,27 @@ async function pushPerms(appId) {
...
+    path: `/dash/apps/${appId}/perms/pull`,
+    debugName: "Perms pull",
+    errorMessage: "Failed to pull perms.",
+  });
+
+  if (!prodPerms.ok) return;
+
+  const diffedStr = jsonDiff.diffString(prodPerms.data.perms, perms)
+  if (!diffedStr.length) {
+    console.log("There are no changes to apply.");
```
dwwoelfel: Will it be obvious to the user that we're talking about perms changes here?

dwwoelfel (approved):
One comment, but LGTM!  This will make pulling much nicer.

client/packages/cli/index.js
stopachka: Ah good point! Will update the text 🫡

PR #557, opened 2024-11-26
[cli] Add a type helper to rules

stopachka:
Previously, rules had no type information. This made it a bit annoying to write it out, because you would get no intellisense. 

I made an `InstantRules` type, which helps with autocomplete when writing out the rules. 

CleanShot 2024-11-26 at 15 51 39@2x

@dwwoelfel @nezaj


dwwoelfel (approved):
Nice! Does this also work on the dashboard?

PR #558, opened 2024-11-27
[cli] Centralize our `.env` messages, and detect the env name

stopachka:
Previously, 

If you ran `instant-cli pull schema`, and we couldn't find an env file, we would ask if the user wanted us to create one. 

This had a few issues: 
1. If the user had vite or next or svelte, the variable name may be need to be specific
2. If the user provided the app specifically -- i.e with --app, then it may not make sense to prompt the user to create a .env file -- they may not want it 
3. We also had a different section which printed information about adding newly created app ids to a .env file 

I now centralize all this, so: 
1. If the user ever creates or imports an app, we enter the .env info flow
2. We detect the ideal env name for the user (by checking for next, vite, etc in their package.json) 
3. If the user does not have a .env file, we prod to create it 
4. If the user does have a .env file, we print out a quick note

**Future considerations** 

Using .env may not be the right path for us. Mainly, nextjs for example use `.env.local`, et al.  There's a hidden feature inside instant, which checks for an `instant.config.ts`, but no one knows about it. 

We can look into standardizing that down the road, but I thought this would already be an improvement. 

---- 

**If you don't have a .env file:** 
CleanShot 2024-11-26 at 16 25 33@2x

**If you have a .env file** 
CleanShot 2024-11-26 at 16 31 05@2x

### DIFF
--- a/client/packages/cli/index.js
+++ b/client/packages/cli/index.js
@@ -51,6 +51,23 @@ const potentialEnvs = {
   vite: "VITE_INSTANT_APP_ID",
 };
 
+async function detectEnvType({ pkgDir }) {
+  const packageJSON = await getPackageJson(pkgDir);
+  if (!packageJSON) {
+    return "catchall";
+  }
+  if (packageJSON.dependencies?.next) {
+    return "next";
+  }
+  if (packageJSON.devDependencies?.svelte) {
+    return "svelte";
+  }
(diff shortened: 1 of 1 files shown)


dwwoelfel (approved):
LGTM!

It's not clear to me why apps like Clerk tell you to put the publishable key in .env.local since that's not supposed to be checked in. Seems like it would make it hard to collaborate?

PR #560, opened 2024-11-27
[strong-init][cli] make i.schema the default

stopachka:
This PR: 
- Deprecates `i.graph` in favor of `i.schema`
- Uses `i.schema` throughout our codebase
- Updates cli's `pull` schema to create a file with `i.schema` rather than i.graph

**Some stuff punted** 

I am punting a few pieces of work in strong init, for a future release: 

1. Better format for links
  a. I was not 100% sure it would be a good idea to remove link names. They indeed aren't necessary, but it would make migrating from i.graph more cumbersome, and it's not clear to me that `links: Array<link>` is the better approach. I think we can look into this as a separate project later
2. Better typing for `where`
  a. @dwwoelfel (and our users) pointed out that right now `where` does not use type info. I think we _can_ make it use type info. I'm thinking we ship this as a separate unit of work 

**Some stuff coming soon**

Right now, if the user does `pull` twice, their room schema will wipe out. I am going to do a follow-on PR, where we make a best effort to keep the existing schema. In the meantime, the user always has git, so they can quickly add it back in.

Once cli stuff is finished, I'll push a new PR, which makes strong init the default.

@dwwoelfel @nezaj

### DIFF
--- a/client/packages/cli/index.js
+++ b/client/packages/cli/index.js
@@ -752,12 +752,7 @@ async function pullSchema(appId, { pkgDir, instantModuleName }) {
   const schemaPath = join(pkgDir, "instant.schema.ts");
   await writeTypescript(
     schemaPath,
-    generateSchemaTypescriptFile(
-      appId,
-      pullRes.data.schema,
-      pullRes.data["app-title"],
-      instantModuleName,
-    ),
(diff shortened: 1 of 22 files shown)


dwwoelfel (approved):
Beautiful!

PR #563, opened 2024-11-27
Add a rate-limited flag for apps that are making too many requests

dwwoelfel:
Adds a new `rate-limited-apps` namespace to instant-config where we can add apps that are making too many requests to ban them until they talk to us.

Error message looks like:

```
sandbox-nextjs:dev:  ⨯ {
sandbox-nextjs:dev:   status: 429,
sandbox-nextjs:dev:   body: {
sandbox-nextjs:dev:     type: 'rate-limited',
sandbox-nextjs:dev:     message: "You're making too many requests. Please email support@instantdb.com or ask for help in the Discord.",
sandbox-nextjs:dev:     hint: null
sandbox-nextjs:dev:   }
sandbox-nextjs:dev: }
```

Right now it just bans them from making admin requests, but we can add extra information to the row latter to be more specific about which actions we're limiting and even add rate info.

### DIFF
--- a/server/src/instant/config.clj
+++ b/server/src/instant/config.clj
@@ -22,6 +22,12 @@
       :test
       :dev)))
 
+(defonce process-id
+  (delay
+    (str (name (get-env))
+         "_"
+         (string/replace (UUID/randomUUID) #"-" "_"))))
+
 (def config-map
   (delay (do
            ;; init-hybrid because we might need it to decrypt the config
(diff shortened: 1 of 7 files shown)


server/src/instant/config.clj
```diff
@@ -95,9 +101,9 @@
 (defn get-aurora-config
   ([] (get-aurora-config {:env (get-env)}))
   ([{:keys [env]}]
-   (let [application-name (uri/query-encode (format "instant server; host: %s, env: %s"
+   (let [application-name (uri/query-encode (format "%s, %s"
```
dwwoelfel: This just makes it easier to identify which instance is making the queries

stopachka (approved):
Lesdoit!

PR #564, opened 2024-11-27
order by any checked and indexed field

dwwoelfel:
Adds order by support for any type-checked+indexed field.

Here's what it looks like:

```clojure
{:tasks {:$ {:limit 5 :order {:priority "desc"}}}}
```

It requires the field to be both checked and indexed. If you try to use it with a field that doesn't meet those requirements, you'll get a helpful error message from the server:

```json
{
  "message": "Validation failed for query",
  "hint": {
    "data-type": "query",
    "input": {
      "goals": {
        "$": {
          "limit": 5,
          "order": {
            "title": "asc"
          }
        }
      }
    },
    "errors": [
      {
        "expected": "supported-order?",
        "in": [
          "goals",
          "$",
          "order"
        ],
        "message": "The `goals.title` attribute is not indexed. Only indexed and type-checked attrs can be used to order by."
      }
    ]
  }
}
```

We treat null and undefined as equal and as smaller than any other value. Postgres treats nulls as larger than any value, but I don't think many developers find that very intuitive. JS and clojure both treat them as smaller:

```
> 1 < null
< false
> 1 > null
< true
```

If people complain, we can add some way to say `nulls first` or `nulls last` like postgres does.


server/src/instant/data/bootstrap.clj
```diff
@@ -83,7 +83,7 @@
                        :cardinality :one
                        :value-type :blob
                        :unique? (boolean (#{"email" "handle" "isbn13"} idn))
-                       :index? (boolean (#{"email" "handle" "title"} idn))}
+                       :index? (boolean (#{"email" "handle" "title" "order"} idn))}
```
stopachka, line 86: awwh yeah!

stopachka (approved):
Beautiful work @dwwoelfel !!

PR #565, opened 2024-11-27
[cli] persist `rooms` info across pulls

stopachka:
**Context** 

We write `rooms` schema inside schema.ts. However, the backend doesn't currently store this schema. 

This could lead a problem, where if the user does a `pull` and chooses to overwrite, they will lose their `rooms` info. 

**Fix** 

We keep the existing schema file, and re-create the 'rooms' section if it's possible.

**Longer-term** 

It's still not 100% ideal, the user would lose comments, or if they wrote particular types, like: 

```typescript
i.entity<MyTypeOverride>(...)
``` 

In the future, we may want to use something like https://github.com/benjamn/recast to modify the code as an ast. But for now, this already an improvement.

@dwwoelfel @nezaj

### DIFF
--- a/client/packages/cli/index.js
+++ b/client/packages/cli/index.js
@@ -739,9 +739,8 @@ async function pullSchema(appId, { pkgDir, instantModuleName }) {
     console.log("Schema is empty. Skipping.");
     return;
   }
-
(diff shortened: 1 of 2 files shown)


dwwoelfel (approved):
LGTM!

PR #568, opened 2024-11-27
[cli][schema] Don't break early when user skips schema

stopachka:
Previously, if a user wrote something like: 

```bash
instant-cli push
``` 

And then there were no schema changes, the whole flow would exit. This didn't make sense though, as the user _may_ have wished to push perms too. 

Similarily, if a user wrote: 

```bash
instant-cli pull
```

And they didn't want to override their schema file for example, the whole flow would exit. But this also didn't make sense, as the user _could_ have wanted to pull perms. 

### Solution
Updated `pushSchema` and `pullSchema` to return an `ok` flag. This returns `true` if the user decides to exit early, or there is no schema. It only returns false where there is a real error and we should break out. 

@dwwoelfel @nezaj

### DIFF
--- a/client/packages/cli/index.js
+++ b/client/packages/cli/index.js
@@ -413,7 +413,7 @@ async function handlePush(bag, opts) {
 
 async function push(bag, appId, opts) {
   if (bag === "schema" || bag === "all") {
-    const ok = await pushSchema(appId, opts);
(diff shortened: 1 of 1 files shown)


dwwoelfel (approved):
LGTM!

PR #569, opened 2024-11-27
Fix how we do per-route middleware

dwwoelfel:
If we wrap the `GET` or `POST` with the middleware, then it gets applied before any route matching is applied.

https://github.com/weavejester/compojure/blob/master/src/compojure/core.clj#L185

If I'm reading how `defroutes` works correctly, we try every handler until we get a match, so my middleware could run on every request (depending on ordering of the routes).

Also fixes how we wrap cookies. Tested the oauth login locally.

### DIFF
--- a/server/src/instant/admin/routes.clj
+++ b/server/src/instant/admin/routes.clj
@@ -435,14 +435,14 @@
         (handler req)))))
 
 (defroutes routes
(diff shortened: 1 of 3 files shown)


stopachka (approved):
_sends high five_ @dwwoelfel !

PR #570, opened 2024-11-28
[cli][schema] use checked-types when creating new attrs, show rev ident

stopachka:
- I realized we can set the checked types for when we create attrs too
- I also fixed up a few bugs in cli (we were printing json twice in verbose mode), and had a bug in `attrRevName`

@dwwoelfel @nezaj


dwwoelfel (approved):
LGTM! Thanks for adding the checked types on create!

PR #575, opened 2024-12-02
[experiment] disable $isNull on refs

stopachka:
This disables $isNull on refs


clarkg:
@stopachka is this live? could you update the docs if so?

We got a honeycomb error when attempting a $isNull ref query, wondering if this was it

clarkg:
oh nvm I can't read, you didn't actually merge this

PR #577, opened 2024-12-02
Add primary keys to all tables

dwwoelfel:
Drop the deprecated_triples table and add a `uuid` primary key to `instant_subscriptions`.

Doesn't require any code changes, postgres will automatically insert a random uuid as the id for the existing rows and for new rows.


stopachka (approved):
Thanks!

PR #578, opened 2024-12-02
[server] add unique identifiers to d/query-nested

stopachka:
Right now it's hard to correlate our slow queries on postgres, to the originating datalog queries. 

Added a quick app-id, query-hash to the beginning of every one of our queries, which we can use as variables to correlate

@dwwoelfel @nezaj @tonsky


dwwoelfel (approved):
Nice!

PR #580, opened 2024-12-03
Add a function that lets us fail over to a new db

dwwoelfel:
*note to self: update base branch*

After merging the PR, the process looks like this:

1. Set up the new replica, start replication, and wait for it to catch up
2. Deploy an update to the config.edn that sets `:next-database-url` to the replica's url
   - This will make the invalidator use the replica, which saves us restarting the invalidator when we fail over
4. Run `do-failover-to-new-db`
5. Shut down the old db
6. Push up a config update that moves `:next-database-url` to `:database-url` and delete `:next-database-url`
7. Done!

The failover function does the following:

1. Pauses any new queries and transactions
2. Waits a little bit for write queries to finish
3. Cancels any writes that are taking too long
4. Moves aurora/conn-pool to the new database
5. Unpauses queries and transactions

This should minimize the effect of upgrading the database.

### DIFF
--- a/server/src/instant/jdbc/failover.clj
+++ b/server/src/instant/jdbc/failover.clj
@@ -1,0 +1,85 @@
+(ns instant.jdbc.failover
+  (:require [instant.config :as config]
+            [instant.db.model.transaction :as transaction-model]
+            [instant.jdbc.sql :as sql]
+            [instant.jdbc.aurora :as aurora]
+            [instant.util.tracer :as tracer]))
+
+(defn start-new-pool [aurora-config]
+  (let [conn-pool-size (config/get-connection-pool-size)]
+    (sql/start-pool
+     (assoc aurora-config
+            :maxLifetime (* 10 60 1000)
+            :maximumPoolSize conn-pool-size
+            :targetServerType "primary"))))
+
+;; Keep this here just in case
+(declare previous-conn-pool)
+
+(defn do-failover-to-new-db []
+  (println "Failing over to new db")
+  (let [next-config (config/get-next-aurora-config)
+        _ (assert next-config "missing next database config")
+        next-pool (start-new-pool next-config)
+        conn-pool-fn-before aurora/conn-pool
+        prev-pool aurora/-conn-pool
+        next-pool-promise (promise)]
+    (tool/def-locals)
+    (println "Started next pool")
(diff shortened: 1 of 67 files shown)


stopachka:
> This will make the invalidator use the replica, which saves us restarting the invalidator when we fail over
🤯 Great idea!

server/src/instant/jdbc/failover.clj
```diff
@@ -0,0 +1,77 @@
...
+
+(defn do-failover-to-new-db []
+  (println "Failing over to new db")
+  (let [next-config (config/get-next-aurora-config)
+        _ (assert next-config "missing next database config")
+        next-pool (start-new-pool next-config)
+        conn-pool-fn-before aurora/conn-pool
+        prev-pool aurora/-conn-pool
+        next-pool-promise (promise)]
+    (tool/def-locals)
```
stopachka, line 26: nit: can remove

server/src/instant/jdbc/failover.clj
```diff
@@ -0,0 +1,77 @@
...
+    ;; caller tell us if they wanted a read-only connection and then we wouldn't
+    ;; have to pause reads until after we waited for writes to complete
+    (alter-var-root #'aurora/conn-pool (fn [_] (fn [] @next-pool-promise)))
+    ;; Give transactions half the receive-timeout to complete
+    (println "Waiting for 2.5 seconds for transactions to complete")
+    (Thread/sleep 2500)
+    (println "Canceling in-progress transactions"
+             (count @(:stmts sql/default-statement-tracker)))
+    (sql/cancel-in-progress @(:stmts sql/default-statement-tracker))
+    ;; Create a transaction we can use as a proxy for everything syncing over to
```
stopachka, line 38: Cool!

stopachka (approved):
Beautiful work @dwwoelfel !

server/src/instant/jdbc/failover.clj
dwwoelfel: I'd like to keep it just in case something weird happens. It's meant to be run from the repl, so this way we can more easily clean up if something throws in the middle of the upgrade.

stopachka: 🫡 makes sense!

PR #581, opened 2024-12-03
$default permissions

tonsky:
I think `"$default"` is more consistent with other places where you have “special case” things. Also easier to google/ask questions about


stopachka (approved):
Heeck yeah @tonsky! Users are going to love this : )

PR #582, opened 2024-12-03
Dev conveniences

tonsky:
Adds:

- [x] `dev` to :extra-paths and `user` namespace that is loaded by default in dev
- [x] `clj-reload` integration to properly reload code and stateful systems without restarting
- [x] `eftest` to run tests and not be overwhelmed by output
- [x] `#p` macro to quickly track values (inspired by https://github.com/weavejester/hashp)
- [x] socket server at 6006 in case anyone needs it


tonsky:
This is how eftest runner looks like

https://github.com/user-attachments/assets/73b8d04b-b888-471e-8d2a-88d595551ae6

server/src/tool.clj
```diff
@@ -168,3 +170,29 @@
 (comment
   (start-portal!)
   (tap> {:hello [1 2 3]}))
+
+(def p-lock
```
stopachka: Nit: I would mark this private, and do `defn-` for p-impl, just to keep these all the under the `p` function.

server/src/tool.clj
```diff
@@ -168,3 +170,29 @@
...
+                     (list? form)
+                     (= 'tool/p-impl (first form)))
+                  (TaggedLiteral/create 'p (nth form 3))
+                  form))
+              form)]
+    (locking p-lock
+      (println (str position " #p " form " => " (pr-str res))))
+    res))
+
+(defn p [form]
```
stopachka: Nit: would add a docstring about this, and a potential usage here.

stopachka (approved):
Heeck yeah!

PR #585, opened 2024-12-03
[db-upgrade 1/3] Add next database url to config

dwwoelfel:
*note to self: do not deploy until step 11*

Updates the next-database-url to the new database so that invalidators can start reading from new database


stopachka (approved):
Lesgoo

PR #588, opened 2024-12-03
Make sure we use the `:av` index on lookup triples

dwwoelfel:
Small improvement for lookups--make sure we're looking at the `av` index when we search for the triple value.

*Note* I discovered we don't handle reverse link lookups properly, so we'll probably have to update this to use the `vae` index on reverse lookups. I'm surprised nobody has complained about this, but I guess they're only looking up in the forward direction or they gave up on lookups.


nezaj (approved):
Nice!

PR #590, opened 2024-12-04
Coerce date data types when we use equality comparisons too

stopachka:
I noticed a curious error in prod: 

https://ui.honeycomb.io/instantdb/environments/prod/datasets/instant-server/result/r4cm97gTDzi/trace/aAmj4xsbSfo?fields[]=s_name&fields[]=s_serviceName&span=432079d52cf13865

It was something along the lines of: 

```bash
clojure.lang.ExceptionInfo: [instant-exception] SQL Exception: undefined-function {:instant.util.exception/type :instant.util.exception/sql-exception, :instant.util.exception/message "SQL Exception: undefined-function", :instant.util.exception/hint {:table nil, :condition :undefined-function, :constraint nil}, :instant.util.exception/pg-error-data {:sql-state "42883", :condition :undefined-function, :table nil, :constraint nil, :server-message "operator does not exist: timestamp with time zone = character varying", :detail nil}
```

Looking into this, I realize it was because:
- The users passed in a date as a str in the query
- We use the value, and compare it to triples_extract_timestamp_value 
- But, the input is not casted as a timestamp 

This ends up breaking and throwing the error above. 

I noticed we _do_ coerce in the comparator case, so I went ahead and started to coerce in the equal case too. 

I am not 100% sure if this is the right approach, so all feedback welcome!

@dwwoelfel @nezaj @tonsky

### DIFF
--- a/server/src/instant/db/model/attr_pat.clj
+++ b/server/src/instant/db/model/attr_pat.clj
@@ -253,15 +253,18 @@
                       (json/->json value)
                       (json/json-type-of-clj value))}]))
 
-(defn throw-on-invalid-value-data-value!
-  "Validates an individual value"
+(defn coerce-value-data-value!
+  "Coerces an individual value"
   [state attr data-type v]
   (case data-type
-    :string (when-not (string? v)
+    :string (if (string? v)
+              v
               (throw-invalid-data-value! state attr data-type v))
-    :number (when-not (number? v)
+    :number (if (number? v)
+              v
               (throw-invalid-data-value! state attr data-type v))
-    :boolean (when-not (boolean? v)
+    :boolean (if (boolean? v)
+               v
                (throw-invalid-data-value! state attr data-type v))
     :date (cond (number? v)
                 (try
@@ -277,7 +280,7 @@
 
                 :else
                 (throw-invalid-data-value! state attr data-type v))
-    nil))
+    v))
 
 (defn assert-like-is-string! [state attr tag value]
   (when (not= tag :string)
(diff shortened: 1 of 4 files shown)


server/src/instant/db/model/triple.clj
```diff
@@ -19,8 +19,8 @@
 ;; In the future, we may want to _retract_ the triple if the value is nil
 (defn value? [x]
   (or (string? x) (uuid? x) (number? x) (nil? x) (boolean? x)
-      (sequential? x) (associative? x)))
-
+      (sequential? x) (associative? x)
+      (instance? java.time.Instant x)))
```
stopachka, line 23: I am not 100% sure if this is a good idea. Alternatively we could have a separate spec, which we use in datalog for the constant portion of the queries

server/src/instant/db/model/attr_pat.clj
```diff
@@ -306,14 +309,16 @@
...
-          :let [v (if (and (map? v)
-                           (contains? v :$not))
-                    (:$not v)
-                    v)]]
-    (throw-on-invalid-value-data-value! state attr data-type v))
-  v)
+(defn coerce-v-single! [state attr data-type v]
+  (if (and (map? v)
+           (contains? v :$not))
+    (update v :$not coerce-value-data-value! state attr data-type)
```
dwwoelfel: I think you might need to do something like `(update v :$not (partial coerce-value-data-value! state attr data-type))` to get the arg in the right position.

dwwoelfel (approved):
One comment, but LGTM!

PR #591, opened 2024-12-04
Extend the query hash

dwwoelfel:
We forgot to call `hsql/format` on the query, so the hash was including the variables in it.

This PR does the hash on the instaql forms, replacing the specific values in the where clause with their type. For example:

```clj
{:users {:$ {:where {:name "Daniel"}}}}
=> {:users {:$ {:where {:name :string}}}}
```

That way we can more easily figure out which query generated the hash.

If we're calling datalog/query from outside instaql, then we'll fall back to calling `hash` on the result of the parametrized  first arg from `(hsql/format query)`.

### DIFF
--- a/server/src/instant/db/datalog.clj
+++ b/server/src/instant/db/datalog.clj
@@ -1695,8 +1695,10 @@
                                                        :match-0-
                                                        app-id
                                                        nested-named-patterns)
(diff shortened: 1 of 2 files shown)


stopachka (approved):
Heeck yeah!

PR #592, opened 2024-12-04
[Backend] Add transaction idx

nezaj:
Adding an index to app_id for the transactions table to speed up our metrics job

### DIFF
--- a/server/resources/migrations/42_add_transactions_app_id_idx.down.sql
+++ b/server/resources/migrations/42_add_transactions_app_id_idx.down.sql
@@ -1,0 +1,1 @@
+DROP INDEX IF EXISTS transactions_app_id_idx;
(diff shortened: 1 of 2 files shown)


server/resources/migrations/42_add_transactions_app_id_idx.down.sql
```diff
@@ -0,0 +1 @@
+DROP INDEX IF EXISTS transactions_app_id_idx;
```
dwwoelfel: I would leave out the `if exists` here because something has gone wrong if the index doesn't exist when this down migration is run.

dwwoelfel (approved):
LGTM, mind putting the `create index concurrently` command you're going to run in the description body with a `[ ]` checkbox to indicate when it's done?

PR #597, opened 2024-12-05
[scratch] add export script

stopachka:
We had a script that let us import apps into dev, to do perf checks and the like.

Putting it up for the team to use here. 

Credit to @dwwoelfel for writing the original version of this, and making debugging better + safer. 

@dwwoelfel @nezaj

### DIFF
--- a/server/src/instant/scratch/export.clj
+++ b/server/src/instant/scratch/export.clj
@@ -71,7 +71,7 @@
         _ (println (format  "Export app_id = %s" prod-app-id))
         exported-data (with-open [pool (sql/start-pool (assoc (config/db-url->config prod-db-uri)
                                                               :maximumPoolSize 1))]
-                        (export-app pool #uuid "a749930e-6737-4dcf-b039-60c7f5e4e2e6"))
+                        (export-app pool prod-app-id))
 
         _ (println "Exported")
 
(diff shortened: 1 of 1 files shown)


server/src/instant/scratch/export.clj
```diff
@@ -0,0 +1,94 @@
...
+(defn import!
+  [{:keys [prod-db-uri local-user-email prod-app-id]}]
+  (let [_ (assert prod-db-uri "prod-db-uri is required")
+        _ (assert local-user-email "local-user-email is required")
+        _ (assert prod-app-id "prod-app-id is required")
+
+        _ (println (format  "Export app_id = %s" prod-app-id))
+        exported-data (with-open [pool (sql/start-pool (assoc (config/db-url->config prod-db-uri)
+                                                              :maximumPoolSize 1))]
+                        (export-app pool #uuid "a749930e-6737-4dcf-b039-60c7f5e4e2e6"))
```
dwwoelfel:
```suggestion
                        (export-app pool prod-app-id))
```

dwwoelfel (approved):
One comment, but LGTM!

server/src/instant/scratch/export.clj
stopachka: Nice catch!

PR #598, opened 2024-12-05
Add script to check data on replica

dwwoelfel:
Adds a script that will run a check that all of the data in the primary is also in the replica.

Right now it takes about an hour or two to run--as the db grows, we may have to do a spot check instead of checking everything.


stopachka (approved):
Lesgoo

PR #599, opened 2024-12-05
[db-upgrade 3/3] Update connection string helper to point to new db

dwwoelfel:
Updates the db connection string script. To update in the future, we'll just need to update the `cluster_id` var in the script.


stopachka (approved):
cool!

PR #600, opened 2024-12-05
remove run opts for backtest

stopachka:
I originally wrote this thinking we would spin up a prod db instance. But I realize this is not needed, and can be done in comments. 

In which case, no need for a separate command. 

@dwwoelfel @nezaj @tonsky


dwwoelfel (approved):
LGTM

PR #601, opened 2024-12-05
[Metrics] Add daily metrics job

nezaj:
This PR does two things 1) speeds up our monthly metrics job and 2) sends us a daily ping with active apps and devs.

**speeding up metrics**
I realized we could speed up our monthly metrics query by changing how we calculated daily app transaction counts. I also thought it would be nice to save these counts to a table to make our metrics generation even faster. Generating the metrics and bar charts is pretty instant now.

**daily metrics ping**
Stopa mentioned earlier that he wanted a more frequent feel for number of users/apps. I played around with sending images to discord, but they get cut off in the previews which makes it less fun to look at. Figured it would be nice to just send some plan numbers in a separate discord channel

**Next steps before merging**
- [x] Get stamp on PR
- [ ] Run CREATE statement in prod + migrate db

### DIFF
--- a/server/src/instant/scripts/daily_metrics.clj
+++ b/server/src/instant/scripts/daily_metrics.clj
@@ -1,0 +1,127 @@
+(ns instant.scripts.daily-metrics
+  "Job to ping discord with daily active metrics. We also use this job to
+  populate the daily_app_transactions table with new transactions."
+  (:require
+   [instant.jdbc.aurora :as aurora]
+   [instant.util.date :as date]
+   [clojure.tools.logging :as log]
+   [instant.discord :as discord]
+   [chime.core :as chime-core]
+   [instant.flags :refer [get-emails]]
+   [instant.config :as config]
+   [instant.jdbc.sql :as sql]
+   [instant.grab :as grab])
+  (:import
+   (java.time Period LocalDate)))
+
+(defn excluded-emails []
+  (let [{:keys [test team friend]} (get-emails)]
+    (vec (concat test team friend))))
+
+(defn get-daily-actives
+  "Returns the number of active devs and apps for a day"
+  ([date-str]
+   (get-daily-actives (aurora/conn-pool) date-str))
+  ([conn date-str]
+   (sql/select-one conn
+                   ["SELECT
+                  dat.date as date_start,
+                  COUNT(dat.count) AS total_transactions,
+                  COUNT(DISTINCT u.id) AS distinct_users,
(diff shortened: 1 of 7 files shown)


server/src/instant/scripts/daily_metrics.clj
```diff
@@ -0,0 +1,127 @@
...
+        periodic-seq (chime-core/periodic-seq
+                      nine-am-pst
+                      (Period/ofDays 1))]
+
+    (->> periodic-seq
+         (filter (fn [x] (.isAfter x now))))))
+
+(defn start []
+  (log/info "Starting daily metrics daemon")
+  (def schedule (chime-core/chime-at (period) daily-job!)))
```
stopachka, line 123: will this work? afaik `daily-job!` will get an argument

server/src/instant/scripts/daily_metrics.clj
```diff
@@ -0,0 +1,127 @@
...
+
+(defn stop []
+  (when schedule
```
dwwoelfel: Since schedule is a `def`, it will always be truthy here.

```suggestion
  (when (bound? #'schedule)
```

Ran into the same problem with https://github.com/instantdb/instant/blob/main/server/src/instant/reactive/invalidator.clj#L425

server/src/instant/scripts/daily_metrics.clj
```diff
@@ -0,0 +1,127 @@
...
+
+(defn send-discord!
+  "Ping the discord channel with the metrics for a specific date"
+  [stats date-str]
+  (let [{:keys [distinct_users distinct_apps]} stats
+        message (str "🎯 Daily metrics for " date-str
+                     ": Active Devs: **" distinct_users
+                     "**, Active Apps: **" distinct_apps
+                     "**")]
+    (discord/send! config/discord-metrics-channel-id message)))
```
dwwoelfel: You might just want to dump this in the `team` channel? It's only 1 message a day.

server/src/instant/scripts/daily_metrics.clj
nezaj: Good catch! Updated to handle date correctly :)

stopachka (approved):
lesgo!

dwwoelfel (approved):
LGTM! 

I just need to finish the db failover before we merge this and run the migrations.

server/src/instant/scripts/daily_metrics.clj
nezaj: Updated!

PR #603, opened 2024-12-06
Fix the db swap script

dwwoelfel:
Accidentally incremented the wrong tx-id when reseting the seq, so we were never going to complete the db swap.

The good news is that we rolled back with minimal issues (there were a few handle-receive timeouts for in-progress queries).


stopachka (approved):
🫡🫡🫡

PR #605, opened 2024-12-06
Send less log data to cloudfront in production

dwwoelfel:
Removes one span entirely. I don't think the `collect-instaql-results-for-client` span was really helpful for anything.

Stops logging a few more spans to cloudfront in production:
- handle-refresh/send-event!
- store/record-datalog-query-finish!
- store/record-datalog-query-start!
- store/swap-datalog-cache-delay!
- store/bump-instaql-version!
- store/add-instaql-query!


stopachka (approved):
🫡🫡🫡

PR #606, opened 2024-12-06
Set the `unique?` flag on attrs when they're created

dwwoelfel:
We weren't making ids unique when they were generated from a cli push or when you created a new namespace from the UI. This PR fixes both.


stopachka (approved):
Ah, thanks @dwwoelfel !

PR #608, opened 2024-12-06
Better error handling for query errors in reactor

dwwoelfel:
This should fix the reactor bug that Aphex reported in Discord: https://discord.com/channels/1031957483243188235/1314344673459965992

The basic issue is that we're too aggressive in deleting querySubs when we experience an error.

For the error in discord, we get a stacktrace that looks like:

```
Reactor.js:483 Uncaught TypeError: Cannot set properties of undefined (setting 'result')
    at eval (Reactor.js:483:39)
    at PersistedObject.set (PersistedObject.js:115:29)
    at Reactor._handleReceive (Reactor.js:482:32)
    at Reactor._wsOnMessage (Reactor.js:291:18)
```

I was able to replicate the error by forcing `queryOnce` to fail with a query that `useQuery` is subscribing to. We delete the `querySub` in the error handler and then we throw an error when we handle the `add-query` for the `useQuery` because the querySub is null.

After this change, we won't delete the `querySub` in the error handler unless the `queryOnce` is the last subscriber to the query, or the `useQuery` unsubscribes (and is the last subscriber).

### DIFF
--- a/client/packages/core/src/Reactor.js
+++ b/client/packages/core/src/Reactor.js
@@ -546,17 +546,11 @@ export default class Reactor {
       return;
     }
 
-    const q = msg.q || msg["original-event"]?.q;
-    if (q) {
+    const q = msg["original-event"]?.q;
+    if (q && msg["original-event"]?.op === "add-query") {
       const hash = weakHash(q);
-
-      // This must be a query error
-      this.querySubs.set((prev) => {
-        delete prev[hash];
-        return prev;
-      });
       this.notifyQueryError(weakHash(q), errorMessage);
-      this.notifyQueryOnceError(hash, eventId, errorMessage);
+      this.notifyQueryOnceError(q, hash, eventId, errorMessage);
       return;
     }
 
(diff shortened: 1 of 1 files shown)


client/packages/core/src/Reactor.js
```diff
@@ -546,17 +546,11 @@ export default class Reactor {
       return;
     }
 
-    const q = msg.q || msg["original-event"]?.q;
-    if (q) {
+    const q = msg["original-event"]?.q;
+    if (q && msg["original-event"]?.op === "add-query") {
```
dwwoelfel, line 550: I updated this to only handle the error if the original op was `add-query`. `remove-query` also sets a `q` field, so this will prevent us from getting ourselves into a loop where `remove-query` fails so we send another `remove-query` which fails, etc.

I also deleted the check for `msg.q` because we never set a top-level `q` field in the error handler.

nezaj (approved):
Nice!!

stopachka (approved):
SGTM!

PR #611, opened 2024-12-06
Remove redundant handle-event

dwwoelfel:
We already have the `handle-receive` span which always wraps the `handle-event` span and includes many of the same fields in handle-event.

Now instead of logging a bunch of fields twice, we'll only log them once in the `handle-receive`.


nezaj (approved):
Niiiice!

PR #612, opened 2024-12-06
[Metrics] Chime fix

nezaj:
I realized the metrics job didn't run today. Turns out `chime-at` converts `ZonedDateTime` to `Instant`, this was causing an error when calling `date/numeric-date-str`


dwwoelfel (approved):
LGTM!

PR #613, opened 2024-12-06
[mutations] apply mutations after transact-ok

stopachka:
## Context

Previously, transactions had 3 phases in the lifetime

**1. pending**

The transaction was created, but hasn't been processed by the server. 

We apply it right away to all queries.

**2. ok**

The server said this transaction was `ok`. We record this information, because we don't want to send it again.

**_However_, we can't delete the transaction from the queue yet**. 

Why? Because this transaction has not fully been reflected in active queries.

**3. refresh -> delete**

The server has refresh queries. At this point, we can delete pending mutations. 

## Problem 

The kosmik folks reported an edge case, where queries were missing pending mutations. This is because we were too eager about _when_ we removed pending transactions. We removed them _even_ during add-query-ok, which mean some queries may not have refreshed yet. 

## Solution 

I got rid of cleanPendingMutations entirely. Instead, we apply the mutation onto every active query, during `transact-ok`. 

@dwwoelfel @nezaj

### DIFF
--- a/client/packages/core/src/Reactor.js
+++ b/client/packages/core/src/Reactor.js
@@ -449,13 +449,14 @@ export default class Reactor {
           return prev;
         })
 
-        this._finishTransaction("synced", eventId);
-
         const newAttrs = prevMutation["tx-steps"]
           .filter(([action, ..._args]) => action === "add-attr")
           .map(([_action, attr]) => attr)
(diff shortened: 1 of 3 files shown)


dwwoelfel (approved):
LGTM!

PR #614, opened 2024-12-06
[Intern] Faster top

nezaj:
Figured we can make the `top` page much faster using the `daily_app_transactions` table


dwwoelfel (approved):
Nice!

PR #616, opened 2024-12-07
[delete-entity] make delete-entity use vae index for lookups

stopachka:
Kosmik reported slow delete transactions. 

Looking into this, I realized it was because we compared to `value-md5`: 

```
;; Delete ref triples where we're the value
 :vae
 [:= :value-md5 [:md5 [:cast [:to_jsonb id] :text]]]
```

This worked, but would end up in a scanning all app triples. This is because our `vae` index actually works on the jsonb value, rather then the md5 value. By switching to the jsonb value, we could now use the `v` based indexes. 

```clojure
(time
   (do
     (permissioned-tx/transact!
      ctx
      (admin-model/->tx-steps!
       ctx
       (map (fn [sticker-id]
              ["delete" "stickers" sticker-id])
            sticker-ids)))
     :done))

;; Before 

; eval (current-form): (time (do (permissioned-tx/transact! ctx (admin-model/->tx-step...
; (out) "Elapsed time: 15090.590791 msecs"
:done

;; After

; eval (current-form): (time (do (permissioned-tx/transact! ctx (admin-model/->tx-step...
; (out) "Elapsed time: 111.3705 msecs"
:done
```

@dwwoelfel @nezaj @tonsky

### DIFF
--- a/server/src/instant/db/model/triple.clj
+++ b/server/src/instant/db/model/triple.clj
@@ -463,10 +463,7 @@
                                   :av
                                   [:= :attr-id (first id)]
                                   [:=
-                                   :value-md5
-                                   [:md5 [:cast
-                                          [:cast (->json (second id)) :jsonb]
-                                          :text]]]]}
+                                   :value [:cast (->json (second id)) :jsonb]]]}
                          id)]
                    (if etype
                      [[:and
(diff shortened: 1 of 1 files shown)


dwwoelfel:
Whoops, I think I added the md5 because I thought it would be faster. 

We should add more tests like this one https://github.com/instantdb/instant/blob/a5acc82029eaba4833df7a4ba4f4a56271d224fb/server/test/instant/db/instaql_test.clj#L2142 to make sure our queries stay efficient.

dwwoelfel (approved):
LGTM!

clarkg:
oh wow we were running into this too, thanks!

stopachka:
Heeck yeah @clarkg !

PR #618, opened 2024-12-09
Docs: AppSchema type export

laurentpayot:
AppSchema type export, as described in https://biomejs.dev/linter/rules/use-export-type/ or https://typescript-eslint.io/rules/consistent-type-exports/ (was giving me an error in VSCode with Biome plugin)


stopachka (approved):
Thank you @laurentpayot

PR #619, opened 2024-12-09
[script] nab the first instance from prod_tunnel.sh

stopachka:
I wanted to ssh into prod to diagnose some query issues. I noticed that prod_tunnel.sh was failing, as we now have multiple instances. 

I updated the script to grab the first instance. 

@dwwoelfel @nezaj @tonsky


nezaj (approved):
🚀

PR #620, opened 2024-12-09
[Metrics] Daily metrics sm

nezaj:
Metrics had an off by one issue where the job runs at date X but we only have date for date X - 1. This PR fixes that, I also removed an unused transaction count and figured we should just run this on the weekdays to not ping #team on the weekends

### DIFF
--- a/server/src/instant/scripts/daily_metrics.clj
+++ b/server/src/instant/scripts/daily_metrics.clj
@@ -26,7 +26,6 @@
    (sql/select-one conn
                    ["SELECT
                   dat.date as date_start,
-                  COUNT(dat.count) AS total_transactions,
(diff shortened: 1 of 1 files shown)


server/src/instant/scripts/daily_metrics.clj
```diff
@@ -91,17 +90,23 @@
 
 (defn daily-job!
   [^Instant date]
-  (let [date-str (date/numeric-date-str (.atZone date date/pst-zone))]
+  (let [date-minus-one (-> date (.minus (Period/ofDays 1)))
+        date-fn (fn [x] (date/numeric-date-str (.atZone x date/pst-zone)))
```
stopachka, line 94: nit: would call this `date-str`

PR #622, opened 2024-12-10
Remove entity_id from some indexes

dwwoelfel:
@stopachka found that this improves the performance of some queries that use the `ave` indexes with nested loop joins when there are multiple `e` in the index. The loop causes postgres to do a bunch of inefficient index scans, slowing down the query. When we take the `e` out of the index, postgres no longer tries to use it in a nested loop.

More context on the underlying issue in @stopachka's stackoverflow post: https://dba.stackexchange.com/questions/344103/postgres-graph-query-gets-slow-because-of-a-nested-loop-join

Ran a backtest and we see some major improvements for a few queries and basically no change for the rest of the queries:

image

Deployment process:
1. Add all indexes concurrently:
```sql
create index concurrently triples_string_trgm_gist_idx_no_e on triples using gist (
    app_id,
    attr_id,
    triples_extract_string_value(value) gist_trgm_ops
  )
  where ave and checked_data_type = 'string';

create index concurrently triples_number_type_idx_no_e on triples (
    app_id,
    attr_id,
    triples_extract_number_value(value)
  )
  where ave and checked_data_type = 'number';

create index concurrently triples_boolean_type_idx_no_e on triples (
    app_id,
    attr_id,
    triples_extract_boolean_value(value)
  )
  where ave and checked_data_type = 'boolean';

create index concurrently triples_date_type_idx_no_e on triples (
    app_id,
    attr_id,
    triples_extract_date_value(value)
  )
  where ave and checked_data_type = 'date';

create index concurrently ave_index_no_e
  on triples(app_id, attr_id, value)
  where ave;
```

2. Run the migration.

### DIFF
--- a/server/resources/migrations/45_remove_entity_id_from_indexes.down.sql
+++ b/server/resources/migrations/45_remove_entity_id_from_indexes.down.sql
@@ -29,6 +29,7 @@ create index triples_boolean_type_idx on triples (
   where ave and checked_data_type = 'boolean';
 drop index triples_boolean_type_no_e_idx;
 
+alter index triples_date_type_idx rename to triples_date_type_no_e_idx;
 create index triples_date_type_idx on triples (
     app_id,
     attr_id,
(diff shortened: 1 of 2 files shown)


stopachka (approved):
Aweesome!

PR #623, opened 2024-12-10
Sign in with apple

tonsky:
Native version seems to work fine with minimal changes.

Provider is created automatically when first accessed, since there are no parameters to setup. For the same reason, `signIn` call expects fixed string `clientName: "apple"`

I’m looking into web version now (we want Sign In with Apple on web too, right?), maybe will need to adjust providers as a result.

Also added the docs.


server/src/instant/auth/oauth.clj
```diff
@@ -131,19 +131,24 @@
...
+                                     ;; for https://account.apple.com/.well-known/openid-configuration
+                                     :headers {"User-Agent" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/17.6 Safari/605.1.15"}})]
     (if (clj-http/success? resp)
       {:date (Instant/now)
        :data (:body resp)}
-      (ex/throw-oauth-err! "Unable to fetch OAuth configuration."))))
+      (ex/throw+ {::ex/type    ::ex/oauth-error
+                  ::ex/message "Unable to fetch OAuth configuration"
+                  :status      (:status resp)
+                  :body        (:body resp)}))))
```
dwwoelfel: I'm a little antsy about putting `body` in an instant exception on the off chance we don't sanitize properly and expose the body to the user. Maybe before the `throw-oauth-err!`, we could add an exception trace?

```clj
(tracer/record-exception-span! (ex-info "Error fetching discovery"
                                        {:status (:status resp)
                                         :body (:body resp)
                                         :endpoint endpoint}
                                        {:name "oauth/fetch-discovery-error"})
```

tonsky: Done

tonsky:
Looks like “popup” workflow works fine without any configuration on Instant side, but for redirect we’ll need to create clients. I’ll add them

dwwoelfel (approved):
LGTM!

The docs are maybe a little confusing because you can get to the end before you realize that you have to select one of the three options. Maybe it would help to preselect one of the options and put a border around the current option to show that it's selected?

stopachka:
Woohoo!

PR #625, opened 2024-12-10
make sure we run `distinct-on` when paginating

stopachka:
Uri noticed an error when running a query like: 

```edn
 {:items
   {:$
    {:order {:serverCreatedAt "desc"},
     :limit 5,
     :where
     {:foo.bar {:$gte 5}}}}}
```

Because of the `foo.bar`, it was possible to get multiple instances of `item` entity ids repeating. When we updated the select statement to insert an `order-by` to the query, we could end up over-filtering. 

This is because of the CTE had:

```
entA
entA
entA 
entB 
```

With a `LIMIT` of 3, we would just return `entA`. 

To fix this, I updated `add-page-info`, to include a `select-distinct-on`

@dwwoelfel @nezaj

### DIFF
--- a/server/src/instant/db/datalog.clj
+++ b/server/src/instant/db/datalog.clj
@@ -1080,12 +1080,16 @@
         order-by-direction (if last?
                              (reverse-direction direction)
                              direction)
-        order-by [[(kw table :- order-col-name)
-                   order-by-direction]
-                  [entity-id-col
-                   order-by-direction]]
+
+        order-col (kw table :- order-col-name)
+        order-by [[order-col order-by-direction]
+                  [entity-id-col order-by-direction]]
+        order-cols (map first order-by)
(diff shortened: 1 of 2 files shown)


dwwoelfel (approved):
LGTM for a bug fix, but I think that there's a deeper issue here. We already have the logic to do the distinct for the or gather, we're probably just adding the page info too early?

I'm pretty deep in that code for the order-by, so I can investigate.

One thing I would double check before shipping is that the cursors are still correct.

stopachka:
> We already have the logic to do the distinct for the or gather, we're probably just adding the page info too early?

Oo, that's interesting! 

> I'm pretty deep in that code for the order-by, so I can investigate.

🫡🫡🫡 Sure thing @dwwoelfel . Will hold off on shipping; curious to hear more about what you find.

PR #628, opened 2024-12-11
[docs] update rn starter

stopachka:
When doing our user session with Zack, I realized that our RN starter was outdated. Mainly: 

1. Expo now creates typescript files by default
2. Expo has a router, so the `App` file is now inside a subdirectory

I also felt like our demo colors app could use a bit of a spruce-up. I added some background on our text, and made the default color white.

image

@dwwoelfel @nezaj @tonsky


dwwoelfel (approved):
LGTM!

PR #631, opened 2024-12-12
Add a benchmark to the tests

dwwoelfel:
Could potentially help us catch performance regressions. Outputs in the test output: https://github.com/instantdb/instant/actions/runs/12303534856/job/34339091865

image


nezaj (approved):
Great idea!

PR #632, opened 2024-12-12
Docs: Permissions: isOwner example

laurentpayot:
Added `auth.id != null` to the `isOwner` example clause to avoid granting permission when user is not signed-in and creatorId is set to null.


laurentpayot:
> No matching version found for @vercel/node@5.0.0

:thinking:

nezaj:
Thanks for pointing this out @laurentpayot ! Went ahead and fixed this in a few other places as well here https://github.com/instantdb/instant/pull/643

PR #633, opened 2024-12-14
Launch-Ready Strong Init

stopachka:
This gets strong init ready for prime time launch! 

The PR is big, but it's divided into four separate commits:

## 1. Update CLI, so `init` works without `login`

If the user wasn't logged in, `npx instant-cli` used to break. This led to a poor initial user experience. 

I updated the command so if we aren't logged in, we prompt the user to log in and continue on.

CleanShot 2024-12-13 at 16 49 54@2x

## 2. Make `init_experimental` the default

I went ahead and all `packages`, so the strong init path is now the only path. 

**What this means for users** 

If a user was using normal init: 

```typescript
import { init } from '@instantdb/react'; 

const db = init({ appId })
```


client/packages/cli/index.js
```diff
@@ -1417,6 +1422,14 @@ async function readConfigAuthTokenWithErrorLogging() {
   return token;
 }
 
+async function readAuthTokenOrLoginWithErrorLogging() {
+  const token = await readConfigAuthToken();
+  if (token) return token; 
+  console.log(`Looks like you are no logged in...`);
```
dwwoelfel:
```suggestion
  console.log(`Looks like you are not logged in...`);
```

client/www/components/docs/Ansi.jsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import AnsiToHtml from 'ansi-to-html';
+
+const converter = new AnsiToHtml();
+
+export function Ansi({ children }) {
```
dwwoelfel, line 17: Cool!

client/www/pages/docs/backend.md
```diff
@@ -55,22 +55,39 @@ query once and returns a result.
...
-console.log("New log entry made for" today, "with tx-id", res["tx-id"])
+console.log("New todo entry made for with tx-id", res["tx-id"])
 ```
 
 `transact` is an async function that behaves nearly identical to `transact`
 from `@instantdb/react`. It returns a `tx-id` on success.
 
+## Typesafety
+
+If you're using typescript, `init` accepts a schema argument:
```
dwwoelfel: Even if you're not using typescript, it's nice to provide a schema argument so that we don't make mistakes when we optimistically generate the attributes.

client/www/pages/docs/instaql.md
```diff
@@ -940,6 +949,118 @@ console.log(data)
...
+
+export type { AppSchema };
+export default schema;
+```
+
+### Intellisense
+
+Instant will start giving you intellisense for your queries. For example, if you're querying for goals, you'll see that only `todos` can be associated:
+
+{% screenshot src="https://paper-attachments.dropboxusercontent.com/s_3D2DA1E694B2F8E030AC1EC0B7C47C6AC1E40485744489E3189C95FCB5181D4A_1734124125680_goals.png" /%}
```
dwwoelfel: Can we move these images into our `public/img` folder?

dwwoelfel (approved):
LGTM!

Some thoughts as I went through the example:

1. The `next` link on the [react docs page](https://instant-www-js-st1-jsv.vercel.app/docs) goes to "React native" instead of "Working with data"

2. It would be nice if you didn't have to provide an empty object for `links` and `rooms` if you're not using those. The types will fail if you leave out `rooms`, but if you leave out `links`, we'll throw a unhelpful "TypeError: Cannot convert undefined or null to object" error.

client/www/pages/docs/backend.md
stopachka: I like it! I'll change it to be a bit more general to schema

PR #637, opened 2024-12-17
Make hazelcast the default

dwwoelfel:
It's already the default in prod (through the config), this will make it the default in dev also.

Updates some of the tests to also test hazelcast. 

Note for later: to do a true end-to-end test of Hazelcast, we'd need to spin up multiple instances of the instant server and check that the data syncs between them.


nezaj (approved):
Nice!

PR #639, opened 2024-12-17
Add support for case-insensitive like

dwwoelfel:
Adds support for `ilike`, which is just like `like`, but it's case-insensitive.

Example:

```javascript
const query = {
  goals: {
    $: {
      where: {
        'todos.title': { $ilike: '%stand%' },
      },
    },
  },
};
const { isLoading, error, data } = db.useQuery(query);
console.log(data)
{
  "goals": [
    {
      "id": standupId,
      "title": "Perform standup!",
    },
    {
      "id": standId,
      "title": "Stand up a food truck.",
    }
  ]
}
```

### DIFF
--- a/client/packages/core/src/instaql.js
+++ b/client/packages/core/src/instaql.js
@@ -88,11 +88,23 @@ function refAttrPat(makeVar, store, etype, level, label) {
   return [nextEtype, nextLevel, attrPat, attr, isForward];
 }
 
-function matchesLikePattern(value, pattern) {
-  if (typeof value !== "string" || typeof pattern !== "string") return false;
+function makeLikeMatcher(caseSensitive, pattern) {
+  if (typeof pattern !== "string") {
+    return function likeMatcher(_value) {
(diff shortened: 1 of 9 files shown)


client/packages/core/src/instaql.js
```diff
@@ -88,11 +88,23 @@ function refAttrPat(makeVar, store, etype, level, label) {
...
+    };
+  }
   const regexPattern = pattern.replace(/%/g, ".*").replace(/_/g, ".");
-  const regex = new RegExp(`^${regexPattern}$`);
-  return regex.test(value);
+  const regex = new RegExp(
+    `^${regexPattern}$`,
+    caseSensitive ? undefined : "i",
+  );
+  return function likeMatcher(value) {
```
dwwoelfel, line 102: This should be a little more efficient, because we reuse the regex across matches instead of creating a new one each time.

stopachka (approved):
awh yeah!

PR #640, opened 2024-12-18
Don't filter spam if the attrs changed

dwwoelfel:
The spam filter is preventing the `useSchemaQuery` on the dashboard from catching changes to attributes.

When you add a namespace or index an attr, you won't see the changes on the dashboard until you refresh.

The PR fixes that problem by embedding the attrs into the query hash we use to determine if the query result changed.


stopachka (approved):
Heeck yeah!

PR #643, opened 2024-12-19
[Docs] Fix permissions typos

nezaj:
@laurentpayot Noticed there was an error in our `isOwner` example where indeed an un-authed user could pass the `isOwner` example. Updating all references to these


dwwoelfel (approved):
LGTM!

PR #644, opened 2024-12-19
fix distinct-on with limit

stopachka:
## Context

Consider the following schema: 

```javascript
entities: {
    conversations: i.entity({
      // ...
    }),
    messages: i.entity({
      // ...
    }),
  },
  links: {
    messageConvo: {
      forward: { on: 'messages', has: 'one', label: 'conversation' }, 
      reverse: { on: 'conversations', has: 'many', label: 'messages' }
    },
  },
```

Say I create 3 convos, with 10 messages each.

## Problem

If I query:  

```edn
{:conversations {:$ {:limit 2 :where {:messages.time {:$gte 0}}}}}
```

**I will get _1_ result back, rather then 2** 

## Investigation 

This is because the datalog query looks something like: 

```
[?m :time (> t)] 
[?m :convo ?c] 
```

When we add our `order` information, we actually add it to `[?m :convo ?c]`: 

```sql
WITH match_0_0 AS MATERIALIZED (
  SELECT
    entity_id AS match_0_0_entity_id,
    attr_id AS match_0_0_attr_id,
    value AS match_0_0_value_blob,
    CASE
      WHEN eav THEN CAST(value ->> 0 AS UUID)
      ELSE null
    END AS match_0_0_value_uuid,
    created_at AS match_0_0_created_at
  FROM
    triples
  WHERE
    (
      app_id = 'fc10c06f-1a4c-496f-bb4d-a3b9ce252cb7' :: uuid
    )
    AND (ave = true)
    AND (
      attr_id = '70387671-fad8-416c-9e7f-226228794c05' :: uuid
    )
    AND (TRIPLES_EXTRACT_NUMBER_VALUE(value) >= 0)
    AND (
      checked_data_type = CAST('number' AS CHECKED_DATA_TYPE)
    )
),
match_0_1 AS MATERIALIZED (
  SELECT
    match_0_0.*,
    entity_id AS match_0_1_entity_id,
    attr_id AS match_0_1_attr_id,
    value AS match_0_1_value_blob,
    CASE
      WHEN eav THEN CAST(value ->> 0 AS UUID)
      ELSE null
    END AS match_0_1_value_uuid,
    created_at AS match_0_1_created_at
  FROM
    triples,
    match_0_0
  WHERE
    (
      app_id = 'fc10c06f-1a4c-496f-bb4d-a3b9ce252cb7' :: uuid
    )
    AND (eav = true)
    AND (
      attr_id = 'ea2bf36e-200a-4b55-a86c-bbfd623ed5da' :: uuid
    )
    AND (entity_id = match_0_0_entity_id)
),
match_0_2 AS MATERIALIZED (
  SELECT
    -- this makes ?m unique, rather than ?c
    DISTINCT ON(order_val, match_0_1_entity_id) CAST(created_at AS BIGINT) AS order_val, 
    match_0_1.*,
    entity_id AS match_0_2_entity_id,
    attr_id AS match_0_2_attr_id,
    value AS match_0_2_value_blob,
    CASE
      WHEN eav THEN CAST(value ->> 0 AS UUID)
      ELSE null
    END AS match_0_2_value_uuid,
    created_at AS match_0_2_created_at
  FROM
    match_0_1
    LEFT JOIN triples ON (
      app_id = 'fc10c06f-1a4c-496f-bb4d-a3b9ce252cb7' :: uuid
    )
    AND (ea = true)
    AND (
      attr_id = 'cdf7582a-4b07-4c36-ab28-5916539d8db4' :: uuid
    )
    AND (entity_id = match_0_1_value_uuid)
  ORDER BY
    order_val ASC NULLS FIRST,
    match_0_1_entity_id ASC
  LIMIT
    2
)
SELECT * FROM match_0_2;
```

## Possible Solution

It's a bit of hack, but here's what I do: 

I detect that _if_ we end up joining `curr-table-entity-id` on `prev-table-value-uuid`, then the `entity-id-col` must be prev-table-uuid. 

By doing this, all tests pass

@dwwoelfel @nezaj @tonsky

### DIFF
--- a/server/src/instant/db/datalog.clj
+++ b/server/src/instant/db/datalog.clj
@@ -1136,7 +1136,7 @@
                                     page-pattern)
         prev-table (kw prefix (dec next-idx))
         entity-id-col (if (= order-col-type :created-at-timestamp)
-                        (kw table :-entity-id)
+                        (kw :entity-id)
                         (kw prev-table :-entity-id))
         sym-component-type (component-type-of-sym named-pattern order-sym)
         sym-triple-idx (get (set/map-invert idx->component-type)
--- a/server/src/instant/db/datalog.clj
+++ b/server/src/instant/db/datalog.clj
@@ -1188,14 +1188,14 @@
                                                      :order-col-type order-col-type
                                                      :cursor after
                                                      :cursor-type :after
-                                                     :entity-id-col entity-id-col})
+                                                     :entity-id-col :entity-id})
                       before (add-cursor-comparisons {:direction direction
                                                       :sym-triple-idx sym-triple-idx
                                                       :order-col-name order-col-name
                                                       :order-col-type order-col-type
                                                       :cursor before
                                                       :cursor-type :before
-                                                      :entity-id-col entity-id-col}))
+                                                      :entity-id-col :entity-id}))
 
         first-row-table (kw table :-first)
         last-row-table (kw table :-last)
--- a/server/src/instant/db/datalog.clj
+++ b/server/src/instant/db/datalog.clj
@@ -1135,7 +1135,9 @@
                                     false
                                     page-pattern)
         prev-table (kw prefix (dec next-idx))
-        entity-id-col (kw prev-table :-entity-id)
+        entity-id-col (if (= order-col-type :created-at-timestamp)
+                        (kw table :-entity-id)
+                        (kw prev-table :-entity-id))
         sym-component-type (component-type-of-sym named-pattern order-sym)
         sym-triple-idx (get (set/map-invert idx->component-type)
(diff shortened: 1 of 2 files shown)


server/src/instant/db/datalog.clj
```diff
@@ -1135,7 +1135,9 @@
                                     false
                                     page-pattern)
         prev-table (kw prefix (dec next-idx))
-        entity-id-col (kw prev-table :-entity-id)
+        entity-id-col (if (= order-col-type :created-at-timestamp)
+                        (kw table :-entity-id)
```
dwwoelfel: Try just :entity-id on this line

PR #646, opened 2024-12-19
Make our Y/N prompt a toggle, and make default answer yes for non-destructive questions

stopachka:
From user sessions with Zack and Cam: 

When users were prompted to open a browser, create an app, etc, they were surprised that the default answer was 'No'. They pressed enter, and saw their CLI session end. 

**I made two changes:**

1. I updated it so the default answer is 'yes' for non-destructive questions
2. I introduced a new kind of toggle, which visually shows the 'yes' or 'no' answer, so it's clearer what is selected: 

CleanShot 2024-12-19 at 15 19 36@2x

@dwwoelfel @nezaj @tonsky


nezaj (approved):
Nice!

PR #647, opened 2024-12-19
[bugfix] fix select-distinct-on

stopachka:
**Context:** 

See #644 for the original root issue. 

When we made the query: 

```edn
{:conversations
  {:$
   {:order {:serverCreatedAt "desc"},
    :limit 5,
    :where
    {:groups "---",
     :messages.time {:$gte 1734485030665}}}}}
```

This broke our `joins-on-value?` check. That check was ultimately too brittle. In the case where we added groups, the join on `value` was actually from an earlier CTE: 

```sql
--- ...
match_0_3 AS MATERIALIZED (
  SELECT
    DISTINCT ON(order_val, entity_id) CAST(created_at AS BIGINT) AS order_val,
    match_0_2.*,
    entity_id AS match_0_3_entity_id,
    attr_id AS match_0_3_attr_id,
    value AS match_0_3_value_blob,
    CASE
      WHEN eav THEN CAST(value ->> 0 AS UUID)
      ELSE null
    END AS match_0_3_value_uuid,
    created_at AS match_0_3_created_at
  FROM
    match_0_2
    LEFT JOIN triples ON (
      app_id = 'a749930e-6737-4dcf-b039-60c7f5e4e2e6' :: uuid
    )
    AND (ea = true)
    AND (
      attr_id = 'e2992693-6d30-4973-9712-4b4c700421ae' :: uuid
    )
    AND (entity_id = match_0_0_entity_id)
    AND (entity_id = match_0_1_value_uuid) -- <-- this doesn't join match_0_2, so our order `join` check failed
```

## Fix

I went with the solution @dwwoelfel suggested: 

In the case of created at, we use the current table, rather then the previous one. We think this will still open up bugs for when an order happens on an attribute, but for now it's a hotfix to unblock Uri. We'll do a bigger fix down the road. 

@dwwoelfel @nezaj @tonsky

### DIFF
--- a/server/src/instant/db/datalog.clj
+++ b/server/src/instant/db/datalog.clj
@@ -1107,13 +1107,6 @@
 (defn has-prev-tbl [table]
   (kw table :-has-prev))
 
-(defn joins-on-value? [prev-table [_ & conds]]
-  (let [conds-set (set conds)
-        prev-value-uuid (kw prev-table :-value-uuid)]
-    (or
-     (contains? conds-set [:= :entity-id prev-value-uuid])
-     (contains? conds-set [:= prev-value-uuid :entity-id]))))
-
 (defn add-page-info
   "Updates the cte with pagination constraints."
   [{:keys [next-idx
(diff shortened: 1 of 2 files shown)


nezaj (approved):
Nice sleuthing!

PR #652, opened 2024-12-20
[types] Add an example that includes links in InstaQLEntity

stopachka:
A few users mentioned they didn't know how to include links in InstaQLEntity. 

This is a small addition (may be we need to add more), but I think we should. at least have 1 example. 

@nezaj @dwwoelfel @tonsky


client/www/pages/docs/strong-init.md
```diff
@@ -1,100 +0,0 @@
----
-title: Strong Init (experimental)
----
-
```
stopachka, line 4: I had forgotten to delete this file

nezaj (approved):
Nice!

Issue #493, opened 2024-11-18
creating generic utils with `InstantEntity` surfaces typing issue


uriva:
```ts
export const db = init_experimental({
    adminToken: "...,
    appId: "...",
    schema: i.graph({ users: { email: i.string() } }),
});

type Collection = keyof typeof db.tx;

type Entity<T extends Collection> = InstantEntity<typeof db, T, {}>;

export const newEntity = async <T extends Collection>(
    type: T,
    props: Omit<Entity<T>, "id">,
) => {
    const theId = id();
    await db.transact(db.tx[type][theId].update(props));
    return theId;
};

// seems good
const alice = await newEntity("users", {    email: "alice@gmail.com",});

// Should be a typing error but is not
const bob = await newEntity("users", {    blabla: "bob@gmail.com",});

// Should be a typing error but is not
const eve = await newEntity("users", {  email: 123,});
```

uriva:
also I wanted to add links parameter, but was not sure how to type it.

stopachka:
Hey @uriva , I just published `0.17.3`, with a new utility type, which could help here: 

Here's a complete example: 

```typescript
import { i, id, init, UpdateParams } from "@instantdb/admin";

const _schema = i.schema({ entities: { users: i.entity({ email: i.string() }) } });
type _AppSchema = typeof _schema; 
interface AppSchema extends _AppSchema {} 
const schema: AppSchema = _schema; 

const db = init({
  adminToken: "...",
  appId: "...",
  schema,
});

type Collection = keyof typeof schema.entities;

type EntityUpdate<T extends Collection> = UpdateParams<typeof schema, T>;

export const newEntity = async <T extends Collection>(
  type: T,
  props: EntityUpdate<T>,
) => {
  const theId = id();
  await db.transact(db.tx[type][theId].update(props));
  return theId;
};

// seems good
const existing_attr_works = await newEntity("users", { email: "alice@gmail.com" });

// @ts-expect-error
const non_existing_attr_errors = await newEntity("users", { blabla: "bob@gmail.com" });

// @ts-expect-error
const wrong_type_errors = await newEntity("users", { email: 123 });

// to silence ts warnings
existing_attr_works;
non_existing_attr_errors;
wrong_type_errors;
```

After looking, I realized that InstaQLEntity doesn't fully satisfy `update`. I went ahead and expose the `update` utility type. I also exposed a `LinkParams` utility type, if you'd like to do the same for links. Let me know how it goes!

[...]

PR #653, opened 2024-12-22
[types] add UpdateParams and LinkParams

stopachka:
https://github.com/instantdb/instant/issues/493 

Uri mentioned issues trying to create generic functions for creating entities. 

I looked deeper into his repro, and the underlying issue was: 

```typescript
Omit<Entity<T>, 'id'>
```

Since technically `InstaQLEntity` can include links, this was not matching up with `UpdateParams`.  Perhaps we can hint better to typescript, that since we gave an empty object, `Entity` would have no links, and that _should_ pass for `UpdateParams`. 

For now I just exposed `UpdateParams` directly, and added Uri's example in our playground. 

@nezaj @dwwoelfel @tonsky

### DIFF
--- a/client/packages/admin/src/index.ts
+++ b/client/packages/admin/src/index.ts
@@ -45,6 +45,8 @@ import {
   type InstaQLEntity,
   type InstaQLResult,
   type InstantRules,
+  type UpdateParams,
+  type LinkParams,
 } from "@instantdb/core";
 
 import version from "./version";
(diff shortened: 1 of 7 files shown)


nezaj (approved):
Nice! Is it worth adding ant docs about this?

stopachka:
Good idea! This may be a bit too 'special purpose', but heck let's adding it in. I'll push a separate PR

PR #654, opened 2024-12-22
[types] Expand should ignore builtins

stopachka:
Nacho mentioned type issues when he put `Date` in his schema. 

I made an example with a `Date` type, and realized that our `Expand` utility broke out the type: 

I updated `Expand` to exclude builtin types, and my hope this will make Nacho's upgrade seamless. 

@dwwoelfel @nezaj @tonsky


nezaj (approved):
👌👌

PR #657, opened 2024-12-23
Protect datalog cache from early cancellation

dwwoelfel:
This prevents in-progress datalog queries from getting cancelled when there are multiple sessions waiting on the query.

Previously, we could get into a situation like this:

1. user-1 opens a session and runs query-a
2. user-2 opens a session and runs query-a
3. user-1 closes their session
4. we cancel all in-progress queries, including query-a
5. user-2 gets an error for query-a because the query was cancelled

This PR moves the query execution to a separate call stack where it won't get canceled if the parent is canceled. 

To keep the behavior where we cancel queries that exceed the `handle-receive-timeout-ms`, we add another future that is in the call stack. If it gets canceled and there are no other threads waiting on the query, then it will cancel the query.

### DIFF
--- a/server/src/instant/reactive/store.clj
+++ b/server/src/instant/reactive/store.clj
@@ -17,10 +17,16 @@
   (:require
    [clojure.string :as string]
    [datascript.core :as d]
-   [instant.util.coll :as ucoll]
+   [instant.jdbc.sql :as sql]
    [instant.lib.ring.websocket :as ws]
+   [instant.util.async :as ua]
+   [instant.util.coll :as ucoll]
+   [instant.util.exception :as ex]
    [instant.util.tracer :as tracer]
-   [instant.util.exception :as ex]))
+   [medley.core :refer [dissoc-in]])
+  (:import
+   (java.lang InterruptedException)
+   (java.util.concurrent CancellationException)))
 
 (declare store-conn)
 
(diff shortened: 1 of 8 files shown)


server/src/instant/reactive/store.clj
```diff
@@ -342,23 +347,112 @@
 ;; ------
 ;; datalog cache
 
-(defn swap-datalog-cache-delay! [conn app-id datalog-query delayed-call]
+(defn swap-datalog-cache! [conn app-id datalog-query-fn ctx datalog-query]
```
stopachka, line 350: 🤯 I salute you sir.

server/src/instant/reactive/store.clj
```diff
@@ -342,23 +347,112 @@
...
+                    (deliver (:cancel-signal @result-delay) true)))]
+
+    (tracer/add-data! {:attributes {:cache-hit (not= this-result-delay result-delay)
+                                    :realized (realized? (:promise @result-delay))}})
+
+    (when (= this-result-delay result-delay)
+      ;; We added it, so we must execute it
+      (let [stmt-tracker (sql/make-top-level-statement-tracker)
+            result-promise (:promise @result-delay)
+            work-fut (binding [ua/*child-vfutures* nil ;; Move future to a new "call-stack"
```
stopachka, line 403: Man, bind + statement-tracker makes for a really powerful abstraction.

server/test/instant/reactive/store_test.clj
```diff
@@ -21,5 +22,109 @@
...
+                                                     (try
+                                                       (deliver started true)
+                                                       @(promise)
+                                                       (catch Throwable t
+                                                         (deliver err t))))
+                                                   nil
+                                                   q))]
+        @started
+        (future-cancel f1)
+        (is (instance? java.lang.InterruptedException (deref err 100 :timeout)))))
```
stopachka, line 75: nit: I looked at the test failure: should this say here: 

```
(is (= :timeout (deref err 100 :timeout)
```

stopachka (approved):
Beautiful work @dwwoelfel !

server/test/instant/reactive/store_test.clj
stopachka, line 75: Ah interesting, so `err` is taking more than 100ms to deliver. 🤔

stopachka (approved):
So darn cool!

PR #658, opened 2024-12-24
Send diffs for presence

tonsky:
Uses editscript to compute a diff between the old and new presence data and sends a diff to the client instead of sending the full presence object.

### DIFF
--- a/client/packages/core/src/Reactor.js
+++ b/client/packages/core/src/Reactor.js
@@ -13,7 +13,7 @@ import { buildPresenceSlice, hasPresenceResponseChanged } from "./presence";
 import { Deferred } from "./utils/Deferred";
 import { PersistedObject } from "./utils/PersistedObject";
 import { extractTriples } from "./model/instaqlResult";
-import { areObjectsDeepEqual } from "./utils/object";
+import { areObjectsDeepEqual, assocIn } from "./utils/object";
 import { createLinkIndex } from "./utils/linkIndex";
 import version from "./version";
 
@@ -460,7 +460,11 @@ export default class Reactor {
         break;
       case "refresh-presence":
         const roomId = msg["room-id"];
-        this._setPresencePeers(roomId, msg.data);
+        if (msg["edits"]) {
+          this._updatePresencePeers(roomId, msg["edits"]);
+        } else {
+          this._setPresencePeers(roomId, msg["data"]);
+        }
         this._notifyPresenceSubs(roomId);
         break;
       case "server-broadcast":
@@ -1676,6 +1680,23 @@ export default class Reactor {
     handler.cb(slice);
   }
 
+  _updatePresencePeers(roomId, edits) {
+    const peers = this._presence[roomId]?.result?.peers || {};
+    let sessions = Object.fromEntries(
+      Object.entries(peers).map(([k, v]) => [k, {data: v}]),
+    );
+    sessions[this._sessionId] = {data: this._presence[roomId]?.result?.user};
+    for (let [path, op, value] of edits) {
+      console.log('edit', path, op, value);
+      if (op === '+' || op === 'r') {
+        sessions = assocIn(sessions, path, value);
+      }
+      // TODO op === '-'
+    }
+
+    this._setPresencePeers(roomId, sessions);
+  }
+
   _setPresencePeers(roomId, data) {
     const sessions = { ...data };
     // no need to keep track of `user`
@@ -1684,9 +1705,7 @@ export default class Reactor {
       Object.entries(sessions).map(([k, v]) => [k, v.data]),
     );
 
-    this._presence[roomId] = this._presence[roomId] || {};
-    this._presence[roomId].result = this._presence[roomId].result || {};
-    this._presence[roomId].result.peers = peers;
+    this._presence = assocIn(this._presence, [roomId, 'result', 'peers'], peers);
   }
 
   // --------
--- a/client/packages/core/src/Reactor.js
+++ b/client/packages/core/src/Reactor.js
@@ -459,16 +459,18 @@ export default class Reactor {
         
         this._finishTransaction("synced", eventId);
         break;
-      case "patch-presence":
+      case "patch-presence": {
         const roomId = msg["room-id"];
         this._patchPresencePeers(roomId, msg["edits"]);
         this._notifyPresenceSubs(roomId);
         break;
-      case "refresh-presence":
+      }
+      case "refresh-presence": {
         const roomId = msg["room-id"];
         this._setPresencePeers(roomId, msg["data"]);
         this._notifyPresenceSubs(roomId);
         break;
+      }
       case "server-broadcast":
         const room = msg["room-id"];
         const topic = msg.topic;
--- a/client/packages/core/src/Reactor.js
+++ b/client/packages/core/src/Reactor.js
@@ -61,6 +61,7 @@ const ignoreLogging = {
   "set-presence": true,
   "set-presence-ok": true,
   "refresh-presence": true,
+  "patch-presence": true,
 };
 
 function querySubsFromJSON(str) {
@@ -458,13 +459,14 @@ export default class Reactor {
         
         this._finishTransaction("synced", eventId);
         break;
+      case "patch-presence":
+        const roomId = msg["room-id"];
+        this._patchPresencePeers(roomId, msg["edits"]);
+        this._notifyPresenceSubs(roomId);
+        break;
       case "refresh-presence":
         const roomId = msg["room-id"];
-        if (msg["edits"]) {
-          this._updatePresencePeers(roomId, msg["edits"]);
-        } else {
-          this._setPresencePeers(roomId, msg["data"]);
-        }
+        this._setPresencePeers(roomId, msg["data"]);
         this._notifyPresenceSubs(roomId);
         break;
       case "server-broadcast":
@@ -1680,7 +1682,7 @@ export default class Reactor {
     handler.cb(slice);
   }
 
-  _updatePresencePeers(roomId, edits) {
+  _patchPresencePeers(roomId, edits) {
     const peers = this._presence[roomId]?.result?.peers || {};
     let sessions = Object.fromEntries(
       Object.entries(peers).map(([k, v]) => [k, {data: v}]),
--- a/client/packages/core/src/Reactor.js
+++ b/client/packages/core/src/Reactor.js
@@ -13,7 +13,7 @@ import { buildPresenceSlice, hasPresenceResponseChanged } from "./presence";
 import { Deferred } from "./utils/Deferred";
 import { PersistedObject } from "./utils/PersistedObject";
 import { extractTriples } from "./model/instaqlResult";
-import { areObjectsDeepEqual, assocIn } from "./utils/object";
+import { areObjectsDeepEqual, assocIn, dissocIn } from "./utils/object";
 import { createLinkIndex } from "./utils/linkIndex";
 import version from "./version";
 
@@ -1687,11 +1687,12 @@ export default class Reactor {
     );
     sessions[this._sessionId] = {data: this._presence[roomId]?.result?.user};
     for (let [path, op, value] of edits) {
-      console.log('edit', path, op, value);
       if (op === '+' || op === 'r') {
         sessions = assocIn(sessions, path, value);
       }
-      // TODO op === '-'
+      if (op === '-') {
+        sessions = dissocIn(sessions, path);
+      }
     }
 
     this._setPresencePeers(roomId, sessions);
--- a/server/deps.edn
+++ b/server/deps.edn
@@ -25,6 +25,7 @@
         cider/cider-nrepl {:mvn/version "0.45.0"}
         amazonica/amazonica {:mvn/version "0.3.167"}
         juji/editscript {:mvn/version "0.6.4"}
+        version-clj/version-clj {:mvn/version "2.0.3"}
 
         org.bouncycastle/bcprov-jdk15on {:mvn/version "1.70"}
         incanter/incanter {:mvn/version "1.9.3"
--- a/server/deps.edn
+++ b/server/deps.edn
@@ -24,6 +24,7 @@
         org.slf4j/slf4j-api {:mvn/version "2.0.13"}
         cider/cider-nrepl {:mvn/version "0.45.0"}
         amazonica/amazonica {:mvn/version "0.3.167"}
+        juji/editscript {:mvn/version "0.6.4"}
 
         org.bouncycastle/bcprov-jdk15on {:mvn/version "1.70"}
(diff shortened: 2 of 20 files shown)


tonsky:
Main functionality is working. Two more things:

- [ ] I do want to build a demo that logs everything so I can check that what I think is happening is actually happening
- [ ] Clean up old in-memory impl (might be a separate PR)

server/src/instant/reactive/ephemeral.clj
```diff
@@ -24,93 +24,99 @@
...
-                           EntryRemovedListener
-                           (entryRemoved [_ event]
-                             (handle-event store-conn event))
-
-                           EntryUpdatedListener
-                           (entryUpdated [_ event]
-                             (handle-event store-conn event)))
+(defn init-hz
+  ([store-conn]
+   (init-hz {}))
```
stopachka: Assuming here we want to do `init-hz store-conn {}` 🫡

server/src/instant/reactive/session.clj
```diff
@@ -546,61 +547,43 @@
...
-                                        (inc (or (:skipped-size last-entry) 0)))}
-                    {:entries-to-process (conj entries-to-process last-entry)
-                     :last-entry entry}))
-                {:entries-to-process []
-                 :last-entry (first batch)}
-                (rest batch))]
-    (conj entries-to-process last-entry)))
-
-(defn straight-jacket-process-receive-q-batch [store-conn eph-store-atom batch metadata]
+(defmulti consolidate
```
stopachka, line 565: cool!

client/packages/core/src/Reactor.js
```diff
@@ -460,7 +460,11 @@ export default class Reactor {
         break;
       case "refresh-presence":
         const roomId = msg["room-id"];
-        this._setPresencePeers(roomId, msg.data);
+        if (msg["edits"]) {
```
stopachka: One issue I realize: what happens to old clients? 

As it stands, I _think_ if the client doesn't upgrade their version of the client SDK, they won't be able to handle the new `refresh-presence` events from the server. 

--- 

Potential solution: 

Clients currently send a `versions` map when they init: 

https://github.com/instantdb/instant/blob/main/client/packages/core/src/Reactor.js#L1056

This would be the first time we use it, but perhaps what we can do is: 
1. We detect if the client is on a version of Instant > x
2. In that case, we start sending the `refresh-presence` events with edits in them. 
  a. Aside, perhaps we send different kinds of events too -- something like `reset-presence` and 'patch-presence', to keep the paths separate, but I don't feel strongly about this idea

--- 

What do you think?

server/src/instant/reactive/ephemeral.clj
```diff
@@ -121,122 +127,45 @@
...
-                                               :app-id app-id
-                                               :room-id room-id
-                                               :data room-data
-                                               :session-id sess-id})))))))
+  (let [room-key                        (.getKey event)
+        {:keys [app-id room-id]}        room-key
+        {:keys [session-ids last-data]} (get-in @room-maps [:rooms room-key])]
+    (when (seq session-ids)
+      (let [room-data (.get (get-hz-rooms-map) room-key)
+            edits     (when last-data
```
stopachka, line 147: One rough thought I had: 

Is it necessary to derive edits? Perhaps we could broadcast `set-presence` events similar, to how we broadcast `server-broadcast` events. 

This would mean our resolution would be a bit less granular, but we wouldn't have to do a calculation to derive the edits. 

--- 

This could be something we layer on down the road though, as am open to either solution.

stopachka (commented):
Awesome! 

Main comment was about handling backwards-compatibility. Rest looks good to me, but deferring to @dwwoelfel's look, as he is the hazelcast expert.

client/packages/core/src/Reactor.js
tonsky: Yes, will do

server/src/instant/reactive/ephemeral.clj
tonsky: What do you mean by derive edits?

tonsky:
TODO:

- [x] Calculate and send deltas
- [x] Apply deltas on a client
- [x] Remove eph-state-atom
- [x] Update current tests
- [ ] Check client version and do not send deltas if too old
- [ ] New event type (patch-presence)
- [ ] Some tests for patch-presence
- [ ] I am not sure I understand how room-maps is updated when events come from network. Need to double-check

server/src/instant/reactive/ephemeral.clj
stopachka: My meaning was: 

Here we run `editscript/diff` to derive the changes between `last-data` and `room-data` 

I had a rough thought along the lines of: 

In order to change `room-data`, we already get events like `set-presence`, which is kind of like a small 'edit'. 

Perhaps instead of diffing the states, we could instead communicate all the different `set-presence` events to sessions. We'd have to do some kind of batching, but the advantage would be, that we wouldn't need to run the diff algo. 

This may require a larger change to the existing code though, so perhaps it's not a good idea to do in this PR, but something to think about.

tonsky: I see. Tricky to do it right now, since changes are no longer local and Hazelcast operates on full room maps, not deltas. We can establish a second channel where we will send original commands but it’s not obvious to me it’s worth it. Let’s go with editscript for now and see if it has any performance implications

tonsky:
Ok I think I’m more or less happy with everything, @dwwoelfel @stopachka please review?

server/deps.edn
```diff
@@ -97,10 +100,19 @@
...
+                  :extra-deps {eftest/eftest {:mvn/version "0.6.0"}}
                   :main-opts ["-m" "instant.test-core"]
-                  :jvm-opts ["-Djdk.tracePinnedThreads=full"
+                  :jvm-opts ["-enableassertions"
+                             "-Djdk.tracePinnedThreads=full"
                              ;; print stack traces instead of "Full report at ..."
-                             "-Dclojure.main.report=stderr"]}
+                             "-Dclojure.main.report=stderr"
+                             ;; hazelcast
+                             "--add-modules" "java.se"
```
dwwoelfel, line 109: Should we also add these to the docker-compose.yml file so they'll run in production? https://github.com/instantdb/instant/blob/28d7ac35600841f9071b62d1385919f6ea1436a3/server/docker-compose.yml#L14

server/src/instant/core.clj
```diff
@@ -160,44 +160,45 @@
...
-  (indexing-jobs/start)
-  (when (= (config/get-env) :prod)
-    (log/info "Starting analytics")
-    (analytics/start))
-  (when (= (config/get-env) :prod)
-    (log/info "Starting daily metrics")
-    (daily-metrics/start))
-  (start)
-  (add-shutdown-hook))
+  (binding [*print-namespace-maps* false]
```
dwwoelfel, line 163: What's this for?

server/src/instant/reactive/session.clj
```diff
@@ -281,54 +283,70 @@
...
+  (assoc version-clj.qualifiers/default-qualifiers "dev" 7))
+
+(defn- handle-refresh-presence! [store-conn sess-id {:keys [app-id room-id data edits]}]
+  (let [version (-> (rs/get-versions @store-conn sess-id)
+                    (get core-version-key))]
+    (cond
+      (and edits (empty? edits))
+      :nop
+
+      (and edits (version/newer? version "v0.17.5" {:qualifiers qualifiers}))
```
dwwoelfel: We have `instant.util.semver/semver-compare` we could use to avoid pulling in another dependency. Usage example at https://github.com/instantdb/instant/blob/28d7ac35600841f9071b62d1385919f6ea1436a3/server/test/instant/util/semver_test.clj#L17

server/src/instant/reactive/ephemeral.clj
```diff
@@ -23,65 +17,81 @@
...
     (.setSerializerConfigs serialization-config
                            hz-util/serializer-configs)
 
     (.setGlobalSerializerConfig serialization-config
                                 hz-util/global-serializer-config)
 
-    (let [hz (Hazelcast/getOrCreateHazelcastInstance config)
-          local-member (.getLocalMember (.getCluster hz))
-          hz-rooms-map (.getMap hz "rooms-v2")
+    (let [hz                 (Hazelcast/newHazelcastInstance config)
```
dwwoelfel: Why the switch from `getOrCreateHazelcastInstance` to `newHazelcastInstance`?

server/src/instant/reactive/ephemeral.clj
```diff
@@ -121,122 +133,44 @@
...
-                                               :session-id sess-id})))))))
+  (let [room-key                        (.getKey event)
+        {:keys [app-id room-id]}        room-key
+        {:keys [session-ids last-data]} (get-in @room-maps [:rooms room-key])]
+    (when (seq session-ids) ;; if we have sessions in that room locally
+      (let [room-data (.get (get-hz-rooms-map) room-key)
+            edits     (when last-data
+                        (editscript/get-edits
+                         (editscript/diff last-data room-data {:algo :a-star :str-diff :none})))]
+        (swap! room-maps assoc-in [:rooms room-key :last-data] room-data)
```
dwwoelfel: Is there an opportunity for a race here where `last-data` changes on us between when we calculate the edits and when we set it to `room-data`? Maybe swap-vals! would be safer?

In general, how can we be sure that the edits are applied in order?

server/src/instant/reactive/ephemeral.clj
```diff
@@ -23,65 +17,81 @@
...
-(defonce hz-ops-q (atom nil))
-
 ;; room-maps keeps track of the rooms each session is in (for easy removal
 ;; on session close) and some info about the rooms we're subscribed to on
 ;; this machine
 ;; {:sessions {<session-id>: #{<room-id>}}
-;;  :rooms {<{app-id: app-id, room-id: room-id}> {:session-ids #{<sess-id>}}}}
-(defonce room-maps (atom {}))
+;;  :rooms {<{app-id: app-id, room-id: room-id}> {:session-ids #{<sess-id>}
+;;                                                :last-data <...>}}}
```
dwwoelfel: It's a bit of a shame to have to store `last-data` here because we'll be duplicating the data that's already in hazelcast. 

If we set the `includeValue` field to `true` in `addEntryListener` (https://github.com/instantdb/instant/blob/28d7ac35600841f9071b62d1385919f6ea1436a3/server/src/instant/reactive/ephemeral.clj#L101), we could use `(.getOldValue event)` and `(.getValue event)` to get the diff.

That might also be less prone to races. 

The downside is that there would be more useless cross-traffic between the instances. Right now, if all sessions are on the same machine, then we don't have to send the map data to every instance (not sure what Hazelcast is actually sending under the hood).

server/src/instant/reactive/ephemeral.clj
```diff
@@ -281,203 +208,63 @@
                  (empty? session-ids) (dissoc-in [:rooms room-key])
                  (seq session-ids) (assoc-in [:rooms room-key :session-ids]
                                              session-ids)))))
-
     (hz-util/remove-session! (get-hz-rooms-map) room-key sess-id)))
 
-(defn clean-old-sessions []
```
dwwoelfel, line 287: We need to keep this in case we somehow miss a leave-room, otherwise there's no way to trim that data from Hazelcast and it will keep growing forever.

server/src/instant/reactive/session.clj
```diff
@@ -42,7 +44,7 @@
 ;; Setup
 
 (declare receive-q-stop-signal)
-(def handle-receive-timeout-ms 5000)
+(def handle-receive-timeout-ms 10000)
```
dwwoelfel: Is this increase intentional?

dwwoelfel (commented):
Could you add a feature flag to enable/disable globally and for an individual app? Just in case there is something unexpected that causes issues with production traffic that we didn't encounter in dev.

server/deps.edn
tonsky: Done

server/src/instant/core.clj
tonsky: I thought namespaced maps are generally disliked and people usually turn them off. Do you want to keep it?

server/src/instant/reactive/session.clj
tonsky: Done

server/src/instant/reactive/ephemeral.clj
tonsky: This is an artefact from when I tried to create isolated HZ instances for test. Works either way, so I returned getOrCreateHazelcastInstance

server/src/instant/reactive/ephemeral.clj
tonsky: You mean manually from REPL? I checked that that fn is not called from anywhere?

server/src/instant/reactive/session.clj
tonsky: Yes, otherwise tests with HZ fail with timeout on CI

server/src/instant/reactive/ephemeral.clj
dwwoelfel: Yeah, manually from the repl.

server/src/instant/reactive/session.clj
dwwoelfel: Is there some other way to fix the tests? We rely on this timeout to prevent our system from getting bogged down by long-running queries. We have to be very careful about increasing it, because if we find it's too long and we want to decrease it, then we'll break all of the queries that relied on the longer timeout.

server/src/instant/core.clj
dwwoelfel: No need to keep it, I just didn't know what it does.

tonsky: Basically it prints maps like `{:ns/key 1 :ns/key2 2}` as `#:ns{:key 1 :key2 2}` and it’s enabled by default in Clojure. The benefit of it is very small, but it breaks grep and overall is very inconvenient to read

server/src/instant/reactive/ephemeral.clj
tonsky: I agree getValue/getOldValue is our best bet. It might also be the case that hazelcast broadcasts deltas anyways, or, if nobody uses map on a machine, not even sends events there. Anyways, let’s run and see how performance is, and react if there are trouble

server/src/instant/reactive/ephemeral.clj
tonsky: Switched to .getOldValue/.getValue, so this is not needed

tonsky:
Ok switched to .getOldValue/.getValue and addressed all other notes

dwwoelfel (approved):
LGTM!

stopachka:
Woohoo! Awesome PR Nikita!

Issue #459, opened 2024-11-12
`InstantEntity` does not allow dropping undefined keys


uriva:
e.g. for

```ts
type MyType = InstantEntity<typeof db, "someType">;
```

```ts
const introEvent: MyType = {
  field: 123,
  optionalField: undefined,
};
```

works
but
 
```ts
const introEvent: MyType = {
  field: 123,
};
```

will be a typing error.

stopachka:
Hey @uriva -- starting in 0.17.4, optional fields should work! Let me know how it goes.

PR #660, opened 2024-12-31
[ts] make optional fields work if the key is ommited

stopachka:
Uri pointed out an issue in https://github.com/instantdb/instant/issues/459 

If you have a `messages` entity like so:

```javascript
messages: i.entity({
  content: i.string(),
  createdAt: i.date().optional(),
}),
```

Then using `InstaQLEntity`: 

```typescript
type CoreMessage = InstaQLEntity<AppSchema, "messages">;

const mWithOptionalFieldWorks: CoreMessage = { 
  id: '1', 
  content: 'hello',
};
```

You end up having an error, because typescript _expects_ `createdAt` to be present, even though it's optional 

## Fix

I updated how we resolved attrs, so that instead of generating: `createdAt: Foo | undefined`, we generate: `createdAt?: Foo | undefined`

@nezaj @dwwoelfel @tonsky

### DIFF
--- a/client/packages/core/src/schemaTypes.ts
+++ b/client/packages/core/src/schemaTypes.ts
@@ -45,13 +45,6 @@ export class DataAttrDef<ValueType, IsRequired extends boolean> {
   // }
 }
 
-type ExtractValueType<T> =
-  T extends DataAttrDef<infer ValueType, infer isRequired>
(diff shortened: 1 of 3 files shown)


nezaj (approved):
Nice!

Issue #471, opened 2024-11-14
`db.useAuth().user` typing is wrong


uriva:
I think `db.useAuth().user;` typing is wrong, should be `User | null`.

stopachka:
Hey @uriva -- right now the type is `User | undefined`. To make sure I understand: is that you would like the difference between `undefined` and `null` to be explicit?

uriva:
It's been awhile since I opened this 

Null and undefined are different things in js

stopachka:
Starting in 0.17.5, the type when isLoading: false, error: null, will be `User | null`!

PR #662, opened 2025-01-02
[ts] improve the AuthState type

stopachka:
Uri mentioned that the typing for `useAuth` was incorrect: 

https://github.com/instantdb/instant/issues/471

In the case where isLoading: false and error: null, `user` should be typed as `User | null` rather than `User | undefined` 

@nezaj @dwwoelfel @tonsky


dwwoelfel (approved):
LGTM!

PR #664, opened 2025-01-02
Ordering fix

dwwoelfel:
Generalizes @stopachka's fix in https://github.com/instantdb/instant/pull/647 to work with ordering on arbitrary fields.

With this change we'll use the symbol map to look up the entity id that we need to use as our secondary ordering key. We could look it up from the page-pattern, but I passed it in directly (as `eid-sym`) to make things a little more explicit.

### DIFF
--- a/server/test/instant/db/instaql_test.clj
+++ b/server/test/instant/db/instaql_test.clj
@@ -4029,6 +4029,7 @@
 
             ;; Create conversations, messages
             convo-id-aid (random-uuid)
+            convo-order-aid (random-uuid)
             group-id-aid (random-uuid)
             convo-group-aid (random-uuid)
             msg-id-aid (random-uuid)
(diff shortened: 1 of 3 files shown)


server/test/instant/db/instaql_test.clj
```diff
@@ -4077,9 +4085,12 @@
                                          :value-type :ref
                                          :cardinality :one}]])
 
+            order (atom 0)
+
             ;; add 1 group, 3 conversations, with 5 messages each
             add-conversation (fn [g-id c-id]
                                [[:add-triple c-id convo-id-aid (str c-id)]
+                                [:add-triple c-id convo-order-aid (swap! order inc)]
```
stopachka, line 4093: gangster 😎

stopachka (approved):
Heeck yeah!

PR #665, opened 2025-01-03
Use the order-eid to create the cursor when the order-val is null

dwwoelfel:
Fixes a bug where we could return a null eid in the `start-cursor` or `end-cursor`.


stopachka (approved):
babam!

PR #666, opened 2025-01-03
Search and sort across pages on Explorer

dwwoelfel:
When we added pagination to the explorer, the sorting and filtering no longer worked across all data and only worked on the current page. Users found this surprising and some thought it was just broken. 

Now that we have custom order-by attributes and `$like` operators, we can sort and filter across all the data. The drawback is that we can only sort and filter typed+indexed data.

This restricts sorting to only typed+indexed attrs and filtering to type=string+indexed attrs. For filtering, we run a query that looks like: `str_attr_1 $like %search-string% || str_attr_2 $like %search-string%`.

One potentially weird thing--sorting by id actually sorts by `serverCreatedAt`.

Here's what it looks like:

https://github.com/user-attachments/assets/bc2e5f0b-12cc-460c-baa1-3586292ca38a


stopachka (approved):
Super cool!

PR #668, opened 2025-01-03
Support ISO 8601 date formats, not just date time

stopachka:
### Context

In postgres, we can save date values as strings:

```
SELECT triples_extract_date_value('"2025-01-03"');
```

### Problem

However, we weren't able to parse "2025-01-03" in java. We would get: 

```
; (err) Text '2025-01-03' could not be parsed at index 10
```

## Solution

This is because we used `java.time.Instant/parse`. I updated our codepath, so we now handle _both_ date time parsing, and date parsing. 

@dwwoelfel @nezaj @tonsky

### DIFF
--- a/server/src/instant/db/model/triple.clj
+++ b/server/src/instant/db/model/triple.clj
@@ -15,7 +15,7 @@
    [clojure.string :as string])
   (:import
    (java.util UUID)
-   (java.time Instant LocalDate ZoneOffset)))
+   (java.time Instant LocalDate ZonedDateTime ZoneOffset)))
 
 ;; (XXX): Currently we allow value to be nil
 ;; In the future, we may want to _retract_ the triple if the value is nil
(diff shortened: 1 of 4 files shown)


server/src/instant/db/model/triple.clj
```diff
@@ -708,8 +709,20 @@
       (tracer/add-data! {:attributes {:total-count @row-count}})
       {:row-count @row-count})))
 
+(defn iso8601-date-str->instant [x]
+  (if (string/includes? x "T")
+    (Instant/parse x)
```
dwwoelfel: We might want to do `(.toInstant (ZonedDateTime/parse x))` because `Instant/parse` only supports UTC time zones. So `(.toInstant (ZonedDateTime/parse "2025-01-02T00:00:00-08"))` works but `(Instant/parse "2025-01-02T00:00:00-08")` doesn't.

dwwoelfel (approved):
One comment, but LGTM!

PR #671, opened 2025-01-04
[lookups] support lookups in the `v` portion

stopachka:
### Context

We can already write queries that have lookups in the `e` position: 

```edn
[[handle-aid "stopa"] :users/email ?email]
```

### Problem

However, we _cannot_ write lookups in the `v` position: 

```edn
[?bookshelf :bookshelves/owner [handle-aid "stopa"]]
```

We particularly need this in cel.clj, as we write a query with ids in the `v` position.

### Solution 

I added support for lookups on the `v` side of our query. 

### Notes 

There are two things of note:

**How we detect lookups in the `v` side** 

Technically, a user _could_ have any jsonb value on the `v` side (However, afaik we don't allow jsonb values in where clauses). Still, we detect a lookup by specifically checking that it is: a) a vector b) first element is a uuid and c) it is count 2. 

**adding a av check** 

Previously, our lookup subquery did a search for entity_id, but did not have the `av` flag. I added this flag, as otherwise the subquery would lead to a table scan. 

@dwwoelfel @nezaj @tonsky

### DIFF
--- a/server/src/instant/db/datalog.clj
+++ b/server/src/instant/db/datalog.clj
@@ -556,7 +556,12 @@
        [:= :checked_data_type [:cast [:inline (name data-type)] :checked_data_type]]
        [:not= [(extract-value-fn data-type) :value] val]])))
 
-(defn- in-or-eq-value [idx v-set]
+(defn- value-lookup? [x]
+  (and (vector? x)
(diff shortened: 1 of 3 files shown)


dwwoelfel (approved):
LGTM!

PR #673, opened 2025-01-06
Don't override name in gc span

dwwoelfel:
Setting `name` in attributes overrode the `name` field, so we weren't seeing any "name=gc" events.


nezaj (approved):
🚢

PR #674, opened 2025-01-06
Send one span for all `invalidator/send-refresh`

dwwoelfel:
Follow-up to https://github.com/instantdb/instant/pull/670, where we noticed that `invalidator/work` is randomly pausing between some `invalidator/send-refresh` calls.

It doesn't appear to be caused by the garbage collector. The only other thing that happens between the calls to `send-refresh` is that we log the span. I've reduced it to a single `send-refreshes` span for all of the `:refresh` messages we enqueue instead of one per message. We'll see if this removes the unaccounted-for time.


nezaj (approved):
Nice!

PR #675, opened 2025-01-06
Calculate hostname once

dwwoelfel:
I think this might fix the issue identified in https://github.com/instantdb/instant/pull/670, where we see random pauses in between spans.

[According to the docs](https://docs.oracle.com/en/java/javase/20/docs/api/java.base/java/net/InetAddress.html#getHostName()), `getHostName` can do some i/o.

We might be pausing in `with-span` while we look up the hostname. The hostname shouldn't ever change, so it's better to just calculate it once in the delay, and then always use the same value.

### DIFF
--- a/server/src/instant/config.clj
+++ b/server/src/instant/config.clj
@@ -8,12 +8,13 @@
    (java.net InetAddress)
    (java.util UUID)))
 
-(defn get-hostname []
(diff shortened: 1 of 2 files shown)


stopachka (approved):
Heeck yeah!

PR #677, opened 2025-01-06
Add created_at index on transactions

nezaj:
What it says on the in! Adding this because the daily metrics job keeps timing out, with an index it can very quickly calculate the metrics!


dwwoelfel (approved):
LGTM!

PR #678, opened 2025-01-06
Add replica id to metrics

nezaj:
We had another migration on 12-06 and wanted to delete some inflated data from this metrics table but got an error due to no replica id being set. This PR enables it!


dwwoelfel (approved):
LGTM!

PR #680, opened 2025-01-06
Reset the number of receive-workers

dwwoelfel:
We saw a spike in timeouts after the increase. Need to figure that out before increasing.


stopachka (approved):
bam!

PR #681, opened 2025-01-07
Trigger connection status on network change

nezaj:
One of our users pointed out that `useConnectionStatus` wasn't updating when directly turning off wifi. Was able to confirm that doing so doesn't trigger the websocket onclose event. This should do the trick!


dwwoelfel (approved):
LGTM!

PR #684, opened 2025-01-07
Tell pnpm where to find the replace-images folder

dwwoelfel:
This lets us call `replace-images` for the newsletter without needing a separate `npm install` step.


nezaj (approved):
Nice!

PR #687, opened 2025-01-07
Use a concurrent linked queue in grouped-queue

dwwoelfel:
We hit a bottleneck with the `LinkedBlockingQueue`, where too many workers would cause throughput to drop.

This replaces the `LinkedBlockingQueue` with a `ConcurrentLinkedQueue`, which doesn't have the same problem with many concurrent workers. In my local testing, it has roughly the same performance with 100 workers as it does with 100,000 workers.

The `ConcurrentLinkedQueue` has no blocking `poll` method, so I created a new `start-grouped-queue-with-workers` helper that will create the grouped queue with a `process-fn` and maintain a set of workers that process items from the queue. The existing code didn't have to change much--just a new `on-add` callback you can pass to `grouped-queue/create` to get notified on `puts`.

The size of the worker pool expands and contracts as the number of items in the queue changes, up to `max-workers` that you set when you create the queue.

Also adds `com.clojure-goes-fast/clj-async-profiler` to dev to make it easier to do quick profiling.


stopachka (approved):
Beautiful!

PR #688, opened 2025-01-07
Handle case when core version is "null"

dwwoelfel:
Follow up to https://github.com/instantdb/instant/pull/686, looks like we sometimes get a `null` string as the version.


stopachka (approved):
🫡

PR #690, opened 2025-01-07
Track some wal latency metrics

dwwoelfel:
Tracks two latency metrics for the invalidator:

1. Queries the database every minute to calculate the difference in bytes between the last available wal entry and where our replication subscription is on the log. We'll send that metric with the rest of the `gauges`.
2. Calculates the number of milliseconds between when we process a transaction in the invalidator and when that transaction was created in the database.

There are still a lot of steps where latency could be introduced, but this gives us a little more insight into the system.

### DIFF
--- a/server/src/instant/core.clj
+++ b/server/src/instant/core.clj
@@ -181,7 +181,7 @@
     (stripe/init)
     (session/start)
     (inv/start-global)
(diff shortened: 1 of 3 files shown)


stopachka (approved):
Nice!

PR #691, opened 2025-01-08
Reduce the number of logs we generate

dwwoelfel:
Removes a few noisy fields from the logs we send to cloudwatch and a few spans that aren't very useful.

### DIFF
--- a/server/src/instant/util/logging_exporter.clj
+++ b/server/src/instant/util/logging_exporter.clj
@@ -54,7 +54,8 @@
                   "detailed_query"
                   "detailed_patterns"
                   "detailed_tx_steps"
-                  "process_id"})
+                  "process_id"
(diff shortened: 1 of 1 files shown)


server/src/instant/util/logging_exporter.clj
```diff
@@ -54,7 +54,8 @@
                   "detailed_query"
                   "detailed_patterns"
                   "detailed_tx_steps"
-                  "process_id"})
+                  "process_id"
+                  "query"})
```
dwwoelfel, line 58: Removes `query` from `datalog-query-reactive!`, which generates a lot of text, and `instaql/permissioned-query` (but we get the same query in `instaql/query-nested.forms`).

nezaj (approved):
Nice!

PR #693, opened 2025-01-08
[lookups] improve how lookups work with perms

stopachka:
# Setup 

Consider a zeneca app, with perms like: 

```json
{ 
   "users": { 
        "allow":  {
           "*":  "'Worldview' in data.ref('bookshelves.name')"
       }
    }
}
```

# Problems

We found three issues with how lookups and perms work together. 

## issue 1: updates cause a warn-io in preload-refs

If we tried to run an update to an existing entity:

```js
tx.users[lookup('handle', 'stopa')].update({fullName: "Stopa"})
```

The permission check would run, but we would end up getting a `io` warning. 

### reason: preload-refs has lookup, ref-fn has uuuid

The reason was that `preload-refs` would keep a mapping of: 

```edn
{ eid: [handle-uuid "stopa"]
  ... }
```

While `ref-fn` would contain the actual uuid for the object. 

This would cause us to skip preload-refs, and make a direct query. 

## issue 2: the create perm check would cause an error

If we used a lookup to create a new entity: 

```js
tx.users[lookup('handle', 'idontexist')].update({fullName: "Idonte Xist"})
```

The transaction would throw an error. 

### reason: parse-uuid in persistent vector

The reason this happened was because in `ref-fn`, we expected the `id` to be a string, but instead we got a vector lookup. This would throw when we tried to do: `(parse-uuid id)`

## issue 3: batch fn broken for lookups

You would think we _could_ fix issue 2, by simply handling the case where `id` is a lookup differently. Something like: 

```clj
parsed-id (if (sequential? id) (parse-lookup id) (parse-uuid id))
```

However, doing this would break our batch functions

### reason: no way to tie lookups to join rows

`get-ref-many` and our preload ref fn take a set of ids, and make a single query for them. We then partition the join rows for each id. 

However, if the input id is a lookup, and the child id is a uuid, there would be no way for us to to associate them together. This caused all batch-fns that included lookups to return empty results. 

## Solution

At a high level, I made the following changes: 

**For Updates: We use `preload-triples` to replace lookups with ids** 

Since we already have triples for all update and delete transactions, we replace the `lookup` eid in the check to the actual uuid. 

By doing this, preload-refs will now work.

**For creates: We query for the ids of all the lookups**

For creates, we run a separate query after insertion, to get all the uuids for the lookups. We then run the same code to replace the `eid` with the uuid. 

In order to do this, I made a change to how our `check` works. Instead of each `check` having an anonymous function, I moved the function out. This way, when we are replacing the `lookup` with the `uuid`, we don't have to worry about closures. 

### alternative 1: have `transact!` return lookup-refs

I considered having `transact!` return all the lookups that were involved. However, if we did this we would need to make the change across insert, merge, and delete. I wasn't sure the efficieny win would be worth it, but can be convinced otherwise! 

### larger alternative: resolve lookups early

Another alternative I considered, was to do the 'resolution' step early. For example, right in the beginning of the transaction, we could 'eagerly' get all lookup uuids, and create the lookups if they don't exist. 

From there, we could map over tx-steps, and replace all lookups with those ids. 

There's some danger here: we would be creating data before running the permission check. However, I _think_ that is fine, as we run create checks after we insert data anyways. 

Ultimately I chose to go against this option, as it would require a much larger change. We would need to insert attrs before triples, and we would need to do the same kind of logic in `transact!`. 

Maybe at some point it could make sense to look into this, but just to fix this bug I don't think it was worth it. 

--- 

@dwwoelfel @nezaj @tonsky

### DIFF
--- a/server/src/instant/db/cel.clj
+++ b/server/src/instant/db/cel.clj
@@ -391,7 +391,7 @@
                  patterns
                  results))))
 
-(def auth-ref-validator ^CelAstValidator 
+(def auth-ref-validator ^CelAstValidator
   (reify CelAstValidator
     (validate [_this ast _cel issues-factory]
       (doseq [^CelNavigableExpr node (-> ast
--- a/server/src/instant/db/model/triple.clj
+++ b/server/src/instant/db/model/triple.clj
@@ -43,6 +43,29 @@
 (s/def ::enhanced-triple
   (s/keys :req-un [::triple ::index ::md5]))
 
+(defn fetch-lookups->eid [conn app-id lookups]
+  (if-not (seq lookups)
+    {}
+    (let [lookups-set (set lookups)
+          triples (sql/execute!
+                   conn
+                   (hsql/format
+                    {:select :*
+                     :from :triples
+                     :where [:and
+                             [:= :app-id app-id]
+                             :av
+                             (list* :or
+                                    (map
+                                     (fn [[a v]]
+                                       [:and [:= :attr-id a] [:= :value [:cast (->json v) :jsonb]]])
+                                     lookups-set))]}))
+
+          lookups->eid (->> triples
+                            (map (fn [{:keys [entity_id attr_id value]}]
+                                   [[attr_id value] entity_id]))
+                            (into {}))]
+      lookups->eid)))
 ;; ---
 ;; insert-multi!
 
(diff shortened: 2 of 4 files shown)


dwwoelfel (approved):
LGTM!

PR #695, opened 2025-01-08
[lookup-perms] make sure view checks work

stopachka:
### Context 

Consider: 

```
orgs 
profiles
  handle ; unique
orgs.members <> profiles.org
```

Now imagine we have an _existing_ org and an _existing_ profile. 

## Problem 

If we write a lookup like so: 

```js
tx.profiles[tx.lookup('handle', 'stopa')].update({ fullName: "Stepanado" }).link({ orgs: orgId }  
```

We would get the error: "Record not found: attr",

### Reason 

This was because the link to `orgs` would generate: 
1. an :update check on org 
2. a :view check on profiles

The `:view` check would fail, because we passed in the `etype` as a keyword, while `ref-fn` expected it as a string. 

### Solution 

Passed in the `etype` as a keyword, that fixed it. 

@dwwoelfel @nezaj @tonsky

### DIFF
--- a/server/src/instant/db/permissioned_transaction.clj
+++ b/server/src/instant/db/permissioned_transaction.clj
@@ -138,7 +138,8 @@
 
 (defn object-view-check-fn [{:keys [etype program data] :as _check}
                             {:keys [current-user] :as ctx}]
-  (let [{:keys [original]} data]
+  (let [{:keys [original]} data
+        etype-str (name etype)]
     (if-not program
       true
       (cel/eval-program!
@@ -151,7 +152,7 @@
         "data"
         (cel/->cel-map {:type :data
                         :ctx ctx
-                        :etype etype}
+                        :etype etype-str}
                        original)}))))
 
 ;; Why do we have to decide whether something is an update or a create?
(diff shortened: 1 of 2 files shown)


server/src/instant/db/permissioned_transaction.clj
```diff
@@ -138,7 +138,8 @@
 
 (defn object-view-check-fn [{:keys [etype program data] :as _check}
                             {:keys [current-user] :as ctx}]
-  (let [{:keys [original]} data]
+  (let [{:keys [original]} data
+        etype-str (name etype)]
```
dwwoelfel, line 142: Why did we make it into a keyword in the first place? Looks like we do that below in `object-view-check`?

dwwoelfel (approved):
LGTM!

server/src/instant/db/permissioned_transaction.clj
stopachka: Good q! We did this previously on the [`check`](https://github.com/instantdb/instant/blob/794c34a1f17d71537c91b12cccf38dd4974faab1/server/src/instant/db/permissioned_transaction.clj#L90) . I had assumed that some other caller was relying on the etype being a keyword, but I'll do a check and remove the keywordiness if all looks good

PR #697, opened 2025-01-09
[sm] (keyword etype) -> etype

stopachka:
In our `view` objects, we converted the `etype` string to a keyword. 

Looking at how we use etype throughout permissioned-tx and beyond, it's generally treated as a string. 

I went ahead and looked at callsites (outside of permissioned-tx, check-results are mainly used for sandbox). Things look good, so let's push it

@dwwoelfel @nezaj @tonsky

### DIFF
--- a/server/src/instant/db/permissioned_transaction.clj
+++ b/server/src/instant/db/permissioned_transaction.clj
@@ -79,8 +79,7 @@
 
 (defn object-upsert-check-fn [{:keys [action program data etype] :as _check}
                               {:keys [current-user] :as ctx}]
-  (let [etype-str (name etype)
-        {:keys [original updated]} data]
+  (let [{:keys [original updated]} data]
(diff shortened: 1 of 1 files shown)


dwwoelfel (approved):
LGTM!

server/src/instant/db/permissioned_transaction.clj
```diff
@@ -341,12 +338,12 @@
 
 (defn attr-delete-check [_ctx _aid]
   {:scope :attr
-   :etype :attrs
+   :etype "attrs"
```
dwwoelfel, line 341: Kind of weird that we put an "attrs" etype here.

stopachka: Agreed -- we may want to move this into a `system` etype of sorts, something like $attrs at some point.

PR #698, opened 2025-01-09
:on-delete :cascade

tonsky:
- [x] Optimize current on-delete impl
- [x] Transform it into single recursive query
- [x] Add delete-entity support to client datalog
- [x] Add option to the explorer
- [x] Add option to CLI schema file

### DIFF
--- a/client/www/pages/docs/modeling-data.md
+++ b/client/www/pages/docs/modeling-data.md
@@ -264,6 +264,22 @@ Our micro-blog example has the following relationship types:
 - **One-to-many** between `comments` and `profiles`
 - **Many-to-many** between `posts` and `tags`
 
+### Cascade Delete
+
+If forward link has cardinality `one` (`one-to-one` and `one-to-many` relationships), it can have `onDelete: "cascade"`. In that case, when referenced entity is deleted, all its connected entities will be deleted too:
+
+```typescript
+postAuthor: {
+  forward: { on: "posts", has: "one", label: "author", onDelete: "cascade" },
+  reverse: { on: "profiles", has: "many", label: "authoredPosts" },
+}
+
+// this will delete profile and all its posts
+db.tx.profiles[user_id].delete();
+```
+
+Without `onDelete: "cascade"`, when deleting user, posts will just be unlinked from it, but keep existing.
+
 ## Publishing your schema
 
 Now that you have your schema, you can use the CLI to `push` it to your app:
--- a/client/www/pages/docs/modeling-data.md
+++ b/client/www/pages/docs/modeling-data.md
@@ -266,7 +266,7 @@ Our micro-blog example has the following relationship types:
 
 ### Cascade Delete
 
-If forward link has cardinality `one` (`one-to-one` and `one-to-many` relationships), it can have `onDelete: "cascade"`. In that case, when referenced entity is deleted, all its connected entities will be deleted too:
+If forward link is defined with `has: "one"`, it can have `onDelete: "cascade"`. In that case, when referenced entity is deleted, all its connected entities will be deleted too:
 
 ```typescript
 postAuthor: {
--- a/server/src/instant/db/model/triple.clj
+++ b/server/src/instant/db/model/triple.clj
@@ -1,6 +1,7 @@
 (ns instant.db.model.triple
   (:require
    [clojure.spec.alpha :as s]
+   [clojure.string :as string]
    [honey.sql :as hsql]
    [instant.data.constants :refer [empty-app-id]]
    [instant.db.model.attr :as attr-model]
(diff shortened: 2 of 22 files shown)


tonsky:
Progress so far:

- [x] Wrote a bench for recursive delete
- [x] Optimized `delete-entity-multi!`

server/src/instant/db/transaction.clj
```diff
@@ -141,36 +146,119 @@
...
+    (concat delete-entity-steps' tx-steps')))
+
+(defn expand-delete-entity-cascade [conn app-id tx-steps]
+  (let [[delete-entity-steps tx-steps'] (coll/split-by #(= :delete-entity (first %)) tx-steps)]
+    (if (empty? delete-entity-steps)
+      tx-steps
+      (let [ids+etypes           (map next delete-entity-steps)
+            query                (str "WITH RECURSIVE entids (entity_id, etype) AS (
+                                       -- start with entid
+                                         VALUES " (string/join ", " (repeat (count ids+etypes) "(?, ?)")) "
```
dwwoelfel: I would use honeysql here to keep things consistent. I know you're only inserting question marks, but seeing `str` around sql queries makes me a little nervous.

server/src/instant/db/transaction.clj
```diff
@@ -181,39 +269,50 @@
...
+                :add-triple
+                (assoc acc :add-triple (triple-model/insert-multi! conn attrs app-id args))
+
+                :deep-merge-triple
+                (assoc acc :deep-merge-triple (triple-model/deep-merge-multi! conn attrs app-id args))
+
+                :retract-triple
+                (assoc acc :retract-triple (triple-model/delete-multi! conn app-id args))))
             {}
             (batch tx-steps))
```
dwwoelfel: Seems like this might be simpler if `batch` did a group-by instead of whatever it's currently doing.

server/src/instant/db/transaction.clj
```diff
@@ -141,36 +146,119 @@
...
+            [attr-id value] eid,
+            ...}"
+  [conn app-id lookups]
+  (if (empty? lookups)
+    {}
+    (let [lookups (map (fn [[a v]] [a (->json v)]) lookups)
+          query   {:with [[[:lookups {:columns [:attr_id :value]}]
+                           {:values lookups}]]
+                   :from :triples
+                   :where [:= :app_id app-id]
```
dwwoelfel, line 139: You probably want to check for `:triples.av` either here or in the join to make the query faster (we know that a lookup has to be on a unique attr, so it gets the `av` index).

server/src/instant/db/transaction.clj
```diff
@@ -141,36 +146,119 @@
...
+                                       -- recurse all 'cascade' refs
+                                       UNION
+                                         SELECT triples.entity_id , forward_ident.etype
+                                         FROM   triples
+                                         JOIN   attrs                   ON triples.attr_id = attrs.id
+                                         JOIN   idents AS forward_ident ON forward_ident.id = attrs.forward_ident
+                                         JOIN   idents AS reverse_ident ON reverse_ident.id = attrs.reverse_ident
+                                         JOIN   entids                  ON to_jsonb(entids.entity_id) = triples.value
+                                         WHERE  triples.app_id = ?
+                                         AND    attrs.on_delete = 'cascade'
```
dwwoelfel: I wonder if you could avoid the join on attrs in the database by using the `attrs` object that we pass in to transact!

server/src/instant/db/transaction.clj
```diff
@@ -181,39 +269,50 @@
...
+                :add-attr
+                (assoc acc :add-attr (attr-model/insert-multi! conn app-id args))
+
+                :delete-attr
+                (assoc acc :delete-attr (attr-model/delete-multi! conn app-id args))
+
+                :update-attr
+                (assoc acc :update-attr (attr-model/update-multi! conn app-id args))
+
+                :delete-entity-no-cascade
```
dwwoelfel: Why have `:delete-entity-no-cascade` as an operation? Do we expect users to use it or is it just for our internal usage?

tonsky: The idea here is that we don’t try to expand `:delete-entity` second time, since some methods call `transact-without-tx-conn!` directly and some `permissioned-transaction/transact!`

tonsky:
To the best of my abilities, I made cascade expansion as a single query but it still sometimes triggers slow path.

One of the culprits, I think, is `[:= [:to_jsonb :entity_id] :value]` join. This could be sped up by introducing `value_ref` column that contains parsed uuid. Might benefit many other places, too, anywhere where we need to follow a reference, in fact.

I left a TODO, maybe plan it as a separate task?

tonsky:
Okay I think store.js support for cascade is in, + tests

server/src/instant/db/model/triple.clj
```diff
@@ -477,58 +478,45 @@
...
+                                   [:= :idents.app-id [:param :system-catalog-app-id]]]]}]]
+                :delete-from :triples
+                :where       [:in :ctid
+                              {:union
+                               [{:nest {:select :* :from :forward_attrs}}
+                                {:nest {:select :* :from :reverse_attrs}}]}]
+                :returning   :*}
+        params {:app-id app-id
+                :system-catalog-app-id system-catalog-app-id}]
+    (tool/time* (str "delete-entity-multi! " (count id+etypes) " entities")
```
dwwoelfel: debug code

server/src/instant/db/model/triple.clj
```diff
@@ -477,58 +478,45 @@
...
+                (sql/do-execute! conn (hsql/format query {:params params})))))
```
dwwoelfel: Whoa, I didn't know you could pass `:params` like that! I wonder if it would prevent sending the same parameter twice if you also set `:numbered true`.

server/src/instant/db/transaction.clj
```diff
@@ -141,79 +125,173 @@
...
+  (if (empty? entity-ids)
+    {}
+    (let [query {:from            :triples
+                 :join            [:attrs [:= :triples.attr_id :attrs.id]
+                                   :idents [:= :attrs.forward_ident :idents.id]]
+                 :where           [:and
+                                   [:= :triples.app_id app-id]
+                                   [:in :triples.entity_id entity-ids]]
+                 :select-distinct [:triples.entity_id :idents.etype]}
+          res (tool/time* "resolve-etypes"
```
dwwoelfel: debug code

server/src/instant/db/transaction.clj
```diff
@@ -141,79 +125,173 @@
...
+                                        [:refs {:select [:triples.entity_id :triples.value :cascade.forward_etype :cascade.reverse_etype]
+                                                :from   :triples
+                                                :join   [:cascade [:= :triples.attr_id :cascade.id]]
+                                                :where  [:and
+                                                         :triples.vae
+                                                         [:= :triples.app_id app-id]]}]]
+                       :from   :entids
+                       :select :*})
+        #__            #_(println (sql/analyze conn query+args))
+        res          (tool/time* "expand-delete-entity-cascade"
```
dwwoelfel: debug code

server/src/instant/db/model/triple.clj
```diff
@@ -11,7 +11,8 @@
    [instant.util.json :refer [->json <-json]]
    [instant.util.spec :as uspec]
    [instant.util.string :refer [multiline->single-line]]
-   [instant.util.tracer :as tracer])
+   [instant.util.tracer :as tracer]
+   [tool])
```
dwwoelfel: debug code

dwwoelfel (approved):
LGTM!

client/www/pages/docs/modeling-data.md
```diff
@@ -264,6 +264,22 @@ Our micro-blog example has the following relationship types:
 - **One-to-many** between `comments` and `profiles`
 - **Many-to-many** between `posts` and `tags`
 
+### Cascade Delete
+
+If forward link has cardinality `one` (`one-to-one` and `one-to-many` relationships), it can have `onDelete: "cascade"`. In that case, when referenced entity is deleted, all its connected entities will be deleted too:
```
nezaj: Nit. Maybe we can just say "If a forward link is defined with `has: "one"` it can have " `onDelete: "cascade"`

I found myself slightly confused with the "one-to-many" call-out because in my mind usually user->posts is a one-many and in this case posts->users feels more like a many-to-one (since posts has only one author).

stopachka (commented):
So cool!

server/src/instant/db/model/triple.clj
tonsky: Unfortunately, no

```
(hsql/format
 {:select :*
  :from :triples
  :where [:or
          [:= :e [:param :p]]
          [:= :a [:param :p]]
          [:= :v [:param :p]]]}
 {:params {:p 1}
  :numbered true})
; => ["SELECT * FROM triples WHERE (e = $1) OR (a = $2) OR (v = $3)" 1 1 1]
```

tonsky: Can also do this though:

```
(hsql/format
 {:select :*
  :from :triples
  :where [:or
          [:= :e :?p]
          [:= :a :?p]
          [:= :v :?p]]}
 {:params {:p 1}})
 ```

PR #699, opened 2025-01-10
[datalog] use `IN` for value comparison

stopachka:
A user mentioned that some of his queries got slower. The query looked something like: 

```edn
{:conversations 
  {:$ {:order {:serverCreatedAt "desc"}, :limit 100, 
         :where {:groups "...", :messages.time {:$gte ...}}}, 
  :messages {}}}
```

After investigating, the problem was like so: 

`messages` expanded out to about 16000 entities. When we tried to resolve permissions, we would do a batched query, which would include all 16K uuids. 

When datalog generated the list of `:or` conditions,  it would take postgres about a minute and a half, just to parse the query. 

Changing the condition to an `IN` improved the perf to under 5 seconds. 

@dwwoelfel @nezaj @tonsky

### DIFF
--- a/server/src/instant/db/datalog.clj
+++ b/server/src/instant/db/datalog.clj
@@ -555,11 +555,6 @@
        [:= :checked_data_type [:cast [:inline (name data-type)] :checked_data_type]]
        [:not= [(extract-value-fn data-type) :value] val]])))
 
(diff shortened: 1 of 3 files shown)


dwwoelfel (approved):
Nice!

PR #700, opened 2025-01-10
fix: logick for next link in bottom article nav

ivklgn:
now:

Screenshot 2025-01-10 at 12 58 50

after fix:

Screenshot 2025-01-10 at 12 57 34


stopachka (approved):
Thank you!

PR #701, opened 2025-01-10
Support exact matches and <, > in filter input

dwwoelfel:
In addition to free-form searching for string-indexed attributes, we also support exact search and (for indexed numbers or dates) `<`, `>` comparison attrs:

https://github.com/user-attachments/assets/4a98b790-7418-44d5-9d2c-00a7dbd8b092

Potential improvements:

It might not be obvious to people what the `:` operator means? If we gave them some examples using real data, e.g. show `pageCount:100, pageCount>100` in the dropdown, maybe that would help? It was a little too complex for me to tackle just now.

We're pretty permissive in the parsing. For example, if you type `number: number:hello number:100`, we'll ignore the incomplete `number:` and the invalid `number:hello` and just show you `number = 100`. Some better error handling so that we could show you where the query is invalid would be nice.

Support negation.

Support AND (we only do OR right now). 

Support free-form search and exact search at the same time.

### DIFF
--- a/client/www/components/dash/explorer/Explorer.tsx
+++ b/client/www/components/dash/explorer/Explorer.tsx
@@ -72,7 +72,6 @@ function parseSearchQuery(s: string): ParsedQueryPart[] {
     }
     if ((OPERATORS as readonly string[]).includes(c)) {
       if (currentPart && valueStart != null) {
-        // XXX: Probably fails if there is no space before?
         currentPart.value = s.substring(valueStart, fieldStart).trim();
         parts.push(currentPart);
       }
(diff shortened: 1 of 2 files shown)


client/www/components/dash/explorer/Explorer.tsx
```diff
@@ -36,75 +42,325 @@ import { DBAttr, SchemaAttr, SchemaNamespace } from '@/lib/types';
...
+            e.preventDefault();
+
+            const active = activeOption.current || comboOptions[0];
+            if (active) {
+              completeQuery(active.display);
+            }
+          }
+        }}
+        placeholder="Filter..."
+      ></ComboboxInput>
```
stopachka: nit: afaik you can complete this element with `/>`, as it has no children

stopachka (approved):
Fricking awesome!

PR #705, opened 2025-01-13
Remove placement group

dwwoelfel:
Removes the placement group so that we can try deploying into another availability zone.

Tested in staging and it did appear to deploy an instance without a placement group.


stopachka (approved):
nice!

PR #708, opened 2025-01-14
Don't treat isNull as a constant

dwwoelfel:
When we do an `$isNull` query, we put something like `{:$isNull {:attr-id aid}}` in the value position. The pattern looks something like `[?e aid {:$isNull {:attr-id aid}}}]`.

When we choose the triples index to look at, we use the `av` index if we have a constant value for an indexed attr. Normally a non-constant is a symbol, e.g. `?v` in `[?e aid ?v]`. We saw a map for `$isNull` instead of a symbol, so we thought it was a constant value and we used the `av` index where the value wasn't actually known.

Now we'll check for `$isNull` before we say something is a constant.

### DIFF
--- a/server/src/instant/db/instaql.clj
+++ b/server/src/instant/db/instaql.clj
@@ -183,14 +183,15 @@
                             (grow-paths path)))})
 
         (and (map? v) (contains? v :$isNull) (= true (:$isNull v)))
-        ;; If the where cond has `$isNull=true`, then we
-        ;; need it should match if any of the intermediate
-        ;; paths are null
-        (let [path (string/split (name k) #"\.")]
-          {:or (concat [[[path v]]]
-                       (map (fn [p]
-                              [[p {:$isNull true}]])
-                            (grow-paths path)))})
+        ;; If the where cond has `$isNull=true`, then we need it to
+        ;; match if any of the intermediate paths are null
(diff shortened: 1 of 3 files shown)


server/src/instant/db/instaql.clj
```diff
@@ -183,14 +183,15 @@
...
-                            (grow-paths path)))})
+        ;; If the where cond has `$isNull=true`, then we need it to
+        ;; match if any of the intermediate paths are null
+        (let [path (string/split (name k) #"\.")
+              conds (map (fn [p]
+                           [[p {:$isNull true}]])
+                         (grow-paths path))]
+          (if (= 1 (count conds))
+            {:and conds}
+            {:or conds}))
```
dwwoelfel, line 194: The changes here fix a bug where we would do an `or` with the same query twice.

stopachka (approved):
awh yeah!

server/src/instant/db/instaql.clj
dwwoelfel: There's still a slowdown when using ors that I'm looking into, though.

stopachka: One hunch I always check: if there's any accidental cross join in the intermediate CTEs

PR #709, opened 2025-01-14
Run the receive-queue workers without bindings

dwwoelfel:
I noticed that our traces for `invalidator/work` were including a lot of extra spans past the end of the top-level trace. This happens because we start the worker in an `on-add` callback to `grouped-queue/put!`, so it was inheriting the parent span from the bindings.

Now we have a new type of virtual future that doesn't propagate bindings. That gives our workers a bit more isolation. I also gave them their own virtual thread executor--doesn't do much now, but it would let us more thoroughly "shutdown" a grouped-queue in the future.

Much easier to review with whitespace off: https://github.com/instantdb/instant/pull/709/files?w=1

### DIFF
--- a/server/src/instant/grouped_queue.clj
+++ b/server/src/instant/grouped_queue.clj
@@ -151,38 +151,41 @@
                                                 process-fn
                                                 max-workers]
                                          :or {max-workers 2}}]
-  (let [workers (atom #{})
+  (let [executor (ua/make-virtual-thread-executor)
+        workers (atom #{})
         ;; Use a promise so we can access it in the `on-add` function
         grouped-queue (promise)
         on-add (fn []
                  (when (< (count @workers) max-workers)
-                   (ua/vfuture
-                     (loop [worker-id (Object.)]
-                       (when (contains? (swap! workers
-                                               (fn [workers]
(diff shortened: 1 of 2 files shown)


server/src/instant/grouped_queue.clj
```diff
@@ -217,15 +222,15 @@
                             (inflight-queue-reserve (max 1 (rand-int 25)) iq))
               :process-fn (fn [_ workset]
                             @started
-                            (tracer/record-info! {:name "workset"
-                                                  :attributes {:workset-count (count workset)
-                                                               :total total-items
-                                                               :worker-count ((:get-worker-count @gq))}})
+                            (clojure.tools.logging/info {:name "workset"
```
stopachka, line 225: Nit: is this intentional?

stopachka (approved):
Nice!!

server/src/instant/grouped_queue.clj
dwwoelfel: Yeah, that's just a function for testing (it's in a comment block) and I didn't want to overload honeycomb with debug traces.

stopachka: Ah, makes sense! 🫡

PR #711, opened 2025-01-14
Send gauges periodically instead of on each span

dwwoelfel:
The main change here is to have each machine send gauges every 30 seconds instead of attaching them to each span. In my testing, that reduces the time it takes to build a span by about half.

The downside is that we'll get a little less granularity on the gauge metrics, but the upside is that the spans create faster and we'll be sending much less data.

There are also some other smaller performance improvements in the logging tracer: 

1. Use a batch processor in prod for the logging exporter (flushes at least every half second)
2. Remove all reflection warnings
3. Use String.replace instead of a regex to remove newlines in prod
4. Use a StringBuilder to build the attrs string

### DIFF
--- a/server/src/instant/gauges.clj
+++ b/server/src/instant/gauges.clj
@@ -2,12 +2,14 @@
   "Repeatedly saves metric information in `tracer/last-calculated-metrics`."
   (:require
    [clojure.string :as str]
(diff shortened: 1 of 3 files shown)


stopachka (approved):
Niice!

PR #712, opened 2025-01-15
[Metrics] Increase metrics job timeout

nezaj:
What it says on the tin. This should give enough time for the daily job to run and calculate new activity for mulitple days (e.g. for monday computing friday, sat, and sun metrics).


dwwoelfel (approved):
LGTM!

PR #713, opened 2025-01-15
Add another date format for triples

dwwoelfel:
Also makes it easier to add new formats, just add another function to the `date-parsers` list.


stopachka (approved):
Niice!

PR #714, opened 2025-01-15
Use the grouped queue in the invalidator

dwwoelfel:
Updates the invalidator to use the grouped queue. This should do two things to give instant users a better experience:

1. Each app gets its own logical queue, so it should isolate apps a bit more so that all apps don't slow down invalidation if one app dumps a ton of transactions in the queue
2. Records in each app's queue are combined. If an app sends a ton of updates faster than they can be processed, then we'll generate less load on the system when we handle their transactions.

The per-app isolation is still incomplete because the store-conn will be a bottleneck and we don't have any queuing in transact. It should still be a big improvement, as long as we don't run into too much contention on the store-conn.

### DIFF
--- a/server/src/instant/reactive/invalidator.clj
+++ b/server/src/instant/reactive/invalidator.clj
@@ -5,6 +5,8 @@
    [instant.config :as config]
    [instant.db.model.attr :as attr-model]
    [instant.db.pg-introspect :as pg-introspect]
+   [instant.gauges :as gauges]
+   [instant.grouped-queue :as grouped-queue]
    [instant.jdbc.aurora :as aurora]
    [instant.jdbc.wal :as wal]
    [instant.model.app :as app-model]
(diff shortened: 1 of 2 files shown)


server/src/instant/reactive/invalidator.clj
```diff
@@ -308,34 +337,61 @@
...
+                        :attributes {:app-id app-id
+                                     :tx-id tx-id
+                                     :wal-record-count record-count
+                                     :wal-latency-ms (wal-latency-ms wal-record)}}
+
+      (try
+        (let [sockets (invalidate! process-id store-conn wal-record)]
+          (tracer/add-data! {:attributes {:num-sockets (count sockets)}})
+          (doseq [{:keys [id]} sockets]
+            (tracer/with-span! {:name "invalidator/send-refresh"
```
stopachka: maybe send-refreshes?

dwwoelfel: Whoops! Thanks for catching!

PR #715, opened 2025-01-15
Add more replication debugging

dwwoelfel:
Makes two changes to our replication to help us keep track of everything:

1. Sets the reporting status interval to 1 second (default seems to be 10 seconds) so that we get a more accurate view of our wal byte latency
2. Logs the size of the transaction in bytes in `invalidator/work`. I'm curious to see how the size of the transaction affects the latency.

Also removes some reflection warnings from wal.clj.


stopachka (approved):
super cool!

PR #716, opened 2025-01-15
Use the zgc garbage collector

dwwoelfel:
We're currently using g1c and see 30-50ms pauses every ~45 seconds.

image

This switches us to zgc, which should reduce the number of stop-the-world pauses.

We'll be able to look at the `gc` events to see how many pauses we get.


stopachka (approved):
Heeck yeah!

PR #717, opened 2025-01-15
Put us back in the placement group

dwwoelfel:
We disabled the placement group while we were transitioning to a different az.

Let's see if this drops the presence latency.


stopachka (approved):
🫡

PR #718, opened 2025-01-15
Attach latency_ms to e2e events

dwwoelfel:
Tracks the latency between when we created the transaction and when we execute a step in the invalidation pipeline.

This will let us create a metric we can use to track the experience of "click on checkbox in one window and watch it update in another window" and create alerts if it's too slow. 

It's not perfect, because we don't know when the trace is fully finished, but tracking it on send-refresh or json-delivered is pretty close.

I attempted to do this with Honeycomb's derived columns, but there doesn't appear to be any way to calculate the difference between a span's timestamp and the root timestamp.


stopachka (approved):
Heeck yeah!

PR #719, opened 2025-01-16
[Storage] Paginate storage tab

nezaj:
Right now the storage tab only fetches the first 1k files from s3 for an app. This caused some confusion when files weren't showing up [[Discord]](https://discord.com/channels/1031957483243188235/1329216151288414352)

This PR adds pagination logic for instances when there are >1k files in a path.

### DIFF
--- a/server/src/instant/storage/s3.clj
+++ b/server/src/instant/storage/s3.clj
@@ -45,7 +45,17 @@
 (defn list-app-objects
   ([prefix] (list-app-objects default-bucket prefix))
   ([bucket-name prefix]
-   (s3/list-objects-v2 {:bucket-name bucket-name :prefix prefix})))
+   (loop [all-objects []
+          continuation-token nil]
+     (let [with-pagination (fn [opts] (if continuation-token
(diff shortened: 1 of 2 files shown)


server/src/instant/storage/s3.clj
```diff
@@ -45,7 +45,16 @@
 (defn list-app-objects
   ([prefix] (list-app-objects default-bucket prefix))
   ([bucket-name prefix]
-   (s3/list-objects-v2 {:bucket-name bucket-name :prefix prefix})))
+   (loop [all-objects []
```
dwwoelfel, line 48: What happens if there are millions of objects, will we return them all?

nezaj: At the moment it will keep going, we could add a limit but we're going to kill the storage tab soon as well with storage being integrated into triples

nezaj: Add a limit of 50k for now!

dwwoelfel (approved):
LGTM!

PR #721, opened 2025-01-16
[rooms][1] move logic out of class, and have it accept args

stopachka:
https://www.dropbox.com/scl/fi/2l37me2hrjldtil896008/Problem-Dynamic-Hooks.paper?rlkey=4evh9t1hsc08nncu7ni629g07&dl=0

Part 1 of our rooms change. 

This moves all the function inside `InstantReactRoom`'s class, into a static spots outside. The functions now take the `room` as an argument.

This change shouldn't make a difference yet. I will start using the static functions in a separate PR.

@dwwoelfel @nezaj @tonsky


client/packages/react/src/InstantReactAbstractDatabase.ts
```diff
@@ -57,6 +57,195 @@ export type TypingIndicatorHandle<PresenceShape> = {
...
+    const unsub = room._core._reactor.subscribeTopic(
+      room.id,
+      topic,
+      (event, peer) => {
+        onEvent(event, peer);
+      },
+    );
+
+    return unsub;
+  }, [room.id, topic]);
```
stopachka, line 85: I was definitely surprised here: looks like when we subscribe, we don't actually care about the room type. This will probably lead to bugs when a user tries to connect to different room types with the same id. 

Noting to look deeper into this

client/packages/react/src/InstantReactAbstractDatabase.ts
```diff
@@ -57,6 +57,195 @@ export type TypingIndicatorHandle<PresenceShape> = {
...
+
+function usePublishTopic<
+  RoomSchema extends RoomSchemaShape,
+  RoomType extends keyof RoomSchema,
+  TopicType extends keyof RoomSchema[RoomType]["topics"],
+>(
+  room: InstantReactRoom<any, RoomSchema, RoomType>,
+  topic: TopicType,
+): (data: RoomSchema[RoomType]["topics"][TopicType]) => void {
+  useEffect(() => room._core._reactor.joinRoom(room.id), [room.id]);
```
stopachka, line 96: Another interesting thing here -- I didn't realize we _always_ sent `join-room` events, whenever we mounted a hook like this. I am not sure it's a problem, but stuff like this could explain why we see so many join-room events

dwwoelfel (approved):
LGTM!

PR #722, opened 2025-01-16
[sm] fix react warning on fences

stopachka:
Noticed this error from react: 

CleanShot 2025-01-15 at 16 55 48@2x

Made a quick update so we don't spread the key. 

@dwwoelfel @nezaj @tonsky


dwwoelfel (approved):
LGTM!

PR #723, opened 2025-01-16
Undo zgc and placement group

dwwoelfel:
Remove a couple of variables while we figure out what is causing the deploys to go haywire.


stopachka (approved):
Sgtm

PR #724, opened 2025-01-16
db.rooms

stopachka:
This PR: 

Moves our functions from `db.room()...` to `db.rooms.` 

We add deprecation notices to `db.room()` callsites, but there should be no breaking changes to the upgrade. 

@dwwoelfel @nezaj @tonsky 

**Some notes:** 

The type signatures on room functions are a bit obtuse right now. We have _a lot_ of generic parameters, which I don't necessarily think we need. I worry changing it would cause breaking changes though, so I wanted to try and tackle that a bit later.


dwwoelfel (approved):
Nice work!

PR #726, opened 2025-01-16
Add memoization in db.rooms hooks

stopachka:
We had a few instances where we would return functions that didn't have stable values. This would cause consumers to re-render unnecesarily. I looked over the our hooks and added `useCallback` and `useMemo` in places that could keep stable values.

@dwwoelfel @nezaj @tonsky


dwwoelfel (approved):
LGTM!

PR #728, opened 2025-01-16
Log the span id in prod

dwwoelfel:
This will make it easier to track down the log from the span attrs we add to the query (added in #725)


stopachka (approved):
Heck yeah!

PR #729, opened 2025-01-16
[react-native] use IndexedDBStorage on web

stopachka:
When using `@instantdb/react-native` with `react-native-web`, a user reported a 'quota exceeded' error: 

CleanShot 2025-01-16 at 14 02 33@2x

This was because AsyncStorage used `localStorage` for web: https://github.com/react-native-async-storage/async-storage/blob/main/packages/default-storage/src/AsyncStorage.ts#L89

I went ahead and made it so in web, instead of using `react-native-async-storage`, we use our tried and true `IndexedDBStorage` interface. 

To do this, we take advantage of React Native's [platform specific extensions](https://reactnative.dev/docs/platform-specific-code#native-specific-extensions-ie-sharing-code-with-nodejs-and-web) 

@dwwoelfel @nezaj @tonsky

### DIFF
--- a/client/packages/react-native/src/NetworkListener.js
+++ b/client/packages/react-native/src/NetworkListener.js
@@ -7,7 +7,6 @@ export default class NetworkListener {
   }
   static listen(f) {
     return NetInfo.addEventListener((state) => {
-      console.log("connection change", state.isConnected);
       f(state.isConnected);
     });
   }
--- a/client/pnpm-lock.yaml
+++ b/client/pnpm-lock.yaml
@@ -291,6 +291,9 @@ importers:
       '@babel/runtime':
         specifier: ^7.20.0
         version: 7.26.0
+      '@expo/metro-runtime':
+        specifier: ~4.0.0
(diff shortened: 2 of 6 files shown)


client/packages/react-native/src/NetworkListener.js
```diff
@@ -7,7 +7,6 @@ export default class NetworkListener {
   }
   static listen(f) {
     return NetInfo.addEventListener((state) => {
-      console.log("connection change", state.isConnected);
       f(state.isConnected);
```
stopachka, line 10: this was a dangling console.log -- removed

client/pnpm-lock.yaml
```diff
@@ -291,6 +291,9 @@ importers:
       '@babel/runtime':
         specifier: ^7.20.0
         version: 7.26.0
+      '@expo/metro-runtime':
```
stopachka, line 294: I made it so our sandbox works on web too -- 

```
pnpm run start
w # to open in web
```

dwwoelfel (approved):
Cool!

PR #731, opened 2025-01-17
Faster hazelcast startup in tests

dwwoelfel:
I had a flaky test run because hazelcast was slow to start. I found this stackoverflow post explaining how to start Hazelcast in standalone mode: https://stackoverflow.com/a/65594539/361969

Now Hazelcast starts in ~50ms for tests instead of ~3 seconds.

### DIFF
--- a/server/src/instant/reactive/ephemeral.clj
+++ b/server/src/instant/reactive/ephemeral.clj
@@ -20,7 +20,8 @@
                                EntryRemovedListener
                                EntryUpdatedListener)
    (com.hazelcast.topic ITopic MessageListener Message)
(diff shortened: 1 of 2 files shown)


server/src/instant/reactive/ephemeral.clj
```diff
@@ -37,19 +38,14 @@
 
 (declare handle-broadcast-message)
 
-(comment
```
stopachka, line 40: oi, thank you!

stopachka (approved):
yass

tonsky:
Nice!

PR #732, opened 2025-01-18
Use threads for cpu-heavy invalidator workers

dwwoelfel:
I noticed a lot of time is spent waiting in `handle-receive` for simple things like `:refresh-presence` or `:server-broadcast` and I have a hunch that it's because the invalidator is doing a lot of cpu-heavy work in the virtual threads. Then the virtual thread we create to call `handle-receive` gets blocked waiting for the cpu-heavy task to finish.

This PR makes the invalidator queue use a fixed set of workers, each inside a future. It seemed as fast as the virtual thread version in my testing, as long as the number of threads wasn't too high.

### DIFF
--- a/server/src/instant/grouped_queue.clj
+++ b/server/src/instant/grouped_queue.clj
@@ -4,7 +4,7 @@
    [instant.util.async :as ua]
    [instant.util.tracer :as tracer])
   (:import
-   (java.util.concurrent ConcurrentLinkedQueue)
+   (java.util.concurrent ConcurrentLinkedQueue Semaphore)
    (java.util.concurrent.atomic AtomicInteger)
    (clojure.lang PersistentQueue)))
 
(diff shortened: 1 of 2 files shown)


server/src/instant/grouped_queue.clj
```diff
@@ -187,6 +187,40 @@
      :get-worker-count (fn [] (count @workers))
      :virtual-thread-executor executor}))
 
+(defn start-grouped-queue-with-cpu-workers [{:keys [group-fn
+                                                    reserve-fn
+                                                    process-fn
+                                                    worker-count]
+                                             :or {worker-count 2}}]
+  (let [semaphore (Semaphore. 0)
```
stopachka, line 195: gangster 😎

stopachka (approved):
Beautiful! TIL about using a semaphore for coordinating threads in this way!

PR #733, opened 2025-01-18
༼ つ ◕_◕ ༽つ Give Storage

nezaj:
There's still a few things to do but wanted to put up a PR for this. I'm going to restructure the commits as well to make this easier to review!

At a high-level these are the new things you'll be be able to do with Storage

## Query for them like you would for other namespaces

image

And then get back urls you can use directly

image

## You can interact with $files in the explorer

https://github.com/user-attachments/assets/c7be7581-9cfc-4c56-ae66-a4525c8ca278

## transactions for $files are locked-down, you can only link/unlink in the reverse direction

**Admin SDK**

image

**Sandbox**

image
image

## Now our storage metrics will use $files data as well

image

More to come, but here's some of the remaining todos

**Remaining todos**
- [x] Go through and clean-up XXX notes and kill old code as needed
- [x] Change how get-signed-upload-url works to be backwards compatible
- [x] Support setting content type and content disposition
- [x] Add new storage example
- [x] Verify working with old storage  
- [x] Add tests for admin sdk
- [x] Split up PR for review
- [x] Create document for testing new changes
- [x] Address feedback
- [ ] Update docs
- [ ] Write migration for existing users
- [ ] Add createdBy reference?


client/packages/admin/src/index.ts
```diff
@@ -617,7 +621,67 @@ class Storage {
...
+   *   const isSuccess = await db.storage.uploadFile('photos/demo.png', buffer);
+   */
+  uploadFile = async (
+    filename: string,
+    file: Buffer,
+    metadata: UploadMetadata = {},
+  ): Promise<any> => {
+    const formData = new FormData();
+    formData.append('path', filename);
+    formData.append('file', new Blob([file]));
```
dwwoelfel: Make sure to append the `file` field last so that our server can parse all of the fields it needs to start the upload before we get to the `file` part of the stream. Otherwise we have to slurp the full file before we can get to the content-type.

server/src/instant/admin/routes.clj
```diff
@@ -366,37 +367,30 @@
...
-        data (storage-util/create-signed-upload-url! app-id filename)]
-    (response/ok {:data data})))
-
-;; Retrieves all files that have been uploaded via Storage APIs
-(defn files-get [req]
+(defn upload-post [req]
   (let [{app-id :app_id} (req->admin-token! req)
-        subdirectory (-> req :params :subdirectory)
-        data (storage-util/list-files! app-id subdirectory)]
+        params (w/keywordize-keys (get-in req [:multipart-params]))
```
dwwoelfel: I wouldn't bother keywordizing keys, we can access them as strings just as easily.

server/src/instant/model/app_file.clj
```diff
@@ -0,0 +1,106 @@
...
+(comment
+  (delete-by-ids! {:app-id #uuid "831355ee-6a59-4990-8ef3-9c9fe7c26031"
+                   :ids [#uuid "cbda1941-d192-4f7d-b0a7-f9d428e1ca0b"]})
+  (let [{app-id :app_id path :path} (get-by-path {:app-id #uuid "831355ee-6a59-4990-8ef3-9c9fe7c26031"
+                                                  :path "circle_red.jpg"})]
+    (delete-by-path! {:app-id app-id :path path})))
+
+(defn get-usage* [conn app-id]
+  {:pre [(or (nil? app-id) (uuid? app-id))]}
+  (let [fm-attr (attr-model/resolve-attr-id $system-attrs "$files" "metadata")
```
dwwoelfel: This should be a lot faster to compute if we put the size as its own separate field and type it as a number. Then you should be able to do `triples_extract_number_value` and it might be able to just read the values directly from the index.

client/packages/admin/src/index.ts
```diff
@@ -592,7 +592,11 @@ class Auth {
   }
 }
 
-type UploadMetadata = { contentType?: string } & Record<string, any>;
+type UploadMetadata = {
+  contentType?: string
+  contentDisposition?: string
+} & Record<string, any>;
```
stopachka: nit: can remove Record<string, any>

client/packages/admin/src/index.ts
```diff
@@ -617,7 +621,67 @@ class Storage {
...
    * @example
    *   const buffer = fs.readFileSync('demo.png');
-   *   const isSuccess = await db.storage.upload('photos/demo.png', buffer);
+   *   const isSuccess = await db.storage.uploadFile('photos/demo.png', buffer);
+   */
+  uploadFile = async (
+    filename: string,
+    file: Buffer,
+    metadata: UploadMetadata = {},
+  ): Promise<any> => {
```
stopachka: nit: would return the a type from this

client/packages/core/src/index.ts
```diff
@@ -314,50 +314,70 @@ class Auth {
   };
 }
 
+export type FileOpts = {
+  contentType?: string;
+  contentDisposition?: string;
+} & Record<string, any>;
```
stopachka: nit: imo can remove

stopachka (commented):
Exciting!

client/packages/admin/src/index.ts
nezaj: Went with doing a `PUT` and killed off `FormData`

server/src/instant/model/app_file.clj
nezaj: Done!

server/src/instant/admin/routes.clj
nezaj: Done!

client/packages/admin/src/index.ts
nezaj: Done!

client/packages/core/src/index.ts
nezaj: Done!

server/resources/migrations/49_add_app_upload_urls.up.sql
```diff
@@ -0,0 +1,7 @@
+CREATE TABLE app_upload_urls (
```
dwwoelfel: Can you add an index on `app_upload_urls (app_id)`. That way we don't have to do a full table scan when an app is deleted.

server/src/instant/admin/routes.clj
```diff
@@ -366,37 +367,30 @@
...
-        subdirectory (-> req :params :subdirectory)
-        data (storage-util/list-files! app-id subdirectory)]
+        params (:headers req)
+        path (ex/get-param! params ["path"] string-util/coerce-non-blank-str)
+        file (ex/get-param! req [:body] identity)
+        content-type (ex/get-optional-param! params [:content-type] string-util/coerce-non-blank-str)
+        data (storage-coordinator/upload-file! {:app-id app-id
+                                                :path path
+                                                :content-type content-type
+                                                :skip-perms-check? true} file)]
```
```suggestion
                                                :skip-perms-check? true}
                                                file)]
```

The last `file` param is hard to notice when it's on the same line as the map.

server/src/instant/dash/routes.clj
```diff
@@ -931,52 +921,26 @@
...
-        data (storage-util/delete-file! app-id filename)]
+        params (:headers req)
+        path (ex/get-param! params ["path"] string-util/coerce-non-blank-str)
+        file (ex/get-param! req [:body] identity)
+        data (storage-coordinator/upload-file!
+              {:app-id app-id
+               :path path
+               :file file
+               :content-type (:content-type file)
+               :skip-perms-check? true} file)]
```
```suggestion
               :skip-perms-check? true}
               file)]
```

Same as above

dwwoelfel (approved):
Couple of minor comments. Looks great to me!

client/packages/admin/src/index.ts
nezaj: Done!

PR #734, opened 2025-01-20
Enable clj-async-profiler in production

dwwoelfel:
Adds what we need to enable clj-async-profiler in production (following the instructions at https://clojure-goes-fast.com/kb/profiling/clj-async-profiler/production/).

Tested in staging and all looks ok.


stopachka (approved):
Lesdothis!

PR #735, opened 2025-01-20
Update clj-async-profiler version

dwwoelfel:
Updates clj-async-profiler version in prod

### DIFF
--- a/server/test/instant/reactive/session_test.clj
+++ b/server/test/instant/reactive/session_test.clj
@@ -51,7 +51,9 @@
           :process-fn (partial session/process-fn store-conn)})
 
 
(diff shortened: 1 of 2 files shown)


server/test/instant/reactive/session_test.clj
```diff
@@ -92,9 +94,8 @@
           (session/on-close store-conn socket)
           (session/on-close store-conn socket-2)
           (finally
-            (when (realized? eph-hz)
```
dwwoelfel, line 95: What might be happening is that the delay is triggered, but when we get here it hasn't finished initializing, so `realized?` is false.

stopachka (approved):
Yess!

PR #736, opened 2025-01-20
Catch PopState Event (goback) and Redirect to popNavStack

mreyesh:
Minor UX enhancement:

Catch PopState Event (goback) and Redirect to popNavStack so that if you are exploring in the explorer and traverse a path of links, you can go back by using the go back browser button or any shortcuts from the keyboard or trackpad/mouse.

### DIFF
--- a/client/www/components/dash/explorer/Explorer.tsx
+++ b/client/www/components/dash/explorer/Explorer.tsx
@@ -487,6 +487,20 @@ export function Explorer({
     }
   }, [namespaces === null]);
 
+  useEffect(() =>{
+    function handlePopState(e: PopStateEvent) {
+        if (selectedNamespace && currentNav && allItems && showBackButton) {
+            e.preventDefault();
+            popNavStack();
+        }
+    }
+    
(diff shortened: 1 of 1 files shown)


nezaj:
Thanks for the PR @mreyesh ! 

I noticed there's a slight bug with this. I can make things work like so:

* Start at home-page, click dashboard, go to explorer, find a namespace with a link, traverse the link
* Going backwards and forwards should work!

However navigation will break like so:

* After traversing a link go backwards enough to end up on the home page
* Now click forward button, you'll be back in explorer, and if you traverse a link, hitting the back button takes you to the home page

I think the problem lies with how we use `router.replace` -- I think we may want to switch to just using `push` throughout instead.

nezaj:
@mreyesh Wanted to let you know we merged in https://github.com/instantdb/instant/pull/945 https://github.com/instantdb/instant/pull/946 which adds better browser navigation for the explorer and deep-linking!

Will go ahead and close this PR, appreciate you flagging this!

PR #737, opened 2025-01-21
Faster store operations

dwwoelfel:
I noticed a lot of time spent doing datalog store operations when looking at the profiler in production.

This improves some of the store operations by replacing `d/find` with `d/datoms` in a few places.

We also store a ref to the `instaql-query` in the `:subscription` entity instead of storing the instaql query itself. That lets us skip an expensive check in `add-instaql-query!` where we have to look up subscription by session-id and query. Now we can just look them up by the instaql-query entity id, since it already has the session-id in it.


stopachka (approved):
This is so exciting!

PR #739, opened 2025-01-21
[do-not-merge] Don't trigger update if triple value doesn't change

dwwoelfel:
Uses merge into to prevent updating triples in `insert-multi!` when the value doesn't change.

We can't use this yet because we need support for `returning` from `merge into`, which was added in Postgres 17.

honey.sql doesn't have support for merge-into, so we add some bare-bones extensions to support it.


stopachka (approved):
Cool!

dwwoelfel:
This is blocked for now because the statement-level trigger is not getting all of the rows when we combine a merge with an insert in the CTE.

Reported the bug to psql-bugs at https://www.postgresql.org/message-id/19380-4e293be2b4007248%40postgresql.org

Here's the repro:

```
CREATE TEMP TABLE merge_bug_test (id INT PRIMARY KEY, val TEXT);

-- Create trigger functions that list the IDs they see
CREATE OR REPLACE FUNCTION report_insert_rows()
RETURNS TRIGGER AS $$
BEGIN
    RAISE WARNING '[AFTER INSERT TRIGGER] Row count: %, Rows: %', (SELECT COUNT(*) FROM newrows), (SELECT COALESCE(json_agg(row_to_json(newrows))::text, 'EMPTY') FROM newrows);
    RETURN NULL;
END;
$$ LANGUAGE plpgsql;

-- Create statement-level triggers for INSERT
CREATE TRIGGER insert_trigger
    AFTER INSERT ON merge_bug_test
    REFERENCING NEW TABLE AS newrows
    FOR EACH STATEMENT
    EXECUTE FUNCTION report_insert_rows();

-- MERGE inserts rows, but INSERT CTE inserts nothing
WITH
  input_triples (id, val) AS (
    VALUES (1, 'a'), (2, 'b')
  ),
  insert_cte AS (
    INSERT INTO merge_bug_test (id, val) values (3, 'c')
    RETURNING id
  ),
  -- Insert two row with merge
  merge_cte AS (
    MERGE INTO merge_bug_test t
    USING input_triples s
    ON t.id = s.id
    WHEN NOT MATCHED THEN
        INSERT (id, val) VALUES (s.id, s.val)
    RETURNING t.id
  )
  -- Insert one row with a regular insert

SELECT id FROM merge_cte
UNION ALL
SELECT id FROM insert_cte;

DROP TABLE merge_bug_test;
DROP FUNCTION report_insert_rows;
```

dwwoelfel:
The fix is in the postgres codebase https://github.com/postgres/postgres/commit/b4307ae2e540847f84147d45b258440eaf570536

Hopefully it gets included with the next minor release on February 12, which might make it to Aurora in March.

dwwoelfel:
No longer needed--now we use a `where` check, which is a lot simpler and doesn't need merge.

PR #740, opened 2025-01-22
Fix flaky session tests

dwwoelfel:
We were getting some extra puts on the websocket channel stub and it was blocking the puts on our test. 

Updated the put to be non-blocking. To make sure we pull the message we're interested in off of the channel, the `blocking-send-msg` test function takes the expected op and it will keep polling the channel until it finds the message it's looking for (or until a timeout).

### DIFF
--- a/server/test/instant/reactive/session_test.clj
+++ b/server/test/instant/reactive/session_test.clj
@@ -97,12 +97,16 @@
             (when @realized-eph?
               (HazelcastInstance/.shutdown (:hz @eph-hz)))))))))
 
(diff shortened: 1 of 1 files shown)


stopachka (approved):
lesgoo

tonsky:
I don’t get it. You read message from the channel, compare it with expected and, if not matched, put it back? But this will break message ordering? As I understand it, websocket guarantees ordering, so it’s okay to rely on it in tests. It feels like test should be written more carefully: e.g. skipping extra messages that can unpredictably be inserted, but checking that expected messages arrive in the right logical order

PR #741, opened 2025-01-22
Only create a subscription if we're interested in the query

dwwoelfel:
Bug fix for https://github.com/instantdb/instant/pull/737.

We can get in a situation where the instaql-query is removed from the store (via session closing or remove-query), but we are still in the process of running datalog queries for the query. In that case we throw a noisy error when trying to insert the subscription because we're referencing an instaql-query that doesn't exist.

Now we won't add a subscription to the datalog topics if we're running the query.


stopachka (approved):
Nice!

PR #743, opened 2025-01-22
Add support for read-only connections

dwwoelfel:
Updates `aurora/conn-pool` to take a param indicating whether you want to use the connection for `:read` or `:write`. If you request a `:read` and try to do a write on it, then you'll get an error like `cannot execute INSERT in a read-only transaction`.

We use this to allow reads while we're in the process of failing over to an upgraded database.

In the future, we could potentially use this to direct reads to a secondary.


stopachka (approved):
Great idea!!

PR #745, opened 2025-01-23
cleanup http logs

dwwoelfel:
Cleans up the http logs a bit:

1. Wraps the stripe route middleware in `compojure.core/wrap-routes` so that it doesn't get applied to all routes (we were logging some things twice)
2. Adds a `:route` field to the span that has the actual route (e.g. `/apps/:app_id` instead of `/apps/a349ea6b-8e4d-4fb8-a563-a7130f015137`)
3. Log app_id from the route params
4. Don't log preflight requests
5. Don't log the body

### DIFF
--- a/server/src/instant/core.clj
+++ b/server/src/instant/core.clj
@@ -4,7 +4,7 @@
    [tool]
    [clojure.java.io :as io]
    [clojure.tools.logging :as log]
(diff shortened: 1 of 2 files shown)


stopachka (approved):
Cleean!

PR #746, opened 2025-01-24
Hazelcast stability

dwwoelfel:
We had a weird deploy last night where one of the instances did a rolling update and created a bunch of `idle-in-transaction` queries and had trouble connecting to hazelcast. Generally, it seemed to be having networking issues and had to be killed manually.

Looking in the logs a couple of things stood out:

1. We have some deserialization errors where it expects a throwable, but gets a clojure.lang.PersistentArrayMap
2. The instance that had problems appeared to be trying to connect to itself over Hazelcast. It looks like the instance didn't get removed from the cluster. I have a theory that the instance was in a loop trying to join itself, which caused the networking issues.

To fix 1, I stopped overriding java serialization in the global serializer. I was able to replicate the error once in staging before the fix by killing the docker container before the fix. I tried the same thing after deploying the fix and didn't see the error again. However, I never could get a reproducible error, so hard to say if this is fixed. I changed the cluster name so that we wouldn't have deserialization errors during the deploy (tested in staging). This will cause a short period during the deploy where users won't get all of the set-presence data until they reconnect after their machine shuts down.

To fix 2, we rotate the port between 6001 - 6008 each time we do a rolling deploy, so that no two consecutive rolling deploys on the same instance will ever use the same port.

I also increased the log level for Hazelcast in production.

Tested in staging and all looks good.


server/.platform/hooks/prebuild/prebuild.sh
```diff
@@ -0,0 +1,18 @@
...
+touch "$file"
+if [ ! -f "$file" ]; then
+  echo "0" > "$file"
+fi
+
+read -r current_count < "$file"
+new_count=$((current_count + 1))
+echo "$new_count" > "$file"
+
+hz_port=$((5701 + (new_count % 8)))
```
stopachka, line 16: Monster

server/src/instant/reactive/ephemeral.clj
```diff
@@ -40,9 +40,13 @@
...
 (defn init-hz [env store-conn {:keys [instance-name cluster-name]
                                :or {instance-name "instant-hz-v3"
-                                    cluster-name "instant-server-v1"}}]
+                                    cluster-name "instant-server-v2"}}]
   (-> (java.util.logging.Logger/getLogger "com.hazelcast")
-      (.setLevel java.util.logging.Level/WARNING))
+      (.setLevel (if (= env :prod)
+                   java.util.logging.Level/INFO
+                   java.util.logging.Level/WARNING)))
+  (.setLevel (java.util.logging.Logger/getLogger "com.hazelcast.system.logo")
```
stopachka, line 48: 😆

stopachka (approved):
Heeck yeah!

PR #748, opened 2025-01-24
Test fewer vfutures in the receive queue

dwwoelfel:
We're seeing it take a long time to start a virtual future in prod, but starting a regular future is still fast.

I'd like to experiment with using fewer virtual futures as a quick bandaid. This will hopefully make ephemeral state a lot faster because it shouldn't have to hit a single vfuture.

On 1/3 of the instances, this will use `thread` for both the receive queue workers and for wrapping `handle-receive`.


stopachka (approved):
Lesdoit!

PR #750, opened 2025-01-27
Simple performance optimizations

tonsky:
Some simple performance optimizations

### DIFF
--- a/server/src/instant/reactive/store.clj
+++ b/server/src/instant/reactive/store.clj
@@ -546,7 +546,7 @@
     (set? dq-part) (intersects? iv-part dq-part)
 
     (map? dq-part)
-    (if-let [{:keys [op value]} (:$comparator dq-part)]
+    (if-some [{:keys [op value]} (:$comparator dq-part)]
       (let [f (case op
                 :$gt >
                 :$gte >=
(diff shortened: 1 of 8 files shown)


stopachka (approved):
Doope!

server/src/instant/reactive/store.clj
```diff
@@ -546,31 +546,46 @@
...
+                    iv-part))
       (when (contains? dq-part :$not)
         (let [not-val (:$not dq-part)]
-          (some (partial not= not-val) iv-part))))))
+          (ucoll/seek (partial not= not-val) iv-part))))))
 
-(defn match-topic?
-  [iv-topic dq-topic]
-  (ucoll/every?-var-args match-topic-part? iv-topic dq-topic))
+(defn match-topic? [iv-topic dq-topic]
```
dwwoelfel: We actually know the topic is 4 parts, so we could just do:


```clojure
(defn match-topic?
  [[iv-idx iv-e iv-a iv-v]
   [dq-idx dq-e dq-a dq-v]]
  (and (match-topic-part? iv-idx
                          dq-idx)
       (match-topic-part? iv-e
                          dq-e)
       (match-topic-part? iv-a
                          dq-a)
       (match-topic-part? iv-v
                          dq-v)))
```

server/src/instant/reactive/store.clj
```diff
@@ -546,31 +546,46 @@
     (set? dq-part) (intersects? iv-part dq-part)
 
     (map? dq-part)
-    (if-let [{:keys [op value]} (:$comparator dq-part)]
+    (if-some [{:keys [op value]} (:$comparator dq-part)]
```
dwwoelfel, line 554: Does `if-some` do something different than `if-let` here?

dwwoelfel (approved):
One suggestion, but LGTM!

server/src/instant/reactive/store.clj
tonsky: Just a habit. if-let/when-let will fail on `false` and I’ve been bit by this in the past. if-some/when-some are just a better versions of if-let/when-let in every regard

PR #753, opened 2025-01-28
[sm] use forward attr in cascade delete dialog

stopachka:
image
- Using the `forward attr name` in the description text
- Changed up the layout a bit, so "Cascade Delete" is on it's own line. 

@dwwoelfel @nezaj @tonsky


dwwoelfel (approved):
LGTM!

tonsky:
But this is not what would happen. In the screenshot above, if we delete `$users` entity, it will cascade delete. But if we only delete `user` attribute, it will just unlink. I think, for clarity, we should use entity name, not the attribute name

PR #755, opened 2025-01-28
Fix for null query

dwwoelfel:
If we get a null query, we'll throw a noisy exception. This PR turns it into a normal error and notifies the client.

It seems like this can happen with `queryOnce`.


stopachka (approved):
lesgo!

PR #757, opened 2025-01-29
Enforce client formatting with a GitHub action

dwwoelfel:
This runs `prettier check` in a GitHub action. To fix, you can run `pnpm run format`.

This PR only includes the code changes, all of the formatting changes are in a separate PR to make it easier to review.

The prettier config is now the following:

```
{
  "semi": true,
  "singleQuote": true
}
```

That's what we were using on www, so it seems like the easiest thing to do. But if people have strong opinions, happy to change the config.

Includes an update to markdoc because it was erroring on a `{% conditional %}` tag that was hiding list items.

Will wait until the storage PR is merged before merging this or running the formatting changes, but they'll look like this: https://github.com/instantdb/instant/commit/cca492d13b96cd88191280ebc3268b2dc88f738c


dwwoelfel:
The failing test is expected because this PR just adds the test, it doesn't do the formatting.

stopachka (approved):
Yess!!!

PR #762, opened 2025-01-29
[namespace-dialog] change the grammar for cascade-delete

stopachka:
Since we changed the dialog label to use the forward-attr-name, it became a bit more confusing: 

Does unlinking also cause cascade-delete?

To remove the confusion, reverting to use the namespace name instead:

image
image

@tonsky @dwwoelfel @nezaj


dwwoelfel (approved):
LGTM!

PR #765, opened 2025-01-29
Add record middleware to nrepl

dwwoelfel:
Adds middleware to our nrepl to log activity so that we have an audit trail. 

Only turned on in production. Logs to honeycomb and to stdout.


stopachka (approved):
Monster!

PR #766, opened 2025-01-29
[client][types] add types to auth api

stopachka:
Kosmik team mentioned they had a bug where they accidentally wrote `refreshToken` instead of `refresh_token`. 

I added some more types to the auth api, so these kind of bugs are easy to catch 

@dwwoelfel @nezaj @tonsky


dwwoelfel (approved):
LGTM!

PR #768, opened 2025-01-29
Fix schema push

dwwoelfel:
Makes sure that we don't treat changing the reverse link from nil -> nil as duplicating the reverse link.

Fixes a bug users are having doing schema push.


stopachka (approved):
Thank you @dwwoelfel

PR #772, opened 2025-01-30
Add support for rotating db secret in secretsmanager

dwwoelfel:
Updates our database config to fetch the db connection config directly from aws. 

Instead of storing the database-url encrypted in the config, we just store the cluster name. This makes it a lot easier to see which database we're connecting to because it doesn't need to be encrypted in the config.edn.

When we create the connection pool, we query rds for the connection information. Then when we go to make a connection, we fetch the username/password from the secretsmanager. The secret gets cached so that we only have to fetch it once per startup.

If the secret rolls over, we'll catch the exception the next time we go to create a new database connection, update our cached secret value, and retry the connection. This way we can rotate secrets without any downtime.

I used the new v2 of the aws java sdk directly instead of amazonica to avoid adding new dependencies on amazonica, since amazonica doesn't seem to have a plan to switch away from v1 before it is deprecated.

Tested in staging.


server/deps.edn
```diff
@@ -19,7 +19,10 @@
         ch.qos.logback/logback-classic {:mvn/version "1.5.6"}
         org.slf4j/slf4j-api {:mvn/version "2.0.13"}
         cider/cider-nrepl {:mvn/version "0.45.0"}
+
         amazonica/amazonica {:mvn/version "0.3.167"}
+        software.amazon.awssdk/rds {:mvn/version "2.30.7"}
```
stopachka, line 24: 🔥

stopachka (approved):
Awesome work @dwwoelfel !

PR #774, opened 2025-01-30
[essay] be more explicit about stop the world

stopachka:
A user asked: wouldn't 'stop the world' also take a few seconds? 

I wrote a minute because I imagined the stop the world approach to be more manual. If we automated it, the function would get very close to what we already wrote for zero downtime. I updated the essay to mention this. 

@dwwoelfel @nezaj @tonsky


dwwoelfel (approved):
LGTM!

PR #776, opened 2025-01-30
Fix handling of database connection strings

dwwoelfel:
Does two things:

1. Renames `username` to `user` so that it works properly with jdbc connection urls
2. Query-decodes password when we get it from the connection string. Then we set the password directly because `next-jdbc/jdbc-url` won't url-encode its params.


nezaj (approved):
Nice!

PR #777, opened 2025-01-30
[Emails] Send welcome email

nezaj:
We sent a few emails to users yesterday to figure out the burst in signups and it was nice to get some feedback!

Figured we'd do well with sending some automated emails. We may want to tweak our strategy in the future but for now the idea is this:

Monday -> Friday we'll send out an email to users who signed up between 2 days ago and the previous day. This way we email people after they've tried us for at least 24 hours.

I also added a flag in our config app to turn this off and just the limit (right now set to 50) for the amount of emails we send

### DIFF
--- a/server/src/instant/scripts/welcome_email.clj
+++ b/server/src/instant/scripts/welcome_email.clj
@@ -1,0 +1,90 @@
+(ns instant.scripts.welcome-email
+  "Send welcome emails to our users"
+  (:require
+   [instant.util.date :as date]
+   [instant.postmark :as postmark]
+   [chime.core :as chime-core]
+   [clojure.tools.logging :as log]
+   [instant.grab :as grab]
+   [instant.jdbc.aurora :as aurora]
+
+   [instant.jdbc.sql :as sql]
+   [instant.flags :as flags])
+  (:import
+   (java.time Period DayOfWeek)))
+
+(defn find-new-users
+  ([] (find-new-users (aurora/conn-pool :read)))
+  ([conn]
+   (sql/select
+    :welcome-email-users
+    conn
+    ["SELECT email
+     FROM instant_users
+     WHERE created_at BETWEEN (CURRENT_TIMESTAMP - INTERVAL '2 days')
+                          AND (CURRENT_TIMESTAMP - INTERVAL '1 day')"])))
+
+(def html-body
+  "<p>Hey there! Welcome to Instant! Full disclosure: this is an automated
+email, but if you respond, a real human (likely Joe or Stopa, the
+founders) will read it and get back to you!</p>
+<p>How’s your experience with Instant been so far? Any feedback to
+share?</p>")
+
+(def text-body
+  "Hey there! Welcome to Instant! Full disclosure: this is an automated email, but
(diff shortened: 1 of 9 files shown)


server/src/instant/scripts/welcome_email.clj
```diff
@@ -0,0 +1,87 @@
...
+   (sql/select
+    :welcome-email-users
+    conn
+    ["SELECT email
+     FROM instant_users
+     WHERE created_at BETWEEN (CURRENT_TIMESTAMP - INTERVAL '2 days')
+                          AND (CURRENT_TIMESTAMP - INTERVAL '1 day')"])))
+
+(def html-body
+  "<p>Hey there! Welcome to Instant! Full disclosure: this is an automated
```
dwwoelfel, line 40: Do we need to add an unsubscribe link?

server/src/instant/scripts/welcome_email.clj
```diff
@@ -0,0 +1,87 @@
...
+  (:import
+   (java.time Period DayOfWeek)))
+
+(defn find-new-users
+  ([] (find-new-users (aurora/conn-pool :read)))
+  ([conn]
+   (sql/select
+    :welcome-email-users
+    conn
+    ["SELECT email
```
dwwoelfel: Maybe we could add a `sent_welcome_email` flag to the user so that we don't accidentally send the same email twice?

server/src/instant/scripts/welcome_email.clj
```diff
@@ -0,0 +1,87 @@
...
+         (let [emails (map :email (find-new-users))
+               shuffled (cond->> (shuffle emails)
+                          limit (take limit))]
+           (run! (fn [to]
+                   (postmark/send! {:from "founders@pm.instantdb.com"
+                                    :to to
+                                    :reply-to "founders@instantdb.com"
+                                    :subject "Welcome to Instant!"
+                                    :html html-body
+                                    :text text-body})) shuffled)))))))
```
dwwoelfel:
```suggestion
                                    :text text-body}))
                                    shuffled)))))))
```

dwwoelfel (approved):
Couple of comments, but LGTM!

server/src/instant/scripts/welcome_email.clj
nezaj: I think for this case we may not need it since this a one-off email. However if folks complain we can def add one in with postmark!

server/src/instant/scripts/welcome_email.clj
```diff
@@ -0,0 +1,101 @@
...
+                      (Period/ofDays 1))]
+    (->> periodic-seq
+         (filter (fn [x] (.isAfter x now)))
+         ;; Only run Monday through Friday
+         (filter (fn [x]
+                   (let [day-of-week (.getDayOfWeek x)]
+                     (not (contains? #{DayOfWeek/SATURDAY DayOfWeek/SUNDAY}
+                                     day-of-week))))))))
+
+(def schedule (atom nil))
```
dwwoelfel:
```suggestion
(defonce schedule (atom nil))
```

Prevents clearing out the schedule if you re-load the file in the repl.

dwwoelfel (approved):
LGTM!

PR #778, opened 2025-01-30
[graphs] move graphs to a page

stopachka:
Previously, we needed to run a script to generate graphs. 

**Previous Steps** 

The previous steps used to be: 

```
    1. Run `make dev-with-prod` to connect to prod data
    2. Run `generate!` at the bottom generate metrics
    3. Run `open resources/metrics/*-usage.png` in the terminal from the /server directory to view the metrics
```

**New Steps**

Now, **you no longer need to run `make dev-with-prod`**

To get the latest version, you can always load instantdb.com/intern/graphs 

If you want to customize the graph, you run instant locally and run the `save-pngs` comment

@nezaj @dwwoelfel @tonsky

### DIFF
--- a/client/www/pages/intern/graphs.tsx
+++ b/client/www/pages/intern/graphs.tsx
@@ -1,0 +1,82 @@
+import Head from 'next/head';
+import React, { useEffect, useState } from 'react';
+import { useIsHydrated } from '@/lib/hooks/useIsHydrated';
+import { useAuthToken } from '@/lib/auth';
+import { jsonFetch } from '@/lib/fetch';
+import config from '@/lib/config';
+
+function fetchGraphData(token: string | undefined) {
+  return jsonFetch(`${config.apiURI}/dash/graphs`, {
+    method: 'GET',
+    headers: {
+      authorization: `Bearer ${token}`,
+      'content-type': 'application/json',
+    },
+  });
+}
+
+function useGraphData(token: string | undefined) {
+  const [state, setState] = useState<{
+    isLoading: boolean;
+    error: undefined | { body: { message: string } };
+    data: any | undefined;
+  }>({
+    isLoading: true,
+    error: undefined,
+    data: undefined,
+  });
+  useEffect(() => {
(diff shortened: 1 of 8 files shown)


client/www/pages/intern/graphs.tsx
```diff
@@ -0,0 +1,82 @@
...
+    return <div>Loading...</div>;
+  }
+
+  if (error) {
+    return <div>Error: <pre>{JSON.stringify(error.body, null, 2)}</pre></div>;
+  }
+  return (
+    <div>
+      <Head>
+        <title>Instant Paid Apps</title>
```
nezaj: Nit. Maybe Instant Metrics?

nezaj (approved):
Neato!

server/src/tool.clj
```diff
@@ -238,3 +239,14 @@
      (require 'criterium.core)
      (criterium.core/quick-bench ~@body)
      (flush)))
+
+(defn prod-conn-str
+  []
+  (str/trim (:out (shell/sh "./scripts/prod_connection_string.sh"))))
+
+(defn prod-conn []
```
dwwoelfel: This is dangerous, because it encourages people to leave connections open forever. How about this:

```clojure
(defmacro with-prod-conn
  "Usage: (with-prod-conn [my-conn]
            (sql/select my-conn [\"select 1\"]))"
  [[conn-name] & body]
  `(let [cluster-id# (-> (clojure.java.io/resource "config/prod.edn")
                        slurp
                        clojure.edn/read-string
                        :database-cluster-id)
         rds-cluster-id->db-config# (requiring-resolve 'instant.aurora-config/rds-cluster-id->db-config)
         start-pool# (requiring-resolve 'instant.jdbc.aurora/start-pool)]
     (with-open [~conn-name (start-pool# 1 (rds-cluster-id->db-config# cluster-id#))]
       ~@body)))
```

And then you also don't need `prod-conn-str` above.

stopachka: Oo, I like it!

PR #781, opened 2025-01-31
Monitor transact steps perf

tonsky:
transact! on store-conn takes suspiciously long time. One hypothesis is that it’s contention on lock. This metric will allow us to check it

Screenshot 2025-01-31 at 17 03 29


dwwoelfel (approved):
LGTM!

PR #782, opened 2025-01-31
Remove vfuture-pmap

tonsky:
For unknown reason, just spinning up vfuture takes significant time on prod. Local dev machine:

Screenshot 2025-01-31 at 16 30 53

Prod:

Screenshot 2025-01-31 at 16 31 47

Times varies greatly, but it’s usually between 70 and 250 ms just to start a virtual thread!

Since spinning up real threads is faster (although more resource-intensive), we can try switching to agent thread pool & physical threads to try to improve latency

### DIFF
--- a/server/src/instant/reactive/session.clj
+++ b/server/src/instant/reactive/session.clj
@@ -201,7 +201,7 @@
         _ (reset! debug-info {:processed-tx-id processed-tx-id
                               :instaql-queries (map :instaql-query/query stale-queries)})
         recompute-results (->> stale-queries
-                               (ua/pmap-all (partial recompute-instaql-query! opts)))
+                               (ua/pmap (partial recompute-instaql-query! opts)))
         {computations true spam false} (group-by :result-changed? recompute-results)
         num-spam (count spam)
         num-computations (count computations)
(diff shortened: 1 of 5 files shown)


server/src/instant/reactive/session.clj
```diff
@@ -199,7 +201,7 @@
         _ (reset! debug-info {:processed-tx-id processed-tx-id
                               :instaql-queries (map :instaql-query/query stale-queries)})
         recompute-results (->> stale-queries
-                               (ua/vfuture-pmap (partial recompute-instaql-query! opts)))
+                               (ua/pmap-all (partial recompute-instaql-query! opts)))
```
dwwoelfel: Can you put this behind a `instant.flags` in case we run into unexpected issue in production?

dwwoelfel (approved):
One comment, but up to try it!

server/src/instant/util/async.clj
```diff
@@ -153,6 +153,11 @@
   (let [futs (mapv #(vfuture (f %)) coll)]
     (mapv deref futs)))
 
+(defn pmap-all [f coll]
+  (->> coll
+    (mapv #(future (f %))) ;; force whole seq
```
dwwoelfel: Just realized that we need to use a `tracking-future` here or the cancelation won't work properly.

tonsky:
@dwwoelfel done. The flag is disabled by default, so merging this will keep vfutures. To test tracking-futures, we’ll need to create `threading` namespace, `use-vfutures` attribute and create an entity that sets it to false

dwwoelfel (approved):
LGTM!

PR #783, opened 2025-01-31
[metrics] set headless

stopachka:
We'd get this extra icon:
image

It's because we now import instant.intern.metrics. Added -Djava.awt.headless=true as per @tonsky 's suggestion and the icon dissappeared. 

@tonsky @dwwoelfel @nezaj


server/deps.edn
```diff
@@ -95,14 +95,17 @@
...
                             "--add-modules" "java.se"
                             "--add-exports" "java.base/jdk.internal.ref=ALL-UNNAMED"
                             "--add-opens" "java.base/java.lang=ALL-UNNAMED"
                             "--add-opens" "java.base/sun.nio.ch=ALL-UNNAMED"
                             "--add-opens" "java.management/sun.management=ALL-UNNAMED"
-                            "--add-opens" "jdk.management/com.sun.management.internal=ALL-UNNAMED"]}
+                            "--add-opens" "jdk.management/com.sun.management.internal=ALL-UNNAMED"
+
+                            ;; headless charts 
+                            "-Djava.awt.headless=true"]}
```
dwwoelfel, line 108: Can you add it to the [docker-compose.yml](https://github.com/instantdb/instant/blob/main/server/docker-compose.yml#L18) also

dwwoelfel (approved):
LGTM

server/deps.edn
dwwoelfel: Maybe should also add it to the test flags in the deps.edn?

stopachka: 🫡 

I ran `make test`, and curiously the little guy did not show up. I will add it in for good measure though 🫡

PR #784, opened 2025-01-31
[investor-updates] graphs -> investor updates

stopachka:
Small changes as I get back into our dashboard work. 

I thought we should have a standalone page for investor updates. This way it's always easy to see MoM growth at the end of the month, paste it into our email, etc. 

1. Changed the route to be called investor_updates
2. Added MoM growth metrics at the top

@dwwoelfel @nezaj @tonsky

### DIFF
--- a/server/src/instant/core.clj
+++ b/server/src/instant/core.clj
@@ -161,7 +161,7 @@
                                                       (indexing-jobs/stop)))]]
                                  (deref fut))))))
 
-(defmacro log-init [operation & body]
+(defmacro with-log-init [operation & body]
   `(do
      (tracer/record-info! {:name (format "init.start.%s" (name ~operation))})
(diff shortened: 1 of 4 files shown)


server/src/instant/core.clj
```diff
@@ -179,59 +179,59 @@
     (tracer/init)
 
     (log-init :uncaught-exception-handler
```
dwwoelfel: Can you undo the whitespace changes in this file?

Your lsp should see that it's a macro, but maybe renaming `log-init` to `with-log-init` will fix it?

dwwoelfel (approved):
Good idea!

server/src/instant/core.clj
stopachka: with-log-init did the trick, thanks @dwwoelfel !

PR #788, opened 2025-01-31
[overview] machine summarizer

stopachka:
For our `overview` field, I wanted to get a ~live view of how many connections we are supporting.  Ideally we'd have a table below that we can scroll through, to see all active apps. 

### Problem 

Since we have multiple machines, we can't just get this info in memory

### Solution

We use Hazelcast's [ExecutorService](https://docs.hazelcast.com/imdg/4.2/computing/executor-service), run a `summarize` function across all machines, and return the result together. 

@dwwoelfel @nezaj @tonsky


dwwoelfel (approved):
LGTM!

PR #789, opened 2025-02-01
Serializer for the Task

dwwoelfel:


server/src/instant/core.clj
```diff
@@ -20,6 +20,7 @@
    [instant.jdbc.aurora :as aurora]
    [instant.jdbc.wal :as wal]
    [instant.lib.ring.undertow :as undertow-adapter]
+   [instant.machine-summaries]
```
dwwoelfel, line 23: Requiring this here ensures that `session-report-task` can be resolved by hazelcast

PR #792, opened 2025-02-03
Send charts with daily discord message

dwwoelfel:


server/src/instant/scripts/daily_metrics.clj
```diff
@@ -102,8 +105,16 @@
      (str "daily-metrics-" date-str)
      (fn []
        (insert-new-activity)
-       (let [stats (get-daily-actives date-minus-one-str)]
-         (send-discord! stats date-minus-one-str))))))
+       (let [stats (get-daily-actives date-minus-one-str)
+             conn (aurora/conn-pool :read)
+             charts (->> (metrics/investor-update-metrics conn)
```
stopachka: If we use investor update metrics, afaik these graphs will not change very much, as they are calculated based on the start of the month. 

However, if we use `overview-metrics`, we could get: 
a. The rolling monthly active updates (i.e MAU as of yesterday - 30 days ago) 
b. How are trending towards out monthly active goal for this month

stopachka (approved):
Niice!

server/src/instant/scripts/daily_metrics.clj
dwwoelfel: Updated!

PR #795, opened 2025-02-03
Prevent files deletes in tx

nezaj:
I noticed after rebasing/merging deletes weren't being blocked in tx. Realized the shape of the tx-steps changes slightly so updated the code to correctly prevent updates and deletes.

### DIFF
--- a/server/src/instant/db/transaction.clj
+++ b/server/src/instant/db/transaction.clj
@@ -127,7 +127,7 @@
 
 (defn prevent-$files-add-retract! [op attrs triples]
   (doseq [t triples
-          :let [etype (let [[_eid aid] t]
+          :let [etype (let [[_op _eid aid] t]
                         (-> (attr-model/seek-by-id aid attrs)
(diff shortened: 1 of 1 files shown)


server/src/instant/db/transaction.clj
```diff
@@ -151,7 +151,7 @@
   [attrs grouped-tx-steps opts]
   (when (not (:allow-$files-update? opts))
     (doseq [batch grouped-tx-steps
-            :let [[op & triples] batch]]
+            :let [[op triples] batch]]
```
dwwoelfel: This should be `tx-steps` and not `triples`, right?

dwwoelfel (approved):
LGTM, but the code would be a lot easier to understand if you used `tx-steps` in a lot of the places it says `triples` right now.

server/src/instant/db/transaction.clj
nezaj: Great point! Updated!

PR #797, opened 2025-02-04
Fix another flaky test

dwwoelfel:
Fixes the `join-room` test, where `refresh-presence` might get delivered before `join-room-ok`.

Example of a failing run here: https://github.com/instantdb/instant/actions/runs/13125160638/job/36619894359


nezaj (approved):
Nice!

PR #798, opened 2025-02-04
[overview] overview dashboard

stopachka:
This adds an Overview Dashboard for us. 

(See screenshots in Team Discord) 

Some things we show: 
1. Rolling Actives
2. Month-To-Date Actives vs Target
3. Number of active connections (refreshed every minute)
4. Connections grouped by app

One bugfix for Month to Date Actives

1. Previously, the "real data" line would expand to the end of the month. Now we only show real data until the day we are looking.

@dwwoelfel @nezaj @tonsky

### DIFF
--- a/server/src/instant/dash/routes.clj
+++ b/server/src/instant/dash/routes.clj
@@ -57,6 +57,8 @@
             [instant.model.schema :as schema-model]
             [instant.intern.metrics :as metrics]
             [medley.core :as medley]
+            [instant.reactive.ephemeral :as eph]
+            [instant.machine-summaries :as machine-summaries]
             [instant.storage.coordinator :as storage-coordinator]
             [instant.model.app-file :as app-file-model])
   (:import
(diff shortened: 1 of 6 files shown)


server/src/instant/dash/routes.clj
```diff
@@ -285,6 +288,27 @@
                                                                                 500 400))))]
     (response/ok {:metrics metrics-with-b64-charts})))
 
+(defn admin-overview-daily-get [req]
+  (let [{:keys [email]} (req->auth-user! req)
+        _ (assert-admin-email! email)
+        conn (aurora/conn-pool :read)
+        target-date (.minusDays (LocalDate/now) 1)
```
dwwoelfel: If you use `Instant` instead of `LocalDate`, then you'll probably get better consistency between prod and dev.

I was surprised when I had to use a localdate here: https://github.com/instantdb/instant/pull/792/files#diff-f585e88b581856cd0618307a5d0dd8138367b01a1fd530af1a25992755744c84R110

server/src/instant/intern/metrics.clj
```diff
@@ -317,42 +350,49 @@
 ;; ---------------- 
 ;; Overview Metrics 
 
-(defn overview-metrics [conn target-date]
-  (let [days-ago-30 (.minusDays target-date 30)
+(defn local-analysis-date [x]
+  (.toLocalDate (:analysis_date x)))
+
+(defn overview-metrics [conn]
+  (let [target-date (get-latest-daily-tx-date conn)
```
stopachka, line 353: cc @dwwoelfel opted to get the actual date by simply querying our `daily_app_transactions`, and fetching the date from there. 

Note, this still returns a `LocalDate`. 

```
instant=# \d daily_app_transactions;
                        Table "public.daily_app_transactions"
   Column    |            Type             | Collation | Nullable |      Default
-------------+-----------------------------+-----------+----------+-------------------
 date        | date                        |           | not null |
 app_id      | uuid                        |           | not null |
 active_date | timestamp without time zone |           |          |
 is_active   | boolean                     |           | not null |
 count       | bigint                      |           |          |
 id          | uuid                        |           | not null | gen_random_uuid()
Indexes:
    "daily_app_transactions_pkey" PRIMARY KEY, btree (id)
    "daily_app_transactions_date_app_id_is_active_key" UNIQUE CONSTRAINT, btree (date, app_id, is_active) REPLICA IDENTITY
```

The type on `date` is Postgres' `DATE` column, which has no time of day / time zone information. My thinking is, by keeping everything as `LocalDate`, we focus the code on the fact that the only thing we can really compare are the actual 'date' strings, without knowing much more. 

Let me know if you have other thoughts

server/src/instant/scripts/daily_metrics.clj
```diff
@@ -107,7 +107,7 @@
        (insert-new-activity)
        (let [stats (get-daily-actives date-minus-one-str)
              conn (aurora/conn-pool :read)
-             charts (->> (metrics/overview-metrics conn (.minusDays (LocalDate/now) 1))
+             charts (->> (metrics/overview-metrics conn)
```
dwwoelfel, line 110: Thanks for looking out!

dwwoelfel (approved):
LGTM!

PR #801, opened 2025-02-04
Use the s3 aws jdk 2

dwwoelfel:
Replaces amazonica with the aws-jdk 2 for s3 operations. Uses the aws crt client for uploads, which is supposed to have high throughput: https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/crt-based-s3-client.html

A couple of incidental changes:

1. I pass the `content-type` from the `req` to `upload-stream-to-s3` in dash. This gave me better content type when I was uploading from the dashboard.

2. I pass along the `content-length` field from the `req`, which allows us to use an optimized upload path when we know the content-length ahead of time.

### DIFF
--- a/server/deps.edn
+++ b/server/deps.edn
@@ -21,8 +21,11 @@
         cider/cider-nrepl {:mvn/version "0.45.0"}
 
         amazonica/amazonica {:mvn/version "0.3.167"}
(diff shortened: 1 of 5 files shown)


nezaj (approved):
Awesome!

PR #802, opened 2025-02-04
[Dash] Add feedback for clear and delete app

nezaj:
We got a few reports that clearing/deleting apps don't give feedback. Noticed it does take some time in prod so figured we can add a loading state to make it clearer to the user.

https://github.com/user-attachments/assets/232b8da3-9b45-4ab9-be10-ff7487e71ebc


dwwoelfel (approved):
LGTM!

PR #806, opened 2025-02-04
Show the first user namespace instead of the system namespace

dwwoelfel:
When you first arrive on the Explorer, even for a new app, you'll land on the `$files` namespace and never see the NUX for creating your first namespace.

Now we'll default to the "create your first namespace" explainer on new apps, and the first user-namespace on apps that have created namespaces:

image


stopachka (approved):
niice!

PR #807, opened 2025-02-04
Hide  for non-enabled apps

nezaj:
dww pointed out that it's weird to see `$files` if you can't use them yet. We already pass an `isStorageEnabled` check to the frontend so figured for now we can just hide the namespace until we fully launch

### DIFF
--- a/client/www/components/dash/explorer/Explorer.tsx
+++ b/client/www/components/dash/explorer/Explorer.tsx
@@ -452,11 +452,9 @@ export function Explorer({
   // data
   const { namespaces: _namespaces } = useSchemaQuery(db);
   // (TODO): When fully launching storage we can remove this check
-  const namespaces =
-    _namespaces &&
-    _namespaces.filter((ns) =>
-      isStorageEnabled ? true : ns.name !== '$files',
-    );
+  const namespaces = isStorageEnabled
+    ? _namespaces && _namespaces.filter((ns) => ns.name !== '$files')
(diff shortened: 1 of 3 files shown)


client/www/components/dash/explorer/Explorer.tsx
```diff
@@ -441,7 +450,13 @@ export function Explorer({
   }
 
   // data
-  const { namespaces } = useSchemaQuery(db);
+  const { namespaces: _namespaces } = useSchemaQuery(db);
+  // (TODO): When fully launching storage we can remove this check
+  const namespaces =
```
stopachka: nit: would write this as: 

```
  const namespaces = isStorageEnabled 
    ? _namespaces.filter((ns) => ns.name !== '$files') 
    : _namespaces;
```

stopachka (approved):
SGTM!

client/www/components/dash/explorer/Explorer.tsx
nezaj, line 455:
```suggestion
const namespaces = isStorageEnabled ? _namespaces.filter((ns) => ns.name !== '$files') : _namespaces;
```

nezaj, line 455:
```suggestion
const namespaces = isStorageEnabled ? _namespaces && _namespaces.filter((ns) => ns.name !== '$files') : _namespaces;
```

PR #810, opened 2025-02-04
Run `pnpm run format`

dwwoelfel:
Runs `pnpm run format`.

Now the formatting test will pass and we'll have to prevent the formatting churn on client PRs.

After this is merged, I'll add the commit to `.git-blame-ignore-revs`


nezaj (approved):
Nice!

PR #813, opened 2025-02-05
Upgrade setup clojure action

dwwoelfel:
Includes a fix for an annoying bug where the cache step can hang for a couple of minutes (https://github.com/DeLaGuardo/setup-clojure/pull/109)


stopachka (approved):
wohoo!

PR #817, opened 2025-02-05
Add Storage sweeper

nezaj:
Adds sweeper logic for storage files. Introduces a table to record files to be swept which is populated by triggers.

* Whenever file triples are deleted we add to the sweep table no-ops if delete is already there)
* Whenever file triples are added we remove from the sweep table (no-ops if nothing there)
* We run the sweeper once a day to clean-up any missed files.

The upshot if this is now `$files` will be properly cleaned up on clearing an app and/or deleting an app. This will also enable proper behavior when using `:on-delete` with `$files` links.

h/t to dww for the idea!

### DIFF
--- a/server/resources/migrations/51_add_file_delete_trigger.up.sql
+++ b/server/resources/migrations/51_add_file_delete_trigger.up.sql
@@ -1,0 +1,51 @@
+CREATE TABLE app_files_to_sweep (
+    id uuid PRIMARY KEY DEFAULT gen_random_uuid(),
+    app_id uuid NOT NULL REFERENCES apps(id),
+    path text NOT NULL,
+    created_at timestamp NOT NULL DEFAULT NOW(),
+    UNIQUE (app_id, path)
+);
+
+-- Whenever we delete file triples we want to ensure they are scheduled for
+-- deletion in S3.
+CREATE FUNCTION create_file_to_sweep()
+RETURNS trigger AS $$
+BEGIN
+    -- This should match the attr_id for $files.path
+    IF OLD.attr_id = '96653230-13ff-ffff-2a34-f04cffffffff' THEN
+        INSERT INTO app_files_to_sweep (app_id, path)
+        VALUES (
+            OLD.app_id,
+            OLD.value #>> '{}'  -- Extract from JSON
+        )
+        ON CONFLICT DO NOTHING;
+    END IF;
+    RETURN OLD;
+END;
+$$ LANGUAGE plpgsql;
+
+CREATE TRIGGER insert_files_to_sweep_trigger
+    AFTER DELETE ON triples
+    FOR EACH ROW
+    EXECUTE FUNCTION create_file_to_sweep();
+
+-- When file triples are created, the path may be similar to one that was
+-- scheduled for deletion previously. We want to remove this from the sweep table
+-- since they are no longer candidates for deletion
+CREATE FUNCTION delete_file_to_sweep()
+RETURNS trigger AS $$
+BEGIN
+    -- This should match the attr_id for $files.path
+    IF NEW.attr_id = '96653230-13ff-ffff-2a34-f04cffffffff' THEN
+        DELETE FROM app_files_to_sweep 
+        WHERE app_id = NEW.app_id 
+        AND path = NEW.value #>> '{}';
+    END IF;
+    RETURN NEW;
+END;
+$$ LANGUAGE plpgsql;
+
+CREATE TRIGGER delete_files_to_sweep_trigger
+    AFTER INSERT ON triples
+    FOR EACH ROW
+    EXECUTE FUNCTION delete_file_to_sweep();
--- a/server/src/instant/db/transaction.clj
+++ b/server/src/instant/db/transaction.clj
@@ -136,18 +136,8 @@
      [op t]
      [{:message (format "update or merge is not allowed on $files in transact.")}])))
 
-(defn prevent-$files-deletes! [op tx-steps]
-  (doseq [t tx-steps
-          :let [[_op _eid etype] t]
-          :when (= etype "$files")]
-    (ex/throw-validation-err!
-     :tx-step
-     [op t]
-     [{:message (format "delete is not allowed on $files in transact.")}])))
-
 (defn prevent-$files-updates
-  "With the exception of linking/unlinking, we prevent most updates unless it's
-  gone through an explicitly allowed code path"
+  "Similar to $users, we don't allow updates to $files"
   [attrs grouped-tx-steps opts]
   (when (not (:allow-$files-update? opts))
     (doseq [batch grouped-tx-steps
@@ -155,9 +145,6 @@
       (case op
         (:add-triple :deep-merge-triple :retract-triple)
         (prevent-$files-add-retract! op attrs tx-steps)
-
-        :delete-entity
-        (prevent-$files-deletes! op tx-steps)
 
         nil))))
 
--- a/server/src/instant/storage/sweeper.clj
+++ b/server/src/instant/storage/sweeper.clj
@@ -1,0 +1,69 @@
+(ns instant.storage.sweeper
+  (:require
+   [chime.core :as chime-core]
+   [instant.flags :as flags]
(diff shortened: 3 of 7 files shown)


server/resources/migrations/51_add_file_delete_trigger.up.sql
```diff
@@ -0,0 +1,51 @@
+CREATE TABLE app_files_to_sweep (
+    id uuid PRIMARY KEY DEFAULT gen_random_uuid(),
+    app_id uuid NOT NULL REFERENCES apps(id),
```
dwwoelfel: What should happen to rows in this table when the app gets deleted, right now it will just prevent deleting the app.

server/resources/migrations/51_add_file_delete_trigger.up.sql
```diff
@@ -0,0 +1,51 @@
...
+
+CREATE TRIGGER insert_files_to_sweep_trigger
+    AFTER DELETE ON triples
+    FOR EACH ROW
+    EXECUTE FUNCTION create_file_to_sweep();
+
+-- When file triples are created, the path may be similar to one that was
+-- scheduled for deletion previously. We want to remove this from the sweep table
+-- since they are no longer candidates for deletion
+CREATE FUNCTION delete_file_to_sweep()
```
dwwoelfel: I think you can still end up with a race here. If the sweeper is in the process of deleting, then the underlying file might get deleted before we can remove the `app_files_to_sweep` row.

Here's an idea for an alternative approach that would avoid this:

When we save a file in s3, pre-generate the `$file`'s `entity-id` and set the `ETag` for the object in s3 to the `entity-id`. Keep track of the `entity_id` in the `app_files_to_sweep` table, then when we do a delete, use [`If-Match`](https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeleteObject.html#API_DeleteObject_RequestSyntax) with the entity_id. That way if someone else uploads the file in the meantime, it will get a new `ETag` and won't get deleted.

The other nice thing about this is that we don't have to run a trigger on inserts into the triples table. That will slow down transactions, so we'd want to minimize that as much as possible.

server/src/instant/storage/sweeper.clj
```diff
@@ -0,0 +1,69 @@
...
+   [instant.util.tracer :as tracer]
+   [instant.util.date :as date]
+   [clojure.tools.logging :as log])
+  (:import
+   (java.lang AutoCloseable)
+   (java.time Period)))
+
+(defn handle-sweep! []
+  (tracer/with-span! {:name "storage/sweeper"}
+    (let [files (app-file-to-sweep/get-all)
```
dwwoelfel: We should limit the number of rows we fetch here. We want to keep the jobs small because the instance could shut down at any time. The smaller the unit of work, the less chance it will be interrupted.

 Here's how I would set it up:

1. Run the job every hour instead of once a day
2. Take only 500 rows so that you only have to send one batch delete to AWS (it accepts 1000, so 500 gives us room to double while we're deleting legacy keys)
3. Do a checkout of the row, similar to how we do indexing jobs (https://github.com/instantdb/instant/blob/main/server/src/instant/db/indexing_jobs.clj#L341). This will allow multiple machines to run the job at the same time without stepping on each other.

nezaj:
Updated this PR to use `location-id` for Storage. This sets up the tables for the sweeper job but I'll put the code for that in a separate PR.

Will put up migration doc in our team chat as well with more details!

server/resources/migrations/51_add_file_delete_trigger.up.sql
nezaj: Great point! Relaxed this constraint

server/resources/migrations/51_add_file_delete_trigger.up.sql
nezaj: From this discussion we went with using `location-id`!

server/src/instant/db/transaction.clj
```diff
@@ -136,18 +136,8 @@
...
-          :when (= etype "$files")]
-    (ex/throw-validation-err!
-     :tx-step
-     [op t]
-     [{:message (format "delete is not allowed on $files in transact.")}])))
-
 (defn prevent-$files-updates
-  "With the exception of linking/unlinking, we prevent most updates unless it's
-  gone through an explicitly allowed code path"
+  "Similar to $users, we don't allow updates to $files"
```
stopachka: Nit: I would update this docstring to explain what exactly is allowed and isn't

server/src/instant/util/s3.clj
```diff
@@ -82,6 +86,19 @@
                               (.contents resp))
       :next-continuation-token (.nextContinuationToken resp)})))
 
+(defn list-objects-paginated
```
stopachka: nit: would call this list-all-objects

server/src/instant/storage/s3.clj
```diff
@@ -42,68 +62,80 @@
...
+        object-key (->object-key app-id location-id)]
     (str (s3-util/generate-presigned-url
           {:method :get
            :bucket-name s3-util/default-bucket
            :key object-key
            :duration duration}))))
 
+(defn create-signed-download-url! [app-id path location-id]
+  (if (migrating?)
+    (path-url app-id path)
```
stopachka, line 136: In case people are caching this, I would not delete the previous url until 7 days from our switch

server/src/instant/system_catalog.clj
```diff
@@ -275,7 +276,7 @@
               :unique? true
               :index? true)
    (make-attr "$files" "path"
-              :unique? false
+              :unique? true
```
dwwoelfel, line 279: This won't be applied automatically, we'll have to create an indexing job for it. I can do that after we deploy the changes.

server/src/instant/storage/s3.clj
```diff
@@ -32,6 +34,24 @@
...
+(defn path-object-key->path
+  "Extract path from our S3 object keys"
+  [object-key]
+  (let [[_app-id _bin & path] (string/split object-key #"/")]
+    (string/join "/" path)))
+
+;; S3 path manipulation
+;; ----------------------
+(defn location-id->bin
+  "We add a bin to the location id to scale S3 performance"
```
stopachka: nit: would add a link about s3's perf

server/src/instant/storage/s3.clj
nezaj: Noted! We won't delete the old files for at least 1 week!

server/resources/migrations/52_add_app_files_to_sweep.up.sql
```diff
@@ -0,0 +1,57 @@
...
+    UNIQUE (app_id, location_id)
+);
+
+-- Whenever we delete file triples we want to ensure they are scheduled for
+-- deletion in S3.
+CREATE FUNCTION create_file_to_sweep()
+RETURNS trigger AS $$
+BEGIN
+    -- This should match the attr_id for $files.location-id
+    IF OLD.attr_id = '96653230-13ff-ffff-2a34-b40fffffffff' THEN
```
stopachka, line 15: Hoping this does not add too much extra overhead to transact.

stopachka (approved):
SGTM!

server/src/instant/util/uuid.clj
```diff
@@ -18,6 +19,13 @@
         (string? x) (parse-uuid (string/trim x))
         :else nil))
 
+(defn str->uuid
```
dwwoelfel, line 22: What do we use this for?

dwwoelfel (approved):
LGTM!

server/src/instant/util/uuid.clj
nezaj: This will be used for deterministically setting initial `location-id` for the current triples

server/src/instant/db/transaction.clj
nezaj: Updated!

server/resources/migrations/52_add_app_files_to_sweep.up.sql
nezaj: Definitely!!

PR #818, opened 2025-02-05
[metrics] add pro subscriptions to metrics

stopachka:
Adds the number of pro subscriptions we have to metrics!

@nezaj @tonsky @dwwoelfel

### DIFF
--- a/server/src/instant/dash/routes.clj
+++ b/server/src/instant/dash/routes.clj
@@ -293,12 +293,14 @@
         _ (assert-admin-email! email)
         conn (aurora/conn-pool :read)
         overview (metrics/overview-metrics conn)
+        rev-subs (dash-admin/get-revenue-generating-subscriptions)
         overview-with-b64-charts
         (update overview :charts (partial medley/map-vals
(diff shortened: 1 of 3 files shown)


server/src/instant/dash/routes.clj
```diff
@@ -293,12 +293,14 @@
...
         overview (metrics/overview-metrics conn)
+        rev-subs (dash-admin/get-revenue-generating-subscriptions)
         overview-with-b64-charts
         (update overview :charts (partial medley/map-vals
                                           (fn [chart] (metrics/chart->base64-png chart
                                                                                  500 400))))]
 
-    (response/ok overview-with-b64-charts)))
+    (response/ok (assoc overview-with-b64-charts
+                        :num-rev-subs (count rev-subs)))))
```
dwwoelfel: Could we add the total monthly revenue here too?

dwwoelfel (approved):
LGTM!

server/src/instant/dash/routes.clj
stopachka: Added!

PR #820, opened 2025-02-05
[Server] Welcome email fix

nezaj:
Noticed welcome emails haven't been going out and realized the issue was due to `chime-at` passing in an argument to send-welcome-email.

I also switched to using `swap!` instead of `reset!` to be thread-safe


dwwoelfel (approved):
LGTM!

PR #821, opened 2025-02-05
Split rs/store-conn by app

tonsky:
Design:

- `rs/store-conn` becomes a record w/ 2 fields
- `:sessions` is a datascript conn that stores session info. Shouldn’t have too high concurrent access as it’s only modified on connect/disconnect
- `:conns` is a `ConcurrentHashMap` of `app-id` → datascript conn that store instaql queries, datalog queries, tx-meta and subscriptions.
- All `instant.reactive.store` methods accept opaque `store` argument instead of explicit `conn`/`db`

This way we solve:

- Opening ws conn without app-id is okay, as it only goes into `:sessions` db and stays there
- When we modify any other data in datascript, we _must_ have app-id, so we can always find correct conn in `:conns`
- Current schema design doesn’t use datascript ref to sessions, it uses `.../session-id` (uuid) instead. That makes splitting dbs very easy
- By making `store` opaque, we can change impl further if we feel we need more concurrency or smth

### DIFF
--- a/server/src/instant/reactive/invalidator.clj
+++ b/server/src/instant/reactive/invalidator.clj
@@ -155,7 +155,7 @@
   [_process-id store-conn {:keys [app-id tx-id] :as wal-record}]
   (let [topics (topics-for-changes wal-record)
         [db session-ids] (rs/mark-stale-topics! store-conn app-id tx-id topics)
-        sockets (keep (partial rs/get-socket db) session-ids)]
+        sockets (keep #(:session/socket (rs/session db %)) session-ids)]
     sockets))
 
 (defn- topics-for-byop-triple-insert [table-info change]
@@ -219,7 +219,7 @@
   [table-info app-id store-conn {:keys [tx-id] :as record}]
   (let [topics (topics-for-byop-changes table-info record)
         [db session-ids] (rs/mark-stale-topics! store-conn app-id tx-id topics)
-        sockets (keep (partial rs/get-socket db) session-ids)]
+        sockets (keep #(:session/socket (rs/session db %)) session-ids)]
     sockets))
 
 ;; ------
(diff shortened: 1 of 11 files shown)


stopachka:
This plan sounds great to me @tonsky !

tonsky:
Okay this kinda works? All tests pass easily, from my quick testing it also went all fine. Although the change is significant, so maybe test it on staging first?

server/src/instant/reactive/invalidator.clj
```diff
@@ -367,7 +367,7 @@
                                                  :tx-created-at tx-created-at}))))
         (catch Throwable t
           (def -wal-record wal-record)
-          (def -store-value @store-conn)
+          (def -store-value store)
```
dwwoelfel: This used to get a snapshot of the store. Can we do something similar here? It make it easier to debug if you can get the state as it was when the error happened.

dwwoelfel: We have the `app-id`, so maybe we could just deref the `sessions` db and the db for the app in the reactive session?

server/src/instant/reactive/store.clj
```diff
@@ -279,89 +256,74 @@
...
-
-;; ------
-;; socket
+  [db sess-id]
+  (for [datom (d/datoms db :avet :subscription/session-id sess-id)]
+    [:db/retractEntity (:e datom)]))
+
+(defn remove-session! [store app-id sess-id]
+  ;; sync so new sessions are not added while we clean up this one
+  (locking (:sessions store)
```
dwwoelfel, line 308: Seems a little dangerous if I understand this correctly. This is relying on an implementation detail of datascript to prevent updates to the session db while we're removing the session. But if the `locking` gets removed from datascript, we might not realize that we need to update our code here.

dwwoelfel (approved):
Couple of comments, but LGTM!

server/src/instant/reactive/store.clj
tonsky: We actually have our own locking in `transact!` so should be okay

PR #822, opened 2025-02-06
improve investor updates and grammar

stopachka:
- re-ordered graph, so it matches how we send investor updates
- changed growth percentage to round to 1 decimal point, to match what we do in investor updates
- fixed some more grammar in our essay 

@nezaj @dwwoelfel @tonsky


dwwoelfel (approved):
LGTM!

PR #826, opened 2025-02-06
$entityIdStartsWith

dwwoelfel:
This is a prerequisite for supporting links in the Add/Edit row dialog in the Explorer.

This gives us a way to let the user search for the id that they want to link.

A better solution would be to support `$startsWith` on uuid fields, but I didn't want to get into adding support for uuid indexes just yet.

Query looks like this:

```clojure
{:posts {:$ {:where {:$entityIdStartsWith "f20"}}}}
```


nezaj (approved):
Nice!

PR #828, opened 2025-02-07
[docs] use tailwind in our react example

stopachka:
We used to use the 'bare' nextjs template. The downside for this, was that there was no tailwind. In order to make copy pasting as easy possible, we added inline styles. 

However, inline styles can be problematic because: 
1. Most people use tailwind, so if they wanted to repurpose things they couldn't
2. Most LLMs are better at tailwind. Having inline styles confuses them and gets them creating more inline styles. 

Recently nextjs has fixed their light mode bug. This means we can use their basic starter. I went ahead and updated our example so: 

1. We use nextjs' normal template, with tailwind. (I added a `--yes` flag so there are no questions when user pastes create-next-app command) 
2. I changed the typescript we used for `onSubmit` (this was causing errors when pasting)
3. I updated the css to use tailwind

--- 

I tried to keep the look and feel as similar as possible:

**Before**

image

**After**

image

**After in dark mode** 
CleanShot 2025-02-06 at 16 47 44@2x

@nezaj @dwwoelfel @tonsky

### DIFF
--- a/client/www/pages/docs/index.md
+++ b/client/www/pages/docs/index.md
@@ -15,7 +15,7 @@ And if you're ready, follow the quick start below to **build a live app in less 
 To use Instant in a brand new project, fire up your terminal and run the following:
 
 ```shell {% showCopy=true %}
-npx create-next-app -e hello-world instant-demo
+ npx create-next-app instant-demo --tailwind --yes
 cd instant-demo
 npm i @instantdb/react
 npm run dev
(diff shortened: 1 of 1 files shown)


client/www/pages/docs/index.md
```diff
@@ -28,8 +28,8 @@ Now open up `app/page.tsx` in your favorite editor and replace the entirety of t
 
 import { id, i, init, InstaQLEntity } from "@instantdb/react";
 
-// Instant app
-const APP_ID = "__APP_ID__";
+// ID for app: Test
+const APP_ID = "0860e7d4-0911-412f-aaf4-fb048ddea100";
```
dwwoelfel: Is this intentional?

stopachka: Oi, sorry! Updated

dwwoelfel (approved):
Cool!

PR #830, opened 2025-02-07
[docs] use tailwind for our magic codes example

stopachka:
**Before** 
image

CleanShot 2025-02-06 at 17 53 10@2x

**After** 
image

image

--- 

Apart from styling, made a few more changes:
1. We now autofocus
2. I made the inputs required, so you can't enter without filling something in 
3. Renamed components to EmailStep and CodeStep 
4. Added some more typescript types 

@nezaj @tonsky @dwwoelfel


dwwoelfel (approved):
LGTM!

PR #831, opened 2025-02-07
RFC: select-keys in queries

dwwoelfel:
This solves the problem where you only want to get a subset of the fields for an object in your query result.

A couple of use cases:
1. Showing the list of posts on a blog. You can just fetch titles instead of fetching the full body.
2. Fetching links on the explorer. We could just fetch the ids instead of fetching the full object for every linked object.

Syntax ideas:

**option-map**
```js
{posts: {$: {attributes: ["title"]}}}
```

This is what's implemented in the PR.

I like `attributes` because it's what we call the fields in the UI. 

Some other options:
1. `only`
2. `select`
3. `fields`

**gql-style**
```js
{posts: {title: true}}
```

The benefit of this syntax is that you could do more expressive things with the fields. For example, if we supported an `Image` type, you could do something like `{posts: {headerImage: {size: 'large', quality: 80}}}` to return a URL.

The downside is that it would be harder for us to tell if you wanted all fields or not when you selected links. For optimistic attrs, it might be hard for us to know if a field is meant to be a link or not.

### DIFF
--- a/server/src/instant/db/instaql.clj
+++ b/server/src/instant/db/instaql.clj
@@ -106,6 +106,7 @@
 (s/def ::before ::cursor)
 (s/def ::after ::cursor)
 (s/def ::aggregate #{:count})
+(s/def ::attributes (s/coll-of string?))
 
 (s/def ::option-map (s/keys :opt-un [::where-conds
                                      ::order
@@ -115,7 +116,8 @@
                                      ::offset
                                      ::before
                                      ::after
-                                     ::aggregate]))
+                                     ::aggregate
+                                     ::attributes]))
 
 (s/def ::forms (s/coll-of ::form))
 (s/def ::child-forms ::forms)
@@ -350,7 +352,9 @@
         aggregate (when-let [aggregate (:aggregate x)]
                     (coerce-aggregate! state aggregate))
 
-        x (dissoc x :where :order :limit :first :last :offset :before :after :aggregate)]
+        attributes (:attributes x)
+
+        x (dissoc x :where :order :limit :first :last :offset :before :after :aggregate :attributes)]
 
     (when (seq x)
       (ex/throw-validation-err!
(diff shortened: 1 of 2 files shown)


server/src/instant/db/datalog.clj
```diff
@@ -596,23 +596,25 @@
...
-                    :where [:and
-                            [:= :app-id app-id]
-
-                            [:=
-                             ;; Make sure it uses the av_index
-                             [:json_null_to_null :value]
-                             [:cast (->json (second lookup)) :jsonb]]
-                            [:= :attr-id [:cast (first lookup) :uuid]]
-                            :av]}])))
+    :e (if (every? uuid? v)
```
nezaj, line 599: Curious, why this change?

nezaj (approved):
Wooohooo!

client/sandbox/react-nextjs/pages/play/fields.tsx
```diff
@@ -0,0 +1,383 @@
...
+      'Content-Type': 'application/json',
+    },
+    body: JSON.stringify({
+      title: 'Comparisons example',
+    }),
+  });
+
+  const res = await r.json();
+  if (res.app) {
+    const db = init({ ...config, appId: res.app.id, schema: schema });
```
nezaj, line 294: Very nice, I like this pattern of seeding examples with data!

client/packages/core/src/queryTypes.ts
```diff
@@ -277,13 +278,18 @@ type InstaQLEntitySubquery<
...
+> = (Extract<keyof S['entities'][K]['attrs'], string> | 'id')[];
+
 type InstaQLQuerySubqueryParams<
   S extends IContainEntitiesAndLinks<any, any>,
   E extends keyof S['entities'],
 > = {
-  [K in keyof S['entities'][E]['links']]?:
-    | $Option
-    | ($Option &
+  [K in keyof S['entities'][E]['links']]?:  // XXX
```
stopachka: nit: is this XXX intended?

dwwoelfel: Nope! I had it there to remind me to update it, but it's already updated.

server/src/instant/db/datalog.clj
dwwoelfel: Postgres used a better index with the in, but I just pushed an update to make it even better.

Now we'll pass a single array parameter instead of one parameter per item in the list. That's good because preloading the entity ids might have a ton of entity ids. Now we won't be limited by the parameter count (I tested with 1M ids, but there's also a test that tests with 100k).

PR #833, opened 2025-02-07
[docs] add a `digging deeper` section in docs

stopachka:
As I was reading auth, it felt a bit confusing that right after the full example, we had more code snippets. I went ahead and made it clear that the remaining code snippets were explaining the full example. Added a `Digging Deeper` section. since our useAuth partial was only used here now, I also moved it back into the same file. 

@nezaj @dwwoelfel @tonsky


client/www/pages/docs/auth/magic-codes.md
```diff
@@ -163,9 +174,28 @@ on the backend during permission checks.
 
 On the client, `useAuth` will set `isLoading` to `false` and populate `user` -- huzzah!
 
-{% partial file="auth/useAuth.md" /%}
+### useAuth
```
nezaj: One note. We currently don't index ### topics. With this change someone won't be able to search for useAuth.

client/www/pages/docs/auth/magic-codes.md
```diff
@@ -176,7 +206,7 @@ db.auth.sendMagicCode({ email }).catch((err) => {
 
 Use `auth.sendMagicCode` to generate a magic code on instant's backend and email it to the user.
 
-## Sign in with Magic Code
+### Sign in with Magic Code
```
nezaj: Similarly this won't be searchable

client/www/pages/docs/auth/magic-codes.md
```diff
@@ -187,7 +217,7 @@ db.auth.signInWithMagicCode({ email: sentEmail, code }).catch((err) => {
 
 You can then use `auth.signInWithMagicCode` to authenticate the user with the magic code they provided.
 
-## Sign out
+### Sign out
```
nezaj: This also won't be searchable

nezaj (approved):
🚀

PR #834, opened 2025-02-07
[devtool] Improve devtool local flow, and add an `X` button 

stopachka:
A few improvements to devtool: 

1. I added a localStorage flag, so when these are on on, we default devtools to open our local dashboard
2. I added an `X` button in the devtools pane. people may not know that you can escape with an `esc` key, so wanted to add the button to be explicit 

CleanShot 2025-02-07 at 10 28 34@2x

@nezaj @tonsky @dwwoelfel


nezaj (approved):
Nice!

PR #835, opened 2025-02-07
Simplify grouped-queue

tonsky:
It looks like there are a few opportunities for `grouped-queue` rewrite:

- reduce code size (2x smaller as of right now)
- better concurrency (lock on put!, but no locks on poll)
- make it easy to swap executors
- join reserve, batch & consolidate logic together
- share metrics logic

TODO:

- [x] Migrate sessions
- [x] Migrate invalidator
- [x] Use “flat” entries (do not wrap into additional `{:item}`)
- [x] Move `put-at` to grouped-queue
- [x] Move `skipped` to grouped-queue
- [x] Move metrics to grouped-queue


server/src/instant/reactive/session.clj
```diff
@@ -722,44 +662,69 @@
...
-  (group-fn {:item {:session-id 1 :op :add-query :q {:users {}}}}))
+(defmethod combine [:refresh :refresh] [event1 event2]
+  (e2e-tracer/invalidator-tracking-step!
+   {:name          "skipped-refresh"
+    :tx-id         (:tx-id event1)
+    :tx-created-at (:tx-created-at event1)})
+  event2)
+
+(defmethod combine [:refresh-presence :refresh-presence] [event1 event2]
+  (update event2 :edits #(concat (:edits event1) %)))
```
stopachka: nit: should we use `into`? (I remember vaguely if we apply concat repeatedly on an item it can cause perf issues, but not 100% sure if it applies for this case

server/src/instant/grouped_queue.clj
```diff
@@ -1,334 +1,162 @@
...
+   How long to wait for existing tasks to finish processing before interrupting."
+  ([q]
+   ((:shutdown-fn q) {}))
+  ([q opts]
+   ((:shutdown-fn q) opts)))
+
+(defn size
+  "~ Amount of items currently in all queues"
+  [q]
+  (AtomicInteger/.get (:items q)))
```
stopachka: Nit: should we call this `num-items` or `size`? Reasoning: when I read `items` I expect it to be a list of something, but this is an AtomicInteger

server/src/instant/grouped_queue.clj
```diff
@@ -1,334 +1,162 @@
...
+
+     :metrics-path :: String | nil
+
+   A string to report gauge metrics to. If skipped, no reporting"
+  [{:keys [group-key-fn combine-fn process-fn executor max-workers metrics-path]}]
+  (let [groups      (ConcurrentHashMap.)
+        accepting?  (atom true)
+        processing? (atom true)
+        items       (AtomicInteger. 0)
+        workers     (AtomicInteger. 0)
```
stopachka: nit: would call this `num-workers` or something too

server/src/instant/grouped_queue.clj
```diff
@@ -1,334 +1,162 @@
...
+            (Map/.put groups key group)
+            (ExecutorService/.submit executor ^Runnable #(process q key group))))
+        (AtomicInteger/.incrementAndGet items)))))
+
+(defn- longest-wait-time [groups]
+  (when-some [items (->> groups
+                         (Map/.values)
+                         (keep Queue/.peek)
+                         not-empty)]
+    (- (System/currentTimeMillis) (transduce (map ::put-at) min items))))
```
stopachka, line 67: damn, transduce. Cool!

server/src/instant/grouped_queue.clj
```diff
@@ -1,334 +1,162 @@
...
+   (java.util Map Queue)
+   (java.util.concurrent ConcurrentHashMap ConcurrentLinkedQueue Executors ExecutorService LinkedBlockingQueue ThreadPoolExecutor TimeUnit)
+   (java.util.concurrent.atomic AtomicInteger)))
+
+(defn- poll
+  "Gets 0..∞ items from group, fetching as many combinable items as possible in a row.
+   Returns 1 (possibly combined) item or nil"
+  [group combine-fn]
+  (loop [item1 (Queue/.poll group)]
+    (clojure+/cond+
```
stopachka, line 17: cool macro

server/.clj-kondo/imports/http-kit/http-kit/httpkit/with_channel.clj
```diff
@@ -0,0 +1,16 @@
+(ns httpkit.with-channel
```
stopachka, line 1: Curious q -- what is the story of these additions? I don't think we use http-kit

stopachka (approved):
LGTM!

server/src/instant/reactive/session.clj
tonsky: Probably won’t matter here, but sure, lazy sequences are a loaded gun

server/src/instant/grouped_queue.clj
tonsky: Good point!

server/src/instant/grouped_queue.clj
tonsky: Right?

server/.clj-kondo/imports/http-kit/http-kit/httpkit/with_channel.clj
tonsky: Yeah, so basically clj-kondo needs to “import” rules from dependencies before it can apply them. I did just that (imported them) by running

```
clj-kondo --lint "$(clj -Spath)" --dependencies --copy-configs --skip-lint
```

If we add new library that uses custom rules, we’ll need to redo this.

http-kit comes from `djblue/portal`, which is in our main dependencies for some reason (if you ask me, it should be on `:dev` profile)

stopachka: SGTM! Re: djblue/portal, I think we keep it in main, because sometimes we do a remote connection to production and use it.

PR #838, opened 2025-02-10
Add `id` import in `Writing data` docs

ovistoica:
Hello, this is a fix in the `Writing data` docs where `id` fn was omitted from import. It is important as it took me 10 minutes to find out where this was imported from. Initially, I thought `id` was something that just works inside `transact` as some kind of remarkable js macro.

Current docs: 
image

Hope this helps 👍


stopachka (approved):
Thank you!

PR #841, opened 2025-02-10
simplify pagination test

stopachka:
Our `pagination` test has been flaky for a while. join-rows tend come back in some slightly different order, and cause issues. 

I think the long term fix is it improve our `query-pretty` function. However for now, just changing this test to be a bit more specific about what we test. 

@dwwoelfel @nezaj @tonsky


dwwoelfel (approved):
LGTM!

IIUC, the bookshelves that we're making the link could be one of many bookshelves, so the join rows that we get back sometimes change for unknown reasons.

stopachka:
> IIUC, the bookshelves that we're making the link could be one of many bookshelves, so the join rows that we get back sometimes change for unknown reasons.

Ah that makes a lot of sense!

PR #842, opened 2025-02-11
[Stripe] sm stripe fix

nezaj:
We got an email about a webhook failing when we manually deleted the subscription for an app that no longer exists.

Longer term fix for this would be to introduce a trigger + sweeper on app deletes for subscribed users


stopachka (approved):
LGTM!

PR #844, opened 2025-02-11
on-delete-reverse

tonsky:
- [x]  backend part
- [x] test
- [x] frontend
- [x] admin UI
- [x] CLI
- [x] docs


tonsky:
Looks like it’s done

UI

Screenshot 2025-02-12 at 21 41 20

Docs

Screenshot 2025-02-12 at 21 59 45

stopachka (approved):
Aweesome!

server/src/instant/model/schema.clj
```diff
@@ -186,7 +190,7 @@
 
 (defn cascade-message [[etype label]]
   (str etype "->" label ": "
-       "Cascade delete is only possible on one-to-one and one-to-many attributes. "
+       "Cascade delete is only possible on cardinality: one attributes. "
```
dwwoelfel: Could we write this in a way that matches what the user writes in their `instant.schema.ts` file? I'm worried they might be confused since we don't use `cardinality` in that file.

Would it be correct to say "only possible on links with `has: "one"`" or "not possible on links with `has: "many"`"?

server/src/instant/db/transaction.clj
```diff
@@ -226,36 +226,70 @@
...
+                                   {:select [[[:cast nil :uuid]] nil nil] :where false}
+                                   {:values reverse-attrs+etypes})]
+                                
+                                [:refs_reverse {:select [:triples.entity_id :triples.value :attrs_reverse.forward_etype :attrs_reverse.reverse_etype]
+                                                :from   :triples
+                                                :join   [:attrs_reverse [:= :triples.attr_id :attrs_reverse.id]]
+                                                :where  [:and
+                                                         :triples.vae
+                                                         [:= :triples.app_id app-id]]}]]
+               :from   :entids_reverse
```
dwwoelfel: Would it be better to do:

```clojure
:union [{:select :* :from :entids_reverse}
        {:select :* :from :entids_forward}]
```

here instead of doing the union in the `entids_reverse` cte?

dwwoelfel (approved):
Couple of comments, but LGTM!

server/src/instant/db/transaction.clj
tonsky: Actually, both my current approach and your suggestion are wrong. It’s not enough to go once in forward direction only and once in reverse direction only. Each step of recursive iteration should go both. There might be a chain of starting entity -> forward -> reverse -> forward -> reverse and we want to capture all that... Back to drawing board

tonsky: Solved

PR #845, opened 2025-02-12
[www] add cached to dash fetch, remove the number of loading screens in devtool

stopachka:
This PR ships a series of different fixes

**asClientOnlyPage: router.isReady issues** 

In Devtool and Dash, we relied on query params using `useRouter`. However, on first load the query param is _always_ going to be empty. We need to wait for router.isReady to actually get the query param. 

This caused an issue in Devtool, where if you really slowed down the page, you would see a "No app id provided" error. 

**Caching /dash fetch** 

Both devtool and instant wait on GET /dash. This makes for a worse Explorer experience: since the Instant DB itself works offline, it would be great if you could load Explorer offline too. 

To solve this, I added a cache around GET /dash, with a 1 week TTL 

**Removing more loading screens**

In Devtool, we had loading screens saying things like "Loading...", but only for a few milliseconds. This led to a jarring experience -- I replaced them with FullscreenLoading.

--- 

@dwwoelfel @nezaj @tonsky

### DIFF
--- a/client/www/components/clientOnlyPage.tsx
+++ b/client/www/components/clientOnlyPage.tsx
@@ -1,0 +1,29 @@
+import { useIsHydrated } from '@/lib/hooks/useIsHydrated';
+import { NextRouter, useRouter } from 'next/router';
+
+export function useReadyRouter(): Omit<NextRouter, 'isReady'> {
+  const router = useRouter();
+  if (!router.isReady) {
+    throw new Error(
+      'Router wasn not ready. Make sure to call this hook somewhere inside an `asClientOnlyPage` component',
+    );
+  }
+  return router;
+}
+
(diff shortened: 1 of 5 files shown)


client/www/components/clientOnlyPage.tsx
```diff
@@ -0,0 +1,29 @@
+import { useIsHydrated } from '@/lib/hooks/useIsHydrated';
+import { NextRouter, useRouter } from 'next/router';
+
+export function useReadyRouter(): Omit<NextRouter, 'isReady'> {
+  const router = useRouter();
+  if (!router.isReady) {
+    throw new Error(
+      'Router wasn not ready. Make sure to call this hook somewhere inside an `asClientOnlyPage` component',
```
nezaj: Minor typo

nezaj (approved):
LGTM!

PR #847, opened 2025-02-12
[Docs] Defer query docs

nezaj:
We got a question in [discord](https://discord.com/channels/1031957483243188235/1338329481961803837) about how to pause/disable queries. We don't support pausing but we do support deferring and just haven't added that into the docs.

image


dwwoelfel (approved):
LGTM!

PR #848, opened 2025-02-12
make delete-row sticky

stopachka:
A user reported that delete row did not show up, when you kept scrolling in the explorer. 

This makes delete row sticky, so the no matter how low you scroll, it shows up at the header. 

@dwwoelfel @nezaj @tonsky 

https://github.com/user-attachments/assets/9ae74de0-0f70-4d53-af21-6420dcd91db2

### DIFF
--- a/server/src/instant/scratch/export.clj
+++ b/server/src/instant/scratch/export.clj
@@ -59,18 +59,13 @@
                                                                            [:cast (json/->json v) :jsonb]))))
                                                     triples)})))))
 
-(defn prod-conn-str []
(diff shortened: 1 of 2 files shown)


server/src/instant/scratch/export.clj
```diff
@@ -59,18 +59,13 @@
                                                                            [:cast (json/->json v) :jsonb]))))
                                                     triples)})))))
 
-(defn prod-conn-str []
```
stopachka, line 62: made a quick fix here to use tool/with-prod-conn (needed to import stopa.io)

dwwoelfel (approved):
LGTM!

PR #849, opened 2025-02-12
[like] handle regex chars

stopachka:
Previously, if you wrote a query like: 

```
{
  "posts": {
    "$": {
      "where": {
        "slug": {
          "$like": "%2)"
        }
      }
    }
  }
}
```

You would get an error: 

```

SyntaxError: Invalid regular expression: /^.*2)$/: Unmatched ')'

Source
new RegExp
<anonymous> (0:0)
../packages/core/dist/module/instaql.js (79:1) @ makeLikeMatcher

  77 |     }
  78 |     const regexPattern = pattern.replace(/%/g, '.*').replace(/_/g, '.');
> 79 |     const regex = new RegExp(^${regexPattern}$, caseSensitive ? undefined : 'i');
     | ^
```

This is because we did not escape special regex chars when creating our like matcher. I updated it so we do. 

@dwwoelfel @nezaj @tonsky

### DIFF
--- a/client/packages/core/__tests__/src/instaql.test.js
+++ b/client/packages/core/__tests__/src/instaql.test.js
@@ -4,7 +4,7 @@ import zenecaAttrs from './data/zeneca/attrs.json';
 import zenecaTriples from './data/zeneca/triples.json';
 import { createStore, transact } from '../../src/store';
 import query from '../../src/instaql';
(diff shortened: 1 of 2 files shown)


dwwoelfel (approved):
LGTM!

PR #850, opened 2025-02-12
Treat json nulls as postgres nulls in the unique index

dwwoelfel:
Part 1 of fast paginated queries across many entities.

Also fixes a bug reported at https://discord.com/channels/1031957483243188235/1304944961380286515

After this change, users will be able to store multiple entities that have null as the value for a unique attribute.

As part of making pagination queries faster, we need all the null values to already be in the index. We'll have to auto-populate the null values ourselves if they're undefined. We don't want to break users' `unique` indexes by inserting nulls, so we treat jsonb nulls like postgres null and postgres automatically treats them as distinct values.

We do this by converting `'null'::jsonb` into a postgres `NULL` in an index expression. That way postgres treats the null just like postgres nulls in the index.

Other alternatives we considered:

1. Make the `value` column nullable. Decided against this because the size of the change is a lot bigger and scarier.
2. Use a partial index that filters "null". Decided against this because it might be handy to be able to look up null things in the index.

Deployment plan:

1. Create the index concurrently:

```
create or replace function public.json_null_to_null(v jsonb) returns jsonb
  language sql immutable
  as $$
    select case
      when v = 'null'::jsonb then null
      else v
  end;
$$;

create unique index concurrently av_ignore_nulls_index
  on triples(app_id, attr_id, json_null_to_null(value))
  include (entity_id)
  where av;
```

3. Deploy the backend and wait for it to fully deploy

4. Run the migration

We need both indexes to be active while both versions of the code are active. The migration replaces the av_index with the new one, so we want to wait until all instances are on the new version before running the migation.

### DIFF
--- a/server/test/instant/db/instaql_test.clj
+++ b/server/test/instant/db/instaql_test.clj
@@ -370,8 +370,8 @@
 
   (testing "makes sure we use distinct"
     (is-pretty-eq? (-> (iq/query @ctx {:users {:$ {:where {:bookshelves {:in
-                                                                         ;; `eid-worldview` and `eid-the-way-of-the-gentleman` 
-                                                                         ;; both match `stopa` 
+                                                                         ;; `eid-worldview` and `eid-the-way-of-the-gentleman`
+                                                                         ;; both match `stopa`
                                                                          ;; limit `3` should _still_ return joe, stopa, alex
                                                                          [(resolvers/->uuid @r "eid-worldview")
                                                                           (resolvers/->uuid @r "eid-the-way-of-the-gentleman")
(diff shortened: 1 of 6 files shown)


server/test/instant/db/instaql_test.clj
```diff
@@ -2639,6 +2639,75 @@
...
+                         first
+                         (get-in ["Plan" "Plans"])
+                         first)
+                ;; Make sure it's using the full index
+                expected-index-cond (format "((triples.app_id = '%s'::uuid) AND (triples.attr_id = '%s'::uuid) AND (CASE WHEN (triples.value = 'null'::jsonb) THEN NULL::jsonb ELSE triples.value END = '\"a\"'::jsonb))"
+                                            (:id app)
+                                            (:handle attr-ids))]
+
+            (is (= expected-index-cond (get plan "Index Cond")))
+            (is (= "av_index" (get plan "Index Name")))))
```
stopachka, line 2689: Dope!

stopachka (approved):
Awesome @dwwoelfel !

PR #854, opened 2025-02-13
Use level-sym instead of default-level-sym

stopachka:
Technically both seem to work, but feels safer to use level-sym

@dwwoelfel @nezaj @tonsky


dwwoelfel (approved):
LGTM!

PR #855, opened 2025-02-13
[ex] be a bit harsher on how we parse magic codes 

stopachka:
Previous we used string/trim to parse magic codes. This threw a 500 error. 

I instead added a `safe-trim` function, which a) guarantees the input is a string, and then trims it. 

I also updated a few other places to use `coerce-non-blank-str` 

@dwwoelfel @nezaj @tonsky

### DIFF
--- a/server/src/instant/util/string.clj
+++ b/server/src/instant/util/string.clj
@@ -24,6 +24,10 @@
     (.update ^CRC32 crc bytes)
     (.getValue ^CRC32 crc)))
 
+(defn ensure-non-blank-str [s]
+  (and (string? s)
+       (string/trim s)))
+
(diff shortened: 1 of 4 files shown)


server/src/instant/util/string.clj
```diff
@@ -24,6 +24,10 @@
     (.update ^CRC32 crc bytes)
     (.getValue ^CRC32 crc)))
 
+(defn safe-trim [s]
+  (and (string? s)
```
dwwoelfel: Maybe this should be `when` so that you return null instead of `false` from safe-trim?

```suggestion
  (when (string? s)
```

dwwoelfel (approved):
One comment, but LGTM!

PR #858, opened 2025-02-13
Add storage sweeper

nezaj:
Adds a sweeper for storage!

Uses the scheme outlined by @dwwoelfel below!


dwwoelfel (commented):
We needed jobs for indexing because it's a long-running process and if you get interrupted in the middle of indexing, you'll leave the user's app in a bad state.

Deleting objects from s3 is much more forgivable--it doesn't matter if the s3 api fails, for example, because you can just try again later. And it doesn't matter if you do it twice because deleting a non-existent file is a no-op.

I don't think we need the overhead of a job queue with multiple job stages here--we can do it all in the `app_file_to_sweep` table.

To get the next batch of files to delete, just do:

```sql
with selected_rows as (
    select id 
    from app_file_to_sweep 
    where process_id is null
       -- if it hasn't been updated in five minutes we can assume it's stuck
       -- no harm if we try to delete something that's already being deleted
       or (process_id is not null and updated_at < :time-minus-five-minutes:)
    for update skip locked
)
update app_file_to_sweep 
set process_id = @config/process-id
where id in (select id from selected_rows)
returning *;
```

Then delete those rows from the db after deleting the files from s3. No need for the overhead of tracking job stages.

I also wouldn't start 16 workers to handle deleting s3 objects because I don't anticipate we're going to have to delete anywhere close to 1M objects per hour. I would just run the `delete` every hour and delete 1000/hour. 

If you're worried about getting behind on deleting, you can do a loop that terminates after it runs and deletes less than 500 items or runs 10 times. And if it runs 10 times, then you know it's getting backed up and you can send a warning into the errors or teams channel.

server/src/instant/config.clj
```diff
@@ -193,6 +193,11 @@
                      :prod "https://api.instantdb.com"
                      "http://localhost:8888"))
 
+(def s3-bucket-name
+  (case (get-env)
+    :prod "instant-storage"
+    "instantdb-test-bucket"))
```
nezaj, line 199: Updated our IAM policy as well so our StorageUser has access to this

server/src/instant/storage/sweeper.clj
```diff
@@ -0,0 +1,155 @@
...
+   :end-at (str (Instant/now))
+   :run-time-ms (ms-between start-ms (Instant/now))})
+
+(defn files-available-wheres []
+  [:or
+   [:is :process-id nil]
+   [:and
+    [:is-not :process-id nil]
+    [:< :updated-at
+     [:- :%now [:raw "interval '5 minutes'"]]]]])
```
stopachka, line 40: dam `:%now` til

server/src/instant/storage/sweeper.clj
```diff
@@ -0,0 +1,155 @@
...
+                         :updated-at :%now}
+                   :where [:in :id {:select [:id]
+                                    :from :to-update}]
+                   :returning [:*]}))))
+
+(defn delete-files!
+  ([params]
+   (delete-files! (aurora/conn-pool :write) params))
+  ([conn {:keys [ids]}]
+   (sql/execute!
```
stopachka: nit: shouldn't this also have a tag? ::delete-files!

server/src/instant/storage/sweeper.clj
```diff
@@ -0,0 +1,155 @@
...
+   [instant.jdbc.sql :as sql]
+   [instant.config :as config]
+   [instant.discord :as discord])
+
+  (:import
+   (java.time Duration Instant)
+   (java.lang AutoCloseable)
+   (java.time.temporal ChronoUnit)))
+
+(defonce max-loops 10) ;; max loops per sweep job
```
dwwoelfel: I would just make these defs, otherwise they won't change when you change the value in the code and re-load the file.

`defonce` is more for things where redefining it would break something, e.g. an atom where you're storing some state.

server/src/instant/config.clj
```diff
@@ -193,6 +193,11 @@
                      :prod "https://api.instantdb.com"
                      "http://localhost:8888"))
 
+(def s3-bucket-name
```
dwwoelfel, line 196: ❤️

server/resources/migrations/54_add_storage_sweeper.up.sql
```diff
@@ -0,0 +1,4 @@
+alter table app_files_to_sweep add column process_id text;
+alter table app_files_to_sweep add column updated_at timestamp;
```
dwwoelfel: I would make this `timestamp with time zone not null default now()` and then you can add

```sql
create trigger update_updated_at_trigger
  before update on app_files_to_sweep for each row
  execute function update_updated_at_column();
```

That way you don't have to manually modify the updated_at timestap (the `update_updated_at_column` function is already defined)

dwwoelfel (approved):
LGTM!

PR #861, opened 2025-02-13
[sm] remove user.email in signOut

stopachka:
User noted this from the docs -- frontend doesn't require the email key.

@dwwoelfel @nezaj @tonsky


dwwoelfel (approved):
LGTM!

PR #867, opened 2025-02-13
[cli] add ability to set custom locations

stopachka:
Previously the CLI explicitly relied on instant.* files living in the root directory. 

I've updated things so now:
1. We are okay if you save it in root, src, or app directories 
2. You can also set specific locations using env vars `INSTANT_SCHEMA_FILE_PATH` and `INSTANT_PERMS_FILE_PATH` 

**Docs: https://instant-www-js-clipaths-jsv.vercel.app/docs/cli#where-to-save-files** 

@dwwoelfel @nezaj @tonsky


client/packages/cli/src/index.js
```diff
@@ -792,16 +793,16 @@ async function pullPerms(appId, { pkgDir, instantModuleName }) {
...
-
-  if (await pathExists(join(pkgDir, 'instant.perms.ts'))) {
+  const prev = await readLocalPermsFile();
+  if (prev) {
     const ok = await promptOk(
       'This will overwrite your local instant.perms file, OK to proceed?',
     );
 
-    if (!ok) return;
+    if (!ok) return { ok: true };
```
nezaj: This line feels a bit strange, I think the intention is to return a valid exit code but it reads strangely

nezaj (approved):
Aside from one nit LGTM!

client/packages/cli/src/index.js
stopachka: I made this similar to what is already written in pullSchema

The reasoning was: 

If a user presses `quit`, it's not really an error for us to stop. There are sometimes reasons for us to stop that are not errors though. 

I'll update both to say something like: 

```
    const wantsToContinue = await promptOk(
      'This will overwrite your local instant.perms file, OK to proceed?',
    );
```

PR #869, opened 2025-02-14
Insert nulls for indexed blob attrs

dwwoelfel:
Part 2 of faster pagination queries.

Easier to review if you ignore whitespace changes: https://github.com/instantdb/instant/pull/869/files?w=1

This ensures that every entity will have a triple in the database for an indexed blob attr. Instead of the value being missing, we add a null. That way all of the entities will be in the index and we can avoid joining onto the rest of the table to generate synthetic nulls when we do pagination queries.

We really only need it for checked and indexed attrs, since those are the ones you can paginate against. But then it gets a lot more complicated to determine when you need to add the nulls. If you started two jobs to check and then index the attr simultaneously, then you might not realize that both were active and you needed to add the nulls. Much easier to just always add them when you index.

There are three places where we handle inserting the null attr values:

1. When you start an indexing job, we'll insert nulls as part of the job
2. When you add a new entity, a CTE will automatically insert nulls for any new entities that were created
3. When you create a new indexed attr for an existing namespace, a CTE will automatically insert nulls for any new entities that were created.

For 3, I originally had it create an indexing job in permissioned-transaction for new attrs. That's better in some ways, because we prevent the invalidator from getting a huge chunk of work if the namespace had many existing entities. The downside was that it injected some unpredictability from the update in the background. That caused some test flakiness and made them a lot more complicated because they all needed to run the job queue.

There is one place where we don't insert the null attrs. When you do `:retract-triple`, we don't insert a null if you retracted the indexed attr. We only use that for unlinking, so it didn't seem worth the trouble. I added a note in the file in case we start using `:retract-triple` for more things in the future.

**Incidental changes**

1. Upgrades our tests to actually use postgres 16. I thought I had done it before, but apparently I didn't update the actual dockerfile and we were still using postgres 13 that was incorrectly labeled as 16.
2. Some of the indexing tests failed with postgres 16. I found a workaround by wrapping the failing tests in `with-zeneca-app`, which added enough triples that postgres used better indexes.
3. Removed the confirmation from `make prod-version`, since it only reads from the db

**Deployment strategy**
1. Fully deploy changes
2. Run the `indexing-jobs/insert-nulls-for-existing-indexed-blob-attrs` function to insert nulls for existing attrs


server/src/instant/db/model/attr.clj
```diff
@@ -257,104 +259,121 @@
...
+                             "idents" [:coalesce
+                                       {:select [[[:json_agg [:row_to_json :ident-inserts]]]]
+                                        :from :ident-inserts}
+                                       [:cast [:inline "[]"] :json]]
+                             "attrs" [:coalesce
+                                      {:select [[[:json_agg [:row_to_json :attr-inserts]]]]
+                                       :from :attr-inserts}
+                                      [:cast [:inline "[]"] :json]]]]]}
+           result (sql/execute-one! ::insert-multi! conn (hsql/format query))]
+       {:attrs (-> result
```
dwwoelfel, line 419: Lots of whitespace changes here, but this is the only real change. I need to get the attrs that were created so that we can add nulls for existing entities if we added an indexed blob attr.

server/src/instant/fixtures.clj
```diff
@@ -26,14 +28,37 @@
       :params {:app_id (:id a)}
       :body {}})))
 
+(defmacro with-indexing-job-queue [job-queue & body]
```
dwwoelfel, line 31: Moved this into fixtures so that we can use it in other tests if needed (didn't end up needing it, but it's nicer to have it here anyway).

server/src/instant/db/transaction.clj
```diff
@@ -55,11 +54,6 @@
 ;; ----
 ;; coerce
 
-(defn- walk-uuids
```
dwwoelfel, line 58: Moved this into the uuid-util ns

server/src/instant/fixtures.clj
```diff
@@ -51,10 +64,21 @@
             r (resolvers/make-zeneca-resolver id)]
         (f app r)))))
 
+(defn with-zeneca-app-no-indexing [f]
```
dwwoelfel, line 67: I use this for the indexing test--turns out the test wasn't really testing what I wanted it to because everything was already indexed.

server/src/tool.clj
```diff
@@ -99,18 +99,22 @@
...
-                                              ""))))))))
+    (-> q
+        (clojure.string/replace #"\?"
+                                (fn [_] (let [i @idx
+                                              v (nth params i)]
+                                          (swap! idx inc)
+                                          (str (cond
+                                                 (int? v) (format "%s" v)
+                                                 (string? v) (format "'%s'" (-> v
+                                                                                (.replace "'" "''")))
```
dwwoelfel, line 110: Apparently you escape a `'` in a postgres string with `''`

server/test/instant/db/indexing_jobs_test.clj
```diff
@@ -257,11 +389,12 @@
                             (every? (fn [{:keys [id]}]
                                       (= "completed" (:job_status (jobs/get-by-id id))))
                                     [job]))
-                          1000)
+                          5000)
```
dwwoelfel, line 392: Had to bump this timeout--I think our test instance runs into noisy neighbor problems against our other test instances.

server/Makefile
```diff
@@ -68,8 +68,6 @@ prod-connect:
 
 prod-version:
 	@DATABASE_URL=$$(./scripts/prod_connection_string.sh); \
-	echo "Checking prod db version... [y/N]" && \
```
dwwoelfel, line 75: Removed this because we're only reading here

server/src/instant/fixtures.clj
stopachka: 🫡

server/src/instant/db/model/triple.clj
```diff
@@ -525,6 +608,9 @@
                 :system-catalog-app-id system-catalog-app-id}]
     (sql/do-execute! conn (hsql/format query {:params params}))))
 
+;; n.b. if we ever use `:retract-attr` for blob attrs (it's currently
```
stopachka: nit: do we mean `retract-triple` here?

server/src/instant/db/model/triple.clj
```diff
@@ -445,37 +446,119 @@
...
+                         :returning :*}]
                        [:remaining-inserts
                         {:insert-into [[:triples triple-cols]
                                        {:select triple-cols
                                         :from :remaining-triples}]
                          :on-conflict [:app-id :entity-id :attr-id :value-md5]
                          :do-nothing true
+                         :returning :*}]
+                       [:indexed-null-triples
+                        {:select [[:needs-null-attr.app-id :app-id]
```
stopachka, line 458: Monster

stopachka (approved):
Masterful work @dwwoelfel !

server/src/instant/db/model/triple.clj
dwwoelfel: I did! Thanks!

PR #870, opened 2025-02-14
Add metadata update helpers

nezaj:
Used these to update content types for Clark! Will use them later to run through all our $files triples to make them in sync with s3


stopachka (approved):
Nice!

PR #879, opened 2025-02-17
Check DB version before deploy

tonsky:
An additional check before running deploy:

```
[/ws/instant/server] ./scripts/eb_deploy.clj
Fetching application versions...
Getting db URL...
Checking prod db version...
[ERROR] Looks like you need to run DB migrations first
  Current prod version: 54
      Latest migration: 55
```


stopachka (approved):
Heeck yeah!

PR #880, opened 2025-02-17
optimize ->cel-map

tonsky:
According to profile, we spend a lot of time in ->cel-map, specifically walking data structures and rebuilding them to stringify keys and values. This patch changes it so that we coerce to string at retrieval time, which should be faster

I am only not sure about CelMap::entrySet. It doesn’t show up in tests, but could it be used somewhere nevertheless?


stopachka (approved):
Nice idea!

PR #888, opened 2025-02-17
[devtool]  allowedHosts

stopachka:
This adds an `allowedHosts` config key. It can be useful if a user has a custom scheme for loading dev pages -- i.e `site.local`, etc 

@nezaj @tonsky @dwwoelfel


nezaj (approved):
Looks great!

PR #892, opened 2025-02-18
Fix match topic for comparators with non-numbers

dwwoelfel:
Our comparators relied on `<`-like operators and those only work on numbers in Clojure.

Created a set of custom functions for comparing each type we support (bool, string, number, and date).

Also fixes how we handle `nil`, which will become more relevant when I merge in https://github.com/instantdb/instant/pull/869

### DIFF
--- a/server/src/instant/reactive/store.clj
+++ b/server/src/instant/reactive/store.clj
@@ -569,7 +569,6 @@
 (defn match-nil
   "nil is always the smallest value and the comparator value can't be nil"
   [op]
-  (println "NIL!" op)
   (case op
     :$gt false
     :$gte false
@@ -621,7 +620,7 @@
                        iv-part))
       (when (contains? dq-part :$not)
         (let [not-val (:$not dq-part)]
-          (ucoll/seek (partial not= not-val) iv-part))))))
+          (ucoll/exists? (partial not= not-val) iv-part))))))
 
 (defn match-topic?
   [[iv-idx iv-e iv-a iv-v]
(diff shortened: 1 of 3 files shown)


server/src/instant/reactive/store.clj
```diff
@@ -525,6 +528,56 @@
...
+(defn bool-lt [^Boolean a ^Boolean b]
+  (neg? (.compareTo a b)))
+
+(defn bool-lte [^Boolean a ^Boolean b]
+  (>= 0 (.compareTo a b)))
+
+(defn match-nil
+  "nil is always the smallest value and the comparator value can't be nil"
+  [op]
+  (println "NIL!" op)
```
nezaj: Did we want to leave this in?

server/src/instant/util/coll.clj
```diff
@@ -24,6 +24,15 @@
                not-found))
            not-found coll)))
 
+(defn exists?
```
nezaj, line 27: Just curious, why like this instead of doing some? Wouldn't this work?

```
(defn exists?
  [pred coll]
  (boolean (some pred coll)))
  ```

nezaj (approved):
Very nice!

server/src/instant/reactive/store.clj
dwwoelfel: Whoops!

server/src/instant/util/coll.clj
dwwoelfel: I'd like it to return a `boolean` instead of whatever `pred` returns or nil.

dwwoelfel: Oh, I see your example coerces. But seems more efficient to write directly?

PR #893, opened 2025-02-18
Add getAuth

nezaj:
We got an ask for one-off `getAuth` and seemed straight-forward / quick-win.

image

### DIFF
--- a/client/packages/core/src/Reactor.js
+++ b/client/packages/core/src/Reactor.js
@@ -1354,6 +1354,14 @@ export default class Reactor {
     };
   }
 
+  async getAuth() {
+    const {user, error} = await this.getCurrentUser();
+    if (error) {
+      throw new Error(error);
+    }
+    return user;
+  }
+
   subscribeConnectionStatus(cb) {
     this.connectionStatusCbs.push(cb);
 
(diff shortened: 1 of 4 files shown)


client/packages/core/src/Reactor.js
```diff
@@ -1354,6 +1354,14 @@ export default class Reactor {
     };
   }
 
+  async getAuth() {
+    const {user, error} = await this.getCurrentUser();
+    if (error) {
+      throw new Error(error);
```
stopachka: nit: If `error` is an object, I am not sure that you can throw it this way. Would try to test this case and make sure

client/www/pages/docs/auth/magic-codes.md
```diff
@@ -228,3 +228,13 @@ db.auth.signOut();
 Use `auth.signOut` from the client to invalidate the user's refresh token and
 sign them out.You can also use the admin SDK to sign out the user [from the
 server](/docs/backend#sign-out).
+
+## Get auth
+
+```javascript
+const user = db.getAuth();
```
stopachka: nit: `await`

stopachka (approved):
Niice!

PR #894, opened 2025-02-18
Fix flags session

dwwoelfel:
Fixes the dummy session we create for flags so that it gets updated in the invalidator


stopachka (approved):
niice!

PR #895, opened 2025-02-18
Warn users on deploy instead of blocking it

tonsky:
Follow-up to  #879

We now allow users to choose whether to ignore the mismatched DB versions or abort. Also do fetch to ensure commit that is deployed is presented locally


dwwoelfel:
Tested locally and works for me!

(looks like a flaky test in the index test--that might be fixed when I merge in https://github.com/instantdb/instant/pull/869)

PR #899, opened 2025-02-18
Increase timeout during indexing-job tests

dwwoelfel:
Seeing some flaky tests where we timeout while the test is still going because simple sql queries take forever (e.g. an update by id will take 200ms). I suspect there's some issue with the test machine that makes it slow, so increasing the timeout. 

If it keeps happening, I'll look into a better way to check that progress is being made.


stopachka (approved):
🚢🚢🚢

PR #900, opened 2025-02-18
[docs] small scroll improvements

stopachka:
- Added bottom padding for sidebar
- Reducing the spacing between sidebar sections. This way in at least a 15 inch macbook with bookmarks enabled, there is no need to scroll the sidebar

@nezaj @tonsky @dwwoelfel


nezaj (approved):
🔥

PR #902, opened 2025-02-19
New Storage docs

nezaj:
What it says on the tin!

[Preview Link](https://instant-www-js-storage-docs-jsv.vercel.app/docs/storage)

### DIFF
--- a/client/sandbox/react-nextjs/pages/play/avatar.tsx
+++ b/client/sandbox/react-nextjs/pages/play/avatar.tsx
@@ -123,7 +123,7 @@ function ProfilePage() {
   );
 }
 
-function Wrapper() {
+function Page() {
   const { isLoading, user, error } = db.useAuth();
   if (isLoading) {
     return <div>Loading...</div>;
@@ -137,4 +137,4 @@ function Wrapper() {
   return <Login auth={db.auth} />;
 }
 
-export default Wrapper;
+export default Page;
--- a/client/sandbox/react-nextjs/pages/play/avatar.tsx
+++ b/client/sandbox/react-nextjs/pages/play/avatar.tsx
@@ -1,0 +1,140 @@
+'use client';
+
+import { init, tx } from '@instantdb/react';
+import React, { useState, useEffect } from 'react';
+import Login from '../../components/Login';
+import config from '../../config';
+
+const db = init(config);
+
+interface AvatarUploadProps {
+  defaultSize?: number;
+}
+
+function AvatarUpload({ defaultSize = 96 }: AvatarUploadProps) {
+  const [isUploading, setIsUploading] = useState(false);
+  const { user } = db.useAuth();
+  const {
+    isLoading: isLoadingAvatar,
+    data,
+    error: avatarError,
+  } = db.useQuery(
+    user
+      ? {
+          profiles: {
+            $: {
+              where: { '$user.id': user.id },
+            },
+            avatar: {},
+          },
+        }
+      : null,
+  );
+  if (isLoadingAvatar) return <div>Loading...</div>;
+  if (avatarError) return <div>Error: {avatarError.message}</div>;
+
+  const profile = data.profiles[0];
+  const { id: profileId } = profile;
+  const avatar = profile.avatar[0];
+
+  const handleFileSelect = async (e: React.ChangeEvent<HTMLInputElement>) => {
+    const file = e.target.files?.[0];
+    if (!file) return;
+
+    try {
+      setIsUploading(true);
+      // We set an explicit path to make sure that when users change
+      // their avatar we upload to the same path. This way we keep
+      // the same URL for the avatar.
+      //
+      // We set the profileId in the path for permission checks. This
+      // way we can write a rule to ensure that only the user can
+      // upload to their own profile.
+      const path = `avatars/${profileId}/avatar`;
+
+      const { data } = await db.storage.uploadFile(path, file);
+
+      // Link the file to the profile
+      await db.transact(tx.profiles[profileId].link({ avatar: data.id }));
+    } catch (error) {
+      console.error('Error uploading avatar:', error);
+    } finally {
+      setIsUploading(false);
+    }
+  };
+
+  return (
+    <div className="flex flex-col items-center gap-4">
+      <div className="relative">
+        <div style={{ width: defaultSize, height: defaultSize }}>
+          {avatar ? (
+            <img src={avatar.url} className="w-full h-full object-cover" />
+          ) : (
+            <div className="w-full h-full bg-gray-400" />
+          )}
+        </div>
+
+        {isUploading && (
+          <div className="absolute inset-0 bg-black bg-opacity-40 rounded-full flex items-center justify-center">
+            <div className="w-8 h-8 border-4 border-white border-t-transparent rounded-full animate-spin" />
+          </div>
+        )}
+      </div>
+
+      <label className="cursor-pointer">
+        <input
+          type="file"
+          accept="image/*"
+          onChange={handleFileSelect}
+          className="hidden"
+        />
+        <span className="py-2 px-4 bg-blue-500 hover:bg-blue-600 text-white rounded text-sm transition-colors">
+          {avatar ? 'Change Avatar' : 'Upload Avatar'}
+        </span>
+      </label>
+    </div>
+  );
+}
+
+function ProfilePage() {
+  return (
+    <div className="box-border bg-gray-50 font-mono min-h-screen p-5 flex items-center flex-col">
+      <div className="tracking-wider text-3xl text-gray-700 mb-8">
+        Profile Settings
+      </div>
+
+      <div className="bg-white rounded-lg shadow-md p-8 max-w-2xl w-full">
+        <h2 className="text-xl mb-6 pb-2 border-b border-gray-200">
+          Profile Picture
+        </h2>
+
+        <div className="flex justify-center">
+          <AvatarUpload defaultSize={120} />
+        </div>
+      </div>
+      <button
+        className="text-sm text-gray-500 mt-2"
+        onClick={() => db.auth.signOut()}
+      >
+        {' '}
+        Sign out
+      </button>
+    </div>
+  );
+}
+
+function Wrapper() {
+  const { isLoading, user, error } = db.useAuth();
+  if (isLoading) {
+    return <div>Loading...</div>;
+  }
+  if (error) {
+    return <div>Uh oh! {error.message}</div>;
+  }
+  if (user) {
+    return <ProfilePage />;
+  }
+  return <Login auth={db.auth} />;
+}
+
+export default Wrapper;
--- a/client/www/pages/docs/storage.md
+++ b/client/www/pages/docs/storage.md
@@ -7,7 +7,8 @@ You can use Storage to store images, videos, documents, and any other file type.
 
 ## Storage quick start
 
-Here's a full example of how to upload and display a grid of images
+Let's use a fresh Next JS app to build a full example of how to upload and
+display a grid of images
 
 ```shell {% showCopy=true %}
 npx create-next-app instant-storage --tailwind --yes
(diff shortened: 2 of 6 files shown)


client/www/pages/docs/storage.md
```diff
@@ -5,131 +5,555 @@ title: Storage
 Instant Storage makes it simple to upload and serve files for your app.
 You can use Storage to store images, videos, documents, and any other file type.
 
-{% callout %}
+## Storage quick start
 
-Storage is still in **beta**, but you can request access [here](https://docs.google.com/forms/d/e/1FAIpQLSdzInffrNrsYaamtH_BUe917EOpcOq2k8RWcGM19XepJR6ivQ/viewform?usp=sf_link)!
+Here's a full example of how to upload and display a grid of images
```
stopachka: Nit: perhaps we add an intro sentence: 

"Let's start with a fresh NextJS app:" or something like this?

client/www/pages/docs/storage.md
```diff
@@ -5,131 +5,555 @@ title: Storage
...
+```javascript {% showCopy=true %}
+'use client';
+
+import { init } from '@instantdb/react';
+import schema from '../instant.schema';
+import React from 'react';
+
+// Types
+// ----------
+export type Image = {
```
stopachka: nit: if we are using `schema`, i.m.o we can just extract the type from it. 

For example: 

```
type InstantFile = InstaQLEntity<AppSchema, '$files'>

File: client/www/pages/docs/storage.md [hunk 3]
```diff
@@ -5,131 +5,555 @@ title: Storage
...
+async function deleteImage(image: Image) {
+  await db.storage.delete(image.path);
+}
+
+function App() {
+  // $files is the special namespace for querying storage data
+  const { isLoading, error, data } = db.useQuery({
+    $files: {
+      $: {
+        order: { serverCreatedAt: 'asc' },
```
stopachka, line 123: nit: is this order necessary? (My guess is that files would order this way naturally anyways?)

client/www/pages/docs/storage.md
```diff
@@ -5,131 +5,555 @@ title: Storage
...
+      },
+    },
+  });
+
+  if (isLoading) {
+    return <div>Fetching data...</div>;
```
stopachka: nit: I would return `null` for isLoading. 

Reasoning: Otherwise this will have a bit of a jarring flash when you first load the page

client/www/pages/docs/storage.md
```diff
@@ -5,131 +5,555 @@ title: Storage
...
+  }
+
+  if (error) {
+    return <div>Error fetching data: {error.message}</div>;
+  }
+
+  // The result of a $files query will contain objects with
+  // metadata and a download URL you can use for serving files!
+  const { $files: images } = data as { $files: Image[] };
```
stopachka: You shouldn't need to write this if you use schema.

client/www/pages/docs/storage.md
```diff
@@ -5,131 +5,555 @@ title: Storage
...
+      nickname: i.string(),
+      createdAt: i.date(),
+    }),
+  },
+  links: {
+    profileUser: {
+      forward: { on: 'profiles', has: 'one', label: '$user' },
+      reverse: { on: '$users', has: 'one', label: 'profile' },
+    },
+    profileUploads: {
```
stopachka, line 300: Maybe a better example would be `profilePhoto`?

client/www/pages/docs/storage.md
```diff
@@ -5,131 +5,555 @@ title: Storage
...
+              where: { '$user.id': user.id },
+            },
+            avatar: {},
+          },
+        }
+      : null,
+  );
+  const [isUploading, setIsUploading] = useState(false);
+
+  if (isLoadingAvatar) return <div>Loading...</div>;
```
stopachka: nit: would return null

client/www/pages/docs/storage.md
```diff
@@ -5,131 +5,555 @@ title: Storage
...
+      >
+        {' '}
+        Sign out
+      </button>
+    </div>
+  );
+}
 
-  return <img src={imageUrl} />;
+function Wrapper() {
```
stopachka: nit: would call this `App` or something

client/www/pages/docs/storage.md
```diff
@@ -5,131 +5,555 @@ title: Storage
...
+  const { isLoading, user, error } = db.useAuth();
```
stopachka: nit: if we already check here for the user, perhaps we just pass in `user` to `ProfilePage`? This way we don't need to do the user ? query : null check there.

stopachka (approved):
Some minor comments, but LGTM! Love the detail

client/www/pages/docs/storage.md
nezaj: It wasn't neccessary but I thought it would be nice to showcase how you can use it like a regular namespace

client/www/pages/docs/storage.md
nezaj: Yea was debating this but thought it might be nice to show off 1) how you don't need to drill props and 2) demonstrate the defer pattern (so folks know about it more readily)

client/sandbox/react-nextjs/pages/play/avatar.tsx
```diff
@@ -0,0 +1,140 @@
...
+        onClick={() => db.auth.signOut()}
+      >
+        {' '}
+        Sign out
+      </button>
+    </div>
+  );
+}
+
+function Wrapper() {
```
stopachka: nit: would write this as `export default function Page` or something

client/www/pages/docs/storage.md
```diff
@@ -5,287 +5,743 @@ title: Storage
...
-The `pathname` determines where the file will be stored, and can be used with permissions to restrict access to certain files.
+Initialize your schema and permissions via the [cli tool](/docs/cli)
 
-The `file` should be a [`File`](https://developer.mozilla.org/en-US/docs/Web/API/File) type, which will likely come from a [file-type input](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file).
+```
+npx instant-cli@latest init
+```
 
-{% callout type="warning" %}
+Now open `instant.perms.ts` and add the following permissions
```
stopachka: nit: you mean `instant.schema.ts`?

client/www/pages/docs/storage.md
```diff
@@ -5,287 +5,743 @@ title: Storage
...
+```javascript {% showCopy=true %}
+import type { InstantRules } from "@instantdb/react";
+import { i } from "@instantdb/react";
+
+const _schema = i.schema({
+  entities: {
+    $files: i.entity({
+      "content-disposition": i.string().indexed(),
+      "content-type": i.string().indexed(),
+      "key-version": i.number(),
```
stopachka: Are these needed?

client/www/pages/docs/storage.md
```diff
@@ -5,287 +5,743 @@ title: Storage
...
-The `file` should be a [`File`](https://developer.mozilla.org/en-US/docs/Web/API/File) type, which will likely come from a [file-type input](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file).
+```
+npx instant-cli@latest init
+```
 
-{% callout type="warning" %}
+Now open `instant.perms.ts` and add the following permissions
+
+```javascript {% showCopy=true %}
+import type { InstantRules } from "@instantdb/react";
```
stopachka: InstantRules are not needed here

client/www/pages/docs/storage.md
```diff
@@ -5,287 +5,743 @@ title: Storage
 Instant Storage makes it simple to upload and serve files for your app.
 You can use Storage to store images, videos, documents, and any other file type.
```
stopachka:
```suggestion
You can store images, videos, documents, and any other file type.
```

client/www/pages/docs/storage.md
```diff
@@ -5,287 +5,743 @@ title: Storage
...
 
-{% callout %}
+## Storage quick start
 
-Storage is still in **beta**, but you can request access [here](https://docs.google.com/forms/d/e/1FAIpQLSdzInffrNrsYaamtH_BUe917EOpcOq2k8RWcGM19XepJR6ivQ/viewform?usp=sf_link)!
+Let's use a fresh Next JS app to build a full example of how to upload and
```
stopachka:
```suggestion
Let's build a full example of how to upload and
```

client/www/pages/docs/storage.md
```diff
@@ -5,287 +5,743 @@ title: Storage
...
+display a grid of images
```
stopachka:
```suggestion
display a grid of images. We'll start with a fresh NextJS app:
```

client/www/pages/docs/storage.md
```diff
@@ -5,287 +5,743 @@ title: Storage
...
+const db = init({ appId: "REPLACE ME", schema });
+
+// `uploadFile` is what we use to do the actual upload!
+// the `$files` will automatically update once the upload is complete
+async function uploadImage(file: File) {
+  try {
+    // Optional metadata you can set for uploads
+    const opts = {
+      // See: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type
+      // Default: 'application/octet-stream'
```
stopachka, line 100: nit: perhaps it's unnecessary to link out to moz? It feels a bit cluttered when looking at the doc

client/www/pages/docs/storage.md
```diff
@@ -5,287 +5,743 @@ title: Storage
...
+      contentType: file.type,
+      // See: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition
+      // Default: 'inline'
+      contentDisposition: 'attachment; filename="moop.jpg"',
```
stopachka: nit: did we mean the filename to be moop.jpg?

client/www/pages/docs/storage.md
```diff
@@ -5,287 +5,743 @@ title: Storage
...
+      })}
     </div>
   );
 }
+
+export default App;
 ```
 
-Then, in your `ImageViewer` component, you can use the `cachedUrl` by default, and handle the expiration when necessary:
+Make sure to update this line with the app id in your `.env` file.
```
stopachka: If they have a .env file, perhaps it makes sense to simply use `process.env.NEXT_PUBLIC_INSTANT_APP_ID` in the example?

client/www/pages/docs/storage.md
```diff
@@ -5,287 +5,743 @@ title: Storage
...
+// or, give the file a custom name
+const path = `${user.id}/avatar.png`;
+await db.storage.uploadFile(path, file);
 
-At the moment, the only available file metadata is `data.path`, which represents the file's path in Storage. (In the future we will likely include metadata such as `size` and `type`.)
+// or, set the content type and content disposition
+const path = `${user.id}/orders/${orderId}.pdf`;
+await db.storage.uploadFile(path, file, {
+  contentType: 'application/pdf',
+  contentDisposition: 'attachment; filename="confirmation.pdf"',
```
stopachka: nit: the `filename` should be dynamic no?

client/www/pages/docs/storage.md
```diff
@@ -5,287 +5,743 @@ title: Storage
...
+      reverse: { on: '$files', has: 'one', label: 'profile' },
     },
-    "bind": [
-      "isImage",
-      "data.path.endsWith('.png') || data.path.endsWith('.jpeg')"
-    ]
+  },
+});
+
+
```
stopachka: nit: perhaps these should be two separate code blocks?

client/www/pages/docs/storage.md
```diff
@@ -5,287 +5,743 @@ title: Storage
...
+        label: 'profile',
+      },
+    },
+  },
+  rooms: {},
+});
+```
 
-The Admin SDK offers the same API for managing storage on the server, plus a few extra convenience methods for scripting.
+Here's a more detailed example showing how you may implement an avatar upload feature:
```
stopachka: This feels a bit jarring -- it's a pretty big example, but feels like it's said nonchalantly. Perhaps we can give this a full subheader? We can show the example, and link out to the full repo? I am not sure but just felt a bit off scrolling here. At the least it's weird that with such a large codeblock there is no copy button

stopachka (commented):
Great work!

client/www/pages/docs/storage.md
nezaj: Nope! Will remove!

client/www/pages/docs/storage.md
nezaj: Was thinking it may be nice to give people a sense? I'm open to removing

client/www/pages/docs/storage.md
nezaj: I did want to show an example of setting a custom filename but perhaps we should be more descriptive!

client/www/pages/docs/storage.md
nezaj: I was actually intending not but we could do that!

client/www/pages/docs/storage.md
nezaj: Sure!

PR #904, opened 2025-02-19
Parse content-type and content-disposition in admin sdk

nezaj:
Clark gave `uploadFile` from the admin SDK a whirl and noticed we weren't parsing content-type correctly. I remember previously I keywordized the params but then got rid of that. Needed to also update the parsing.

Doing that here!


stopachka (approved):
LGTM!

PR #906, opened 2025-02-20
Enable storage for all

nezaj:
PR for enabling storage for all! This PR kills the storage tab and adds a `storage-block-list` we can manage for turning off storage as needed per app


stopachka (approved):
Nice!

PR #911, opened 2025-02-20
[monaco] improve height resize, loading state

stopachka:
When we had an error, we would show an error pane. But our Monaco editor would not resize. I fixed this by adding a few `min-h-0`s.

I also updated our Loading state, so it's a bit less jarring. It used to flash "Loading...". Now it's white background, and does a pulse animation. In most cases users won't see the pulse.

@nezaj @tonsky @dwwoelfel


dwwoelfel (approved):
LGTM!

PR #912, opened 2025-02-20
Simplify grouped queue 2

tonsky:
Looks like I figured out why grouped-queue refactor didn’t work. Executor was only spawning 1 thread

Fixed by following this answer https://stackoverflow.com/a/8125505

See #835 #884


server/test/instant/grouped_queue_test.clj
```diff
@@ -70,6 +70,28 @@
           (testing group
             (is (= (count ids) processed))))))))
 
+(deftest thread-pool-size-test
+  (let [input (for [group [1 2 3 4 5]
+                    id    (range 100)]
+                {:group group :id id})
+        latch (CountDownLatch. 500)
```
stopachka, line 77: nice test!

stopachka (approved):
I am in!

PR #914, opened 2025-02-20
Lint tests & filter sources to pre-commit hook

tonsky:
- Tests code is now linted too (and it even caught a bug!)
- pre-commit hook now correctly only checks sources from src/ and test/ and applies .clj-kondo config


stopachka (approved):
Niice!

PR #918, opened 2025-02-21
Fix nested files

nezaj:
Clark noted a bug that once we were at least three levels deep we wouldn't return `$files` urls. After some sleuthing with @stopachka we saw we were losing some metadata in our child forms. Inspecting `child-patterns` revealed our missing data.


stopachka (approved):
Nice!

PR #919, opened 2025-02-21
[docs] include content in our indexer

stopachka:
Previously, we only indexed h2 tags for our docs. This was because if we indexed all of the markdown, we'd have a lot of noisy text (like code blocks). 

I updated our index script so: 

1. Instead of manipulating text, we parse the markdown into an AST
2. Then we only count text that are in headings, paragraphs, or lists
3. I changed how we wrote our hierarchy in algolia objects, so that for all searches, we get content highlights
4. I also added a `--dry-run` flag, so we can take a look at how the algolia objects look in dev. 

**To check out how the index works:**

**Load https://instant-www-js-roar1-jsv.vercel.app/docs** and make some searches 

@nezaj @dwwoelfel @tonsky


client/www/components/docs/Search.jsx
```diff
@@ -7,7 +7,8 @@ import '@docsearch/css';
 const docSearchConfig = {
   appId: '98PPX6H1AS',
   apiKey: 'ee52f4bc250c519ea97596da07560d82', // search only API key
-  indexName: 'docs',
+  // TODO: replace before landing
+  indexName: 'docs_dev',
```
nezaj: Dope. Maybe we can make this config so that in local it uses dev

nezaj (approved):
Nice job!

PR #920, opened 2025-02-21
[Storage] Enable json uploads

nezaj:
We got a report that json uploads were blocked and indeed it was because of our middleware which would parse the upload as json instead of an input stream.

This PR updates our middleware to conditionally `wrap-json-body`


stopachka (approved):
SGTM!

PR #922, opened 2025-02-21
[index-docs] use typescript

stopachka:
I wanted to upgrade our script to use typescript. TS is especially useful in scripts. We rarely run these scripts, so some extra types to help gain context can be great. 

@nezaj @tonsky @dwwoelfel


nezaj (approved):
Sounds good!

PR #923, opened 2025-02-21
[Docs] Add lookups in link section

nezaj:
Got some feedback using lookups in link entity ids was hard to parse, making this more explicit!


dwwoelfel (approved):
LGTM!

PR #924, opened 2025-02-21
[Docs] Add pattern for finding entities without links

nezaj:
We got some feedback it wasn't clear how to query for entities without links. I've seen this questions asked quite a bit so figured it would worth adding it as a pattern


dwwoelfel (approved):
LGTM!

PR #925, opened 2025-02-21
Make zeneca normal

dwwoelfel:
Zeneca in the tests does a weird thing where it makes ids into refs. No other app behaves like this and it's causing problems for the select-keys PR https://github.com/instantdb/instant/pull/831

This updates the zeneca fixtures to create ids like a normal app and fixes all of the tests.


stopachka (approved):
Thank you @dwwoelfel !!

PR #926, opened 2025-02-21
[Docs] Be explicit that transact is atomic

nezaj:
Got feedback that it wasn't clear that transacts are atomic. I realized we don't mention this in the docs and thought it would be nice to be more explicit.


dwwoelfel (approved):
LGTM!

PR #929, opened 2025-02-21
[CEL] Parse false values correctly

nezaj:
We've been getting error reports when folks do things like `!data.isFavorite`

Turns out we weren't parsing `false` values correctly and they would get coerced into null values. And doing a logical not on null throws an error.

Now we check whether our cel-map contains the key and if it does we extract the value from the map. Adding in a test too to verify this behavior

### DIFF
--- a/server/src/instant/db/cel.clj
+++ b/server/src/instant/db/cel.clj
@@ -154,6 +154,12 @@
                          (.format ^Date x))
     :else              x))
 
+(defn get-cel-value [m k]
+  (stringify
+   (if (contains? m k)
+     (get m k)
+     (get m (keyword k)))))
+
 (deftype CelList [xs]
   java.util.List
   (get [_ i]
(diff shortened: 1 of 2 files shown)


stopachka (approved):
LGTM!

server/src/instant/db/cel.clj
```diff
@@ -170,7 +170,9 @@
   java.util.Map
   (get [_ k]
     (stringify
-     (or (get m k) (get m (keyword k)))))
+     (if (contains? m k)
+       (get m k)
+       (get m (keyword k)))))
```
stopachka: You know for `entrySet` below, we may want to do something like: 

Maybe we should have a function like `(get-cel-value m k)`

Then, for entrySet we can use it too: 

```clojure
(->> (keys (or m {})) 
         (map (fn [k] (get-cel-value m k))
         set) 
```

Reasoning: 

This way when you print the map in your repl, you see the same values you would have seen if you used `get`

PR #930, opened 2025-02-22
[Temp] Hotfix for user apps

nezaj:
Pushing up a quick fix for a user to unblock them for now. Seems like turning off materialize results in a 10x slow-down. For now just enabling it for their apps


nezaj:
Since this PR just forces`materialize` for these specific apps will merge in. We can follow-up later to figure a more general solution

PR #933, opened 2025-02-24
[sm][docs] order app picker by created at

stopachka:
I realized that our docs app picker did not list apps in the same order as our dashboard. To fix, I extracted out the `createdAtComparator`, and made sure they have the same order.

@dwwoelfel @nezaj @tonsky

### DIFF
--- a/client/www/pages/dash/index.tsx
+++ b/client/www/pages/dash/index.tsx
@@ -57,6 +57,7 @@ import { useForm } from '@/lib/hooks/useForm';
 import useLocalStorage from '@/lib/hooks/useLocalStorage';
 import { useDashFetch } from '@/lib/hooks/useDashFetch';
 import { asClientOnlyPage, useReadyRouter } from '@/components/clientOnlyPage';
+import { createdAtComparator } from '@/lib/app';
 
 // (XXX): we may want to expose this underlying type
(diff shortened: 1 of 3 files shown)


client/www/pages/dash/index.tsx
```diff
@@ -270,11 +271,8 @@ function Dashboard() {
     });
   }, [token]);
 
-  const apps = useMemo(() => {
```
stopachka, line 273: I read through the code here, and I think the useMemo in this case is not necessary. 

1. This just keeps the identity of the `apps` array stable, but imo nothing depends on the list identity being stable

dwwoelfel: Is it not there to reduce the amount of re-rending that happens every time swr refetches?

dwwoelfel (approved):
LGTM!

client/www/pages/dash/index.tsx
stopachka: Afaik I don't think this will save us much. Here's my reasoning: 

1. This memo was saying: if _anything_ in  `data.apps` changes, then re-calculate `apps`
2. `swr` returns the states `data`, `error`, `isValidating`, `isLoading`. 
3. But if I understand correctly,  most of the time `data` is the thing that is going to be changing

PR #934, opened 2025-02-24
Only use `not materialized` when we need it in an ordered query

dwwoelfel:
Looked for patterns we could use to determine when to use `not-materialized` and when to use `materialized`, but couldn't come up with anything definitive.

With this PR, we'll only use `not materialized` when we need it for ordering.


stopachka (approved):
Nice!!

PR #936, opened 2025-02-24
Materialize the pagination cte that we include in the result

dwwoelfel:
We know we'll include the final pagination CTE in the json_build_object result from datalog, so we can materialize it ahead of time.


nezaj (approved):
🔥

PR #943, opened 2025-02-25
Fix import script when creator does not exist

dwwoelfel:
The script would previously fail with a foreign key violation if the `creator_id` didn't already exist locally. Now we insert into a temp table so that we can filter out the creator_id.


stopachka (approved):
🫡🫡🫡🫡

PR #944, opened 2025-02-26
Use a left-join for ors when possible

dwwoelfel:
When we're joining to a previous CTE in the or gather, we can use a left join instead of a full join, which reduces the number of rows that postgres has to filter.

There are a couple of other small improvements:

1. We check if offset is 0 instead of just if it's nil to allow us to skip the has-prev query
2. We materialize any CTE that shows up in the `json_build_object`. This prevents postgres from running some queries twice.
3. If we see multiple ors on the same field, we'll combine them into an in `{:or [{field: value}, {field: value}]}` -> `{:or [{field: {in: [value, value]}]}`. This had huge improvement for one of the queries in production.

For the backtesting, I did a distinct by query_hash, only compared queries whose sql query changed, and got the time from `explain` instead of running the query. I think that made the backtesting much more consistent:

| :old-ms | :new-ms | :improvement |
|---------|---------|--------------|
|     908 |     922 |          -14 |
|      83 |      87 |           -3 |
|      73 |      76 |           -3 |
|       1 |       4 |           -3 |
|       4 |       5 |            0 |
|       2 |       3 |            0 |
|       5 |       5 |            0 |
|      22 |      22 |            0 |
|       2 |       1 |            0 |
|       3 |       2 |            1 |
|      29 |      27 |            1 |
|     327 |     321 |            5 |
|      24 |      16 |            8 |
|     332 |     317 |           15 |
|      43 |      26 |           17 |
|      91 |      63 |           28 |
|      82 |      47 |           35 |
|     327 |     236 |           91 |
|    4641 |     889 |         3751 |
|    4661 |     877 |         3783 |
|    9415 |    1435 |         7979 |

### DIFF
--- a/server/src/instant/db/datalog.clj
+++ b/server/src/instant/db/datalog.clj
@@ -963,15 +963,25 @@
                 ;; Include the previous cte if we're not the first
                 [prev-table])
               [(kw prefix (first join-idxes))])}
-      (when full-join-idxes
-        {:full-join (mapcat (fn [i]
-                              ;; Ensures everything is included
-                              [(kw prefix i) [:= :0 :1]])
-                            full-join-idxes)})
-      (when-not (neg? prev-idx)
+      (if-not prev-table
+        (merge {:from (kw prefix (first join-idxes))}
+               (when full-join-idxes
+                 {:full-join (mapcat (fn [i]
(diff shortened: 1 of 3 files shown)


stopachka (approved):
Biyutiful!

PR #945, opened 2025-02-26
Make explorer deep-linkable

nezaj:
This PR enables browser navigation for different namespaces, filters, and sorts for the explorer.

https://github.com/user-attachments/assets/42b8f989-a984-4e34-8e14-8a40f8a16c82

### DIFF
--- a/client/www/components/dash/explorer/Explorer.tsx
+++ b/client/www/components/dash/explorer/Explorer.tsx
@@ -519,6 +519,11 @@ export function Explorer({
   }
 
   function pushNavStack(_nav: ExplorerNav) {
+    const currentNamespace = navStack[navStack.length - 1]?.namespace;
+    if (currentNamespace !== _nav.namespace) {
+      setSearchFilters([]);
+    }
+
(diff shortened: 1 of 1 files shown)


client/www/components/dash/explorer/Explorer.tsx
```diff
@@ -418,35 +451,88 @@ export function Explorer({
...
+    }
+
+    // Add search filters
+    if (current.filters && current.filters.length > 0) {
+      queryParams.search = filtersToQueryString(current.filters);
+    } else {
+      delete queryParams.search;
+    }
+
+    // Add sort
```
stopachka, line 483: Perhaps we can have nav take an additional argument -- "replace or not" or something. This way for some places, were we only change the sort, we don't make that a part of browser history

stopachka (approved):
Nice!

client/www/components/dash/explorer/Explorer.tsx
nezaj: Done, made it so sort doesn't push browser history either

PR #946, opened 2025-02-26
[Explorer] Deep link pagination

nezaj:
Realized it would be nice to deep-link pagination too in the explorer so adding that here!


stopachka (approved):
LGTM!

PR #947, opened 2025-02-27
Add flags to set query settings for slow queries

dwwoelfel:
Occasionally we have a query that's slow, but we could fix it with a postgres setting `set enable_nestloop = off`.

This adds a facility where we can set those config settings for individual queries by their query hash.

To set it, you just have to add a new row in `instant-config` with the query_hash, the setting, and the value.

There's no validation on the config to ensure they're valid settings. We complain loudly if the db throws when we apply the setting, but we won't break the query.

Inspired by a query that takes 8 seconds with a nested loop, but takes 80ms with `set enablenestloop = off`.

### DIFF
--- a/server/src/instant/jdbc/sql.clj
+++ b/server/src/instant/jdbc/sql.clj
@@ -130,7 +130,7 @@
          ~@body))))
 
 (defn- postgres-config-span-attrs [postgres-config]
-  (reduce(fn [acc {:keys [setting value]}]
+  (reduce (fn [acc {:keys [setting value]}]
            (assoc acc (str "postgres-config." setting) value))
          {}
          postgres-config))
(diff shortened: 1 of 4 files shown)


server/src/instant/jdbc/sql.clj
```diff
@@ -126,12 +129,19 @@
        (with-open [~conn-name (.getConnection ~conn-pool)]
          ~@body))))
 
-(defn- span-attrs [conn query tag]
+(defn- postgres-config-span-attrs [postgres-config]
+  (reduce(fn [acc {:keys [setting value]}]
```
nezaj: Nit. Need to add a space after `reduce`

nezaj (approved):
Aside from one nit LGTM!

server/src/instant/jdbc/sql.clj
dwwoelfel:
```suggestion
  (reduce (fn [acc {:keys [setting value]}]
```

stopachka (commented):
Awesome!!

PR #948, opened 2025-02-27
[cli] do not infer system catalog attrs, add checked data types to system catalog attrs

stopachka:
I noticed we would write: 
'We inferred 1 attribute' 

This was because $files.url was not typed. 

This PR: 

1. Adds the checked data type to the system_catalog.clj file
2. Makes sure we ignore system catalog attrs in cli inference

@nezaj @tonsky @dwwoelfel 

Once landed, I will run the following sql on prod, to update the files system catalog to have a checked data type :

```sql
UPDATE attrs 
SET checked_data_type = 'string'
WHERE attrs.id = '96653230-13ff-ffff-2a35-48afffffffff'::uuid;

UPDATE triples
SET checked_data_type = 'string'
WHERE triples.attr_id = '96653230-13ff-ffff-2a35-48afffffffff'::uuid;
```

### DIFF
--- a/client/packages/cli/src/index.js
+++ b/client/packages/cli/src/index.js
@@ -1677,8 +1677,9 @@ export default rules;
   `.trim();
 }
 
-function inferredType(config) {
-  const inferredList = config['inferred-types'];
+function inferredType(attr) {
+  if (attr.catalog === 'system') return null;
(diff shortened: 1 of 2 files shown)


nezaj (approved):
LGTM

PR #949, opened 2025-02-27
Disable the av index test because it's flaky

dwwoelfel:
Still trying to figure out how we can make it consistent


stopachka (approved):
_cheers!_

PR #950, opened 2025-02-27
Support params in queries and transactions

tonsky:
`ruleParams` let you write permission rules of type  “allow access to anyone with a link”. E.g.

```
{
  docs: {
    view: "ruleParams.secret in data.ref('docLinks.secret')"
  },
  comments: {
    view: "ruleParams.secret in data.ref('parent.docLinks.secret')"
  }
}
```

Use in queries:

```
db.useQuery({docs: {}}, { secret: secret })
```

and transactions:

```
db.transact(db.tx.docs[id].ruleParams({ secret: secret }).update({ title: 'eat' }));
```

Updated APIs:

- [x] `db.queryOnce(q, opts?)`
- [x] `db.subscribeQuery(q, cb, opts?)`
- [x] `db.useQuery(q, opts?)`
- [x] `db.tx.ns[id].ruleParams(...).upd...`

where opts are `{ ruleParams: {...}}`


stopachka:
(One note: afaik we'll want to support this for admin sdk too)

client/packages/core/src/index.ts
```diff
@@ -489,11 +490,11 @@ class InstantCoreDatabase<Schema extends InstantSchemaDef<any, any, any>>
    *    console.log(resp.data.goals)
    *  });
    */
-  subscribeQuery<Q extends InstaQLParams<Schema>>(
-    query: Q,
-    cb: (resp: InstaQLSubscriptionState<Schema, Q>) => void,
-  ) {
-    return this._reactor.subscribeQuery(query, cb);
+  subscribeQuery<Q extends InstaQLParams<Schema>>(query: Q, cb: (resp: InstaQLSubscriptionState<Schema, Q>) => void);
```
stopachka: Nit: I would suggest putting opts as the third arg, mainly to avoid overloads. 

Afaik overloads don't show up well in intellisense. For example, if you hover over the function, it will now choose the first overload, and then say (+ 1 overload), without explaining the other overloads. 

(Learned about this when looking into using overloads for schema here : https://stackoverflow.com/questions/79187112/what-is-the-idiomatic-way-to-make-a-generic-type-optional#comment139640779_79189035)

tonsky: Yeah. My concern was that callback works best when it’s the last argument. But I agree it’s better when arguments don’t change their position in overloads

server/src/instant/util/s3.clj
```diff
@@ -180,7 +180,7 @@
   (delay
     (let [access-key (config/s3-storage-access-key)
           secret-key (config/s3-storage-secret-key)
-          region (.toString (.region (.serviceClientConfiguration @default-s3-client*)))]
+          region (.toString (.region (.serviceClientConfiguration (default-s3-client))))]
```
stopachka, line 183: Thank you!

server/src/instant/db/permissioned_transaction.clj
```diff
@@ -418,10 +340,27 @@
   [{:keys [attrs]} attr-changes]
   (into attrs (get-new-attrs attr-changes)))
 
-(defn get-check-commands [ctx attr-changes preloaded-triples]
-  (let [attr-checks (attr-checks ctx attr-changes)
-        object-checks (object-checks ctx preloaded-triples)]
-    (into attr-checks object-checks)))
+(defn attr-checks [ctx attr-changes]
```
stopachka, line 343: nice!

client/www/pages/docs/permissions.md
```diff
@@ -291,3 +291,85 @@ delete to only succeed on todos associated with a specific user email.
 ```
 
 See [managing users](/docs/users) to learn more about that.
+
+### ruleParams
```
stopachka, line 295: Took a quick stab at editing here: 

https://gist.github.com/stopachka/3d902de5abdb558e9bc01a381a3b83e5

Main changes: 
1. Added syntax highlighting. 
2. Changed the structure of some sentences.

stopachka (approved):
Great work! Love the cleanup along the way

client/packages/admin/src/index.ts
```diff
@@ -367,11 +370,15 @@ class InstantAdmin<
...
   debugQuery = async <Q extends Query>(
     query: Exactly<Query, Q>,
-    opts?: { rules: any },
+    opts?: { rules?: any; ruleParams?: { [key: string]: any } },
   ): Promise<{
     result: QueryResponse<Q, Schema, WithCardinalityInference>;
     checkResults: DebugCheckResult[];
   }> => {
+    if (query && opts && 'ruleParams' in opts) {
+      query = { $$ruleParams: opts['ruleParams'], ...query };
```
stopachka, line 379: I wonder if we should push $$ruleParams into `$`. This way, the backend only needs to worry about ruleParams in there. 

It _is_ weird, in that `$` can defined in child queries, and `$$ruleParams` only makes sense at the top level, but since we are the ones inserting the option, perhaps not a big deal. 

Am open to either way, but raising in case I am missing something

tonsky: The idea here was that query can have multiple namespaces (can it?) like `{docs: {}, comments: {}}`, but we only define `ruleParams` once for the whole query. But if we only allow single top-level ns, then for sure we should move it inside $

stopachka: Ah, good point! Forgot about multiple namespaces case

PR #951, opened 2025-02-27
[Docs] Standardize db.tx

nezaj:
Doing `db.tx` is the recommended way now so updating our docs and doc-strings to consistently reflect that!


dwwoelfel (approved):
LGTM!

stopachka (commented):
Nice!

PR #952, opened 2025-02-27
Don't rely on existing data in the db for instaql tests

dwwoelfel:
Follow-up to https://github.com/instantdb/instant/pull/925

A lot of the instaql tests rely on the zeneca app that we insert during oss bootstrap. This PR uses `with-zeneca-app` for all of them.

Also includes a helpful `(update-snapshot #'my-test-var)`, which will replace `()` in `(is-pretty-eq? expected '())` with whatever `expected` returns. Very useful when you're writing new tests or updating existing tests. It acts kind of like snapshot testing. The idea was inspired by @tonsky's https://tonsky.me/blog/clojure-macros/

### DIFF
--- a/server/test/instant/db/instaql_test.clj
+++ b/server/test/instant/db/instaql_test.clj
@@ -101,7 +101,6 @@
   (let [pad (apply str (repeat offset " "))]
     (->> s
          string/split-lines
-         ;; XXX: Need to skip first line?
          (map (fn [s] (str pad s)))
          (string/join "\n"))))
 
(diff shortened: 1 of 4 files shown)


server/test/instant/db/instaql_test.clj
```diff
@@ -60,6 +62,114 @@
...
+         (-> (ex-data e)
+             ::ex/hint
+             :errors
+             first))))
+
+(defn pad-block [offset s]
+  (let [pad (apply str (repeat offset " "))]
+    (->> s
+         string/split-lines
+         ;; XXX: Need to skip first line?
```
stopachka: nit: is this comment intended?

stopachka (approved):
Super cool!

server/test/instant/db/instaql_test.clj
dwwoelfel: Turns out I didn't need to skip! Will remove.

PR #953, opened 2025-02-27
[cli] simplify schema comments

stopachka:
When I was playing with instantgram, found myself put off by the amount of comments we print out when we generate the code. 

I thought it would be a bit cleaner to just have a `// Docs: ` link at the top, similar to what we do in instant.perms.ts 

Here's how the current output looks for a fresh app:

CleanShot 2025-02-27 at 14 08 53@2x

@nezaj @tonsky @dwwoelfel


dwwoelfel (approved):
LGTM!

PR #954, opened 2025-02-27
Fix $entityIdStartsWith

dwwoelfel:
The `or` batching from https://github.com/instantdb/instant/pull/944 broke it


stopachka (approved):
Babaam!

PR #956, opened 2025-02-27
[cli] add onDelete: cascade annoation

stopachka:
I realize we didn't annotate when a user added onDelete cascade rules. 

Now we do: 

CleanShot 2025-02-27 at 14 56 24@2x

Note: this has the same issue that prod currently has, where if a user links to the same etype, the onDelete rule will look ambiguous. Going to tackle that in a separate PR 

@tonsky @dwwoelfel @nezaj

### DIFF
--- a/client/packages/cli/src/index.js
+++ b/client/packages/cli/src/index.js
@@ -1011,6 +1011,18 @@ async function waitForIndexingJobsToFinish(appId, data) {
   }
 }
 
+function linkOptsPretty(attr) {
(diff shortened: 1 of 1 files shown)


dwwoelfel (approved):
LGTM!

I wish there was a clearer way to write what deletes what, but I can't think of anything better than x cascades y.

stopachka:
Yeah I was noodling on other options too, but nothing too clear

PR #958, opened 2025-02-27
Only fetch blob attrs

dwwoelfel:
Easier to review with whitespace turned off: https://github.com/instantdb/instant/pull/958/files?diff=split&w=1

Stops fetching "ref" attrs when we build the CTE that gets all of the attrs for an entity. We know they won't be `:ea`, so it just adds extra work for postgres and the invalidator that we don't need to do.

Part of select-keys for queries: https://github.com/instantdb/instant/pull/831

I pulled this out on its own to make the select-keys PR easier to review and because it has some benefits of its own:

1. We include fewer attrs in the `ea` CTE, which makes postgres pick a better query plan.
2. We only pick attrs that are actually :ea
3. We stop invalidating on ref attrs that wouldn't affect the query


stopachka (approved):
Niice!

PR #961, opened 2025-02-28
Fail tests on unexpected io

dwwoelfel:
Fails tests if `warn-io` is triggered during the test. This would have prevented the bug fixed by https://github.com/instantdb/instant/pull/960

In order to make the test errors track the correct position in the file, I had to make all of the `with-X-app` functions into macros. That binds the `*file*` variable to the test file and we can search the stack trace when we encounter io to get the nearest line in the test file.


stopachka (approved):
Biyutiful!

PR #966, opened 2025-03-01
[presign] provide a custom presign url function

stopachka:
This implements our own `presign-s3-url` function. Crucially, this function takes in a `signing-instant`. By let us set the `signing-instant`, we can start to generate stable urls. This will be huge for for helping browsers cache these urls. 

I don't actually use a custom `signing-instant` yet. I wanted to first get the underlying system up. 

@dwwoelfel @tonsky @nezaj


server/src/instant/util/s3.clj
```diff
@@ -199,39 +177,49 @@
...
           (mapcat #(delete-objects bucket-name (vec %)))))))
 
+(def signer-creds*
+  (delay
+    (let [access-key (config/s3-storage-access-key)
+          secret-key (config/s3-storage-secret-key)]
+      (if (and access-key secret-key)
+        {:access-key access-key
+         :secret-key secret-key
+         :region "us-east-1"}
```
stopachka: I am not 100% sure this is the best way. I am assuming the previous S3 client somehow figured out regions for us -- I am not sure how though. For now I could add the region as a config value too, if we like that approach

server/src/instant/util/aws_signature.clj
```diff
@@ -0,0 +1,270 @@
...
+  [zone]
+  (str (get zone->endpoints zone) ".amazonaws.com"))
+
+(defn s3-host [region bucket]
+  (str bucket "." (zone->host region)))
+
+(defn s3-presigned-url
+  [{:keys [region access-key secret-key method bucket ^String path
+           signing-instant ^Duration expires-duration]
+    :or {signing-instant (Instant/now)}}]
```
stopachka: the key idea is this ^. We can set it specifically in a follow-on PR

server/src/instant/util/aws_signature.clj
```diff
@@ -0,0 +1,270 @@
...
+    (str
+     "AWS4-HMAC-SHA256\n"
+     timestamp "\n"
+     short-timestamp "/" region "/" service "/aws4_request" "\n"
+     (sha-256 (to-utf8 cannonical-request-str)))))
+
+(defn- amz-credential [{:keys [access-key short-date region service]}]
+  (str access-key "/" short-date "/" region "/" service "/aws4_request"))
+
+(defn generate-signature
```
stopachka: The aws-sign folks _do_ have a pretty serious test suite: 

https://github.com/joseferben/clj-aws-sign/tree/master/resources/suite

We could import those too, but was a bit conflicted, as it's a lot of files

server/src/instant/util/aws_signature.clj
```diff
@@ -0,0 +1,352 @@
...
+       (str/join "/")
+       (#(if (str/blank? %) "/" %))))
+
+(defn- append-slash
+  [path raw]
+  (str path (and (re-matches #".*/$" raw) "/")))
+
+(defn- replace-double-slash
+  [path]
+  (str/replace path #"//" "/"))
```
dwwoelfel: `(.replace path "//" "/")` would avoid a regex (but you probably need to add a `^String path` to prevent reflection.

server/src/instant/util/aws_signature.clj
```diff
@@ -0,0 +1,352 @@
...
+(defn- encode-path
+  [path]
+  (->> (str/split path #"/")
+       (map url-encode)
+       (str/join "/")
+       (#(if (str/blank? %) "/" %))))
+
+(defn- append-slash
+  [path raw]
+  (str path (and (re-matches #".*/$" raw) "/")))
```
dwwoelfel: A comment here would be helpful. It seems like it's only appending a slash if it already had a slash? If so, would it be more efficient to do

`(when (.endsWith raw "/") "/")`

server/src/instant/util/aws_signature.clj
```diff
@@ -0,0 +1,352 @@
...
+        (fn [s]
+          (or (str/blank? s)
+              (= "." s))))
+       (change-directory)
+       (str/join "/")
+       (str "/")))
+
+(defn- encode-path
+  [path]
+  (->> (str/split path #"/")
```
dwwoelfel: `(.split path "/")` would avoid a regex.

dwwoelfel (approved):
Couple of comments, but LGTM!

server/src/instant/util/aws_signature.clj
stopachka: nixed it!

PR #967, opened 2025-03-03
Add react native storage example

iansinnott:
Was looking at the docs for an example using storage with React Native and didn't find one. This is how I ended up doing it.


client/www/pages/docs/storage.md
```diff
@@ -474,6 +474,38 @@ const _schema = i.schema({
 [Check out this repo](https://github.com/jsventures/instant-storage-avatar-example)
 for a more detailed example showing how you may leverage links to implement an avatar upload feature
 
+## React Native
```
nezaj:
```suggestion
## Using Storage with React Native
```
Our search docs index based on header title so updating this to be more explicit that this is about using storage with react native

nezaj:
LGTM! Pushed up a lint fix here https://github.com/instantdb/instant/pull/983

PR #971, opened 2025-03-03
Only fetch ids for nested fields in explorer

dwwoelfel:
Updates the explorer to fetch ids instead of the full object for fetching links.

I had to update the link editor to fetch the object by its id when you click on the info icon to see the full object. 

Also fixes a bug where the search input for editing links wouldn't show up if the modal content overflowed.


nezaj (approved):
Nice!

PR #972, opened 2025-03-03
[Docs] Update split attributes pattern

nezaj:
Updating this pattern now that we have select fields. Also thought it might be nice to explicitly call out in select fields that users still need to be aware of permissions


dwwoelfel (approved):
LGTM!

PR #973, opened 2025-03-03
Use array param to pass attrs

dwwoelfel:
Passes a single array with all attr ids instead of a big list of ORs. Uses fewer parameters, sends less data over the wire, and is the slightest bit faster for honeysql to generate.

Did some backtesting and didn't actually see any improvements one way or the other. This should help prevent the query planner from reverting back to the created_at index, though.


nezaj (approved):
Nice!

PR #976, opened 2025-03-04
Fix the return type of query data when using fields

dwwoelfel:
Updates the return type to pick from the fields param (if the fields param is provided):

image

### DIFF
--- a/client/packages/core/src/queryTypes.ts
+++ b/client/packages/core/src/queryTypes.ts
@@ -110,6 +110,10 @@ type Remove$<T> = T extends object
   ? { [K in keyof T as Exclude<K, '$'>]: Remove$<T[K]> }
   : T;
 
+type Remove$NonRecursive<T> = T extends object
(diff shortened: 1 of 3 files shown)


dwwoelfel:
Looks like I still have a bug to fix in the types...

client/packages/core/src/queryTypes.ts
```diff
@@ -110,6 +110,10 @@ type Remove$<T> = T extends object
   ? { [K in keyof T as Exclude<K, '$'>]: Remove$<T[K]> }
   : T;
 
+type Remove$NonRecursive<T> = T extends object
```
dwwoelfel, line 113: Previously we removed the `$` recursively at the top-level `InstaQLResult` when we built up the result type. But I need the `$` when I'm figuring out which fields you selected. Now we remove it at each level as we process the query.

stopachka (approved):
Niice!

PR #977, opened 2025-03-04
[aws] stable presign urls

stopachka:
AWS URLs depend on the signing-instant.  
  
We want to keep URLs stable: repeated calls to the same object should return 
the same URL, so that browsers can cache the object. 
To do this, we bucket dates to the start of the day. 

This gives about 24 hours where URLs are stable

@nezaj @dwwoelfel @tonsky


dwwoelfel (approved):
I'm a little worried that we tell people the urls are good for 7 days, but then they could actually be half of that. Maybe we should set the signing instant to the start of the day? 

Otherwise, LGTM

PR #980, opened 2025-03-05
[presence] ensure we don't mutate `this._presence`

stopachka:
**Bug** 

If a user published:

```
publishPresence({ item: { n: 1 } })
publishPresence({ item: { n: 2 } })
```

The `patch-presence` for `n: 2` would not get propagated to `subscribePresence`. 

**Root cause** 

We received a `patch-presence` command. We would run: 

```
sessions = assocIn(sessions, path, value);
```

The path would look something like `['item', 'n']`

This would mutate the original `item`. 

Then, when `hasPresenceResponseChanged` would run, we would see that in fact nothing had changed, and skip the update. 

**This only happened when the user had a deeply nested presence.** 

**Fix** 

1. Updated it so that `assocIn` and `dissocIn` are explicit about being mutative
2. Then, used `mutative` to generate new `session` objects.

--- 

@dwwoelfel @nezaj @tonsky

### DIFF
--- a/client/packages/core/__tests__/src/instaql.test.js
+++ b/client/packages/core/__tests__/src/instaql.test.js
@@ -1203,22 +1203,6 @@ test('fields', () => {
     ],
   });
 
-  console.log(
-    JSON.stringify(
-      query(
-        { store },
-        {
-          users: {
-            $: { where: { handle: 'alex' }, fields: ['handle'] },
-            bookshelves: { $: { fields: ['name'] } },
-          },
-        },
-      ).data,
-      null,
-      2,
-    ),
-  );
-
   expect(
     query(
       { store },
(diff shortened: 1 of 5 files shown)


client/packages/core/__tests__/src/instaql.test.js
```diff
@@ -1203,22 +1203,6 @@ test('fields', () => {
     ],
   });
 
-  console.log(
```
stopachka, line 1206: (Noticed this when running and test and thought it wasn't intentional)

dwwoelfel, line 1206: whoops!

stopachka: 🤜🤛

dwwoelfel (approved):
Nice debugging!

tonsky:
Oops. Mutability is hard!

PR #981, opened 2025-03-05
[Explorer] Enable shift-click

nezaj:
Got a user ask for this and thought it was another nice win.

https://github.com/user-attachments/assets/247d3e28-7e9b-4e53-b9fe-d9c8c7467769

### DIFF
--- a/client/www/components/dash/explorer/Explorer.tsx
+++ b/client/www/components/dash/explorer/Explorer.tsx
@@ -782,16 +782,6 @@ export function Explorer({
     const allItemIds = allItems.map((i) => i.id as string);
     const currentIndex = allItemIds.indexOf(currentId);
     const lastSelectedIndex = allItemIds.indexOf(lastSelectedIdRef.current!);
-
-    // No-op if no previous selection or no current selection
-    if (
-      currentIndex === -1 ||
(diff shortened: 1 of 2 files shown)


client/www/components/dash/explorer/Explorer.tsx
```diff
@@ -777,6 +778,39 @@ export function Explorer({
     }
   };
 
+  const handleRangeSelection = (currentId: string, checked: boolean) => {
+    const allItemIds = allItems.map((i) => i.id as string);
```
stopachka, line 782: nit: perhaps we should turn this into a set?

client/www/components/ui.tsx
```diff
@@ -321,7 +321,10 @@ export function Checkbox({
   checked: boolean;
   className?: string;
   labelClassName?: string;
-  onChange: (checked: boolean) => void;
+  onChange: (
+    checked: boolean,
+    event?: React.ChangeEvent<HTMLInputElement>,
```
stopachka: nit: imo `event` is guaranteed, so we shouldn't have a `?`

stopachka (approved):
Doope!

PR #984, opened 2025-03-05
Convert cel to instaql (behind a feature flag)

dwwoelfel:
Converts cel to instaql where clauses.

The general idea is that we take a rule like `data.name == 'Daniel'` and instead of running the rule on every object from the database, we convert the rule into a where clause and apply it at the db-level. Then we only have to evaluate the rule once to get the where clause instead of once per object.

`data.name == 'Daniel'` becomes `{where: {name: 'Daniel'}}`

We use a custom instance of the cel compiler with overloads that let it return a `WhereClause` instead of a boolean.

The nice thing about this approach is that it works with expressions that can be turned into where clauses and expressions that can't. For example, if you have `auth.id != null && data.name == 'Daniel'`, we'll return `false` when the user isn't logged in and we know that we can just skip that branch of the query.

This PR enables the cel -> where clause conversion behind a feature flag. You can enable it for either an entire app or a query hash. It will still run the cel rules on every entity until we are comfortable that we're not leaking data (it logs if it leaks data).

There are still a few things to implement before it is 100%, but I'll roll those out separately:

1. Handle recursive rules.
  If you a rule on books `data.hidden != true` and a rule on bookshelves `auth.id in data.ref('books.owner.id')`. Then when you query bookshelves, you should get a query that looks like`{"books.owner.id": auth.id, "books.hidden" {:$not true}}`. Right now we only add the first-level `{"books.owner.id": auth.id}`
2. Handle short-circuit?
  If you have a rule like `auth.id != null`, then we don't have to run that query at all. It's a little tricky because we might be in a nested query, e.g. `books` in `{:bookshelves {:books {}}`. And then it's extra tricky with recursive rules. I think we might need a special operator for it.
3. Display rules in the sandbox
  I do send the rule-wheres in the response when you do a query check in the sandbox, but don't do anything with them in this PR yet
4. Feedback for the user about whether their rules can convert to where clauses.

Deploy plan:
1. Merge into main
2. Push the flag schema changes
3. Create the rule-wheres flag with empty arrays
4. Deploy the change


server/deps.edn
```diff
@@ -57,7 +57,7 @@
         io.grpc/grpc-netty-shaded {:mvn/version "1.68.1"}
         com.github.steffan-westcott/clj-otel-api {:mvn/version "0.2.4"}
 
-        dev.cel/cel {:mvn/version "0.5.2"}
+        dev.cel/cel {:mvn/version "0.9.1"}
```
dwwoelfel, line 60: I looked through the release notes and didn't see any changes that would affect our usage of CEL: https://github.com/google/cel-java/releases

server/src/instant/db/instaql.clj
```diff
@@ -1901,46 +1914,106 @@
...
+                              (cel/get-where-clauses ctx etype (:code program))
+                              ;; convert to where conds here in case this throws
+                              where-conds (rule-wheres->where-conds etype where-clauses)]
+                          (tracer/add-data! {:attributes {:wheres where-clauses
+                                                          :short-circuit? short-circuit?}})
+                          (assoc acc etype {:short-circuit? short-circuit?
+                                            :wheres where-clauses
+                                            :where-conds where-conds}))
+                        (catch Exception e
+                          (tracer/add-data! {:attributes {:exception (.getMessage e)}})
```
stopachka, line 1977: Curious q: is the reason we don't use `tracer/add-exception!` here, so we don't count this as an exception just yet?

server/src/instant/db/cel.clj
```diff
@@ -420,6 +1105,13 @@
                  patterns
                  results))))
 
+(def ^CelUnparser unparser (CelUnparserFactory/newUnparser))
+
+(defn unparse
```
stopachka, line 1110: cool!

server/src/instant/db/cel.clj
```diff
@@ -265,8 +337,621 @@
...
+     :cel-return-type whereclause-cel-type
+     :java-args [DataKey String]
+     :impl (fn [[^DataKey x ^String y]]
+             (let [attr (.instant_attr x)]
+               (if (and (= :string (:checked-data-type attr))
+                        (:index? attr))
+                 (WhereClause. {(.data_key x) {:$like (str "%" y "%s")}})
+                 (throw (ex-info "Invalid attr" {:x x})))))}]))
+
+;; Overloads for `type` to prevent `type(data.key)` from succeeding
```
stopachka, line 841: Realized, in the future I guess we could have these return _actual_ underlying types! (i.e if data.title is a string etc)

server/src/instant/db/cel.clj
```diff
@@ -265,8 +337,621 @@
...
+
+                                        :else
+                                        {:$not v})))
+                   {}
+                   c)))
+
+(def not-overloads
+  (global-overload
+   (.getFunction Operator/LOGICAL_NOT)
+   [{:overload-id "_not_datakey"
```
stopachka, line 784: so cool!

server/src/instant/db/cel.clj
```diff
@@ -265,8 +337,621 @@
...
+                         ;; elements, so null in data.ref('x') is always false
+                         (= NullValue/NULL_VALUE x)
+                         false
+
+                         :else (throw (ex-info "invalid inputs for in" {:x x :y y})))
+
+                   ;; Hard to know what to do here. Since we set `containsKey` to true
+                   ;; in CelMap, this will always return true for the rules.
+                   ;; If we didn't do that, then it would return true unless the key
+                   ;; was misisng from the map (even null returns true). We don't have
```
stopachka:
```suggestion
                   ;; was missing from the map (even null returns true). We don't have
```

server/src/instant/db/instaql.clj
stopachka, line 1977: One thing I am uncertain about here: 

- What happens if two programs have an error -- would one message override the others?

Second question is: 

- Is the reason we don't do something like `tracer/add-exception!`, because this isn't _quite_ a breaking exception? (if we fail, we simply move onto the old way of running perms. Do I understand correctly?

server/src/instant/db/cel.clj
```diff
@@ -265,8 +337,621 @@
...
+
+(defmethod and-overload-fn [:whereclause :bool]
+  ([_]
+   (fn [[^WhereClause x ^Boolean y]]
+     (and y x))))
+
+(defmethod and-overload-fn [:datakey :bool]
+  ([_]
+   (fn [[^DataKey x ^Boolean y]]
+     (and y x))))
```
stopachka, line 590: Neat! So this effectively "passes" through the DataKey, to the next thing, which may then convert it into a where clause. Mind is being blown 🤯

server/src/instant/db/cel.clj
```diff
@@ -265,8 +337,621 @@
...
+                          [:datakey :bool]
+                          [:bool :datakey]) SimpleType/DYN
+                         [:bool :bool] SimpleType/BOOL
+                         SimpleType/DYN)
+      :java-args (map type->java args)
+      :impl (get-or-overload-fn args)})))
+
+;; Overloads for `AND`
+;; We overload the existing AND function to handle our custom types
+;; cel won't let us replace the `AND` function unless we rewrite te ast
```
stopachka:
```suggestion
;; cel won't let us replace the `AND` function unless we rewrite the ast
```

server/src/instant/db/instaql.clj
dwwoelfel: > would one message override the others? 

It does a `record-info` for each error, which will create a new span.

> reason we don't do something like tracer/add-exception!

You have it right--I don't want to create a bunch of error alerts for rules that we can't handle. I just want to have a log so that we can identify the rules we're not able to handle.

server/src/instant/db/cel.clj
```diff
@@ -199,60 +219,112 @@
 
 (def ^ListType type-ref-return (ListType/create SimpleType/DYN))
 
-(def ref-fn
+(defn make-overload
```
stopachka, line 222: nice!

stopachka (approved):
_makes martial artist bow_ 

Such a hard problem so beautifully cut!

PR #985, opened 2025-03-05
[presence] handle arrays when doing `+` and `-` patches

stopachka:
Consider a presence state that uses arrays: 

```
{ selectedIds: []
```

We had two bugs:

**1. `+` operations**

If machine1 sent: 

```
// Machine1
set-persence ['sticker-5']
set-persence ['sticker-4', 'sticker-5']
```

machine2 would receive: 

```
// Machine2
patch-presence  [peerId, 'data', 'selectedIds'], 'r', ['sticker-5']
patch-presence  [peerId, 'data', 'selectedIds', 0], '+', 'sticker-4' // Uh oh
```

In the case where we had a `+`, we actually want to _insert_ at index 0, but instead we replaced it: 

```
// Desired: ['sticker-4', 'sticker-5']
// Got ['sticker-5']
```

Fixed this by created an `insertInMutative` function, which splices at index N rather than replace

**1. `-` operations**

Now imagine if `machine-1` sent: 

```
set-persence ['sticker-5']
set-persence ['sticker-4', 'sticker-5']
set-persence ['sticker-5']
```

Machine two would receive: 

```
patch-presence  [peerId, 'data', 'selectedIds'], 'r', ['sticker-5']
patch-presence  [peerId, 'data', 'selectedIds', 0], '+', 'sticker-4'
patch-presence  [peerId, 'data', 'selectedIds', 0], '-', 
```

In this case dissocInMutative would set the value to undefined:

```
// Desired: ['sticker-5']
// Got [undefined, 'sticker-5']
```

Fixed this by updating `dissocInMutative` to handle arrays. 

--- 

Imported the great repro user sent over here! 

image

@tonsky @dwwoelfel @nezaj

### DIFF
--- a/client/packages/core/__tests__/src/utils/object.test.js
+++ b/client/packages/core/__tests__/src/utils/object.test.js
@@ -1,6 +1,10 @@
 import { test, expect, describe } from 'vitest';
 
-import { assocInMutative, dissocInMutative } from '../../../src/utils/object';
+import {
+  assocInMutative,
+  dissocInMutative,
+  insertInMutative,
+} from '../../../src/utils/object';
 
 describe('assocInMutative', () => {
   test('adds value at a shallow path', () => {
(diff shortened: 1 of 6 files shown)


dwwoelfel (approved):
Cool!

PR #986, opened 2025-03-05
[Explorer] Enable setting null

nezaj:
A user pointed out that you couldn't set json values as `null` I thought this would be straightforward fix but it was a bit more involved. This does the following:

* Introduce a checkbox for setting a value to null
* Restricts fields types when attr is type checked
* Upgrade reset functionality to restore previous values even if you changed the input type

https://github.com/user-attachments/assets/eb45111d-0f8f-422d-a7d7-f00de9090544

### DIFF
--- a/client/www/components/dash/explorer/EditRowDialog.tsx
+++ b/client/www/components/dash/explorer/EditRowDialog.tsx
@@ -1,9 +1,4 @@
-import {
-  id,
-  InstantReactWebDatabase,
-  InstaQLParams,
-  tx,
-} from '@instantdb/react';
+import { id, InstantReactWebDatabase, tx } from '@instantdb/react';
 import { useMemo, useRef, useState } from 'react';
 
 import {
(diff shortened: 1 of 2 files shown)


client/www/components/dash/explorer/EditRowDialog.tsx
```diff
@@ -567,6 +643,13 @@ export function EditRowDialog({
     setUpdatedBlobValues((prev) => {
       const current = prev[field] || {};
 
+      if (value.trim() === 'null') {
```
stopachka: Will this mean that if someone had set string, but then they write `null`, it will change the viewer into a json value? I am not 10% sure what this is doing

stopachka (approved):
One question but LGTM!

client/www/components/dash/explorer/EditRowDialog.tsx
nezaj: Ah good catch, meant to clean this up!

PR #989, opened 2025-03-06
Fix lots of nested ands and ors

dwwoelfel:
Ran into this bug while working on cel -> instaql, where we can generate lots of nested ORs.

Three fixes in this PR:

1. If there's a single AND or OR, then we just strip it out (`[{:or [{handle: 'dww'}]}]` is the same as `{handle: 'dww'}`)
2. Collapse the ors properly, so that `[{:or [{:or [cond-a, cond-b]}, cond-c]]}]` generates `{:or [cond-a, cond-b, cond-c]}` instead of `{:or [[cond-a, cond-b], cond-c]}`.
3. Fix in datalog.clj to handle ORs in the destination path. I simplified the logic to only do recursion for generating the join conds in one place.

There is still a remaining issue where we could potentially include the same cte twice, leading to postgres complaints about ambiguous columns. I can't reproduce it without disabling the `or` collapsing, so I'm ignoring it for now. I think it might have something to do with how we generate join syms for the different groups of ors and ands. We might need to increment the level?

### DIFF
--- a/server/src/instant/db/instaql.clj
+++ b/server/src/instant/db/instaql.clj
@@ -130,6 +130,27 @@
   (and (= "and" (name k))
        (sequential? v)))
 
+(defn- collapse-or-where-conds
+  "Converts {:or [{:or [{:handle \"Jack\"}]}]} -> {:or [{:handle \"Jack\"}]}"
+  [conds]
+  (reduce (fn [acc [_k v :as c]]
+            (println (or-where-cond? c) c)
+            (if (or-where-cond? c)
+              (apply conj acc (mapcat collapse-or-where-conds v))
+              (conj acc c)))
+          []
+          conds))
+
+(defn- collapse-and-where-conds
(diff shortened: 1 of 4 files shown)


server/src/instant/db/instaql.clj
```diff
@@ -177,32 +187,37 @@
...
-                                          (collapse-or-where-conds conds)))))]
-               (if (seq conds)
-                 conds
-                 (ex/throw-validation-err!
-                  :query
-                  (:root state)
-                  [{:expected 'non-empty-list?
-                    :in (conj (:in state) :or)
-                    :message "The list of `or` conditions can't be empty."}])))}
+        (let [conds (->> v
```
dwwoelfel, line 190: This is still overly complicated--I think it would be a lot easier to reason about if we got rid of the spec coercion, but that's a bigger project.

stopachka: Oo, that could be interesting! Would make it nicer too when we print stuff deep inside the codebase

PR #992, opened 2025-03-07
[Perms] Allow bool rules

nezaj:
I was testing some perms things and forgot that we need to wrap bools in a string. Unfortunately the error we give is not helpful. Seems like it would be nice to just make the bool a string for users.

CleanShot 2025-03-07 at 11 30 09@2x


dwwoelfel (approved):
LGTM!

PR #994, opened 2025-03-07
[apps] mark for deletion

stopachka:
A user had a really large app, and our "delete" button was starting to time out. 

Changed our system to schedule apps for deletion. The way it works: 
1. Added an `deletion_marked_at` flag on apps 
2. When a user deletes an app, we only mark this flag
3. A worker runs every day. we pick up the app and run the operation in the background, with a 5 minute timeout 
4. You can disable this using the `app-deletion-sweeper` flag in instant-flags 

Deploy plan:
1. push instant-config
2. push migration
3. deploy

@dwwoelfel @nezaj @tonsky

### DIFF
--- a/client/www/pages/dash/index.tsx
+++ b/client/www/pages/dash/index.tsx
@@ -385,7 +385,9 @@ function Dashboard() {
   }
 
   async function onDeleteApp(app: InstantApp) {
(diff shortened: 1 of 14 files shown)


dwwoelfel (approved):
LGTM!

PR #995, opened 2025-03-07
[Perms] Null-safe operations

nezaj:
Makes `!`, `||`, `&&` null safe in CEL rules. This works by 1) treating `nulls` as `false` and 2) rewriting any explicit null checks (`data.owner == null` or `null == data.owner`) into checking for false.

Originally I wanted to see if we could overload the `!` operator to safely handle nulls

```
(def null-safe-not-operator
  {:decl (CelFunctionDecl/newFunctionDeclaration
          "!"
          (ucoll/array-of
           CelOverloadDecl
           [(CelOverloadDecl/newGlobalOverload
             "null_safe_logical_not"  ; This is a descriptive name for the overload
             SimpleType/BOOL
             (ucoll/array-of CelType [SimpleType/DYN]))]))
   :runtime (let [impl (reify CelFunctionOverload$Unary
                         (apply [_ val]
                           (if (or (nil? val) (= val NullValue/NULL_VALUE))
                             true  ; if null, treat as falsy
                             (not (Boolean/valueOf (str val))))))]
              ;; Use the CelFunctionBinding for unary functions - adjust based on your CEL version
              (CelRuntime$CelFunctionBinding/from
               "null_safe_logical_not"  ; Function name
               Object         ; Parameter type (can handle any object)
               impl))})
```

I could get this working but there were issues with nulls for `&&` and `||` -- those operators expect booleans on both sides.

> ; (out) Caused by: dev.cel.runtime.CelEvaluationException: evaluation error at <input>:24: expected boolean value, found: NULL_VALUE

I tried overloading those but it would still fail

```
(def null-safe-or-operator
  {:decl (CelFunctionDecl/newFunctionDeclaration
          "||"
          (ucoll/array-of
           CelOverloadDecl
           [(CelOverloadDecl/newGlobalOverload
             "null_safe_logical_or"
             SimpleType/BOOL
             (ucoll/array-of CelType [SimpleType/DYN SimpleType/DYN]))]))
   :runtime (let [impl (reify CelFunctionOverload$Binary
                         (apply [_ left right]
                           ;; If either is truthy, return true
                           (if (or (and (not= left NullValue/NULL_VALUE)
                                        (not= left nil)
                                        (Boolean/valueOf (str left)))
                                   (and (not= right NullValue/NULL_VALUE)
                                        (not= right nil)
                                        (Boolean/valueOf (str right))))
                             true
                             false)))]
              (CelRuntime$CelFunctionBinding/from
               "null_safe_logical_or"
               Object
               Object
               impl))})

(def null-safe-and-operator
  {:decl (CelFunctionDecl/newFunctionDeclaration
          "&&"
          (ucoll/array-of
           CelOverloadDecl
           [(CelOverloadDecl/newGlobalOverload
             "null_safe_logical_and"
             SimpleType/BOOL
             (ucoll/array-of CelType [SimpleType/DYN SimpleType/DYN]))]))
   :runtime (let [impl (reify CelFunctionOverload$Binary
                         (apply [_ left right]
                           ;; Both must be truthy for true result
                           ;; Null values are treated as falsy
                           (if (and (not= left NullValue/NULL_VALUE)
                                    (not= left nil)
                                    (Boolean/valueOf (str left))
                                    (not= right NullValue/NULL_VALUE)
                                    (not= right nil)
                                    (Boolean/valueOf (str right)))
                             true
                             false)))]
              (CelRuntime$CelFunctionBinding/from
               "null_safe_logical_and"
               Object
               Object
               impl))})
```

However by treating `null` as `false` all the operations are happy as is.

I checked rules to see if there were some instances where folks explicitly check against null. Indeed there are so added some logic to rewrite null checks into false checks

### DIFF
--- a/server/src/instant/db/cel.clj
+++ b/server/src/instant/db/cel.clj
@@ -255,13 +255,6 @@
       (.addFunctionBindings (ucoll/array-of CelRuntime$CelFunctionBinding custom-fn-bindings))
       (.build)))
 
-(defn transform-not-expr [expr-str]
-  (clojure-string/replace expr-str #"(?<!\!)!(?:\(([^)]+)\)|([a-zA-Z0-9_.]+))"
-                          (fn [[_ paren-expr prop-expr]]
-                            (if paren-expr
-                              (str "safeNot(" paren-expr ")")
-                              (str "safeNot(" prop-expr ")")))))
-
 (defn transform-null-expr [expr-str]
   (-> expr-str
       ;; Replace "prop == null" with "prop == false"
--- a/server/src/instant/db/cel.clj
+++ b/server/src/instant/db/cel.clj
@@ -138,7 +138,7 @@
 
 (defn stringify [x]
   (cond
-    (nil? x)           NullValue/NULL_VALUE
+    (nil? x)           false
     ;; For some reason, cel-java only supports longs when determining
     ;; type. We convert ints to longs to prevent type(data.param) from
     ;; throwing a NPE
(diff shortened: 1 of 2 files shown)


nezaj:
Will close this for now to take a different approach!

PR #997, opened 2025-03-08
Add further note about useQuery(null).isLoading being true

iansinnott:
This was not what I expected, so I think its worth elaborating a bit more. My interpretation of `isLoading` was "in flight" which is not the case here.


nezaj (approved):
LGTM! Thanks @iansinnott !

PR #1001, opened 2025-03-10
[client] Disable devtool in non-DOM contexts (React Native)

ibejoeb:
The devtool uses the Browser Object Model API, which is not available in react-native. This change tests for the existance of `document` in the global namespace chain and aborts devtool creation if it does not exist.

This allows usage in react native without explicitly passing `devtool: false` to `init`.


ibejoeb:
Example init failure due to devtool on Expo development build:

dne

stopachka (approved):
Thank you!

stopachka:
In production starting from 0.17.29!

PR #1003, opened 2025-03-10
[cli] exiting -> skipping

stopachka:
image

Noticed a little inconsistency -- when we skip schema, we say "skipping", when we skip perms, we say "exiting". But we are not necessarily exiting. 

Thought I'd be fair and just call both "skipping". 

@nezaj @dwwoelfel @tonsky


dwwoelfel (approved):
LGTM!

PR #1005, opened 2025-03-11
[s3-uploads] pass content-length

stopachka:
## Context

We were getting some errors with our async s3 crt client: 

> The service request was not made within 10 seconds of doBlockingWrite being invoked. Make sure to invoke the service request BEFORE invoking doBlockingWrite if your caller is single-threaded.

## What I learned

storage-routes/upload-put was not sending `content-length` to `upload-stream-to-s3!`.

This forced us to use the CRT client, as it can send streams without content-length. You may be wondering: how does the CRT client send the stream without a content length? It chunks the stream, and uses S3's multipart API to sends chunks (with content-length) over. 

**But, we _do_ receive the content-length header.** 

Our http handler receives a content-length header. If we passed it along, we could skip the more error-prone 'no content-length' branch of our code. Plus, it lets us use the Netty client again without many issues. Netty is not amazon specific, and has years of testing behind it. 

## Fix

I first went ahead and 

1. Extract the content-length header, and passed it in 
2. Added logging around stream-to-s3 

I am going to watch stream-to-s3, to make sure that we provided content-length in all scenarios. Once we do, we can remove the "no-content length" branch of the code, and potentially use Netty too if we want. 

@nezaj @dwwoelfel @tonsky 

## FAQ

Some questions that may come to mind: 

**Wait, is `content-length` _always_ provided?** 

In the case of multi-part forms, the content-length header will come in as `0`. **However, our upload API does not handle multi-part forms**. If we tried to send with multipart forms before, we would store the _whole_ multipart stream into one file.

@nezaj @tonsky @dwwoelfel

### DIFF
--- a/server/src/instant/storage/routes.clj
+++ b/server/src/instant/storage/routes.clj
@@ -9,16 +9,22 @@
             [instant.storage.coordinator :as storage-coordinator]
             [clojure.walk :as w]))
 
+(defn req->content-length! [{:keys [content-length] :as _req}]
+  (when content-length
+    (long content-length)))
+
 (defn req->app-file! [req params]
   (let [app-id (ex/get-param! params [:app_id] uuid-util/coerce)
         refresh-token (http-util/req->bearer-token req)
         current-user (app-user-model/get-by-refresh-token
                       {:app-id app-id
-                       :refresh-token refresh-token})]
+                       :refresh-token refresh-token})
+        content-length (req->content-length! req)]
     {:app-id app-id
      :current-user current-user
      :path (ex/get-some-param! params [[:path] [:filename]] string-util/coerce-non-blank-str)
      :content-type (ex/get-optional-param! params [:content-type] string-util/coerce-non-blank-str)
+     :content-length content-length
      :content-disposition (ex/get-optional-param! params [:content-disposition] string-util/coerce-non-blank-str)}))
 
 (defn upload-put [req]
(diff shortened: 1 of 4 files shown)


server/src/instant/storage/routes.clj
```diff
@@ -18,6 +26,7 @@
     {:app-id app-id
      :current-user current-user
      :path (ex/get-some-param! params [[:path] [:filename]] string-util/coerce-non-blank-str)
+     :content-length (ex/get-param! params [:content-length] coerce-long)
```
nezaj: I don't think we can make this required. For example it breaks "delete" in `localhost:4000/play/storage-v2`

nezaj: This also breaks legacy uploads, you can see in `localhost:4000/play/storage`

stopachka: Updated!

stopachka: (As an aside -- all those endpoints come with a content-length. This didn't parse it from the correct place)

server/src/instant/storage/routes.clj
```diff
@@ -9,16 +9,22 @@
             [instant.storage.coordinator :as storage-coordinator]
             [clojure.walk :as w]))
 
+(defn req->content-length! [{:keys [content-length] :as _req}]
```
nezaj, line 12: qq, I notice we don't explicitly call `(long content-length)` in the other `upload-put` functions for admin and dash, should we do that too?

nezaj (approved):
LGTM w/ one small note!

server/src/instant/storage/routes.clj
stopachka: GP, I will push this into the s3/stream function itself. afaik fromInputStream expects a long -- will just have it coerce directly there

PR #1007, opened 2025-03-11
[instatx] proxy should return undefined with unknown keys

stopachka:
consider: 

```js
tx.users[id].update(...).this_is_an_unsupported_command
```

Right now, we would have returned a `chunk` object. But `this_is_an_unsupported_command` is not a valid command. 

I updated our logic so unless the user calls a valid function command, we return undefined. 

@nezaj @dwwoelfel @tonsky


nezaj (approved):
Nice!

PR #1011, opened 2025-03-11
[sm] make sure html bg matches content bg

stopachka:
**Before**
We had this weird bug, where if you 'overscrolled' our marketing and docs pages, you would see a jarring white bg:

https://github.com/user-attachments/assets/2658435e-1079-4b08-bf25-753e7edd0634

**After**
This was because our html element had a white background. Updated so for those pages, we set html to have the correct background color. 

https://github.com/user-attachments/assets/60023eb6-1551-4700-bab5-f4d9bcee882d

@nezaj @dwwoelfel @tonsky


nezaj (approved):
Nice!

PR #1014, opened 2025-03-12
Instant OAuth apps

dwwoelfel:
Adds support for OAuth apps that will allow 3rd parties to perform actions on behalf of an Instant user.

General flow looks like this:

1. Company creates a new instant app which will hold their OAuth app
  a.  It would be nice to tie these to an org instead of an app, but apps is all we have for now
2. Company creates a new OAuth app on the instant dashboard. By default any app is created with `is_public: false` and only the members of the app can grant it access to their instant accounts (we need to manually flip the flag in the db on our end).
3. The instant user goes through an OAuth flow similar to the ones we currently have to grant access to their app
  a. From company's site, they redirect the user to `https://api.instantdb.com/platform/oauth/start`
  b. That stores a new OAuth flow in the db and redirects to `https://instantdb.com/platform/oauth/start`
  b. We show the user an authorization screen (Do you want to allow "Company" to access your Instant account?") after they log in to their Instant account
  c. User says yes, we redirect back to `https://api.instantdb.com/platform/oauth/grant` and check the information matches up with the OAuth flow we created
  d. We redirect back to the company's redirect url they provided in a.
  d. Company exchanges the code for a token and refresh token
  e. Company uses the token to hit the superadmin API for `create-app` and friends
  d. When the token expires, company can use their refresh token to get a new one
  e. If the company doesn't want to store anything on their side, they can use the client-only flow with pkce. The only difference is that they won't get a refresh token and will have to re-auth every two weeks.

We have two scopes "apps-read" and "apps-write". "apps-write" implicitly grants you "apps-read".  With apps-read you can list all apps, rules, and permissions. With apps-write, you can create new apps, delete existing apps, and update rules and permissions.

Still TODO (in follow-up PRs):
1. Docs page
2. Show users which OAuth apps they've enabled and let them revoke access


stopachka (approved):
Niice!

client/www/components/dash/Billing.tsx
```diff
@@ -12,7 +12,6 @@ import { errorToast } from '@/lib/toast';
 import clsx from 'clsx';
 import confetti from 'canvas-confetti';
 
-const stripePromise = loadStripe(stripeKey);
```
dwwoelfel, line 15: I updated this so that Stripe only loads when you load the billing page.

They keep a singleton internally, so it's fine to call `loadStripe` multiple times.

client/www/components/dash/Auth.tsx
```diff
@@ -119,6 +122,7 @@ function EmailStep(props: {
             href={url(config.apiURI, `/dash/oauth/start`, {
               ticket: props.ticket,
               redirect_to_dev: isDev ? 'true' : undefined,
+              redirect_path: router.asPath,
```
dwwoelfel, line 125: This will redirect the user back to the page they were on when they started the auth flow

server/src/tool.clj
```diff
@@ -278,6 +280,15 @@
   [form]
   `(p-impl (p-pos) '~form ~form))
 
+(defmacro inspect
```
stopachka, line 283: oo!

server/src/instant/util/token.clj
```diff
@@ -0,0 +1,23 @@
+(ns instant.util.token
+  (:require [clojure.string :as string]
+            [instant.util.uuid :as uuid-util]))
+
+(def platform-refresh-token-prefix "prt_")
+(def platform-access-token-prefix "pat_")
+
+(deftype PlatformAccessToken [value]
+  Object
+  (toString [_] "<secret>"))
```
stopachka, line 10: cool!

server/src/instant/superadmin/routes.clj
```diff
@@ -1,30 +1,49 @@
...
-
-(defn req->superadmin-user-and-app! [req]
-  (let [{user-id :id :as user} (req->superadmin-user! req)
+(defn req->superadmin-user!
+  "We're reusing some of the superadmin routes for the OAuth app platform routes.
+   The scope is only applicable to requests that authenticate with a
+   token generated by the OAuth app"
+  [scope req]
+  (let [token (http-util/req->bearer-token! req)]
+    (if (token-util/is-platform-access-token? token)
```
stopachka, line 28: nice idea to prefix it!

server/src/instant/model/oauth_app.clj
```diff
@@ -0,0 +1,786 @@
...
+   (create-redirect (aurora/conn-pool :write) params))
+  ([conn {:keys [redirect-id
+                 client-id
+                 state
+                 cookie
+                 redirect-uri
+                 scopes
+                 code-challenge-method
+                 code-challenge]}]
+   (let [lookup-key (crypt-util/uuid->sha256 redirect-id)
```
stopachka, line 517: 🫡

server/src/instant/jdbc/sql.clj
```diff
@@ -300,7 +300,13 @@
                   (apply-postgres-config postgres-config# create-connection?# c#)
                   (with-open [ps# (next-jdbc/prepare c# query# opts#)
                               _cleanup# (register-in-progress create-connection?# ~rw c# ps#)]
-                    (~query-fn ps# nil opts#))
+                    (let [res# (~query-fn ps# nil opts#)]
+                      (when (false? (:return-keys opts#))
```
stopachka: Curious q: why add attributes only when this is false?

stopachka (approved):
Beautiful work @dwwoelfel !

server/src/instant/jdbc/sql.clj
dwwoelfel: When return-keys is false, then next-jdbc will return its metadata as the result. Otherwise it returns the result-set and there's no :update-count to add to the query attrs.

I'll look and see if there are some general query stats we can tell next-jdbc to give us or there might be a way to grab them from the prepared-statement (`ps#` in the code).

stopachka: cool, good to know!

dwwoelfel: Updated to get use `.getUpdateCount` on the prepared statement

PR #1015, opened 2025-03-12
"Was this page helpful?"

stopachka:
I wanted to test out ruleParams, so I built out a little "Was this page helpful?" component for our docs : ). 

Additional changes: 
1. I added a `useLocalId` hook to our react SDK. It's a small syntax sugar so the user can extract the value from our `getLocalId` promise, without having to write a hook. 

https://github.com/user-attachments/assets/d53aabb9-3ebe-4c9f-bb91-38d329fedc2c

@dwwoelfel @nezaj @tonsky


client/www/lib/feedback/RatingBox.tsx
```diff
@@ -0,0 +1,165 @@
...
+      <div className="text-right">
+        <Button
+          disabled={value === savedValue}
+          onClick={() => {
+            onSave(value);
+          }}
+          size="mini"
+          type="submit"
+        >
+          Save
```
dwwoelfel: Maybe we could make this say `Saved!` if `value && value === savedValue`?

When I clicked "Save", it felt like it didn't do enough and I wasn't 100% sure it actually saved.

An alternative would be to let the user add many comments on the page and when you click "submit" it clears the input and shows the comment at the bottom of the page. Then they can leave multiple comments on the same page.

dwwoelfel (approved):
LGTM!

client/www/lib/feedback/RatingBox.tsx
stopachka: Ah, good idea! Will start with `value && value === savedValue` 🫡

PR #1017, opened 2025-03-13
[rule-params] fix tx.foo[newId].ruleParams

stopachka:
This fixes a bug, where ruleParams would fail if we created an entity. 

(Co-authored by @tonsky and me!) 

@dwwoelfel @nezaj

### DIFF
--- a/client/packages/react/src/InstantReactAbstractDatabase.ts
+++ b/client/packages/react/src/InstantReactAbstractDatabase.ts
@@ -290,10 +290,11 @@ export default abstract class InstantReactAbstractDatabase<
    */
   queryOnce = <Q extends InstaQLParams<Schema>>(
     query: Q,
+    opts?: InstaQLOptions,
   ): Promise<{
     data: InstaQLResponse<Schema, Q>;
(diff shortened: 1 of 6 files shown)


tonsky:
Hopefully fixed. But I removed test with create + lookup. I think you shouldn’t be able to create with lookup, no?

> Updates with lookups can only update the lookup attribute if an entity with the unique attribute value already exists.

```
db.tx.users(["handle", "alyssa"]).update({"handle": "alyssa"})
```

↑ this seems wrong to me. Lookups should be for updates

stopachka:
Heeck yeah! 

>  I think you shouldn’t be able to create with lookup, no?

Right now lookups create entities if they don't exist: 

https://www.instantdb.com/docs/instaml#:~:text=If%20no%20entity%20has%20the%20value%2C%20then%20a%20new%20entity%20with%20a%20random%20id%20will%20be%20created%20with%20the%20value.

It does feel a little off, but we chose this behavior, is it was consistent with the rest of `update`. (i.m.o we may want to let users have more specific verbs in the next tx api)

PR #1021, opened 2025-03-14
Fix rule params create with lookup

tonsky:
Fixed `ruleParams` for creates with lookups like

```
tx.docs[lookup('key', key)]
  .ruleParams({ ... })
  .update({ ... });
```

### DIFF
--- a/server/src/instant/db/permissioned_transaction.clj
+++ b/server/src/instant/db/permissioned_transaction.clj
@@ -638,7 +638,7 @@
                 tx-data
                 (tx/transact-without-tx-conn-impl! tx-conn (:attrs ctx) app-id grouped-tx-steps {})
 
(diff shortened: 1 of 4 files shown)


server/src/instant/db/permissioned_transaction.clj
```diff
@@ -638,7 +638,21 @@
                 tx-data
                 (tx/transact-without-tx-conn-impl! tx-conn (:attrs ctx) app-id grouped-tx-steps {})
 
-                create-checks-resolved (resolve-lookups-for-create-checks tx-conn app-id create-checks)
+                ;; udpate lookups with newly created triples
```
stopachka:
```suggestion
                ;; update lookups with newly created triples
```

stopachka (approved):
Niice!

PR #1023, opened 2025-03-14
[docs] fix inconsistency when loading from cache

stopachka:
When we matched from local cache, we did not check if the app still existed. This meant that if a user deleted an app, we'd get in a broken state. 

Fixed by making sure that we match to an actual app in our `apps` list. 

@nezaj @dwwoelfel @tonsky


dwwoelfel (approved):
LGTM!

PR #1024, opened 2025-03-14
[explorer] wrap delete row button in tr

stopachka:
I noticed in dev we get: 

```
Warning: validateDOMNesting(...): <div> cannot appear as a child of <thead>. Error Component Stack
```

This is because `<div>` is not a proper child of `<thead>`. I went ahead and updated it to be a `<tr>` 

@dwwoelfel @nezaj @tonsky


dwwoelfel (approved):
LGTM!

PR #1025, opened 2025-03-14
[reactor] consider adminToken in `globalInstantCoreStore`.

stopachka:
We cache reactor instantiations, when users write `const db = init`. 

This was causing a subtle bug with how our feedback component was interacting with our Explorer. 

Here's what would happen: 
1. `feedback/clientDB.ts` was instantiating a db for appId `A` [1]
2. If you viewed Explorer of app `A`, we would try to instantiate a db for `A`, with adminToken `B`. 
  a. _But_ this would hit the cache, and get the previously instantiated Reactor, _without_ the adminToken. 

I updated our code, so we now use a combo of appId, websocketURI, apiURI, and adminToken to cache inits

[1] you may be wondering: why is `clientDB.ts` called at all? It doesn't seem like it should be included in the bundle for dash. Will explore this in a separate PR. 

@dwwoelfel @nezaj @tonsky

### DIFF
--- a/client/packages/core/src/index.ts
+++ b/client/packages/core/src/index.ts
@@ -193,6 +193,12 @@ function initGlobalInstantCoreStore(): Record<string, any> {
   return globalThis.__instantDbStore;
 }
 
+function reactorKey(config: InstantConfig<any>): string {
+  // @ts-expect-error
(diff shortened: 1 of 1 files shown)


nezaj (approved):
Sounds good!

PR #1027, opened 2025-03-17
[landing] improve our social previews

stopachka:
Made a series of improvements to our social link improvements. 

## Main Change: Title + Custom OG Image Preview for Blog Posts

image

- I use the actual blog post title for the og title 
- I leave off the description, so og properly loads a preview of the text
- I made a custom og image generator

## Other change: removed defaults

In `_app.js` we set defaults like:  `title`, `description`, and `url`. But this doesn't quite make sense. If you loaded pricing for example: 

**Before**

CleanShot 2025-03-17 at 11 02 31@2x

- You got a non-descriptive title and description
- The og:url was wrong here, saying it was the same as the home page. 

**After** 

image

@tonsky @nezaj @dwwoelfel

### DIFF
--- a/client/www/pages/_app.tsx
+++ b/client/www/pages/_app.tsx
@@ -140,14 +140,14 @@ function AppHead() {
         content="width=device-width, initial-scale=1, maximum-scale=1"
       />
       <meta name="theme-color" content="#ffffff" />
-      <meta property="og:type" content="website" />
-      <meta property="og:url" content="https://www.instantdb.com" />
-      <meta property="og:title" content="InstantDB: A Modern Firebase" />
+      <meta key="og:type" property="og:type" content="website" />
       <meta
-        property="og:description"
-        content="We make you productive by giving your frontend a real-time database."
+        key="og:url"
+        property="og:url"
+        content="https://www.instantdb.com"
       />
       <meta
+        key="og:image"
         property="og:image"
         content="https://www.instantdb.com/img/og_preview.png"
       />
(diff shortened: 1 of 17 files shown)


client/www/pages/_app.tsx
```diff
@@ -140,14 +140,14 @@ function AppHead() {
         content="width=device-width, initial-scale=1, maximum-scale=1"
       />
       <meta name="theme-color" content="#ffffff" />
-      <meta property="og:type" content="website" />
-      <meta property="og:url" content="https://www.instantdb.com" />
-      <meta property="og:title" content="InstantDB: A Modern Firebase" />
+      <meta key="og:type" property="og:type" content="website" />
```
stopachka, line 143: I added a `key` to these; this lets child components define a `Head` that overwrites the values when they want too.

client/www/pages/api/og/essay.tsx
```diff
@@ -0,0 +1,91 @@
...
+};
+
+export default async function handler(request: NextRequest) {
+  const { searchParams } = new URL(request.url);
+  const title = searchParams.get('title');
+  if (!title) {
+    return new Response('Title is required', { status: 400 });
+  }
+
+  const res = await fetch(
```
dwwoelfel, line 15: Is it possible to put the `ttf` in a local file and import it from there instead of fetching it each time?

dwwoelfel (approved):
Cool!

client/www/pages/api/og/essay.tsx
stopachka: There's two pieces preventing us from doing this: 

1. I spoke to the Berkeley Mono team about using the font for web. The rule they said was, that the file should not be included in source control, on github.com et al. 
2. I don't think we can import binary data if we want to run this on the 'edge' runtime. The example for custom fonts on nextjs show this curious pattern too: https://vercel.com/guides/using-custom-font

PR #1028, opened 2025-03-17
[docs][bugfix] app id from query params should work

stopachka:
Nicole caught a bug, where if you clicked a docs link from our dashboard, and it included the `app` param, our logic to select the app would break. This was due to a rename issue. 

@dwwoelfel @nezaj @tonsky

### DIFF
--- a/client/www/components/docs/Layout.jsx
+++ b/client/www/components/docs/Layout.jsx
@@ -27,14 +27,16 @@ function useSelectedApp(apps = []) {
     const cachedAppData = getLocal(cacheKey);
     const { app: queryAppId, ...remainingQueryParams } = router.query;
 
-    const matchFromQueryParam =
-      queryAppId && apps.find((a) => a.id === queryAppId);
-    const matchFromCache =
+    const fromParams = queryAppId && apps.find((a) => a.id === queryAppId);
+    const fromCache =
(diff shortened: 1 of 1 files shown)


client/www/components/docs/Layout.jsx
```diff
@@ -27,14 +27,16 @@ function useSelectedApp(apps = []) {
...
-      queryAppId && apps.find((a) => a.id === queryAppId);
-    const matchFromCache =
+    const fromParams = queryAppId && apps.find((a) => a.id === queryAppId);
+    const fromCache =
       cachedAppData && apps.find((a) => a.id === cachedAppData.id);
     const first = apps[0];
-    if (matchFromQueryParam) {
+    if (fromParams) {
       // We got a match for from a query param. Let's cache it and use it
-      const data = { id: match.id, title: match.title };
```
nezaj, line 37: I'm guessing the problem was previously this should have just been?

```
const data = { id: matchFromQueryParam.id, title: matchFromQueryParam.title };
```

nezaj (approved):
Sounds good!

client/www/components/docs/Layout.jsx
stopachka: Yes!

PR #1029, opened 2025-03-18
Encode instant in json

dwwoelfel:
`refresh` timeouts are getting "can't encode Instant as JSON" errors because we inject a `java.time.Instant` as `tx-created-at` into the event that we try to encode as the ":original-event" when a `refresh` fails.

This PR tells the json encoder how to handle Instants as a quick fix.

A better fix would be to refactor `session.clj` to not put our internal stuff inside the event itself, but that's a bigger project.


stopachka (approved):
Aweesome!

PR #1031, opened 2025-03-18
[files] fix `url` in fields

stopachka:
## Bug 

If you explicitly asked for a `url` in your fields: 

```js
{ $files: { $: {  fields: ['url'] }  }
```

You would get a null value in the url: 

```js
{ id: '...', url: null }
```

## Cause

This was because in order for us to generate a `url`, we need the `location-id` field. But since that was omitted, we weren't generating a proper url. 

## Fix

1. In `etype-attr-ids`, I include the `location-id`, if the user asks for `url`
2. When we transform results for files, I now: 
  a. Only add `url` if it was requested 
  b. Remove location-id if it was not requested

I also added some tests in instaql.test.clj 

--- 

@nezaj @dwwoelfel @tonsky


server/src/instant/db/instaql.clj
```diff
@@ -1018,19 +1036,29 @@
...
-                          [etype field]
-                          attrs)]
-                (if (= :one (:cardinality attr))
-                  (conj acc (:id attr))
-                  acc)))
-            #{}
-            ;; Make sure we give them the id or else the client
-            ;; won't be able to find the entity
-            (conj fields "id"))
+    (let [fields' (cond-> fields
```
dwwoelfel: An alternative way to do this that's a little bit faster:


```clojure
(def $file-location-id-attr-id (system-catalog/get-attr-id "$files" "location-id"))
(def $file-url-attr-id (system-catalog/get-attr-id "$files" "url"))

(defn etype-attr-ids [{:keys [attrs]} etype fields]
  (if fields
    (let [attr-ids (reduce (fn [acc field]
                             (let [attr (attr-model/seek-by-fwd-ident-name
                                         [etype field]
                                         attrs)]
                               (if (= :one (:cardinality attr))
                                 (conj acc (:id attr))
                                 acc)))
                           #{}
                           ;; Make sure we give them the id or else the client
                           ;; won't be able to find the entity
                           (conj fields "id"))]
      (if (contains? attr-ids $file-url-attr-id)
        (conj attr-ids $file-location-id-attr-id)
        attr-ids))
    (attr-model/ea-ids-for-etype etype attrs)))
```

dwwoelfel (approved):
LGTM!

server/src/instant/db/instaql.clj
stopachka: Good idea, done!

PR #1033, opened 2025-03-19
[Dashboard] Reverse sort icons

nezaj:
User pointed out our icons in explorer are confusing. Seems like the convention is indeed to be reverse of what we currently have. [[Discord]](https://discord.com/channels/1031957483243188235/1148284450992574535/1351718429924659294)


dwwoelfel (approved):
LGTM. Feels wrong to me, but it seems to be what everyone else does.

PR #1034, opened 2025-03-19
create entities from link lookups

stopachka:
Consider a transaction like: 

```js
db.tx.users[lookup('handle', 'bobby_newuser')].link({ bookshelves: bookshelfId })
```

If a `user` with handle `bobby_newuser` did not exist, we would: 

1. add triple [user-id :handle "bobby-new-user"]
2. add triple [user-id :bookshelves bookshelfId]

## Problem

In this case, we never added the `id` triple. This means that when writing queries like: 

```
{ users: { } }
```

This newly created `bobby_newuser` would not show up. 

## Solution

I went ahead and inserted the id triple even when we are making links 

@dwwoelfel @nezaj  @tonsky

### DIFF
--- a/client/packages/core/__tests__/src/instaml.test.js
+++ b/client/packages/core/__tests__/src/instaml.test.js
@@ -585,6 +585,58 @@ test('lookup creates unique ref attrs for ref lookup in link value', () => {
   }
 });
 
+test('lookups create entities from links', () => {
+  const bookshelfId = uuid();
+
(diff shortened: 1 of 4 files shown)


dwwoelfel (approved):
LGTM, but maybe we should only be doing it for `link` and not `unlink`?

stopachka:
> LGTM, but maybe we should only be doing it for link and not unlink?

My thinking was: 

In `unlink`, we technically still create the underlying lookup triple. 

To be consistent, perhaps we should do this on both sides?

PR #1037, opened 2025-03-20
[perms] handle ruleParams in the view check on link

stopachka:
### Problem: view checks on links

If a user had a rule like: 

```js
{ 
  "books": { 
     "allow": { 
        "view": "data.id == ruleParams.knownBookId"
     }
}
```

And wrote a transaction like: 

```js
tx.bookshelves[id].ruleParams({ knownBookId: bookId }).update({ ... }).link({ books: bookId })
```

This would lead to a failed transaction. 

**The reason?: view check did not receive any ruleParams** 

**Fix** I made the following change: 

Now, when we insert the `view` check, we also make sure to copy the ruleParams from the transaction that was causing it, onto that check. 

@tonsky @nezaj @dwwoelfel

### DIFF
--- a/server/src/instant/db/permissioned_transaction.clj
+++ b/server/src/instant/db/permissioned_transaction.clj
@@ -236,16 +236,18 @@
 
    And we group them by `eid`, `etype`, and `action`.
 
-   {
-     {:eid joe-eid
+   :groups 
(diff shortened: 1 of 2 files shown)


dwwoelfel (approved):
LGTM!

PR #1039, opened 2025-03-21
[admin][rooms] get presence

stopachka:
Let's let users get presence info from the admin API!
 
Added a `GET /admin/presence?room-type=A&room-id=B` endpoint. This: 
1. Gets presence info
2. Enhances user data with things like email, et al

@nezaj @dwwoelfel @tonsky

### DIFF
--- a/server/src/instant/admin/routes.clj
+++ b/server/src/instant/admin/routes.clj
@@ -488,8 +488,8 @@
         ;; rooms. 
         ;; Adding this as a required field so once we do use it we won't have a breaking 
         ;; issue here.
-        _room-type (ex/get-param! req [:params "room-type"] string-util/coerce-non-blank-str)
-        room-id (ex/get-param! req [:params "room-id"] string-util/coerce-non-blank-str)
+        _room-type (ex/get-param! req [:params :room-type] string-util/coerce-non-blank-str)
+        room-id (ex/get-param! req [:params :room-id] string-util/coerce-non-blank-str)
         room-data (eph/get-room-data app-id room-id)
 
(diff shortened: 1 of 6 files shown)


server/src/instant/admin/routes.clj
```diff
@@ -477,6 +480,28 @@
                   files)]
     (response/ok {:data data})))
 
+(defn presence-get [req]
+  (let [{app-id :app_id} (req->admin-token! req)
+        _room-type (ex/get-param! req [:params "room-type"] string-util/coerce-non-blank-str)
```
dwwoelfel: Could use a comment about why we require this field but don't use it.

server/src/instant/admin/routes.clj
```diff
@@ -477,6 +480,28 @@
...
+                          vals
+                          (keep (comp :id :user))
+                          set)
+
+        id->user (when (seq user-ids)
+                   (app-user-model/get-by-ids {:app-id app-id :ids user-ids}))
+
+        enhanced-room-data (medley/map-vals
+                            (fn [sess]
+                              (ucoll/update-when sess :user
```
dwwoelfel: medley has `update-existing`

dwwoelfel (approved):
Cool!

server/src/instant/admin/routes.clj
stopachka: _makes salute_ on it!

server/src/instant/admin/routes.clj
stopachka: Oo, nice, will use!

PR #1041, opened 2025-03-21
[cel] throw when `view` or `delete` has `newData`

stopachka:
Previously it was possible to write a `view` or `delete` rule, with a `newData` var. We never passed this value in though, so it would cause an error. 

I update dour cel-compiler, so we use one without the `newData` field for view and delete checks 

CleanShot 2025-03-21 at 16 57 22@2x

I checked to see if there are any rules in prod that does this, and found none: 

```sql
SELECT 
  app_id,
  r.rule_name,
  r.rule_obj->'allow' AS allow_rules
FROM rules,
  jsonb_each(code) AS r(rule_name, rule_obj)
WHERE 
  r.rule_obj ? 'allow'
  AND (
       (r.rule_obj->'allow'->>'view' ILIKE '%newData%')
       OR (r.rule_obj->'allow'->>'delete' ILIKE '%newData%')
  );
```

@dwwoelfel @nezaj @tonsky

### DIFF
--- a/server/src/instant/db/cel.clj
+++ b/server/src/instant/db/cel.clj
@@ -304,16 +304,24 @@
 
 ;; n.b. if you edit something here, make sure you make the
 ;;      equivalent change to iql-cel-compiler below
-(def ^:private ^CelCompiler cel-compiler
+
+(def ^:private runtime-compiler-builder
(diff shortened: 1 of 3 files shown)


dwwoelfel (approved):
LGTM!

PR #1044, opened 2025-03-22
[hotfix] add newData back to view-delete rules

stopachka:
We automatically prepend all binds to cel rules. This caused an error when using our view-delete compiler. 

@dwwoelfel @nezaj @tonsky


dwwoelfel (approved):
LGTM!

PR #1046, opened 2025-03-24
[email-providers] switch most auth emails to sendgrid

stopachka:
Postmark has been having some deliverability issues.  This PR: 

1. Adds sendgrid as an alternative emailer sender. 
  a. Right now, I authenticated `auth-sg.instantdb.com` for sendgrid 
2. Switched sending magic codes for instantdb.com to sendgrid
3. Switched sending magic codes for users _who do not have custom email signatures_ to sendgrid
  a. (People who have custom email signatures have verified on Sendgrid.)

**Note** 
The only difference in the API is that sendgrid uses arrays maps and maps for `from` and `to` fields: 

```clj
;; postmark 
:from "Instant <auth@pm.instantdb.com>
:to "stepan.p@example.com"
;; sendgrid
:from {:name "Instant" :email "auth@pm.instantdb.com"}
:to [{:email "stepan.p@example.com"}]
```

I added a helper that takes the "structured" request style of sendgrid, and can convert it to postmark too. I think eventually we should switch to this format, as it's easier to work with, and will let us switch providers more quickly in the future.

@nezaj @dwwoelfel @tonsky


nezaj (approved):
Nice!

PR #1049, opened 2025-03-25
Add validation for required attrs

tonsky:
This PR only adds required attr to the schema and validates it during tx. Since no attrs will be marked as required at the start, this shouldn’t affect anything and is safe to merge. I suggest we merge this first to minimise merge conflicts later down the road

### DIFF
--- a/server/src/instant/db/model/triple.clj
+++ b/server/src/instant/db/model/triple.clj
@@ -146,14 +146,14 @@
   "Given [{:entity_id ..., :attr_id ...} ...] of entity attributes that were
    added or removed during this tx, checks that all affected entities that are still
    alive have all required attributes set"
-  [conn app-id all-inserts]
-  (let [all-inserts
-        (sql/recordset all-inserts
+  [conn app-id eid+attr-ids]
+  (let [eid+attr-ids
+        (sql/recordset eid+attr-ids
                        {'entity_id {:type :uuid, :as 'entity-id}
                         'attr_id   {:type :uuid, :as 'attr-id}})
 
         eid+etypes
-        {:from 'all-inserts
+        {:from 'eid+attr-ids
          :join ['attrs  [:= 'attr-id 'attrs/id]
                 'idents [:= 'attrs/forward-ident 'idents/id]]
          :select-distinct ['entity-id 'etype]}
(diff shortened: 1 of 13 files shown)


server/src/instant/db/model/triple.clj
```diff
@@ -123,152 +124,234 @@
...
+
+        res (sql/do-execute! conn (hsql/format query))]
+    (when (seq res)
+      (ex/throw-validation-err!
+       :required-attrs
+       res
+       (for [{entity-id :entity_id
+              etype     :idents/etype
+              label     :idents/label} res]
+         {:message (str (str/capitalize etype) " " entity-id " is missing required attribute '" label "'")})))))
```
dwwoelfel: It might be weird to see the etype capitalized in the error message, I would just leave it lowercase.

Would also be handy to put the `etype`,`entity-id`, and `label` in the hint field.

server/src/instant/db/transaction.clj
```diff
@@ -349,7 +349,13 @@
...
              {}
              grouped-tx-steps)
-
+            eid+attr-ids (distinct
+                          (concat
+                           (:delete-entity results)
+                           (:add-triple results)
+                           (:deep-merge-triple results)
+                           (:retract-triple results)))
+            _  (triple-model/validate-required! conn app-id eid+attr-ids)
```
dwwoelfel: I'm worried about the extra queries that we're adding to `transact`. 

We have the attrs here, could we at least skip this query if there are no entities with required attrs that are affected by this transaction?

server/src/instant/db/model/triple.clj
```diff
@@ -123,152 +124,234 @@
...
+                                            )"]]]})))
+
+(defn validate-required!
+  "Given [{:entity_id ..., :attr_id ...} ...] of entity attributes that were
+   added or removed during this tx, checks that all affected entities that are still
+   alive have all required attributes set"
+  [conn app-id eid+attr-ids]
+  (let [eid+attr-ids-cte
+        (sql/recordset eid+attr-ids
+                       {'entity_id {:type :uuid, :as 'entity-id}
```
stopachka: It does look visually better!

server/src/instant/jdbc/sql.clj
```diff
@@ -124,6 +124,99 @@
...
+               [col-name as]
+               col-name))
+   :from   [[[:json_to_recordset [:cast (->json rs) :json]]
+             [[:raw (str "("
+                         (string/join ", "
+                                      (for [[col-name {:keys [type]}] cols]
+                                        (str (name col-name) " " (name (or type "text")))))
+                         ")")]]]]})
+
 (defn span-attrs-from-conn-pool [conn]
```
stopachka, line 220: niice!

PR #1050, opened 2025-03-27
[storage] delay storage apis

stopachka:
OS devs reported that they get this error: 

"Unable to load region from any of the providers in the chain software.amazon.awssdk.regions.providers.DefaultAwsRegionProviderChain@375bc55f: [software.amazon.awssdk.regions.providers.SystemSettingsRegionProvider@6dbb85bd: Unable to load region from system settings. Region must be specified either via environment variable (AWS_REGION) or  system property (aws.region)., software.amazon.awssdk.regions.providers.AwsProfileRegionProvider@4a038130: No region provided in profile: default, software.amazon.awssdk.regions.providers.InstanceProfileRegionProvider@37ac16a5: Unable to contact EC2 metadata service.]"}}

I have a hunch this may be because we eagerly load s3 configuration when we start storage/s3. I wrapped the config objects into a delay again, so OS devs don't have to deal with this, unless they want to enable storage.

@dwwoelfel @tonsky @nezaj

### DIFF
--- a/server/src/instant/core.clj
+++ b/server/src/instant/core.clj
@@ -48,8 +48,7 @@
    [ring.middleware.json :refer [wrap-json-body wrap-json-response]]
    [ring.middleware.keyword-params :refer [wrap-keyword-params]]
    [ring.middleware.multipart-params :refer [wrap-multipart-params]]
-   [ring.middleware.params :refer [wrap-params]]
-   [instant.storage.s3 :as storage-s3])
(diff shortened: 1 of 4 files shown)


dwwoelfel (approved):
LGTM, but if we're going to put it in a `delay`, we could do that where we declare it instead and get rid of the `start` function.

stopachka:
Good point @dwwoelfel ! Will update ya : )

PR #1051, opened 2025-03-31
[admin][cleanup] remove `InstantAdmin`

stopachka:
We have two admin classes: 
- `InstantAdmin`
- `InstantAdminDatabase`

This happened because we were upgrading our `init` to support schemas. Eventually InstantAdminDatabase started to handle all cases. 

@dwwoelfel @nezaj @tonsky


dwwoelfel (approved):
LGTM!

PR #1055, opened 2025-03-31
[admin][query] add fetchOpts to query

stopachka:
For NextJS users who want to use admin.query in ISR:

If they use `admin.query`, it will explicitly fail, because we pass in `cache: 'no-store'` into fetch headers.  

I added a `fetchOpts` option, so users can explicitly set a different option: 

```js
admin.query({ ... }, { fetchOpts: { cache: 'force-cache' } })
```

@dwwoelfel @nezaj @tonsky


dwwoelfel (approved):
LGTM!

PR #1057, opened 2025-03-31
[rules] Only include binds that are used

dwwoelfel:
Only adds a `bind` when the rule actually uses the bind.

For this rule, we don't need to include the `isOwner` bind:

```json
{"myns": {"bind": ["isOwner", "data.ownerId == auth.id"], "allow": {"view": "true"}}}
```

Before:
```
cel.bind(isOwner, data.ownerId == auth.id, true)
```

After:
```
true
```

This should fix the problem we had with `newData` showing up in `view` rules.

I also added a cache for the converting the rule to a program, since we're doing a little extra work each time we parse the rule now.

### DIFF
--- a/server/src/instant/db/cel.clj
+++ b/server/src/instant/db/cel.clj
@@ -23,7 +23,8 @@
                        CelExpr$CelComprehension
                        CelExpr$ExprKind$Kind
                        Expression$Map$Entry)
-   (dev.cel.common.navigation CelNavigableExpr)
+   (dev.cel.common.navigation CelNavigableAst
+                              CelNavigableExpr)
    (dev.cel.common.types CelType
                          ListType
                          MapType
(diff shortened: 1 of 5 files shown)


server/src/instant/db/cel.clj
```diff
@@ -343,6 +344,34 @@
...
+               (when (= CelExpr$ExprKind$Kind/IDENT (.getKind n))
+                 (.name
+                  (.ident
+                   (get-expr n)))))
+             (-> (.parse compiler expr-str)
+                 (.getAst)
+                 (CelNavigableAst/fromAst)
+                 (.getRoot)
+                 (.allNodes)
+                 (stream-seq!)))))
```
stopachka, line 373: TIL stream-seq!

server/src/instant/model/rule.clj
```diff
@@ -68,14 +69,34 @@
...
+  (clojure.set/intersection bind-keys
+                            (cel/ident-usages compiler expr)))
+
+
+(defn with-binds [rule etype action expr]
+  (let [binds (get-in rule [etype "bind"])]
+    (if (empty? binds)
+      expr
+      (let [compiler (cel/action->compiler action)
+            bind-map (apply hash-map binds)
```
stopachka, line 82: This is really cool! 

One other win: 

Because of this change `binds` doesn't have to be an array anymore! It _could_ be a hash-map, which imo would make it a bit tighter visually! (Originally I chose the array so that we compiled 'in order', and skirted this)

Maybe down the road we can update our rule lang to accept bind as an object

stopachka (approved):
Monster @dwwoelfel !

PR #1060, opened 2025-04-01
[client] no op of undefined

stopachka:
Made the following changes to instaml: 

**Updating with `undefined` is a no-op:** 

Previously, writing: 

```js
tx.items[id].update({ foo: 1, bar: undefined })
```

Would generate: 

```js
{ foo: 1, bar: null }
```

Now we no-op undefined: 

```js
{ foo: 1 } 
```

**Undefined is ignored in deepMerge**

Previously `deepMerge` would remove keys that came in as undefined: 

```js
// original
{ a: 1, b: 2 }
// toMerge
{ b: undefined } 
// produced: 
{ a: 1 }
```

Now we ignore undefined. 

--- 

@dwwoelfel @nezaj @tonsky

### DIFF
--- a/client/packages/core/__tests__/src/store.test.js
+++ b/client/packages/core/__tests__/src/store.test.js
@@ -470,3 +470,58 @@ test('ruleParams no-ops', () => {
 
   checkIndexIntegrity(newStore);
 });
+
+test('deepMerge', () => {
+  const gameId = uuid();
+  const gameStore = transact(
+    store,
+    instaml.transform(
+      { attrs: store.attrs },
+      tx.games[gameId].update({
+        state: {
+          score: 100,
+          playerStats: { health: 50, mana: 30, ambitions: { win: true } },
+          inventory: ['sword', 'potion'],
+          level: 2,
+        },
+      }),
+    ),
+  );
+  const updatedStore = transact(
+    gameStore,
+    instaml.transform(
+      { attrs: gameStore.attrs },
+      tx.games[gameId].merge({
+        state: {
+          // Objects update deeply
+          playerStats: {
+            mana: 40,
+            stamina: 20,
+            ambitions: { acquireWisdom: true, find: ['love'] },
+          },
+          // arrays overwrite
+          inventory: ['shield'],
+          // null removes the key
+          score: null,
+          // undefined is ignored
+          level: undefined,
+        },
+      }),
+    ),
+  );
+  const updatedGame = query(
+    { store: updatedStore },
+    { games: { $: { where: { id: gameId } } } },
+  ).data.games[0];
+  expect(updatedGame.state).toEqual({
+    playerStats: {
+      health: 50,
+      mana: 40,
+      stamina: 20,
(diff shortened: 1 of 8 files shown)


client/packages/core/src/instaml.js
```diff
@@ -186,7 +186,8 @@ function expandUnlink(attrs, [etype, eidA, obj]) {
   return withIdAttrForLookup(attrs, etype, eidA, retractTriples);
 }
 
-function expandUpdate(attrs, [etype, eid, obj]) {
+function expandUpdate(attrs, [etype, eid, _obj]) {
+  const obj = immutableRemoveUndefined(_obj);
```
dwwoelfel:
```suggestion
function expandUpdate(attrs, [etype, eid, obj_]) {
  const obj = immutableRemoveUndefined(obj_);
```

`_obj` makes me think that the variable isn't going to be used.

client/packages/core/src/instaml.js
```diff
@@ -203,12 +204,12 @@ function expandDelete(attrs, [etype, eid]) {
   return [['delete-entity', lookup, etype]];
 }
 
-function expandDeepMerge(attrs, [etype, eid, obj]) {
+function expandDeepMerge(attrs, [etype, eid, _obj]) {
+  const obj = immutableRemoveUndefined(_obj);
```
dwwoelfel:
```suggestion
function expandDeepMerge(attrs, [etype, eid, obj_]) {
  const obj = immutableRemoveUndefined(obj_);
```

client/packages/core/__tests__/src/store.test.js
```diff
@@ -470,3 +470,58 @@ test('ruleParams no-ops', () => {
...
+  const gameId = uuid();
+  const gameStore = transact(
+    store,
+    instaml.transform(
+      { attrs: store.attrs },
+      tx.games[gameId].update({
+        state: {
+          score: 100,
+          playerStats: { health: 50, mana: 30, ambitions: { win: true } },
+          inventory: ['sword', 'potion'],
```
dwwoelfel, line 484: What happens if you put `undefined` as an item in the array?

dwwoelfel (approved):
LGTM!

client/packages/core/__tests__/src/store.test.js
stopachka: Oo good question! 

Say we merge: 

```js
{ inventory: ['sword', undefined, 'potion'], }
```

**On the client**

Since we overwrite arrays, we copy the array over as is. The optimistic update would produce:

```js
{ state { ... , inventory: ['sword', undefined, 'potion'], } }
```

`JSON.stringify(['sword', undefined, 'potion'])` produces: 

```js
['sword', null, 'potion'], }
```

So eventually it will become: 

```
{ state { ... , inventory: ['sword', null, 'potion'], } }
```

--- 

In this case I _think_ it's reasonable. Added a test to be explicit about this. Let me know!

dwwoelfel: LGTM!

PR #1062, opened 2025-04-01
[cli] detect packages in devDependencies too

stopachka:
Sometimes folk put the instant module in devDependenices. This could happen if they are writing a script for example. 

Updated it so when detecting an instant module, we check both dependencies and devDependencies

@dwwoelfel @nezaj @tonsky


dwwoelfel (approved):
LGTM!

PR #1065, opened 2025-04-01
Make apps select box searchable

dwwoelfel:
Uses a combobox for the apps select input to make it easier to search for an app.

Direct link to preview: https://instant-www-js-search-apps-jsv.vercel.app/dash

image

Filtered:
image


stopachka (approved):
Niice!

PR #1066, opened 2025-04-01
Add llms.txt and llms-full.txt

nezaj:
This PR adds the following:

* A script for generating a public `llms.txt` and `llms-full.txt` based on our docs
* A doc with common mistakes and corrections to help LLMS.

These files follow the format from https://llmstxt.org/#format, you can see an example of this with anthropic's [llms.txt](https://docs.anthropic.com/llms.txt) and [llms-full.txt](https://docs.anthropic.com/llms-full.txt)

**Question:** Do we want to include the common mistakes in our doc navigation? This format was made more for LLMs than for humans.

## Background

I learned quite a few things the last few weeks exploring vibe coding. Here’s what the evolution of these `llms.txt` files looked like

### Trying Cursor

I first tried using Cursor to make a blog app

* I tried adding a url to our docs as context, that worked sometimes, but sometimes not.
* Even with the docs indexed, it seemed like Cursor would hallucinate quite a bit, common hallucinations include
    * calling `create` instead of `update`
    * Not understanding entity ids needs to be uuids
    * Using `orderBy` instead of `order`
* The above were minor and could easily be fixed, but one thing that would be a common theme was the struggle understanding
    * how to properly make profiles after creating a user
    * writing valid permission rules
    * writing team oriented permission rules

### Using Claude Web

After battling Cursor a bit I decided to try Claude web instead. Also instead of trying to one-shot the blog. I made a [plan](https://github.com/nezaj/instant-llm-test/blob/main/notes/blog_plan.md) to break down the task in smaller chunks. I recorded that experience in some [notes](https://github.com/nezaj/instant-llm-test/tree/main/notes)

* I had Claude do the tasks in chunks, copying/pasting the artifacts locally.
* I asked Claude to convert our docs into a format more understandable to LLMs
* I gave Claude subset of the docs for each steps, that worked well up until permissions. It seemed like Claude needed more context about the other guides
* Claude did noticeably better once I had several guides. Using the project feature in the web and syncing Claude with the GitHub repo for the blog code with the guides seemed like it really helped

I was then able to one-shot an [Evernote-clone](https://github.com/nezaj/instant-evernote-clone) with the combined guides in Claude Web which came out ~5K lines or ~30K tokens

### Using Claude Code

To my surprise, I found that Claude Code (the cli tool) did not do as well as Claude Web. It would either spend a really long time trying to read the combined guides or it would load some of the file but not all of it. Claude Code is still a bit early so it may not perform the same as Claude Web

### Using Cursor again

I tried the combined guides with Cursor as well, when I could get it to load the full context of guides it seemed to do better, but sometimes it also would forget the context unless I explicitly reminded it. This made me think about a more robust way of testing the guides.

### Exploring Evals

I spent a day learning about evals, my hope was that we could put in a framework to programmatically measure how these prompts perform vs no prompt or just the current docs. The most common evals I found were either using strict comparison (e.g. translate a word from one language to another, it should be an exact match), fuzzy comparison (When asking an llm what’s the meaning of life it should contain the answer 42), llm-as-a-judge (use another llm to judge the output of the current llm) or introduce a human in the loop. 

* Doing fuzzy matches seemed like a bad idea for us since one character off could be the difference from good code and wrong code
* llm-as-a-judge could work but that seemed like project itself.
    * It also felt like to make llm-as-a-judge really useful we need to first come up with patterns with the hard problems (e.g. unlinking a team member from a project) so the llm knows what to judge against.
* Strict evals could work, but it feels like it would be useful for more trivial cases like correctly using `useQuery` or `transact` which didn’t seem like an issue from my app building exploration
* Human-in-the-loop could involve building some tool to compare outputs between prompts and score them, but that definitely seems like overkill. Spot checking for now seems reasonable

So in regards to an eval framework, it could be nice to have but probably overkill for now

### Building Slack

I tried Cursor and Claude Code again but this time I gave it a plan with the context. The goal was to make a slack clone. I used Claude Web to generate a plan for me this time which I honestly don’t think was as good as the hand crafted blog plan. Cursor did better than Claude Code but getting profiles and workspaces to sing was a real chore. I wrote up a [friction document](https://www.dropbox.com/scl/fi/6my3lp32fxxgs6je499y6/Slack-Clone-Friction-Log.paper?rlkey=tjwy1yfugf4vsa5dmv1sogfcc&dl=0)

### Refining the guides

I went through the guides again and cleaned up a few things. I also tried to cut down some fluff. I was able to use these to get close to one-shotting a calorie tracker app with Claude Code. The fully refined guides are [here](https://github.com/nezaj/instant-llm-guides)

### Switching to generating `llms.txt` and `llms-full.txt`

I didn’t like having two types of docs (ones for users and ones for LLMs). It seems better to maintain one source of docs. I noticed also how Anthropic themselves have their `llms.txt` and `llms-full.txt` be based on their docs.

I'm not sure if the guides are actually much better than the docs, but one thing that does seem helpful is providing examples of common mistakes and corrections.

### Putting it all together

I settled on 1) generating our `llms.txt` and `llms-full.txt` based on our docs and 2) extracting the common mistakes from the guides and adding that as part of the gen process.


client/www/pages/docs/common-mistakes.md
```diff
@@ -0,0 +1,706 @@
...
+```
+// ❌ Bad: data.ref always returns a CEL list. This will throw an error!
+"view": "auth.id == data.ref('owner.id')"
+```
+
+✅ **Correction**: Extract the first element from the list when using `==`
+
+```
+// ✅ Good: Extracting the first element from a one-to-one relationship
+"view": "auth.id == data.ref('owner.id')[0]"
```
dwwoelfel: Can we still point people to use `in` here? 

The cel -> where clauses translator doesn't know how to translate `data.ref('owner.id')[0]`, but it can handle the `in` case.

client/www/pages/docs/common-mistakes.md
```diff
@@ -0,0 +1,706 @@
...
+```
+
+✅ **Correction**: Correct ways to check for an empty list
+
+```
+// ✅ Good: Extracting the first element from a CEL list to check if it's empty
+"view": "data.ref('owner.id')[0] != null"
+
+// ✅ Good: Checking if the list is empty
+"view": "data.ref('owner.id') != []"
```
dwwoelfel, line 161: Similarly with checking for empty lists--this one is the only one that could potentially be translated (doesn't quite work yet).

client/www/pages/docs/common-mistakes.md
```diff
@@ -0,0 +1,706 @@
...
+```
+// ❌ Bad: auth.ref returns a list! This will throw an error!
+"create": "auth.ref('$user.role.type') == 'admin'"
+```
+
+✅ **Correction**: Extract the first element from `auth.ref`
+
+```
+// ✅ Good: Extracting the first element from auth.ref
+"create": "auth.ref('$user.role.type')[0] == 'admin'"
```
dwwoelfel, line 205: This one will actually translate from cel -> where clauses, because we evaluate this part of the cel.

dwwoelfel (approved):
Couple of comments, but LGTM!

Should we add llms.txt to the sitemap?

client/www/scripts/gen-llms-txt.ts
```diff
@@ -0,0 +1,246 @@
...
+}
+
+function getDocumentUrl(href: string): string {
+  return `https://instantdb.com${href}`;
+}
+
+async function processMarkdownFile(filePath: string): Promise<Document | null> {
+  try {
+    const content = await fs.readFile(filePath, 'utf-8');
+    const { frontmatter } = parseFrontmatter(content);
```
stopachka, line 117: If you end up needing to convert the markdown. (For example, to change the {% callout %} calls to some other thing, you can also use `Markdoc.parse`, and transform the ast. I think with that, you may be able to get frontmatter info too, without a regex.

(You can check out some inspo for this in index-docs.ts)

client/www/pages/docs/common-mistakes.md
nezaj: Updated!

client/www/pages/docs/common-mistakes.md
nezaj: Updated!

client/www/pages/docs/common-mistakes.md
nezaj: Updated!

client/www/scripts/gen-llms-txt.ts
nezaj: I tried using `Markdoc.parse` but was a bit hairy. I was able to do it easily though with a regex so pushing that up now

nezaj:
@dwwoelfel applied example changes, added to `sitemap.xml` and removed the non-standard tags! https://g.co/gemini/share/7c36b808fd7e

PR #1067, opened 2025-04-01
Separate sandbox value per app

dwwoelfel:
Creates a separate local store for each sandbox so that you don't lose your work when switching between apps.


stopachka (approved):
Aah so nice!

PR #1069, opened 2025-04-02
Couple of cel improvements

dwwoelfel:
Two changes in this PR:

1. Use the attrs to validate that the ref path from `data.ref` is a valid path in the cel->where-clauses compiler
2. Support `data.ref('a.id') == []` and `size(data.ref('a.id')) == 0`. We convert those to `{a: {$isNull: true}}`.
    - This allows us to support all of the rules for one user that is using this pattern


nezaj (approved):
Nice!

PR #1071, opened 2025-04-02
[Format] Run prettier

nezaj:
I've been using this command for fixing lint issues

```
prettier --check --ignore-path ../.gitignore --config --write ./.prettierrc "**/*.{ts,tsx,js,jsx,json,md}"
```

This often picks up issues we have in other files that I guess we haven't run prettier on? Thought it would be nice to run format those.


nezaj:
Realized was running prettier from the wrong directory!

PR #1072, opened 2025-04-02
improve db.signOut

stopachka:
**client**

We now accept a `db.signOut({ invalidateToken })` flag. This is useful for when folks want to integrate with other auth providers, and want to finely control the lifetime of the refresh token

**server** 

We previously only accepted signing out by email. We now accept `{ email }` `{ id }` and `{ refresh_token }` 

@dwwoelfel @nezaj @tonsky

### DIFF
--- a/client/packages/admin/src/index.ts
+++ b/client/packages/admin/src/index.ts
@@ -419,6 +419,9 @@ class Auth {
       | { id: string }
       | { refresh_token: string },
   ): Promise<void> {
+    // If input is a string, we assume it's an email.
+    // This is because of backwards compatibility: we used to only
+    // accept email strings. Eventually we can remove this
     const params = typeof input === 'string' ? { email: input } : input;
     const config = this.config;
     await jsonFetch(`${config.apiURI}/admin/sign_out`, {


client/packages/core/src/Reactor.js
```diff
@@ -1509,18 +1509,31 @@ export default class Reactor {
...
-      } catch (e) {}
+    if (!refreshToken) {
+      return;
+    }
+    const wantsToSkip = opts.invalidateToken === false;
+    if (wantsToSkip) {
+      log.info('[auth-invalidate] skipped invalidateToken');
+      return;
     }
+    authAPI
```
stopachka, line 1522: Removed the await here -- my reasoning was that we may not want this to block the signout speed

dwwoelfel: I think we should block here because the request might fail, in which case you should try again.

I guess we just ignore the error currently, but that seems like a bad idea to me.

client/packages/admin/src/index.ts
```diff
@@ -387,20 +387,44 @@ class Auth {
...
+  async signOut(email: string): Promise<void>;
+
+  async signOut(
+    input:
+      | string
+      | { email: string }
+      | { id: string }
+      | { refresh_token: string },
+  ): Promise<void> {
+    const params = typeof input === 'string' ? { email: input } : input;
```
nezaj, line 425: Nit. might be nice to add a note this is for backwards compat

nezaj (approved):
LGTM aside from one nit

dwwoelfel (approved):
One comment, but LGTM!

client/packages/admin/src/index.ts
stopachka: done!

client/packages/core/src/Reactor.js
stopachka: (Discussed offline. We can't quite do this, as it would block offline sign out)

daveycodez:
Looks great! Can we also get createTokenById in addition to createTokenByEmail ?

PR #1075, opened 2025-04-03
Support required params: migrations

tonsky:
- [x] a fn to check if attr is effectively required
- [ ] a fn to mark attr as required
- [ ] a fn to mark attr as not required

### DIFF
--- a/server/src/instant/db/model/attr.clj
+++ b/server/src/instant/db/model/attr.clj
@@ -735,6 +735,50 @@
              (seek-by-rev-ident-name n wrapped-attrs)))))
 
 ;; ------
+;; required?
+
+(defn check-required
+  "Checks if attribute is effectiveliy required (all entities with the same etype have it set).
+   Returns map with :count and first 10 :entity-ids that miss that attribute"
+  [conn attr-id]
+  (let [query "WITH inputs_cte AS (
+                 SELECT idents.app_id, idents.etype
+                   FROM attrs
+                   JOIN idents on attrs.forward_ident = idents.id
+                  WHERE attrs.id = ?attr-id
+                  LIMIT 1
+               ),
+               attr_ids_cte AS (
+                 SELECT attr_id
+                   FROM inputs_cte, idents
+                   JOIN attrs ON idents.id = attrs.forward_ident
+                  WHERE idents.etype = inputs_cte.etype
+               ),
+               triples_cte AS (
+                 SELECT DISTINCT entity_id
+                   FROM inputs_cte, triples
+                  WHERE triples.attr_id IN (SELECT attr_id FROM attr_ids_cte)
+                    AND triples.app_id = inputs_cte.app_id
+               )
+               SELECT triples_cte.entity_id, count(triples_cte.entity_id) OVER ()
+                 FROM triples_cte
+                 LEFT JOIN triples ON triples_cte.entity_id = triples.entity_id
+                                  AND triples.attr_id = ?attr-id
+                WHERE triples.ctid IS NULL
+                LIMIT 10"
+        params {"?attr-id" attr-id}
+        res    (sql/execute! conn (sql/format query params))
+        count  (:count (first res))
+        eids   (map :entity_id res)]
+    (when count
+      {:count      count
+       :entity-ids eids})))
+
+(comment
+  (check-required (aurora/conn-pool :read) #uuid "6eebf15a-ed3c-4442-8869-a44a7c85a1be")
+  (check-required (aurora/conn-pool :read) #uuid "275e5c3c-d565-4f5a-b832-4290cc6de915"))
+
+;; ------
 ;; play
 
 (comment
--- a/server/src/instant/jdbc/sql.clj
+++ b/server/src/instant/jdbc/sql.clj
@@ -1,4 +1,5 @@
 (ns instant.jdbc.sql
+  (:refer-clojure :exclude [format])
   (:require
    [clojure.string :as string]
    ;; load all pg-ops for hsql
@@ -27,13 +28,13 @@
 (defn ->pg-text-array
   "Formats as text[] in pg, i.e. {item-1, item-2, item3}"
   [col]
-  (format
+  (clojure.core/format
    "{%s}"
    (string/join
     ","
-    (map (fn [s] (format "\"%s\""
-                         ;; Escape quotes (but don't double esc)
-                         (string/replace s #"(?<!\\)\"" "\\\"")))
+    (map (fn [s] (clojure.core/format "\"%s\""
+                                      ;; Escape quotes (but don't double esc)
+                                      (string/replace s #"(?<!\\)\"" "\\\"")))
          col))))
 
 (defn ->pg-uuid-array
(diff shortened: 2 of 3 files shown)


tonsky:
Progress update:

- [x] indexing job
- [x] Edit attr UI
- [x] Validation error & mesasge
- [ ] Add attr UI
- [ ] Validation when adding/editing record without required attr

Screenshot 2025-04-11 at 21 39 03

Screenshot 2025-04-11 at 21 22 48

tonsky:
Progress:

- [x] Validate `:add-attr`
- [x] Validate `:update-attr`
- [x] Validation when adding/editing record without required attr
- [x] Add blob attr UI
- [ ] Add ref attr UI

tonsky:
Some preliminary testing on Aura app (12M triples):

| fn | time |
|------------------------------------|--------|
| `indexing-jobs/required--validate` | ~6 sec |
| `attr/validate-add-required!`      | ~0.1 sec |
| `attr/validate-update-required!`   | ~6 sec |
| `triple/validate-required!`        | ~2 sec |

Should be mostly fine, except for the last one, which bacisally adds extra 2 seconds to every update or insert to triples. I’ll see what can be done, ideally it should go down to <0.1 sec

client/Makefile
```diff
@@ -5,6 +5,12 @@ dev:
 	@echo "Booting up frontend..."
 	pnpm run dev
 
+lint:
+	pnpm run check-format
+
+lint-fix:
```
stopachka, line 11: nice!

server/src/instant/db/model/attr.clj
```diff
@@ -257,6 +257,25 @@
...
+  (doseq [attr  attrs
+          :when (:required? attr)
+          :let  [[_ etype label] (:forward-identity attr)
+                 query "SELECT *
+                          FROM idents
+                          JOIN attrs ON attrs.forward_ident = idents.id
+                          JOIN triples ON triples.attr_id = attrs.id
+                         WHERE idents.etype = ?etype
+                           AND idents.app_id = ?app-id
+                           AND triples.app_id = ?app-id
```
stopachka, line 270: You may be able to make this faster by writing `AND ea` (as all of these triples will live on that index)

server/src/instant/db/model/attr.clj
```diff
@@ -257,6 +257,25 @@
...
+                         LIMIT 1"
+                 res   (sql/execute! conn (sql/format query {"?etype" etype, "?app-id" app-id}))]
+          :when (seq res)]
+    (ex/throw-validation-err!
+     :attributes
+     attr
+     [{:message (str "Can't set attribute `" label "` as required because `" etype "` already have entities")}])))
```
stopachka: nit: perhaps we say something like 'can't create new attribute ` label

(set could imply this is an update, when really we are creating)

server/src/instant/db/model/attr.clj
```diff
@@ -451,6 +471,58 @@
...
+                 ) AS etype_cnt"
+          params {"?attr-ids" (mapv :id attrs)
+                  "?app-id"   app-id}
+          res    (sql/execute! conn (sql/format query params))]
+      (doseq [{:keys [id etype label attr_count etype_count]} res
+              :when (not= attr_count etype_count)]
+        (ex/throw-validation-err!
+         :attributes
+         {:id id}
+         [{:message (str "Can't set attribute `" label "` as required because `" etype "` already have entities without it")}])))))
```
stopachka: nit: perhaps here we can use the word `update`

server/src/instant/db/model/attr.clj
```diff
@@ -451,6 +471,58 @@
...
+                     JOIN attrs ON attrs_cte.id = attrs.id
+                     JOIN idents ON attrs.forward_ident = idents.id
+                 )
+
+                 SELECT attrs_etype_cte.*,
+                        attr_cnt.cnt AS attr_count,
+                        etype_cnt.cnt AS etype_count
+                 FROM attrs_etype_cte,
+
+                 LATERAL (
```
stopachka, line 493: some serious sql going on here 🫡

server/src/instant/jdbc/sql.clj
```diff
@@ -203,20 +204,38 @@
...
                col-name))
-   :from   [[[:json_to_recordset [:cast (->json rs) :json]]
+   :from   [[[:jsonb_to_recordset [:cast (->json rs) :jsonb]]
              [[:raw (str "("
                          (string/join ", "
                                       (for [[col-name {:keys [type]}] cols]
                                         (str (name col-name) " " (name (or type "text")))))
                          ")")]]]]})
 
+(defn format
```
stopachka, line 221: oh cool!

stopachka (approved):
Beautiful work @tonsky !

server/src/instant/db/model/triple.clj
```diff
@@ -175,17 +176,20 @@
                          [:and
                           [:= 'eid+required-attrs/entity-id 'triples/entity-id]
```
stopachka, line 177: perhaps ea index hint could help with the speed here

stopachka (approved):

server/src/instant/db/transaction.clj
```diff
@@ -356,6 +356,8 @@
                            (:deep-merge-triple results)
                            (:retract-triple results)))
             _  (triple-model/validate-required! conn attrs app-id eid+attr-ids)
```
stopachka, line 358: I do worry about this. I am not 100% sure how our average perf looks for transactions (A quick look shows P50 at around [20ms](https://ui.honeycomb.io/instantdb/environments/prod/datasets/instant-server/result/m9ZniL9kJn4)), so even 100ms extra would be meaningful increase in latency. 

Some ideas I am thinking could work too: 

1. **Deferred trigger in postgres.** We could add a 'deferred' that checks all inserted and deleted triples at once. This would remove the extra round-trip. Plus if multiple insert-multi happens for a triple, this could help batch it all together. I am not sure how much more that will help though 
1. **Some auxillary table:** If we can't get the query to be efficient enough, I wonder if we can keep track of the existence of an entity, and it's required columns, in some other table. This would be updated with a trigger on triples. It could have it's own laws about when it throws. This is a bit vague though; I am not sure what it would look like
1. **Doing it in clojure alone:** This would be the fastest option. But as you mentioned, it can get hairy, and it won't guarantee things at the db level. 

Admittedly these are all hazy ideas. Let me know how the perf investigation goes -- happy to riff together too

server/src/instant/db/model/triple.clj
```diff
@@ -164,7 +165,7 @@
             {:from 'eid+etypes-cte
              :join ['idents [:and
                              [:= 'idents/app-id app-id]
-                             [:= 'idents/etype 'idents/etype]]
+                             [:= 'idents/etype 'eid+etypes-cte/etype]]
                     'attrs [:= 'idents/id 'attrs/forward-ident]]
              :where [:= 'attrs/is_required true]
```
stopachka, line 170: Perhaps we can add an index to `is_required`.

tonsky: By this point we already narrowed down attributes to specific etype, so we are dealing with 10-ish attributes or so? I don’t think index here will help with anything

server/src/instant/db/model/attr.clj
tonsky: Not necessarily. What I’m checking here is that type has at least 1 entity (thus `LIMIT 1`). But if it only has entity with cardinality many it won’t be on `ea`?

Also, because of LIMIT 1, we don’t really need index here, it’s a simple existence check

server/src/instant/db/model/triple.clj
stopachka: Makes sense 🫡

server/src/instant/db/model/attr.clj
stopachka: Ah, I originally assumed required would only work on attributes, but I realize it can work for links too! Good point. 

Perhaps based on the attr we could infer if it's cardinality-one, and use ea, or value-type ref, and use eav. I am not sure if it would help though.

server/src/instant/db/transaction.clj
tonsky: Okay, I figured it out. It was missing `app-id = '...'` when querying triples, I guess without it indexes couldn’t kick in because app-id is part of all of them.

Re: speed, I’m now getting consistent 110-120ms, most of which seem to be my own latency, because I get the same numbers on simple queries to prod, too. So we should be good here

stopachka: Niice! Okay, let's shiip!

PR #1076, opened 2025-04-03
skip null check for `$not` with indexed attrs

dwwoelfel:
When we do `$not`, we also have to check that the attr is not undefined since there might be entities that are missing a triple for the attr. But with indexed attrs, we already insert a `null` for every triple so we can skip that step.

This PR skips adding the `$isNull true` where cond when we know we're checking not on an indexed attr.

### DIFF
--- a/server/deps.edn
+++ b/server/deps.edn
@@ -44,7 +44,7 @@
 
         nrepl/nrepl {:mvn/version "0.9.0"}
         jarohen/chime {:mvn/version "0.3.3"}
-        com.github.seancorfield/honeysql {:mvn/version "2.6.1270"}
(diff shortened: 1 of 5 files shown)


server/deps.edn
```diff
@@ -44,7 +44,7 @@
 
         nrepl/nrepl {:mvn/version "0.9.0"}
         jarohen/chime {:mvn/version "0.3.3"}
-        com.github.seancorfield/honeysql {:mvn/version "2.6.1270"}
+        com.github.seancorfield/honeysql {:mvn/version "2.7.1295"}
```
dwwoelfel, line 47: Update honeysql for `is distinct from` support.

stopachka (approved):
Heeck yeah!

PR #1077, opened 2025-04-03
[sendgrid] back to postmark

stopachka:
Now that postmark is back, let's standardize back to it. I'll keep sendgrid around, as an emergency backup. 

@nezaj @dwwoelfel @tonsky


dwwoelfel (approved):
LGTM!

PR #1078, opened 2025-04-03
[RN] Export InstantReactNativeDatabase

nezaj:
We got a request to export this so adding it here.

Looks like we export `InstantReactWebDatabase` in the react package too


dwwoelfel (approved):
LGTM!

PR #1081, opened 2025-04-04
Add pg statistics that help pg choose better indexes

dwwoelfel:
Adds most common value statistics for attr_id, ave, checked_data_type, and value, which help postgres to choose a better join strategy for nested loops.

https://www.postgresql.org/docs/current/planner-stats.html#PLANNER-STATS-EXTENDED-MCV-LISTS

I chose 2500 as the `set statistics` value because it allows the db to finish analyze in under 30 seconds.

### DIFF
--- a/server/Makefile
+++ b/server/Makefile
@@ -101,6 +101,12 @@ psql-dump:
 	echo "Exporting prod db..." && \
 	pg_dump "$${DATABASE_URL}" -f dump.sql
 
+psql-dump-dir:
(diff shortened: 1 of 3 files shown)


nezaj (approved):
Nice!

server/Makefile
```diff
@@ -101,6 +101,12 @@ psql-dump:
 	echo "Exporting prod db..." && \
 	pg_dump "$${DATABASE_URL}" -f dump.sql
 
+psql-dump-dir:
```
stopachka, line 104: niice!

PR #1082, opened 2025-04-04
[Docs] Add example prompt

nezaj:
I remember @stopachka mentioned an example would be nice so thought I'd add one in here


stopachka (approved):
Love it!

PR #1083, opened 2025-04-04
[emails] custom domain for dash magic codes

stopachka:
Set a custom domain for when we send magic codes for dashboard. 

@dwwoelfel @nezaj @tonsky


dwwoelfel (approved):
Cool!

PR #1084, opened 2025-04-04
[dash] make cmd+click work on tabs

stopachka:
A user pointed out that it's a bit annoying you can't cmd + click on our dash menu items. 

I went ahead and upgraded how we use `ToggleCollection` and `TabBar`, so that these are treated as links. 

@nezaj @dwwoelfel @tonsky 

**Note: we _may_ want to do this for our selector, as well as our explorer tabs too. But first, wanted to ship what users actually asked for**

### DIFF
--- a/client/www/pages/dash/index.tsx
+++ b/client/www/pages/dash/index.tsx
@@ -298,10 +298,17 @@ function Dashboard() {
         return {
           id: t.id,
           label: t.title,
-          link: app ? `/docs?app=${app.id}` : '/docs',
+          link: {
+            href: app ? `/docs?app=${app.id}` : '/docs',
+            target: '_blank',
+          },
         };
       }
-      return { id: t.id, label: t.title };
+      return {
+        id: t.id,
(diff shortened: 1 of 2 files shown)


client/www/pages/dash/index.tsx
```diff
@@ -291,17 +290,24 @@ function Dashboard() {
...
-          link: app ? `/docs?app=${app.id}` : '/docs',
+          link: {
+            href: app ? `/docs?app=${app.id}` : '/docs',
+            target: '_blank',
+          },
         };
       }
-      return { id: t.id, label: t.title };
+      return {
+        id: t.id,
```
stopachka, line 307: note: this _does_ mean that for normal tabs, we no longer call the `nav` function. 

But the only extra thing the nav function did was to save the app id locally. This is not needed for the menu tabs.

dwwoelfel (approved):
This is so much nicer! LGTM!

PR #1085, opened 2025-04-04
[google] skip nonce checks

stopachka:
- in IOS, you can't always provide a `nonce` check. At least not with `react-native-google-signin/google-signin`, the currently popular google sign in module. 
- I went ahead added an option to disable the nonce check. This lets you sign in with Google. 

Coming up, I'll: 
1. Update the UI, to perhaps make it a bit more nicer to flow (ios vs web, etc) 
2. Update the docs 
3. Mini announcement 

@dwwoelfel @nezaj @tonsky


dwwoelfel (approved):
LGTM!

PR #1088, opened 2025-04-05
Parse RFC 1123 date times

dwwoelfel:
Parses dates in RFC 1123 / RFC 822 format, commonly used in headers.

They look like this `Tue, 3 Jun 2008 11:05:30 GMT`


nezaj (approved):
Heck yeah!

PR #1089, opened 2025-04-05
[dev] Make it easier to view sql queries in portal

dwwoelfel:
Adds a custom viewer for sql queries and a command to copy them to the clipboard in portal.

There's a line in sql.clj you can uncomment to send the queries to portal.

See it in action:

https://github.com/user-attachments/assets/be4d936a-f0c6-4e84-a249-80f267dea366


nezaj (approved):
Neat!

PR #1090, opened 2025-04-06
Small cel refactor

dwwoelfel:
Refactors the cel code a bit in preparation for changing how we do prefetching.

Main changes:

1.  We no longer expect callers to construct `CelMap`s. Now that is all internal to `cel.clj`.

    **Before**

    ```clojure
    (cel/eval-program!
      program
      {"auth"       (cel/->cel-map {:ctx ctx, :type :auth, :etype "$users"} current-user)
       "data"       (cel/->cel-map {:ctx ctx, :type :data, :etype etype} updated)
       "newData"    (cel/->cel-map {} updated)
       "ruleParams" (cel/->cel-map {} rule-params)})
    ```
    **After**

    ```clojure
    (cel/eval-program!
      ctx
      program
      ;; Don't have to provide auth, cel.clj will pull it from `ctx` for you
      {:data updated
       :new-data updated
       :rule-params rule-params})
    ```

2.  Introduces a new `severed-vfuture` to fix a bug in dataloader where the dataload might get canceled if a parent `vfuture` is canceled.

### A note about upcoming prefetching change

Right now our prefetching works like this: We walk the cel ast and find any `data.ref` and `auth.ref` calls, collecting the `path-str` from each call. Before running the rules, we fetch all of the `ref` calls in a single big query. Then the `ref` function looks up the prefetched data when the rule executes.

I want to change the prefetching to use cel's unknown tracking feature. The new strategy works like this: First we run the rules for all of the entities with unknown tracking enabled and return `unknown` from any `data.ref` or `auth.ref` calls. Then we fetch all of the data in one big query. Then we run the rules through another pass and this time the ref function will find the data we fetched. Usually it will just take one pass, but if not we'll keep making passes until all of the data is resolved.

The new approach has a couple of benefits:

1. We will fetch less data that we don't need. If you have a rule that would short-circuit before getting to a `data.ref`, right now we'll still fetch that data even though we won't use it. If you had `true || data.ref('owner.id') == []`, then with the unknown tracking enabled, cel will just return true and we won't have to fetch anything. 
1. It works better with the rule-wheres
  i. We won't have to special-case the prefetching to just fetch auth data--we can use the same functions to prefetch for both paths
  ii. It makes it easier to re-use the prefetched data if we still need to run the rules on the entities we get back

### DIFF
--- a/server/src/instant/db/dataloader.clj
+++ b/server/src/instant/db/dataloader.clj
@@ -25,9 +25,13 @@
                  {:requests (conj requests request)
                   :schedule-delay (or schedule-delay
                                       (delay
-                                        (ua/vfuture
-                                         (Thread/sleep (long delay-ms))
-                                         (run-batch! dataloader-opts k))))}))
+                                        ;; store/swap-datalog-cache! has a future
+                                        ;; that will be canceled if all of its "watchers"
+                                        ;; are canceled. Something like that would be a
+                                        ;; nice improvement over severed-vfuture here.
+                                        (ua/severed-vfuture
+                                          (Thread/sleep (long delay-ms))
+                                          (run-batch! dataloader-opts k))))}))
         schedule-delay (get-in v [k :schedule-delay])]
     @schedule-delay))
 
--- a/server/src/instant/db/permissioned_transaction.clj
+++ b/server/src/instant/db/permissioned_transaction.clj
@@ -83,7 +83,7 @@
    tx-steps))
 
 (defn object-upsert-check-fn [{:keys [action program etype eid data] :as _check}
-                              {:keys [current-user rule-params] :as ctx}]
+                              {:keys [rule-params] :as ctx}]
   (let [{:keys [original updated]} data
         rule-params (get rule-params {:eid eid :etype etype})]
     (cond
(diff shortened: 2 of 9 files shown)


stopachka (commented):
Biiyutiful!

server/src/instant/db/dataloader.clj
```diff
@@ -25,9 +25,13 @@
...
                   :schedule-delay (or schedule-delay
                                       (delay
-                                        (ua/vfuture
-                                         (Thread/sleep (long delay-ms))
-                                         (run-batch! dataloader-opts k))))}))
+                                        ;; store/swap-datalog-cache! has a future
+                                        ;; that will be canceled if all of its "watchers"
+                                        ;; are canceled. Something like that would be a
+                                        ;; nice improvement over severed-vfuture here.
+                                        (ua/severed-vfuture
```
stopachka, line 32: 🤩

server/src/instant/db/permissioned_transaction.clj
```diff
@@ -83,53 +83,49 @@
...
       (= :create action)
-      (cel/eval-program!
-       program
-       {"auth"       (cel/->cel-map {:ctx ctx, :type :auth, :etype "$users"} current-user)
-        "data"       (cel/->cel-map {:ctx ctx, :type :data, :etype etype} updated)
-        "newData"    (cel/->cel-map {} updated)
-        "ruleParams" (cel/->cel-map {} rule-params)})
+      (cel/eval-program! ctx
+                         program
+                         {:data updated
```
stopachka, line 96: 🔥

stopachka (approved):
Exciting!

PR #1091, opened 2025-04-07
[docs] pattern for composite keys

stopachka:
A user asked us about supporting composite keys. We don't support composite keys as first-class options, but you can use columns to make them work. I wanted to add a pattern showing how to do this. 

@nezaj @dwwoelfel @tonsky


dwwoelfel (approved):
LGTM!

PR #1092, opened 2025-04-07
[do-not-review] import test data as a migration

stopachka:
## Context

When we run `make docker-compose`. `Dockerfile-dev` runs: 

```dockerfile
    RUN clojure -P -M:dev
    RUN clojure -P -M:oss-bootstrap
```

And `oss-bootstrap` runs: 

```clojure
    (defn migrate-database []
      (config/init)
      (let [database-url (-> (config/get-aurora-config)
                             (jdbc-url)
                             (jdbc-url->postgres-url {:sslmode "disable"}))]
        (process/exec "migrate"
                      "-database" database-url
                      "-path" "resources/migrations"
                      "up")
        (process/exec
         {:env {"DATABASE_URL" database-url}}
         (or (System/getenv "SHELL") "bash")
         "dev-resources/import_test_data.sh")))
```

## Problems

Problem 1
    server-1    | dev-resources/import_test_data.sh: line 7: psql: command not found

`psql` is not defined in the server dockerfile. 

Problem 2

Even if we did install it. Then we’ll get an error, as import_test_data will run every time the user writes `make docker-compose`, and unique constraints will throw errors

## Solution

I moved the script into a migration instead. This way, the script only ever runs once. 

@nezaj @dwwoelfel @tonsky


stopachka:
#1094 is a better solution

PR #1093, opened 2025-04-07
Use cel's unknown tracking to batch ref data prefetching

dwwoelfel:
Replaces our current prefetching that parses the ast with a more targeted system that uses cel's unknown tracking.

When we encounter a rule like `auth.id in data.ref('owner.id')`, we have to fetch the list of owners for the entity to resolve the rule.

The existing method parses the ast and pulls out the ref strings (`owner.id` in this case) then we send a big query to the database that gets the refs for all of the entities at once. Once we have all of the ref data, we execute the rules and our `ref` function looks them up in the cache.

The new method uses unknown tracking in cel to determine which refs we need to fetch. We run the rules for each entity once and return a `CelUnknownSet` from our `ref` function. We run the rule for each entity and collect all of the missing refs. Then we fetch the refs in one big query and run the rules again. 

This will reduce some unnecessary ref fetches. For example, if you're not logged in then this rule won't do any extra fetching: `auth.id != null && auth.id in data.ref('owner.id')`. cel will evaluate that to `false && unknown`. Since no value for unknown can possibly make that true, cel can just return `false.`

This is also easier to make work with the new cel-rules->iql-where-clauses pipeline.


stopachka (approved):
So frickin cool!

PR #1094, opened 2025-04-07
[oss] import test user in migration

stopachka:
## Context

When we run `make docker-compose`. `Dockerfile-dev` runs: 

```dockerfile
    RUN clojure -P -M:dev
    RUN clojure -P -M:oss-bootstrap
```

And `oss-bootstrap` runs: 

```clojure
    (defn migrate-database []
      (config/init)
      (let [database-url (-> (config/get-aurora-config)
                             (jdbc-url)
                             (jdbc-url->postgres-url {:sslmode "disable"}))]
        (process/exec "migrate"
                      "-database" database-url
                      "-path" "resources/migrations"
                      "up")
        (process/exec
         {:env {"DATABASE_URL" database-url}}
         (or (System/getenv "SHELL") "bash")
         "dev-resources/import_test_data.sh")))
```

## Problems

Problem 1
    server-1    | dev-resources/import_test_data.sh: line 7: psql: command not found

`psql` is not defined in the server dockerfile. 

Problem 2

Even if we did install it. Then we’ll get an error, as import_test_data will run every time the user writes `make docker-compose`, and unique constraints will throw errors

## Solution

I added a migration which adds the users that are required for tests. I deleted the import script. We no longer need all the `app` data, as the tests create apps on the fly. 

@nezaj @dwwoelfel @tonsky


dwwoelfel (approved):
LGTM!

Do you want to delete all of the `csv`s in dev-resource too?

stopachka:
deleted csvs too!

verveguy:
Nice!!!

PR #1097, opened 2025-04-07
[clean] replace calls to zeneca-app-id, movies-app-id

stopachka:
We no longer have a zeneca-app and movies-app built into the db. This makes some of our comment blocks out of date. 

What I did:
- Added a new `instant.comment` ns, which has a handy `zeneca-app!` function
  - This is an ns that doesn't require anything (so can be imported anywhere). It assumes that it will always be called inside comment blocks
  - I use ephemeral-apps to create zeneca apps on the fly for comment blocks
- I deleted some older comment blocks that I doubt we'd need. 

@dwwoelfel @nezaj @tonsky


server/src/instant/model/ephemeral_app.clj
```diff
@@ -0,0 +1 @@
+()
```
dwwoelfel: What's this?

dwwoelfel (approved):
LGTM!

server/src/instant/model/ephemeral_app.clj
stopachka: oops!

PR #1100, opened 2025-04-09
Rule wheres with refs

dwwoelfel:
Updates rule wheres to work with refs and creates a testing pipeline where we can find bugs before rolling it out to everyone.

Two main changes to how cel->rule-wheres work in this PR:

1. We notice when you reference a link in your query and we add the wheres for that link.
    Example:
    Original query:
    ```clojure
    {:users {:$ {:where {:books.title "Land of Lisp"}}}}
    ```
    Rules:
    ```json
    {"books": {"allow": {"view": "data.hidden == false"}}}
    ```
    
    Query with rules applied:
    
    ```clojure
    {:users {$: {:where {:books.title "Land of Lisp", :books.hidden false}}}}
    ```
    
    I also changed where we apply the rules--now we apply it to the actual query before passing the forms to `query` instead of adding the `where-conds` inside of `query`. This will make it easier for us to see the actual query that was run in honeycomb because it will show up in the `forms` attribute for `instaql/query`.
    
2. We'll execute the rule code for any entities brought in by links in the rules.
   For example, if you have a rule on posts like `auth.id in data.ref('owner.id')`, we'll add `{:owner.id auth.id}` to the query and we might pull in a user in the join rows. In order to make sure we filter those properly, we'll still run the rules for the user. We accomplish that by marking which entities are checked by the rule-wheres when we construct the permission helpers after running the query. Then we run the rules for any entity that is not marked as checked.
   
  My first approach was to apply the rules recursively (so if posts references owners, all of the owner wheres get added, and if owner references team, then it also adds those, etc.), but it exploded pretty quickly and I ran into a postgres limitation where we created too many columns.
   
I added a testing pipeline that will put every query with rules into a sliding-buffer queue and will run the query with and without rule-wheres. It checks how long each method takes to complete and if the results match. Then we can look in the logs to find any queries that didn't match and get an idea of how it will affect query times.

The testing queue just runs one at a time, so it shouldn't add too much load to the system.

### DIFF
--- a/server/src/instant/model/rule.clj
+++ b/server/src/instant/model/rule.clj
@@ -128,7 +128,7 @@
     (let [compiler (cel/action->compiler action)]
       (if (and (= "$users" etype)
                (= "view" action))
-        (let [code "auth.id != null && auth.id == data.id"
+        (let [code "auth.id == data.id"
               ast (cel/->ast compiler code)]
           {:etype etype
            :action action


server/src/instant/db/instaql.clj
```diff
@@ -1967,53 +1977,163 @@
...
+        rule-wheres))))
+
+(defn update-where-with-rule-refs [initial-etype attrs rule-wheres where k v]
+  (loop [[segment & rest-path] (string/split (name k) #"\.")
+         current-path [segment]
+         etype initial-etype
+         where (assoc where k v)]
+    (if-let [segment-etype (link-etype attrs etype segment)]
+      (let [rule-where (if (-> rule-wheres (get segment-etype) :short-circuit?)
+                         {:id (str (random-uuid))}
```
dwwoelfel, line 2037: This is kind of a hack--we know that the rule should always return false, so we add a where clause that we know won't match (id == random-uuid). We might want to add some kind of special `never` operator instead, which could save us some querying.

server/src/instant/db/rule_where_testing.clj
```diff
@@ -0,0 +1,80 @@
...
+        (let [ctx (assoc ctx :datalog-query-fn d/query)
+              without-rule-wheres (future
+                                    (tracer/with-span! {:name "test-rule-wheres/without-rule-wheres"}
+                                      (run-test (assoc ctx :use-rule-wheres? false)
+                                                permissioned-query-fn o)))
+              with-rule-wheres (future
+                                 (tracer/with-span! {:name "test-rule-wheres/with-rule-wheres"}
+                                   (run-test (assoc ctx :use-rule-wheres? true)
+                                             permissioned-query-fn o)))
+              attrs {:without.ms (:ms @without-rule-wheres)
```
stopachka, line 39: Forgot you could add dots -- looks nice!

server/src/instant/db/rule_where_testing.clj
```diff
@@ -0,0 +1,80 @@
...
+    (a/put! @process-chan [ctx permissioned-query-fn o])))
+
+(defn start []
+  (reset! process-chan (a/chan (a/sliding-buffer 10)))
+  (a/go
+    (loop []
+      (when-let [args (a/<! @process-chan)]
+        (try (apply test-rule-wheres args)
+             (catch Exception e
+               (def -e e)
```
stopachka, line 71: Should we log this too?

server/src/instant/db/instaql.clj
```diff
@@ -1967,53 +1977,163 @@
...
+                                              v)))
+             {}
+             where))
+
+(defn add-rule-wheres-to-query
+  ([ctx rule-wheres o]
+   (add-rule-wheres-to-query nil ctx rule-wheres o))
+  ([etype ctx rule-wheres o]
+   (reduce-kv (fn [new-o k v]
+                (let [etype (if etype
```
stopachka, line 2078: Wanted to share a chain of thought: 

For self-referencing relationships, i.e : blocks have children blocks. 

When a query is run like: 

```js
{ blocks: { children: { } } }
```

How would the where clauses attach? 

I noticed we use the `etype` to get the where clauses. So say it had something like: 

rule: !data.parent || data.parent == "some-uuid" 

The enhanced query would be: 

```
{ blocks: { 
  $: { where: { or: [{parent: null}, parent: "some-uuid"] } }
  children: {
       $: { where: { or: [{parent: null}, parent: "some-uuid"] } }
   } 
} }
```

Which, though surprising, I _think_ is what it should it do! Cool. 

--- 

Do I understand this correctly?

server/src/instant/db/instaql.clj
```diff
@@ -1967,53 +1977,163 @@
...
+              o)))
 
 (defn permissioned-query [{:keys [app-id current-user admin?] :as ctx} o]
   (tracer/with-span! {:name "instaql/permissioned-query"
                       :attributes {:app-id app-id
                                    :current-user (pr-str current-user)
                                    :admin? admin?
                                    :query (pr-str o)}}
+    (when-not (or admin?
+                  ;; Prevent infinite cycle
```
stopachka, line 2116: Nice!

stopachka (approved):
Biyutiful!!

server/src/instant/db/instaql.clj
dwwoelfel: Yep, that looks right to me!

server/src/instant/db/rule_where_testing.clj
dwwoelfel: It shouldn't actually throw, everything that would throw is in a try/catch--just being extra cautious.

PR #1101, opened 2025-04-09
Refactor indexing_jobs

tonsky:
As a preparation for adding new job for required fields, a little refactoring on indexing_jobs.

Main motivation was to decouple scheduling/dispatching from job body itself. After this PR, job body just changes what needs to be changed, but doesn’t know what stage is next or how to handle error. This is decided outside, from data-first description instead of imperative code. Functional core, imperative shell (although our “functional” core still runs updates over SQL, I still feel like scheduling and execution should be separated)

Nothing functionally changed, just moved code around.

Also net negative on code size: 530 insertions(+), 808 deletions(-)

### DIFF
--- a/server/src/instant/db/indexing_jobs.clj
+++ b/server/src/instant/db/indexing_jobs.clj
@@ -177,16 +177,9 @@
          stage  (-> stages first :stage)]
      (assert app-id)
      (assert attr-id)
-
-     (when-not (contains? jobs job-type)
-       (ex/throw+ {::ex/type ::ex/param-malformed
-                   ::ex/mesasge (str "Unexpected job type: " job-type)
-                   ::ex/hint args}))
-
-     (when (and (= "check-data-type" job-type)
-                (nil? checked-data-type))
-       (ex/throw+ {::ex/type ::ex/param-missing
-                   ::ex/message "checked-data-type must be provided if job type is check-data-type"}))
+     (assert (contains? jobs job-type) (str "Unexpected job type: " job-type))
+     (when (= "check-data-type" job-type)
+       (assert checked-data-type))
 
      (sql/execute-one! ::create-job!
                        conn (hsql/format {:insert-into :indexing-jobs
@@ -800,10 +793,7 @@
                                     (get (:job_type job))
                                     :stages
                                     (->> (drop-while #(not= (:stage %) (:job_stage job)))))
-          _                     (when (nil? stage)
-                                  (ex/throw+ {::ex/type ::ex/param-malformed
-                                              ::ex/mesasge (str "Unexpected job stage: " (:job_stage job))
-                                              ::ex/hint {:job job}}))
+          _                     (assert stage (str "Unknown stage: " (:job_type job) " " (:job_stage job)))
           res                   ((:fn stage) conn job)
           [res-type res-value]  (when (vector? res) res)]
       (cond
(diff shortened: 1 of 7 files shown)


server/test/instant/db/indexing_jobs_test.clj
```diff
@@ -122,7 +126,7 @@
             (is (every? (fn [{:keys [checked-data-type]}]
                           (nil? checked-data-type))
                         handle-triples)))
-          (= 5 (count (jobs/invalid-triples 100 (:id handle-job))))
+          (is (= 4 (count (jobs/invalid-triples 100 (:id handle-job)))))
```
dwwoelfel, line 129: Whoops--maybe somebody will write a clj-kondo rule for that because I'm sure it's not the only place in the codebase where I forgot an `is`.

server/src/instant/db/indexing_jobs.clj
```diff
@@ -214,119 +166,49 @@
...
+                 checked-data-type]} args
+         {:keys [serial-key stages]} (get jobs job-type)
+         stage  (-> stages first :stage)]
+     (assert app-id)
+     (assert attr-id)
+
+     (when-not (contains? jobs job-type)
+       (ex/throw+ {::ex/type ::ex/param-malformed
+                   ::ex/mesasge (str "Unexpected job type: " job-type)
+                   ::ex/hint args}))
```
dwwoelfel: The `::ex/hint` might get sent to the client. There might be stuff in the args at some point that we don't want to send or that won't serialize. I would be explicit and just provide `{:job-type job-type}` as the hint.

dwwoelfel: But we should probably just do asserts both here and when we check for "check-data-type". 

I'd prefer to give the user a 500 over an `::ex/param-x` error for two reasons. 1. We're more likely to get notified on an assert and 2. if we hit the error, it will be a coding error on our part and there's nothing the user can do to fix it.

server/src/instant/db/indexing_jobs.clj
```diff
@@ -808,274 +551,276 @@
...
+
+(defn run-next-stage [conn job]
+  (tracer/with-span! (job-span-attrs "run-next-stage" job)
+    (assert (= "processing" (:job_status job)) (:job_status job))
+    (let [[stage & next-stages] (-> jobs
+                                    (get (:job_type job))
+                                    :stages
+                                    (->> (drop-while #(not= (:stage %) (:job_stage job)))))
+          _                     (when (nil? stage)
+                                  (ex/throw+ {::ex/type ::ex/param-malformed
```
dwwoelfel: I would just do an assert here

dwwoelfel (approved):
Couple of comments, but LGTM!

server/src/instant/db/indexing_jobs.clj
tonsky: But we don’t run asserts on prod. Do you think we should start?

tonsky: Damn, I was under impression Clojure’s `assert` works the same as Java’s `assert` (which are disabled by default), but apparently it’s completely different and _is_ enabled by default. Happy to use asserts then

server/test/instant/db/indexing_jobs_test.clj
tonsky: I only found 1 more, so maybe it’s not as bad as it seems. Also, this one was already detected by clj-kondo

PR #1104, opened 2025-04-09
[landing] improve og image

stopachka:
Updated our OG image to also generate a kind of "section" title: 

docs-ex
blog-ex

I realized this would be nice, when I noticed we had a `-- Instant Docs` in our og images. 

Also used these for our pricing page, etc

@nezaj @dwwoelfel @tonsky


dwwoelfel (approved):
Cool!

PR #1105, opened 2025-04-10
Bug fix for ors and and in rule-wheres

dwwoelfel:
Forgot to provide an argument.

Also removed the call to `partial`, since clj-kondo won't find invalid arity errors with `partial`.

Includes a test.


stopachka (approved):
🫡

PR #1107, opened 2025-04-10
Fix rule-where-testing compare on errors

dwwoelfel:
Two changes:

1. Fixes a bug where I tried to call `instaql-nodes->object-tree` on an Exception
2. Puts the user's where clauses first in hopes we'll get a more efficient match.


nezaj (approved):
LGTM!

PR #1108, opened 2025-04-10
Better error for record-not-unique errors

dwwoelfel:
Creates a better error for `record not unique: triple`.

**Before**

image

**After**

image

Full error:

```json
{
    "op": "error",
    "status": 400,
    "type": "record-not-unique",
    "message": "`handle` is a unique attribute on `users` and an entity already exists with `users.handle` = \"daniel\"",
    "hint": {
        "record-type": "triple",
        "attr-id": "d985fdee-8848-46d5-98fa-580625c43c3e",
        "etype": "users",
        "label": "handle",
        "value": "\"daniel\""
    }
}
```

### DIFF
--- a/server/src/instant/util/exception.clj
+++ b/server/src/instant/util/exception.clj
@@ -130,14 +130,13 @@
                  (add-char c)
                  (recur (inc open-parens)
                         in-quote?))
-            \" (do
-                 (let [escaped-quote? (= (safe-char s (dec @i))
-                                         \\)]
-                   (add-char c)
-                   (recur open-parens
-                          (if escaped-quote?
-                            in-quote?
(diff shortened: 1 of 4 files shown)


nezaj (approved):
Huge win!

server/src/instant/util/exception.clj
```diff
@@ -59,6 +60,21 @@
   (throw+ {::type ::record-not-found
            ::message "hey!"}))
 
+;; -------
+;; Helpers
+
+(defonce get-attr-details* (atom nil))
+
+(defn define-get-attr-details
```
stopachka, line 68: Monster!

PR #1110, opened 2025-04-11
Make it easier to create play pages with ephemeral apps

dwwoelfel:
Creates a new `EphemeralAppPage` component that takes optional `schema` and `perms` props. You can use it to quickly create ephemeral apps for your play page. 

Also updates the ephemeral app endpoint to accept a schema.

You can even seed it with data using the `onCreateApp` prop.

Hopefully this will make it easier to quickly create new play pages. And if you forget how it works, there's a play page for that at http://localhost:4000/play/ephemeral-demo

Here's what the code for a page looks like:

```jsx
import { i, id, InstantReactAbstractDatabase, tx } from '@instantdb/react';
import EphemeralAppPage, {
  ResetButton,
} from '../../components/EphemeralAppPage';

const schema = i.schema({
  entities: {
    docs: i.entity({
      title: i.string(),
    }),
  },
});

const perms = {
  docs: {
    allow: {
      view: 'true',
    },
  },
};

function App({ db }: { db: InstantReactAbstractDatabase<typeof schema> }) {
  const { isLoading, error, data } = db.useQuery({ docs: {} });

  if (isLoading) {
    return <div>Loading</div>;
  }

  if (error) {
    return <div>Error: {error.message}</div>;
  }

  return (
    <div>
      <ResetButton className="bg-black text-white m-2 p-2" />
      <pre>{JSON.stringify(data, null, 2)}</pre>;
      <button
        className="bg-black text-white m-2 p-2"
        onClick={() =>
          db.transact(db.tx.docs[id()].update({ title: 'New doc' }))
        }
      >
        Add doc
      </button>
    </div>
  );
}

export default function Page() {
  return <EphemeralAppPage schema={schema} perms={perms} Component={App} />
}
```

### DIFF
--- a/client/sandbox/react-nextjs/components/EphemeralAppPage.tsx
+++ b/client/sandbox/react-nextjs/components/EphemeralAppPage.tsx
@@ -1,0 +1,206 @@
+import { useRouter } from 'next/router';
+import config from '../config';
+import {
+  EntitiesDef,
+  init,
+  InstantReactAbstractDatabase,
+  InstantSchemaDef,
+  LinksDef,
+} from '@instantdb/react';
+import { useEffect, useState } from 'react';
+import { RoomsDef, TransactionChunk } from '../../../packages/core/dist/module';
+import { IContainEntitiesAndLinks } from '../../../packages/core/dist/module/schemaTypes';
+
+async function provisionEphemeralApp({
+  perms,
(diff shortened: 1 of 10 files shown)


stopachka (approved):
Oo heck yeah!

PR #1115, opened 2025-04-14
Use custom `Error` instances in `Promise.reject` for improved stack traces

Ngalstyan4:
Previously, promise rejections used plain objects, losing debugging context. Now, using a custom `Error` instance ensures proper (though possibly truncated [^1] ) stack traces, enhancing clarity when debugging failed `await db.transact` calls.

Using `Error` instances in promise rejections is considered a best practice: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/reject

[^1]: (My understanding is), to get full stack traces in npm-installed version, [current compile target](https://github.com/instantdb/instant/blob/main/client/tsconfig.base.json#L8) has to be changed from `es2015` to `es2017`

## Example

When an InstantDB transaction such as:

```typescript
const p = await db.transact(memberTxs);
```
fails during development, the error output currently lacks a proper stack trace:

Before this PR:
<div>
current non-Error object
</div>

With the updated version of the library (this PR), errors become instances of a custom Error, providing clearer debugging information:

After this PR (compiled with target=es2015):
<div>
Custom Error object with es2015
</div>

After this PR (compiled with target=es2017):
<div>
Custom Error object with es2017
</div>

## Note

The type definition of `InstantIssueBody` may be incomplete or incorrect. I just noticed that it is missing e.g. "permission-denied" option for `body.type`. I copied the current definition from [here](https://github.com/instantdb/instant/blob/b16b8d1ba2a3c1945e9bb1e41b363cb0dcd56152/client/www/lib/types.ts#L245-L272)

### DIFF
--- a/client/packages/admin/src/index.ts
+++ b/client/packages/admin/src/index.ts
@@ -6,6 +6,8 @@ import {
   id,
   txInit,
   version as coreVersion,
+  InstantAPIError,
+  type InstantIssue,
   type TransactionChunk,
   type AuthToken,
   type Exactly,
@@ -177,7 +179,7 @@ async function jsonFetch(
   const json = await res.json();
   return res.status === 200
     ? Promise.resolve(json)
-    : Promise.reject({ status: res.status, body: json });
+    : Promise.reject(new InstantAPIError({ status: res.status, body: json }));
 }
 
 /**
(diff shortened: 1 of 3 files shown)


stopachka (approved):
Great idea @Ngalstyan4 !

PR #1112, opened 2025-04-11
Single exception log and java in src

dwwoelfel:
This does two things:

1. It sets us up to be able to include java code in our codebase. There have been a few times where it would be easier to just write something in Java, but it was too much effort to get clojure to consume the java file. Now we can easily add new java files in src/java. You still have to restart if you make changes to the java file, though, so it's not super easy (but maybe somebody knows how to set up a reload workflow?). 

2. It only logs a single exception in a set of nested spans.
  Right now, we'll log the exception in every single span, since each span will catch the exception, log it, then rethrow the exception.
  We accomplish this by adding some data to the throwable. You can add a "suppressed" exception to any throwable, so we add a custom `SpanTrackException` that puts the spanid in the message. We want the `SpanTrackException` to be as cheap as possible, so I created an exception that doesn't attach a stack trace (hence the java changes).
  
```clojure
(tracer/with-span! {:name "a"}       
  (tracer/with-span! {:name "b"}     
    (tracer/with-span! {:name "c"}   
      (throw (Exception. "oops"))))) 
```

**After**:

```
[4a8d] 1ms [c] clojure.error.class=java.lang.Exception exception.type=java.lang.Exception exception.message=oops clojure.error.phase=:execution exception.stacktrace=java.lang.Exception: oops
	at instant.db.instaql$eval77986.invokeStatic(NO_SOURCE_FILE:1936)
	at instant.db.instaql$eval77986.invoke(NO_SOURCE_FILE:1934)
	at clojure.lang.Compiler.eval(Compiler.java:7700)
	at clojure.lang.Compiler.eval(Compiler.java:7655)
	...
        Suppressed: instant.SpanTrackException: d4c6875f93a31567
 clojure.error.line=1936 clojure.error.symbol=instant.db.instaql/eval77986 exception.escaped=true clojure.error.cause=oops 
[4a8d] 2ms [b] child-threw-exception=true 
[4a8d] 3ms [a] child-threw-exception=true 
```

**Before**:

```
[ae32] 0ms [c] clojure.error.class=java.lang.Exception exception.type=java.lang.Exception exception.message=oops clojure.error.phase=:execution exception.stacktrace=java.lang.Exception: oops
	at instant.db.instaql$eval78087.invokeStatic(NO_SOURCE_FILE:1942)
	at instant.db.instaql$eval78087.invoke(NO_SOURCE_FILE:1940)
	at clojure.lang.Compiler.eval(Compiler.java:7700)
	...
[...]

PR #1116, opened 2025-04-14
Fixes :dev alias and remove reflection warnings

dwwoelfel:
Fixes a couple of dev-time annoyances introduced in https://github.com/instantdb/instant/pull/1112

1. Uses the correct alias to pull in the dev deps (`:dev` instead of "dev")
2. Fixes reflection warnings in logging_exporter.clj


tonsky:
Damn, I started investigating this too before I arrived at this PR

PR #1117, opened 2025-04-14
Add support for pg_hint_plan

dwwoelfel:
Adds pg_hint_plan to tell postgres which index to use for each CTE.

I originally didn't think we could use it because you can only have a single hint comment per query and we need a hint per CTE. But if we just alias the `triples` table in each CTE, we can individually address each CTE with a hint.

It looks like this:

```sql
with match_0_0 as (
  select * 
  from triples t0 -- alias triples as t0
  where av
), 
match_0_1 as (
  select *
  from triples t1 -- alias triples as t1
  where ave
)
-- Here's the pg_hint_plan that tells postgres to use the indexes
/*+
IndexScan(t0 av_index)
IndexScan(t1 ave_index)
*/
select json_build_object(...)
```


server/src/instant/util/pg_hint_plan.clj
```diff
@@ -0,0 +1,189 @@
...
+;; We'll wait until we can validate these params before setting them
+#_(defn set-guc
+    "Sets GUC parameter to the value defined while planner is running."
+    [param value]
+    [:'Set param value])
+
+
+
+
+nil
```
stopachka: nit: should this nil be here?

dwwoelfel: Removed! Originally had it there because `honey.sql/register-clause!` was the last thing in the file and it fills the minibuffer when the file is loaded in emacs.

PR #1120, opened 2025-04-14
Narek/error type fix

Ngalstyan4:
A few changes missed in #1115 

1.  Add 'permission-denied' subtype to the API error type (needed and used in our tests)
2. Add a bunch of other subtypes to the error type LLM-extracted from [`exceptions.clj`](https://github.com/instantdb/instant/blob/d79ff56d2abe2ede82fb20bdf702167025a9299f/server/src/instant/util/exception.clj) (not used by us, happy to change or revert)
3. Use the created `InstantIssue` type from core in `client/www` in stead of redefining a similar type (happy to revert)

### DIFF
--- a/client/packages/core/src/utils/fetch.ts
+++ b/client/packages/core/src/utils/fetch.ts
@@ -21,6 +21,11 @@ type InstantIssueBody =
       message: string;
       hint: { 'record-type': string };
     }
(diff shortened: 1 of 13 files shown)


stopachka (approved):
Awesome!

PR #1121, opened 2025-04-14
[Dash] Reverse parent <-> children self-link

nezaj:
A user pointed out that the default had this reversed. Likely folks will want `parent` in the forward direction


stopachka (approved):
Nice!

PR #1127, opened 2025-04-16
Faster aurora failover

dwwoelfel:
This PR makes us recover from aurora failover faster.

Instead of connecting to the cluster, which routes us to the writer via DNS, we look up the writer from the RDS API and connect to it directly. There's a process running in the background that will poll the RDS API for changes to the writer. When it notices that the writer changes, it will close all open connections to the old instance and update the config.

In my testing, it took about 15-20 seconds for the failover to succeed.

### DIFF
--- a/server/src/instant/jdbc/aurora.clj
+++ b/server/src/instant/jdbc/aurora.clj
@@ -1,6 +1,6 @@
 (ns instant.jdbc.aurora
   (:require
-   [instant.aurora-config :refer [secret-arn->db-creds]]
+   [instant.aurora-config :refer [rds-cluster-id->db-config secret-arn->db-creds]]
    [instant.config :as config]
    [instant.util.lang :as lang]
    [instant.util.tracer :as tracer]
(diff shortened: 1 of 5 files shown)


server/src/instant/jdbc/aurora.clj
```diff
@@ -164,28 +252,42 @@
...
 (defn start []
   (let [conn-pool-size (config/get-connection-pool-size)]
-    (tracer/record-info!
-     {:name "aurora/start-conn-pool" :attributes {:size conn-pool-size}})
     (lang/set-var! -conn-pool
       (start-pool conn-pool-size (config/get-aurora-config)))))
 
 (defn stop []
-  (lang/clear-var! -conn-pool HikariDataSource/.close))
+  (lang/clear-var! -conn-pool (fn [^HikariDataSource d]
```
stopachka, line 288: oo

server/src/instant/jdbc/aurora.clj
```diff
@@ -164,28 +252,42 @@
...
+        pool (if (:cluster-id config)
+               (let [weak-conn-tracker (atom {})
+                     track-conn-lock (Object.)
+                     add-conn-to-tracker (fn [instance-id conn]
+                                           (locking track-conn-lock
+                                             (swap! weak-conn-tracker
+                                                    update
+                                                    instance-id
+                                                    (fn [m]
+                                                      (let [^WeakHashMap m (or m (WeakHashMap.))]
```
stopachka, line 264: oo

PR #1133, opened 2025-04-21
[client] make instatx introspectable

stopachka:
**Problem**

consider a chunk like: 

```js
const chunk = db.tx.posts[id()].update({ title: 'Foo' }); 
```

Today, if you write: 

```js
console.log(chunk)
// or 
JSON.stringify(chunk)
```

You'll see an empty object. 

**Solution** 

I added `__etype` and `__ops` onto the `target` part of our proxy. This makes it so now console.log will show the keys, and stringify will stringify them too: 

CleanShot 2025-04-21 at 11 51 57@2x
image

@dwwoelfel @nezaj @tonsky

### DIFF
--- a/client/packages/core/src/instatx.ts
+++ b/client/packages/core/src/instatx.ts
@@ -18,6 +18,7 @@ export interface TransactionChunk<
   EntityName extends keyof Schema['entities'],
 > {
   __ops: Op[];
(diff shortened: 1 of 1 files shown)


dwwoelfel (approved):
Handy!

PR #1137, opened 2025-04-21
[flicker repro] start with a UI to try and repro flicker bug

stopachka:
I started building out a UI to repro the flicker bug. 

I _think_ I caught it! 

CleanShot 2025-04-21 at 16 57 16@2x

I don't think ^ this state should be possible, given we _constantly_ check todos 0, 1, 2 in order. 

I was only able to repro this when running local server, with 10ms frequency. 

Will start digging, but wanted to get the repro up! 

@dwwoelfel @nezaj @tonsky

### DIFF
--- a/client/packages/core/__tests__/src/Reactor.test.js
+++ b/client/packages/core/__tests__/src/Reactor.test.js
@@ -181,3 +181,160 @@ test('rewrite mutations works with multiple transactions', () => {
     expect(rewrittenMutations.get(k)['tx-steps']).toEqual(serverSteps);
   }
 });
+
+test('optimisticTx is not overwritten by refresh-ok', async () => {
+  const appId = uuid();
+  const reactor = new Reactor({ appId });
+  reactor._initStorage(IndexedDBStorage);
+  reactor._setAttrs(zenecaAttrs);
+  const q = { users: {} };
+  const joe_id = 'ce942051-2d74-404a-9c7d-4aa3f2d54ae4';
+
+  await reactor.querySubs.waitForLoaded();
+
+  let data = null;
+
+  reactor.subscribeQuery(q, (res) => {
+    data = res;
+  });
+
+  // Initialize the store
+  reactor._handleReceive(0, {
+    op: 'add-query-ok',
+    q,
+    'processed-tx-id': 0,
+    result: [
+      {
+        data: {
+          'datalog-result': {
+            'join-rows': [zenecaTriples],
+          },
+        },
+        'child-nodes': [],
+      },
+    ],
+  });
+
+  await reactor.querySubs.waitForSync();
+
+  expect(data.data.users.map((x) => x.handle)).toEqual([
+    'joe',
+    'alex',
+    'stopa',
+    'nicolegf',
+  ]);
+
+  // first optimistic
+  const ops2 = [
+    instatx.tx.users[joe_id].update({
+      handle: 'joe2',
+    }),
+  ];
+
+  reactor.pushTx(ops2);
+
+  await reactor.querySubs.waitForSync();
+
+  expect(data.data.users.map((x) => x.handle)).toEqual([
+    'joe2',
+    'alex',
+    'stopa',
+    'nicolegf',
+  ]);
+
+  const [eventId2] = reactor.pendingMutations.currentValue.keys();
+
+  // second optimistic
+  const ops3 = [
+    instatx.tx.users[joe_id].update({
+      handle: 'joe3',
+    }),
+  ];
+
+  reactor.pushTx(ops3);
+
+  await reactor.querySubs.waitForSync();
+
+  expect(data.data.users.map((x) => x.handle)).toEqual([
+    'joe3',
+    'alex',
+    'stopa',
+    'nicolegf',
+  ]);
+
+  const [_, eventId3] = reactor.pendingMutations.currentValue.keys();
+
+  // confirmation from server for first optimistic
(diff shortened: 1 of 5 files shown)


financialvice:
not sure if this is related or not, but there is a "flicker" bug if you hook a text attr directly to an input onChange and just type super fast; might be a different bug, but it was never an issue until ~3-4 months ago and has progressively gotten worse

stopachka:
I think that could be related! Will check, thank you @financialvice !

tonsky:
Simplified test code a little bit. Now it just bumps a single number every 10ms and checks that local state values it observes are sequential.

The problem seem to be in how server’s :refresh-ok are applied on a client. You can see sequence breakages coincide with :refresh-ok sent from a server.

BTW speed at which updates are run doesn’t seem to be that important. If I run it every 100ms instead of 10ms, I can see anomalies still. 

Screenshot 2025-05-07 at 18 40 17

client/packages/core/src/Reactor.js
```diff
@@ -431,33 +431,72 @@ export default class Reactor {
       case 'refresh-ok':
```
stopachka, line 447: one thing to be aware of: 

Imagine if a user writes: 

`add-query` `queryA`
`transact` `txA` 

Imagine if `txA` is not related to `queryA`. 

In this case, no `refresh-ok` will come in, and hence transactions will stay in pendingMutations. 

--- 

One way to fix this: 

We may want to have some similar logic in `add-query-ok`

client/packages/core/src/Reactor.js
```diff
@@ -431,33 +431,72 @@ export default class Reactor {
...
+          let store = s.createStore(
             this.attrs,
             triples,
             enableCardinalityInference,
             this._linkIndex,
           );
+
+          // apply pendingMutations that this query has not yet seen
+          for (const mut of Object.values(mutations)) {
+            if (!mut['tx-id'] && mut['tx-id'] > processedTxId) {
```
stopachka: nit: you may have meant `||` here

client/packages/core/src/Reactor.js
```diff
@@ -431,33 +431,72 @@ export default class Reactor {
...
+
+        const processedTxId = msg['processed-tx-id'];
+
         const updates = computations.map((x) => {
           const q = x['instaql-query'];
           const result = x['instaql-result'];
           const hash = weakHash(q);
           const triples = extractTriples(result);
-          const store = s.createStore(
+          let store = s.createStore(
```
stopachka, line 465: One PR you may find interesting: 

https://github.com/instantdb/instant/pull/613/files

Previously we did something similar and cleaned up mutations in refresh. I think at that point we were too eager about when we cleaned pending mutations. I _think_ your current solution is more robust, but wanted to share the previous PR, in case it helps with context

client/packages/core/src/Reactor.js
tonsky: Damn! You are right. This is a conundrum. I have two ideas:

1. If we see _any_ query with processed-tx X, we can assume that all later queries will see even later DB. That gives us a permission to clear up to X even after a single query sees it. Problem with it could be that query A starts earlier (and observes earlier DB) but takes longer. Query B starts later, finishes quickly and got sent. Then, when A arrives, transactions between A and B will be cleared. I think this is realistic?
2. Just clear pendingMutations on a reasonable timeout. Like 10 seconds or so. M?

stopachka: I do think the race condition you mention in `1.` is real. I think I noticed this happening in our previous solution, because folks sometimes have a single query that takes more than a second, and other much faster queries.

stopachka: I'd be down for `2.` as a pragmatic solution to move us forward. I _think_ that would only happen in the case of: 

a. There are no queries at all (so no opportunity for refresh-ok at all)
b. All the queries are unrelated to the pending mutation (mutation does not affect the refresh-ok)

**Hazy thoughts on other solutions**

- Right now `processed-tx-id` only reports the `tx-id` which invalidated the session. Perhaps we can include a new `query-seen-tx-id`, which is returned directly from postgres, saying "what" the latest tx-id the query has actually seen. 
  - If we had this, we would still need some way to know during init: "Okay, I have added N queries, and got them all back, and now I _know_ this is the minimum query-seen-tx-id"

--- 

cc @dwwoelfel if you have thoughts too

client/packages/core/__tests__/src/Reactor.test.js
```diff
@@ -181,3 +181,160 @@ test('rewrite mutations works with multiple transactions', () => {
     expect(rewrittenMutations.get(k)['tx-steps']).toEqual(serverSteps);
   }
 });
+
+test('optimisticTx is not overwritten by refresh-ok', async () => {
```
stopachka, line 185: This is awesome @tonsky !!

server/src/instant/reactive/session.clj
```diff
@@ -223,6 +223,16 @@
     (tracer/with-span! {:name "handle-refresh/send-event!"
                         :attributes tracer-attrs}
       (when (seq computations)
+        (doseq [{q :instaql-query
```
stopachka: nit: assuming we will remove this + transaction.clj in final pr

stopachka (commented):
(Looks like I can't approve a PR I originally created -- buut LGTM. Beautiful work @tonsky !)

PR #1138, opened 2025-04-23
Room key validation

dwwoelfel:
Validates that the room key is a string.

Here's what it looks like when you provide something that's not a string:
image

Here's what it looks like when you provide null:
image

And here's what it looked like before:
image


stopachka (approved):
Niice!

PR #1140, opened 2025-04-23
Drop old ref_values_are_uuid constraint 

dwwoelfel:
Part 2 of https://github.com/instantdb/instant/pull/1139

Shouldn't be merged until the migration from #1139  is run and the constraint is checked.


stopachka (approved):
Niice!

Issue #1114, opened 2025-04-13
[OSS] Local docker server / database fails to upgrade to latest code


verveguy:
Error when running make docker-compose against existing containers

Using latest codebase off main branch commit `b16b8d1`

Not sure how to work around this one ... rolling back to previous revision but the database is now dirty. Server reports:

```
server-1    | Booting up backend w/ local db...
server-1    | Migrating database
server-1    | Using config at resources/config/override.edn
server-1    | error: Dirty database version 63. Fix and force version.
```

How do I "fix and force version" ?

Here's the full log from the failed update that put me in this situation.

```
Booting up backend w/ local db...
Migrating database
Using config at resources/config/override.edn
61/u seed_test_user (17.304209ms)
62/u seed_ephemeral_app_creator (27.465096ms)
error: migration failed: insert or update on table "apps" violates foreign key constraint "apps_creator_id_fkey", Key (creator_id)=(6f0f64c9-207f-4b35-aec7-82739dcde58e) is not present in table "instant_users". in line 0: insert into
  instant_users (id, email)
values
  (
    '6f0f64c9-207f-4b35-aec7-82739dcde58e',
    'stopa@instantdb.com'
  ) on conflict do nothing;

insert into 
  apps (id, creator_id, title) 
values (
  'fc5a4977-910a-43d9-ac28-39c7837c1eb5', 
  '6f0f64c9-207f-4b35-aec7-82739dcde58e',
  'homepage'
) on conflict do nothing;

insert into 
  rules (app_id, code) 
values (
  'fc5a4977-910a-43d9-ac28-39c7837c1eb5', 
  '{ "$default": { "allow": { "$default": "false" } } }'
) on conflict do nothing;
 (details: pq: insert or update on table "apps" violates foreign key constraint "apps_creator_id_fkey")
{:clojure.main/message
 "Execution error at clojure.java.process/exec (process.clj:172).\nProcess failed with exit=1\n",
 :clojure.main/triage
 {:clojure.error/class java.lang.RuntimeException,
  :clojure.error/line 172,
  :clojure.error/cause "Process failed with exit=1",
  :clojure.error/symbol clojure.java.process/exec,
  :clojure.error/source "process.clj",
  :clojure.error/phase :execution},
 :clojure.main/trace
 {:via
  [{:type java.lang.RuntimeException,
    :message "Process failed with exit=1",
    :at [clojure.java.process$exec invokeStatic "process.clj" 172]}],
  :trace
[...]

PR #1141, opened 2025-04-23
[db] fix home page migration

stopachka:
I got an error running my migration locally:

error: migration failed: insert or update on table "apps" violates foreign key constraint "apps_creator_id_fkey", Key (creator_id)=(6f0f64c9-207f-4b35-aec7-82739dcde58e) is not present in table 

I think this could happen when stopa@instantdb.com is inserted, but the id is not 6f0... 

Afaik @verveguy had this issue here too https://github.com/instantdb/instant/issues/1114 

This updates the migration so we do all the transactions in one query, and get the id from the db. 

@nezaj @dwwoelfel @tonsky

### DIFF
--- a/server/resources/migrations/63_seed_homepage_app.up.sql
+++ b/server/resources/migrations/63_seed_homepage_app.up.sql
@@ -1,22 +1,40 @@
+with u_ins as (
+  insert into
+    instant_users (id, email)
+  values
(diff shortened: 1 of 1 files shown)


nezaj (approved):
Sounds reasonable!

PR #1142, opened 2025-04-24
Autocomplete users in the sandbox

dwwoelfel:
Adds a combobox to the auth.email field in the sandbox that pulls data from the `$users` table. 

You can still enter a email that's not in the $users table and there's a special `<none>` option for selecting no user.

We'll also keep the last user that you selected in localstorage so you can navigate in the dashboard without losing your place.

I moved it to the middle of the sandbox to give a place for the autocomplete to go and to make it easier to see what user was set.

image


nezaj (approved):
Beautiful!

PR #1148, opened 2025-04-28
Required params: update schema & types

tonsky:
- Added `.required()` to schemaTypes (blobs)
- Added `required: true` to forward links
- Added support for both to CLI (pull & push)


client/packages/cli/src/index.js
```diff
@@ -401,6 +401,7 @@ program
     '-a --app <app-id>',
     'App ID to push to. Defaults to *_INSTANT_APP_ID in .env',
   )
+  .option('-o --override', 'Override existing schema')
```
stopachka: I _think_ you don't need to do this. 

You can pass in `--yes` in the command, and it will override the existing schema by default.

client/packages/cli/src/index.js
```diff
@@ -497,16 +498,16 @@ async function handlePull(bag, opts) {
     opts,
   );
   if (!ok) return;
-  await pull(bag, appId, pkgAndAuthInfo);
+  await pull(bag, appId, pkgAndAuthInfo, opts);
```
stopachka: Similarily, I think you can skip passing an `opts` here, because this library has a global way to access the bag of passed in opts: 

Something like what is happening in `promptOk`: 

https://github.com/instantdb/instant/blob/main/client/packages/cli/src/index.js#L1316-L1318

server/src/instant/model/schema.clj
```diff
@@ -29,98 +29,116 @@
...
-                    vec)]
+        blob-ops (for [[ns-name attrs] new-blobs
+                       [attr-name new-attr] attrs
+                       :let [current-attr      (get-in current-schema [:blobs ns-name attr-name])
+                             changed-type?     (and check-types?
+                                                    (not= (get new-attr :checked-data-type)
+                                                          (get current-attr :checked-data-type)))
+                             changed-unique?   (not= (get new-attr :unique?) (get current-attr :unique?))
+                             changed-index?    (not= (get new-attr :index?) (get current-attr :index?))
+                             changed-required? (not= (get new-attr :required?) (get current-attr :required?))
```
stopachka, line 40: **Bug: `instant-cli push` does not seem to pick up `required`**

I was doing some dogfooding, but wasn't able to get an `add required` to trigger:

https://github.com/instantdb/instant/compare/required-params-3...required-params-dogfood?expand=1

Steps: 
1. I added `required` to a column in `cli-nodejs/db/instant.schema.ts`
2. Ran npx instant-cli push (this uses the latest local version)
3. The backend does not seem to pick this up

client/packages/core/src/schemaTypes.ts
```diff
@@ -115,6 +122,7 @@ export type LinkDef<
     on: FwdEntity;
     label: FwdAttr;
     has: FwdCardinality;
+    required?: RequirementKind;
```
stopachka, line 139: **Discussion: required links in queries**

Consider a link between `posts` and `users`, like:

```js
{ forward: { on: 'posts', has: 'one', label: 'owner', required: true } 
  ... }
```

Now if a user writes a query like:

```js
const { isLoading, error, data }  = useQuery({ posts: { owner: { } })
if (isLoading || error) return
const post = posts[0]; 
if (!post) return 
const owner = posts.owner 
```

**What would be the type of `owner`?**

Currently, I _think `owner` will be typed as `User | undefined` 

### Ideal solution

Ideally, it would be great if `owner` was just typed as `User`. 

We _could_ change that type in this PR, but I think we'd have to do one more change: 

1. Even with required links, it can be possible to return a `post` without an `owner`

This can happen if we have local data for a post, but not yet for a user.

### Current leaning

I say we push this work to a separate PR. I did want to raise it here though, to make sure we're all on the same page.

client/packages/core/src/schemaTypes.ts
```diff
@@ -70,6 +72,11 @@ export type ValueTypes = 'string' | 'number' | 'boolean' | 'date' | 'json';
 
 export type CardinalityKind = 'one' | 'many';
 
+// true      - force required
+// undefined - required in types, not required in backend
```
stopachka: **Discussion: tri state**

I am definitely on the fence for the tri-state here.

Here's my thinking. 

Let's consider this specific column: 

```ts
posts: i.entity({ title: i.string() })
```

We have 3 options: 

### 1. Tri-state: `undefined` means required in types, but not backend

**Pro:**
1. This is the best for backwards-compatibility. By default, schema works _as it did before_ 

**Con:**

1. This encourages a buggy behavior for _new_ users. 
   1. By default, new users will write `i.string()`.
      1. Typescript will think they're required, but they won't actually be
1. It makes the api a bit more complicated for new users: 
   1. It's surprising that you can write both `i.string().required()` _and_ `i.string().optional()`

## 2. string() is required by default

**Pro:**
1. This "matches" how users think about typed objects in Typescript
1. The `.optional` modifier works in a backwards compatible way

**Con:**
1. When existing users push, we will have to run a giant job which turns most columns into `required`, 
2. When existing users _add_ a new column, they will have to make sure to add it as `optional`
   1. Otherwise, push will error, as _all previous_ records _must_ have that value. 

## 3. string() is optional by default

**Pro**
1. This "matches" how optionality works in databases
1. Adding new columns "works" by default.

**Con**
1. If users upgrade, they will have to update the schema to:
   1. Add `.required` to their required fields, 
   1. Remove `.optional` for their optional fields
2. Most typescript users _want_ objects with stronger types -- this may end up having them write `.required` more frequently than they would `.optional`.


### Current Leaning

Reading this through, I lean towards `string()` required by default. Here's my reasoning: 

The cons in a way _are_ useful themselves: 
1. Yes, the inital push will cause a lot of work on the backend, converting columns to required columns
   1. _But_ this was the original intention of the user (they _wanted_ that field to be required)
      1. Perhaps this can be a good thing for them, to really validate their data model
        1. We could also test this out with an experimental build -- give it to our bigger users, and see how onerous the push is for them
3. Yes, new columns will be a bit more verbose
   1. i.e _must_ add `.optional` first
      1. But perhaps this is a good thing, as it forces the user to think about old rows explicitly. 

The main worry with above is, if we have _a lot_ of jobs that spin up for `required`, I hope we don't time out for the user or something.

--- 

As I was on a walk, another fear: if users automatically upgrade most of their types to required, I wonder if it will cause more runtime errors for them.

--- 

What do you guys think?

stopachka (commented):
Excited to ship this! Opened up some discussions

client/packages/core/src/schemaTypes.ts
nezaj: From our previous discussion I leaned more towards option 3 initially but I do think @stopachka has a good point that option 2 is more in-line with what frontend devs / TS users expect. Perhaps we can do option 2 as an experimental release, get some of our power users to try it, and see if folks have issues pushing.

If it's smooth then we can go forward with option 2, otherwise we'll need to think on it more

server/src/instant/model/schema.clj
stopachka: Dohp, @tonsky ignore this. 

For some reason I thought this change was already in the backend. Switching to local backend, everything worked great 👍 👍

client/packages/cli/src/index.js
tonsky: You are right!

client/packages/core/src/schemaTypes.ts
tonsky: > Currently, I _think owner will be typed as User | undefined

Confirmed that.

Extracted into https://github.com/instantdb/instant/issues/1150

client/packages/admin/package.json
```diff
@@ -9,8 +9,8 @@
   "scripts": {
     "test": "vitest",
     "build": "rm -rf dist; npm run build:main && npm run build:module",
-    "dev:main": "tsc -p tsconfig.json -w --skipLibCheck --preserveWatchOutput",
-    "dev:module": "tsc -p tsconfig.module.json -w --skipLibCheck --preserveWatchOutput",
+    "dev:main": "tsc -p tsconfig.json -w --skipLibCheck --preserveWatchOutput | grep -v '^$'",
```
stopachka, line 12: Nice

client/packages/cli/src/index.js
```diff
@@ -1118,6 +1125,22 @@ async function pushSchema(appId, opts) {
         );
         break;
       }
+      case 'required': {
+        console.log(
+          '%s to %s',
```
stopachka, line 1130: I know this was in the code, but TIL! I didn't know you could add %s and have console.log format for you...

Cool

client/packages/core/src/schemaTypes.ts
```diff
@@ -11,8 +11,20 @@ export class DataAttrDef<ValueType, IsRequired extends boolean> {
     } = { indexed: false, unique: false },
   ) {}
 
+  clientRequired() {
```
stopachka, line 19: nit: I would add a `@deprecated` docstring here, explaining that you can use this as a temporary measure.

client/packages/core/src/schemaTypes.ts
```diff
@@ -70,6 +82,11 @@ export type ValueTypes = 'string' | 'number' | 'boolean' | 'date' | 'json';
 
 export type CardinalityKind = 'one' | 'many';
 
+// true           - force required
+// clientRequired - required in types, not required in backend
+// false          - optional, not required
+export type RequirementKind = true | 'clientRequired' | false;
```
stopachka: I don't think you need the type `RequirementKind` to know about `clientRequired`. On the typescript level, imo we can keep it as `true | false`. 

You can do something like:

```js
  clientRequired() {
    return new DataAttrDef<ValueType, true>(
      this.valueType,
      false as unknown as true,
      this.config,
    );
  }
```

The idea here being, typescript doesn't need to know about the third state. 

Maybe it's a bit too crazy though -- happy to go either way, but wanted to bring up the option

client/packages/cli/src/index.js
tonsky: Oh, I didn’t notice it either. Ha!

client/packages/core/src/schemaTypes.ts
tonsky: Okay. Interesting. One reason why I extracted RequirementKind is because I wanted links to have it as well. And for links we don’t have such nice builder methods, we have to type values. That’s why it’s `required: true | false`, even though in the code I’d prefer to see smth like `'required' | 'optional'`. But, now, as all links are treated as optional in our type system, I guess it should be fine for now.

PR #1149, opened 2025-04-29
[Growth] Send welcome email to all business users

nezaj:
We thought it would be nice to follow up with all non-personal email signups. Feels like it could be an easy win to get more engagement from business users!

CleanShot 2025-04-29 at 15 03 11@2x


stopachka (approved):
Sgtm!

stopachka (approved):

PR #1151, opened 2025-05-01
[sm] better error for admin api

stopachka:
I realize we threw an `ex-info` when an unsupported action was provided. I updated this to be a 400 error. 

**Why?**

Previously, we made this a 500, because only our client SDKs provided the `steps`. But now some users call the api directly. In this case, they _could_ provide invalid steps. 

@nezaj @tonsky @dwwoelfel


nezaj (approved):
Nice!

PR #1152, opened 2025-05-01
[routes] default 404 in our api routes

stopachka:
I noticed that if I made api calls to misspelled routes, I would get an empty 200 response.

Updated so we get a 404. 

@dwwoelfel @nezaj @tonsky


nezaj (approved):
LGTM. (I guess this included some linting too?)

stopachka:
(Yes, I think we need to tell kondo about set-var! Noting to get to this)

PR #1153, opened 2025-05-01
[cli] give proper exit codes when we return with an error

stopachka:
Users pointed out that if the CLI stopped with an error, we would still return an exit code with a value of 0. 

I updated our script, so when we fail, we return an exit code of 1. 

@nezaj @dwwoelfel @tonsky

### DIFF
--- a/client/packages/cli/src/index.js
+++ b/client/packages/cli/src/index.js
@@ -362,7 +362,7 @@ program
   .description('Push schema and perm files to production.')
   .action(async function (arg, inputOpts) {
     const ret = convertPushPullToCurrentFormat('push', arg, inputOpts);
-    if (!ret.ok) return;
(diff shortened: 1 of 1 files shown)


client/packages/cli/src/index.js
```diff
@@ -362,7 +362,7 @@ program
   .description('Push schema and perm files to production.')
   .action(async function (arg, inputOpts) {
     const ret = convertPushPullToCurrentFormat('push', arg, inputOpts);
-    if (!ret.ok) return;
+    if (!ret.ok) return process.exit(1);
```
stopachka, line 365: We don't technically need to write `return` with process.exit. 

Chose to specifically write it this way, because it's really jarring to see a function that that will implicitly stop control flow.

nezaj (approved):
sgtm!

PR #1156, opened 2025-05-02
[client] Schema updates should propagate in a hot reload

stopachka:
## Bug

Users reported a bug that worked like so: 

1. Take a schema
2. Change it
3. Make a transaction, which _should_ include the changes to your schema
4. But turns out, **it does not**

## Reason

We cache calls to `init`. This means that when hot-reload triggered a new `init` call, we would return the previously cached `db` instance. The rub? The previous db includes the _old_ schema

## Solution

Now we detect if the schema has changed. If it _did_ change, we patch the existing client, with the new schema.

## Alternatives considered

You may be wondering: 

**Why not _replace_ the reactor?** 

There's a _lot_ of state in the existing reactor instance. For example, we include current subscriptions there. If we shut down and start a new reactor instance, we'd have to be extra careful to make sure that previous callers don't accidentally end up stranded

## Cached states

One known issue, is that we currently pass `linkIndex` around to stores, and that info is cached in a bunch of places. Thinking it through, I don't think it matters for our case. _However_, I think it would be so much nicer to remove that extra caching in store.js. 

I will look into that in a follow-on PR. 

@dwwoelfel @nezaj @tonsky

### DIFF
--- a/client/packages/core/src/Reactor.js
+++ b/client/packages/core/src/Reactor.js
@@ -224,6 +224,15 @@ export default class Reactor {
     }
   }
 
+  updateSchema(schema) {
+    this.config = {
+      ...this.config,
+      schema: schema,
+      cardinalityInference: Boolean(schema),
+    };
+    this._linkIndex = schema ? createLinkIndex(this.config.schema) : null;
+  }
+
   _initStorage(Storage) {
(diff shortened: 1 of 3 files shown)


nezaj (approved):
Nice!

PR #1157, opened 2025-05-05
Hide exception logging outside asserts during tests

tonsky:
Sometimes tests intentionally run broken/incorrect code to assert on exception. We currently log these as exception when, in fact, they are not—they are expected to break during test.

If some exception is unhandled, it will be logged by test runner as before.

### DIFF
--- a/server/src/instant/util/tracer.clj
+++ b/server/src/instant/util/tracer.clj
@@ -3,6 +3,7 @@
   (:gen-class)
   (:require
    [clojure.main :as main]
+   [clojure.test :as test]
    [instant.config :as config]
    [instant.util.logging-exporter :as logging-exporter]
    [instant.util.coll :as ucoll]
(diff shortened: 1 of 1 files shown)


server/src/instant/util/tracer.clj
```diff
@@ -119,10 +120,9 @@
    {:keys [exception escaping? attributes]
     :or   {attributes {}}}]
   (when (and (not logging-exporter/log-spans?)
+             (nil? test/*report-counters*)
              (logging-exporter/exception-belongs-to-span? exception (-> span .getSpanContext .getSpanId)))
-    ;; helps to capture output in tests
-    (binding [*out* (clojure.lang.Var/.getRawRoot #'*out*)]
```
stopachka, line 124: curious q: how did the var previously help in capturing output?

stopachka (approved):
Great improvement!

server/src/instant/util/tracer.clj
tonsky: I messed it up, I though this logging happens from another thread that is not affected by rebinding of *out*. It wasn’t helping

PR #1158, opened 2025-05-05
Too many params error

dwwoelfel:
Give a better error when we get a transaction with too many params. This can happen when you do a transaction that writes a lot of entities at the same time.

```json
{
  "name": "InstantAPIError",
  "status": 400,
  "body": {
    "type": "parameter-limit-exceeded",
    "message": "There are too many parameters in the transaction or query.",
    "hint": {
      "message": "Consider batching transactions to reduce the number of writes in a single transaction.",
      "doc-urls": [
        "https://www.instantdb.com/docs/instaml#batching-transactions"
      ]
    },
    "trace-id": "64972e6c3cef4051940b96c5268905a3"
  }
}
```

### DIFF
--- a/server/src/instant/util/exception.clj
+++ b/server/src/instant/util/exception.clj
@@ -436,7 +436,7 @@
                      (string/trim (subs detail (inc start) end)))
                    (partition 2 1 borders))))))
 
-(defn default-psql-throw! [e {:keys [server-message condition table] :as data} hint]
+(defn default-psql-throw! [e {:keys [condition] :as data} hint]
   (throw+ {::type ::sql-exception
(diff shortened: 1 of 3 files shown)


server/src/instant/util/exception.clj
```diff
@@ -46,6 +47,24 @@
 (s/def ::trace-id string?)
 (s/def ::instant-exception (s/keys :req [::type ::message ::trace-id]))
 
+(def bad-request-types #{::record-not-found
```
dwwoelfel, line 50: Moved these from `instant.util.http` so that they'll be easier to update when you update the set of `::types`

PR #1160, opened 2025-05-05
[reactor] make sure you can't start a shut down reactor

stopachka:
While debugging in the dashboard, I noticed that we started _two_ Reactor instances. 

**Why?** 

https://github.com/instantdb/instant/blob/main/client/www/pages/dash/index.tsx#L352-L368

In dev, React's useEffect quickly runs twice. We would start two reactor instances. The first one would shut down immediately. But, `startSocket` happened after it was shut down. 

**Fix**

I made sure in startSocket, we give up if the Reactor is shut down. 

@dwwoelfel @nezaj @tonsky

### DIFF
--- a/client/packages/core/src/Reactor.js
+++ b/client/packages/core/src/Reactor.js
@@ -976,6 +976,7 @@ export default class Reactor {
   };
 
   shutdown() {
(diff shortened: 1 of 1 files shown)


dwwoelfel (approved):
LGTM!

PR #1163, opened 2025-05-05
Handle non-json error payloads in admin sdk

dwwoelfel:
Returns error messages as text if we can't parse it as JSON instead of throwing a JSON parse error.

Before:

```
 ⨯ SyntaxError: Unexpected token 'w', "whoops" is not valid JSON
    at JSON.parse (<anonymous>) {
  digest: '936232253'
}
```

After:

```
 ⨯ Error [InstantAPIError]: whoops
    at eval (../../src/index.ts:191:6)
    at Generator.next (<anonymous>)
    at fulfilled (webpack-internal:///(rsc)/../../packages/admin/dist/module/index.js:22:32)
  189 |   } catch (_e) {
  190 |     return Promise.reject(
> 191 |       new InstantAPIError({
      |      ^
  192 |         status: res.status,
  193 |         body: { type: undefined, message: body },
  194 |       }), {
  status: 400,
  body: { type: undefined, message: 'whoops' },
  digest: '301343937'
}

```

### DIFF
--- a/client/packages/admin/src/index.ts
+++ b/client/packages/admin/src/index.ts
@@ -176,10 +176,24 @@ async function jsonFetch(
     'Instant-Core-Version': coreVersion,
   };
   const res = await fetch(input, { ...defaultFetchOpts, ...init, headers });
-  const json = await res.json();
-  return res.status === 200
-    ? Promise.resolve(json)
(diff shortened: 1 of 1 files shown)


stopachka (approved):
Niice!

PR #1166, opened 2025-05-06
Add config to sample logs

nezaj:
Adds a config so we sample logs for heavier apps. The hunch is sampling one or a few apps could significantly reduce our cloudwatch bill!

Steps to make it live:

1) Merge code
2) Push config 
3) Add app id(s)
4) Deploy


stopachka (approved):
Let's try!

PR #1167, opened 2025-05-06
CLI: skip system namespaces during schema push

tonsky:
Users who didn’t pull make have $files/$users be marked as required from before, which breaks on push. Solution: do not try to push system namespaces at all

Screenshot 2025-05-06 at 17 26 40

### DIFF
--- a/client/packages/cli/src/index.js
+++ b/client/packages/cli/src/index.js
@@ -1040,7 +1040,7 @@ function linkOptsPretty(attr) {
 async function pushSchema(appId, opts) {
   const res = await readLocalSchemaFileWithErrorLogging();
   if (!res) return { ok: false };
-  var { schema } = res;
+  const { schema } = res;
   console.log('Planning schema...');
 
   const planRes = await fetchJson({
--- a/client/packages/cli/src/index.js
+++ b/client/packages/cli/src/index.js
@@ -1040,7 +1040,19 @@ function linkOptsPretty(attr) {
 async function pushSchema(appId, opts) {
   const res = await readLocalSchemaFileWithErrorLogging();
   if (!res) return { ok: false };
-  const { schema } = res;
+  var { schema } = res;
+
+  // skip system namespaces
+  if (schema && schema.entities) {
+    const entities = { ... schema.entities };
+    Object.keys(entities).forEach(key => {
+      if (key.startsWith('$')) {
+        delete entities[key];
+      }
+    });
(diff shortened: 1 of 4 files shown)


client/packages/cli/src/index.js
```diff
@@ -1040,7 +1040,19 @@ function linkOptsPretty(attr) {
 async function pushSchema(appId, opts) {
   const res = await readLocalSchemaFileWithErrorLogging();
   if (!res) return { ok: false };
-  const { schema } = res;
+  var { schema } = res;
+
+  // skip system namespaces
+  if (schema && schema.entities) {
```
stopachka: Perhaps this would be better to do in the backend? 

(Main reasoning: in the future we may have other callers that push a schema, and we'd have to make sure to do this validation there too)

tonsky: Yes, already changing it, had the same thought

client/packages/cli/src/index.js
```diff
@@ -1040,7 +1040,7 @@ function linkOptsPretty(attr) {
 async function pushSchema(appId, opts) {
   const res = await readLocalSchemaFileWithErrorLogging();
   if (!res) return { ok: false };
-  const { schema } = res;
+  var { schema } = res;
```
stopachka: Nit: intentional?

server/src/instant/dash/routes.clj
```diff
@@ -1008,10 +1009,18 @@
 ;; ---
 ;; CLI
 
+(defn- remove-system-namespaces [entities]
+  (ucoll/filter-keys
+   #(not (string/starts-with? (name %) "$"))
```
stopachka: Nit: I think you can use all-etypes from system-catalog.clj

stopachka (approved):
Sgtm!

server/src/instant/dash/routes.clj
tonsky: It feels like we should prevent modifying any namespaces that start with `$`, no? So that user won’t accidentally create `$folder` or smth that we later won’t be able to use

client/packages/cli/src/index.js
tonsky: Damn. An artifact of rolling back original impl

server/src/instant/dash/routes.clj
stopachka: Fair point! Down

PR #1171, opened 2025-05-06
Remove cloudflare-specific header

dwwoelfel:
Removes the cloudflare-specific "cf-connecting-ip" in favor of taking the last ip from `x-forwarded-for`. Should be the same and it's dependent on Cloudflare.

Also fixes a bug where I did `::type` instead of `::ex/type`, using the wrong namespace prefix.

And returns a JSON payload from our not found handler.


nezaj (approved):
Nice!

PR #1172, opened 2025-05-07
[tsc][1] add a nuxt playground

stopachka:
This is a barebones playground running Nuxt. 

The primary purpose right now is to test how our export looks in Nuxt's backend. When you load `localhost:3060`, you'll see that `id` comes in as an object rather than a function. 

Follow-on PRs will fix this issue, but first I wanted to get this out in it's own PR. 

@dwwoelfel @nezaj @tonsky


dwwoelfel (approved):
LGTM!

PR #1174, opened 2025-05-07
Better errors duplicate ident creation

nezaj:
I noticed we don't give a helpful error message if someone un-intentionally tries to create an attr that already exists. Thought it would be nicer to have a more specific error message

**Before**

before_error

**After**

after_error

### DIFF
--- a/server/src/instant/util/exception.clj
+++ b/server/src/instant/util/exception.clj
@@ -213,14 +213,12 @@
        (let [details (parse-unique-detail (:detail pg-data))
              etype (get details "etype")
              label (get details "label")]
-         (cond (and etype label)
-               {:message (format "`%s` already exists on `%s`"
-                                 label
-                                 etype)
-
-                :hint {:etype etype
-                       :label label}}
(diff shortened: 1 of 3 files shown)


server/src/instant/util/exception.clj
```diff
@@ -198,40 +198,70 @@
...
+
+(defn extract-duplicate-ident-data [pg-data]
+  (when (and (= "idents" (:table pg-data))
+             (= "app_ident_uq" (:constraint pg-data)))
+    (safely-extract-data
+     (fn [pg-data]
+       (let [details (parse-unique-detail (:detail pg-data))
+             etype (get details "etype")
+             label (get details "label")]
+         (cond (and etype label)
```
dwwoelfel: You could just use an `if` here.

dwwoelfel (approved):
Nice!

server/src/instant/util/exception.clj
nezaj: Updated to use `when`!

PR #1176, opened 2025-05-07
Set initial presence when you join the room

dwwoelfel:
Based on a user request: https://discord.com/channels/1031957483243188235/1148284450992574535/1363590329386795079

This prevents a flicker from doing `join-room` followed by `set-presence`. Now you can set your initial presence data when you join the room using the `initialData` field in the opts. It looks like this:

```ts
const {
    user: myPresence,
    peers,
    publishPresence,
  } = db.rooms.usePresence(
    room,
    // Publish your presence when you join the room
    { initialData: { name: user.name } },
  );
```

This is a change to the public API, so please weigh in if you have a better idea of how it should be expressed!

### New flag
I added a new "toggles" namespace to flags that can hold all of the boolean flags. That way you don't have to mess with the flags config every time you want to add a simple on-off switch. Just use `(flags/toggled? :my-flag)` and then add a new row to the "toggles" namespace in instant-config!

### Deployment strategy

This adds a new record type to hazelcast so we can only enable it after all of the old hazelcast nodes cycle out.

1. Merge and deploy to eb
2. Push up the flags `pnpm run push`
3. Wait for deploy to complete
4. Flip the "join-room-v2" toggle in flags to true
5. Push up a new pr that removes join-room-old


client/www/pages/docs/presence-and-topics.md
```diff
@@ -136,12 +136,20 @@ const user = {
...
 function App() {
-  const { user: myPresence, peers, publishPresence } = db.rooms.usePresence(room);
+  const {
+    user: myPresence,
+    peers,
+    publishPresence,
+  } = db.rooms.usePresence(
+    room,
+    // Publish your presence when you join the room
+    { initialData: { name: user.name } },
```
stopachka, line 146: nit: I _think_ we may need to also update `db.rooms.useSyncPresence`, so it passes the whatever it initially gets to `initialData` as well.

server/flags-config/instant.schema.ts
```diff
@@ -45,40 +85,37 @@ const graph = i.graph(
...
-      "disableLegacy?": i.boolean(),
-      "useLocationId?": i.boolean(),
-    }),
-    "storage-block-list": i.entity({
-      appId: i.string().unique().indexed(),
-      isDisabled: i.boolean(),
-    }),
-    "query-flags": i.entity({
-      "query-hash": i.number(),
+    toggles: i.entity({
```
stopachka, line 94: awwh yeah!

stopachka (approved):
Soo awesome!!

client/www/pages/docs/presence-and-topics.md
dwwoelfel: Already done! https://github.com/instantdb/instant/pull/1176/files#diff-5350afe3652729d266c03240b7ebd1708380c05a31efc68c547968f819f31f90L208

PR #1179, opened 2025-05-07
[bundling] use tshy to bundle

stopachka:
**This PR:** 
1. Upgrades our `package.json` for modern bundlers too. Added `type` field, as well as an `exports` map.
1. Uses `tshy` to build our **CommonJS** and **ESModule** builds.

--- 

To explain the change here, I thought I would get a bit deeper on the state of bundling typescript libraries. 

# Package.json

### 1) CommonJS

First, we started with commonjs. You would write your code, and generally create one `dist` bundle file:

```js
{
  “main”: “dist/index.js”
}
```

When node would import this library, it would look inside `dist/index.js`. This could either be bundled in one file, or include other `require` calls: 

```js
// dist/index.js
const uuid = require('uuid'); 
const utils = require('./utils');
```

### 2) ESModules

Eventually JS added support for `import`, a.k.a ESModules. You could specify that you wanted Node to pick up ESM. To do that, you could add a `module` key: 

```js
{
  “main”: “dist/index.js”, 
  "module" "dist/esm/index.js"
}
```

In ESModules, you could use `import` statements: 

```js
import uuid from uuid;
import utils from './utils.js';
```

### 3) Typescript types

And let’s not forget about typescript. You could include a `types` field, which would tell typescript about how to pick up the `d.ts` file: 

```js
{
  “main”: “dist/index.js”, 
  "module" "dist/esm/index.js", 
  "types": "dist/esm/index.d.ts"
}
```

## 4) Export 

But we don’t stop there. Eventually `exports` field was introduced. Instead of looking at `main` or `module`, modern runners now look at the `exports` field: 

```js
{
  “main”: “dist/index.js”, 
  "module" "dist/esm/index.js", 
  "types": "dist/esm/index.d.ts", 
  "exports": { 
    ".": { 
      "require": { 
         "types": "dist/index.d.ts", 
         "default": "dist/index.js",
      },
      "import": { 
         "types": "dist/index.d.ts", 
         "default": "dist/index.js",   
      }
    }
  }
}
```

## 5) Type

And one more value was introduced. A `package.json` now includes the `type` field: 

```js
{
  “main”: “dist/index.js”, 
  "module" "dist/esm/index.js", 
  "types": "dist/esm/index.d.ts", 
  "type": "commonjs",
  "exports": { 
    ".": { 
      "require": { 
         "types": "dist/index.d.ts", 
         "default": "dist/index.js",
      },
      "import": { 
         "types": "dist/index.d.ts", 
         "default": "dist/index.js",   
      }
    }
  }
}
```

The `type` field tells node what it can expect if it sees a `.js` file. If it’s `commonjs`, it expects requires, if it’s `module`, it expects imports.

# B) What is the story with file extensions?

Imagine have a typescript file like this:

```js
// index.ts
import utils from './utils'
```

## CommonJS works

If this compiled to CommonJS, the import would work:

```js
// dist/index.js
const utils = require('./utils'); 
```

Node looks up `./utils.js`. 

## ESM…fails

But if you compiled it directly to an ESModule, Node would fail to resolve the module: 

```
// dist/esm/index.js
import utils from './utils'
```

This is because Node *requires* a specific file extension in relative imports.

## Fix 1: one bundle

One way to get around this issue, is to bundle the library into one file. This removes the need for an relative imports: 

```js
// dist/esm/index.js
// code for utils is inside this file
```

### Downside for Fix 1: Why bundle?

This works, but it’s arguable that you don’t *need* to bundle all the files together. 

Node already knows how to resolve files right?

## Fix 2: .js files

Another fix you can do is, to add a `.js` file to the extension: 

```js
// index.ts
import utils from './utils.js'
```

Typescript ignores the `.js` extension. This passes through to your dist file, and Node is happy again! 

### Downside for Fix 2: What about TS bundlers?

Well, this ends up breaking other bundlers. Some bundlers read typescript directly. They go ahead and read: 

```js
    // index.ts
    import utils from './utils.js'
```

And they break, because `utils.js` does not exist. 

## Fix 3: .ts files

The fix to that, is to write `.ts` in your file extension: 

```js
// index.ts
import utils from './utils.ts'
```

If you set a config in Typescript 5.7+, in dist this will convert to a `.js` extension. Modern bundlers that read typescript will continue to work. 

# C) How does a *purely* modern package.json look?

Now, technically, you can remove `module`, `types`, *and* our `commonjs` bundles.  Modern bundlers now *all* support `import`. So for example we could write: 

```js
{
  "type": "module",
  “main”: “dist/esm/index.js”,  
  "exports": { 
    ".": { 
      "types": "dist/index.d.ts", 
      "default": "dist/index.js",   
    }
  }
}
```

And things would work. Of course, any time you remove backwards compatibility…who knows what can break. Plus, remember you have to handle the issue with file extensions even in this modern version.

# So, how do people make dist files?

## 1) Direct tsc + module resolution

Some folks just rely on tsc, to remove typescript, and make the JS backwards compatible. They *do not* bundle everything into one file. 

This works well, but you have to be really careful about the module resolution. **If you don’t include full extensions, your library could break in Node.**

To make this work, you need to run two `tsc` commands. If you want to have the nifty `type` declaration in your package.json, you also have to be careful about how you run `tsc` (since tsc itself read from that `type` key and produces different javascript)

## 2) tshy (easy tsc)

To make using tsc easier, folks have built a library called `tshy`. This runs tsc, but with some convenience helpers: 

1. It runs both build steps
2. It creates a “package.json” file in `dist/commonjs` and `dist/esm`, which helps node resolve commonjs and module bundles
3. It makes sure you set the right kind of tsconfig for commonjs and esm

## 3) esbuild

Others use esbuild directly. Esbuild produces a single bundle, or can can produce multiple entrypoints. It’s super fast, but requires some configuration.

One downside of using esbuild, is that if you don’t run `tsc`, you may not get all the typescript benefits when you build. You could incidentally release some buggy build.

## 4) tsup (easy esbuild)

`tsup` is a wrapper over esbuild, that is great at building both a commonjs and an esmodule bundle.

# So, what should we do?

I think we have two decisions to make: 

## 1) Backwards compatible or not?

Some folks have decided to eschew commonjs alltogether. Most modern bundlers *do* understand ESModules. 

I don’t think we should do this though. I am still worried there’s going to be tools that don’t know the “latest and greatest” way, and would cause us to break. 

## 2) How to build?

I think the decision is between `tsup` and `tshy`: 

If we went with `tsup`, we would bundle everything into 1 file, so node wouldn't have to worry about module resolution for relative imports. But `tsup` doesn't run a full `tsc` check, so we may inadvertently have some bad build escape us.

If we went with tshy, we would still compile with tsc. tshy could handle some of the trickier parts of the build, for making esm and cjs work. 

I think the best bet is to move to `tshy`. `tshy` does what we do today, but with some tricks that make for a good dual bundle. 

## 3) What about file extensions?

To make our `module` build work for all bundlers, we'll need to have full extensions in our import. If you set `module` to `nodenext`, typescript makes sure you include the extension.  I went ahead and did this.

## 4) How to check things are well? 

I added a `@arethetypeswrong/cli` in the build step, which checks that our package.json looks right.

--- 

## Anything next? 

We currently transpile down to ES2015. I _don't_ think we ned to do that though. We can likely safely raise to ES2020 at least. If we did, the code would look nicer to the user -- for ex we wouldn't need generators for async/await. I'll work on that next. 

--- 

@dwwoelfel @nezaj @tonsky


dwwoelfel (approved):
LGTM!

PR #1181, opened 2025-05-07
Support join-and-set-presence for @instantdb/core

dwwoelfel:
Followup to https://github.com/instantdb/instant/pull/1176

In core, `db.joinRoom` now takes a third argument with the initial presence data:

```ts
db.joinRoom('roomType', 'room-id', {name: user.name})
```


stopachka (approved):
(discussed re opts map) -- soo cool!

PR #1184, opened 2025-05-07
Clone App

stopachka:
```
USAGE:
  ./scripts/clone_app.sh --env {prod|dev} --app-id APP_UUID --new-email EMAIL --new-title TITLE
```

A quick script that lets you clone an app. Wanted to make this to help one of our users test a dangerous change. 

@nezaj @dwwoelfel @tonsky

### DIFF
--- a/server/scripts/clone_app.sql
+++ b/server/scripts/clone_app.sql
@@ -1,0 +1,102 @@
+\set ON_ERROR_STOP on
+BEGIN;
+
+-- Create the new root app
+SELECT gen_random_uuid() AS new_app_id  \gset
+
+INSERT INTO apps (id, creator_id, title)
+SELECT  :'new_app_id',
+        CASE
+          WHEN NULLIF(:'creator_email', '') IS NULL        -- no email supplied
+          THEN creator_id                                  -- keep original creator
+          ELSE (
+            SELECT id
+            FROM   instant_users
+            WHERE  email = :'creator_email'                -- use supplied email
+          )
+        END,
+        COALESCE(NULLIF(:'new_title', ''), title || ' (clone)')
+FROM    apps
+WHERE   id = :'old_app_id';
+
+-- Build mapping tables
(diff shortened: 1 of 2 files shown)


server/scripts/clone_app.sql
```diff
@@ -0,0 +1,96 @@
...
+BEGIN;
+
+-- Create the new root app
+SELECT gen_random_uuid() AS new_app_id  \gset
+
+INSERT INTO apps (id, creator_id, title)
+SELECT  :'new_app_id',
+        (SELECT id
+         FROM   instant_users
+         WHERE  email = :'creator_email'),
```
dwwoelfel: Could we make `creator_email` optional and have it set the email by default to be the original app's creator email?

server/scripts/clone_app.sql
```diff
@@ -0,0 +1,96 @@
...
+
+-- Clone rules
+INSERT INTO rules (app_id, code)
+SELECT :'new_app_id', r.code
+FROM   rules r
+WHERE  r.app_id = :'old_app_id';
+
+-- Clone attrs
+INSERT INTO attrs (
+    id, app_id, value_type, cardinality, is_unique, is_indexed,
```
dwwoelfel, line 41: It would be nice if there was some way to use `*` instead of listing each column. It's going to be annoying to keep the script in sync with the database schema.

dwwoelfel (approved):
LGTM!

server/scripts/clone_app.sql
stopachka: Nice idea!

stopachka: Done!

server/scripts/clone_app.sql
stopachka: Yeah. ChatGPT is convinced this isn't easily possible, if we want to replace some column values. Going to look deeper into it.

PR #1185, opened 2025-05-08
Create a more robust staging environment

dwwoelfel:
Makes it easier to test things in staging by creating a more separate staging environment:

1. The backend now knows the difference between staging and prod
  a. It does many things like prod, but it sends metrics to staging in honeycomb, won't send messages to discord, uses the stripe test env, and uses a separate storage bucket and user
3. Vercel has a staging environment that deploys to `staging.instantdb.com`
4. You can deploy code to staging by appending the string `[staging]` to your commit message

This will let us test the full oauth login flow in staging and will help prevent us from doing something silly (e.g. sending welcome messages to users from staging).


stopachka (approved):
Biyuutiful!

PR #1187, opened 2025-05-08
fix react-native build

stopachka:
## Bugs 

We had two bugs with our react native build: 

1. Module resolution. 

React Native has it's own special module resolution. When you write: 

```js
import './Storage' 
```

Metro will look up: 

1. Storage.(platform).js 
2. Storage.js 

**We used this pattern to conditionally import async storage for native, and indexeddb for web** 

However, since we had explicitly written `./Storage.js` I _think_ this caused tsc to ignore our `Storage.native.js` file. Metro then picked up the wrong library

2. CLI picks up a file from dist

Our CLI specifically imports from `react-native/dist/cli`. This file directory changed with what we wrote. 

## Fixes

I rolled back our build for RN. 

## Future steps

I am _pretty_ sure Metro never uses our ESM build in React Native. I'll do some more research down the road about what the "right" way to build libraries for React Native is. 

@nezaj @dwwoelfel @tonsky

### DIFF
--- a/client/packages/core/src/Reactor.js
+++ b/client/packages/core/src/Reactor.js
@@ -178,9 +178,13 @@ export default class Reactor {
     if (typeof BroadcastChannel === 'function') {
       this._broadcastChannel = new BroadcastChannel('@instantdb');
       this._broadcastChannel.addEventListener('message', async (e) => {
-        if (e.data?.type === 'auth') {
-          const res = await this.getCurrentUser();
-          this.updateUser(res.user);
+        try {
+          if (e.data?.type === 'auth') {
+            const res = await this.getCurrentUser();
+            this.updateUser(res.user);
+          }
+        } catch (e) {
+          this._log.error('[error] handle broadcast channel', e);
         }
       });
     }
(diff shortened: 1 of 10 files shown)


client/packages/core/src/Reactor.js
```diff
@@ -1102,20 +1106,24 @@ export default class Reactor {
...
+          // If an admin token is provided for an app, we will
+          // skip all permission checks. This is an advanced feature,
+          // to let users write internal tools
+          // This option is not exposed in `Config`, as it's
+          // not ready for prime time
+          '__admin-token': this.config.__adminToken,
+        });
+      })
+      .catch((e) => {
+        this._log.error('[socket][error]', targetWs._id, e);
```
stopachka, line 1125: Added this logging, as Reactor was failing silently. In this case `getCurrentUser` failed, but the error was swallowed

client/packages/core/src/index.ts
```diff
@@ -663,10 +663,8 @@ function schemaHash(schema?: InstantSchemaDef<any, any, any>): string {
   }
 
   if (schemaHashStore.get(schema)) {
-    console.log('getting cached schema');
```
stopachka, line 666: Oops

client/sandbox/react-native-expo/app/config.js
```diff
@@ -2,4 +2,5 @@ export default {
   appId: process.env.EXPO_PUBLIC_INSTANT_APP_ID,
   apiURI: 'http://localhost:8888',
   websocketURI: 'ws://localhost:8888/runtime/session',
+  verbose: true,
```
stopachka, line 5: Added the nifty `verbose` flag, so we automatically see our logs in RN. H/t @dwwoelfel

dwwoelfel (approved):
Nothing's ever easy!

PR #1189, opened 2025-05-09
More judicious use of index hints

dwwoelfel:
Removes the hint where we aren't sure if postgres can make good use of the index.

There are a few places where we were telling postgres to use an index that wasn't efficient:

1. Exact matches on strings don't work well with the trgm index (it's better at `like`)
2. `is distinct from` (what we use with `not` to catch nils) doesn't seem to use the index very well. Postgres has to filter afterwards instead of putting it in the index condition.

Now we'll skip putting any hints for `:function` clauses and when we're matching on a string index. Hopefully postgres will chose the index if it makes sense.

This closed the gap for the 4 "bad" queries I was testing and preserved the improvement for the 3 "good" queries I tested. We'll see in the pg-hint-plan testing what other outliers there are.

### DIFF
--- a/server/src/instant/db/datalog.clj
+++ b/server/src/instant/db/datalog.clj
@@ -926,7 +926,18 @@
                :materialized
                :not-materialized)]]
     {:cte cte
-     :pg-hints [(pg-hint/index-scan triples-alias (pg-hint-index (:idx named-p)))]}))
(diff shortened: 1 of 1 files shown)


raiderss:
Great optimization techniques! We use similar approaches for our Eyes Store FiveM Tebex resources.

stopachka (approved):
Wonderful!

PR #1190, opened 2025-05-09
[www] Remove dead konami code

nezaj:
Noticed we had some dead code here so thought it would be nice to remove


stopachka (approved):
Niice!

PR #1191, opened 2025-05-09
[llms] Enable md extension to get raw docs

nezaj:
With this PR you can add `.md` to the end of our doc urls to get the extracted markdown. This can be useful for customizing context in an llm setting.

Also updates our llms.txt to point directly to these urls

See [Getting Started](https://instant-www-js-enable-md-jsv.vercel.app/docs.md) as an example

or [explore the docs ](https://instant-www-js-enable-md-jsv.vercel.app/docs) and add `.md` at the end

### DIFF
--- a/client/www/lib/markdoc.ts
+++ b/client/www/lib/markdoc.ts
@@ -1,0 +1,54 @@
+// Extract frontmatter from markdoc content and separate it from the
+// markdoc content
+function parseFrontmatter(content: string): {
+  frontmatter: any;
+  content: string;
+} {
+  const frontmatterRegex = /^---\s*\n([\s\S]*?)\n---\s*\n([\s\S]*)$/;
+  const match = content.match(frontmatterRegex);
+
+  if (!match) {
+    return { frontmatter: {}, content };
+  }
+
+  const frontmatterStr = match[1];
+  const remainingContent = match[2];
+
+  const frontmatter: Record<string, string> = {};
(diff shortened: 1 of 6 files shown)


client/www/lib/markdoc.ts
```diff
@@ -0,0 +1,54 @@
+// Extract frontmatter from markdoc content and separate it from the
+// markdoc content
+function parseFrontmatter(content: string): {
+  frontmatter: any;
```
stopachka, line 4: nit: This should be Record<string, any> ?

client/www/lib/markdoc.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  return { frontmatter, content: remainingContent };
+}
+
+// Remove `{% ... %}` tags from markdown content
+function sanitizeMarkdown(content: string): string {
+  return content.replace(/{%[\s\S]*?%}/g, '');
+}
+
+// Transform markdoc content into a markdown format that's sanitized for
```
stopachka, line 36: nit: I don't think the comments in this file are necessary

stopachka (approved):
Cool!

client/www/lib/markdoc.ts
nezaj: @stopachka I actually wanted to leave this in to remind myself why I'm bothering to do this (since our files are written in markdown, but actually we need to clean them up since they use markdoc tags which are not used by LLMs)

PR #1192, opened 2025-05-09
Made sandbox package names match folder names, made logs quieter

tonsky:
It was unnecessarily hard to see `issue-demo-idbinwisr` in logs and figure out that it corresponds to `sandbox/vanilla-js-nuxt`


stopachka (approved):
Thank you @tonsky !!

PR #1193, opened 2025-05-12
[Sandbox] Enable db.tx

nezaj:
Was using the sandbox to repro a potential bug and noticed we get type warnings on `db.tx` when in fact that should work and is what we recommend folks use in the docs.

This PR makes it so `db.tx` works in the Sandbox

**Before**

CleanShot 2025-05-12 at 08 16 49@2x

**After**

CleanShot 2025-05-12 at 08 31 03@2x


dwwoelfel (approved):
LGTM!

PR #1194, opened 2025-05-12
Ignore data in join-room if it is null

dwwoelfel:
Fixes a bug where we would reset the presence to null in `@instantdb/core` if you called:

```js
  const room = db.joinRoom('name', 'id', { initialPresence: { example: 'value' } });
  room.subscribePresence({}, (response) => console.log(response));
```

`room.subscribePresence` also calls `joinRoom` with `null` for initialData (unless you call `room.subscribePresence({data: initialDataHere}, cb)`). That would set the presence data back to null until you called `setPresence` again. 

Now we'll ignore that `null` state in our join room handler.

### DIFF
--- a/client/pnpm-lock.yaml
+++ b/client/pnpm-lock.yaml
@@ -397,6 +397,9 @@ importers:
       '@instantdb/admin':
         specifier: workspace:*
         version: link:../../packages/admin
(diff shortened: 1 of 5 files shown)


nezaj (approved):
Nice!

PR #1195, opened 2025-05-13
Remove num tx from intern signup

nezaj:
Fetching tx counts causes the recent sign-ups query to timeout. I tried re-working the query a bit but looks like to make it faster we'll need to add some additional indexes.

For now thought it would be nice to get this page working again to have a quicker overview into how recent users are finding us / what they're trying to build


stopachka (approved):
SG!

PR #1196, opened 2025-05-14
[rn-google-signin] remove scopes

stopachka:
You don't need the `drive.readonly` scope to get auth to work. Originally included this, from perusing rn-google-signin's docs. 

One of our users mentioned that webClientId is required as well. I tried running this in my iOS simulator, and it worked, but maybe it's necessary in some other context. I will get some more info from them and push up a fix. 

@dwwoelfel @nezaj @tonsky


dwwoelfel (approved):
LGTM!

PR #1198, opened 2025-05-14
Show errors in query inspector

dwwoelfel:
Currently we just show "{}", [which can be confusing to users](https://discord.com/channels/1031957483243188235/1370341848304058500/1372154115609853952).

Now we'll show the error message:

image


nezaj (approved):
Nice!

PR #1200, opened 2025-05-15
Some help debugging CLI indexing jobs reporting

tonsky:
Reported by Kosmik. Does not fix the issue but should print better reporting

image

### DIFF
--- a/client/packages/cli/src/index.js
+++ b/client/packages/cli/src/index.js
@@ -14,6 +14,7 @@ import { loadConfig } from 'unconfig';
 import { packageDirectory } from 'pkg-dir';
 import openInBrowser from 'open';
 import ora from 'ora';
+import semver from 'semver';
 import terminalLink from 'terminal-link';
 import { exec } from 'child_process';
 import { promisify } from 'util';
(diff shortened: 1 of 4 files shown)


client/packages/cli/src/index.js
```diff
@@ -847,6 +847,7 @@ function indexingJobCompletedActionMessage(job) {
   if (job.job_type === 'remove-required') {
     return `removing required constraint from ${job.attr_name}`;
   }
+  return `${job.job_type} (${job.id})`;
```
nezaj: Curious, is the reason to return job.id so users can report this error to us?

tonsky: Yes, and we can look it up in a DB and see what’s wrong

nezaj:
```suggestion
  return `unexpected job type ${job.job_type} - please ping us on discord with this job id (${job.id})`;
```

nezaj: Gotcha! Made a suggestion for it to be a bit more explicit

nezaj (approved):

client/packages/cli/src/index.js
tonsky: Applied

tonsky:
Added version check to CLI and to the backend. I guess we need to release backend first :)

stopachka (approved):
SGTM!

client/packages/cli/src/index.js
```diff
@@ -412,7 +413,37 @@ program
...
+      console.log(
+        `${chalk.red('ERROR')} Your CLI version ${chalk.red(cliVersion)} is too old to use with Instant backend, requires ${chalk.red(minVersion)} or later.`,
+      );
+      process.exit(1);
+    }
+  }
+}
+
 async function handlePush(bag, opts) {
+  await checkVersion();
```
stopachka, line 446: I worry about the extra latency here, but probably not too big a deal. We can see how it rolls 🫡

PR #1201, opened 2025-05-15
Include state in oauth error response

dwwoelfel:
Errors should include the state param so that they can be tied back to the original request.


nezaj (approved):
Nice!

PR #1205, opened 2025-05-16
Prefix personal access tokens

dwwoelfel:
Prefixes the personal access tokens with `per_` so that we can dispatch on the token type.

I want to allow platform tokens to do `instant-cli push` and `instant-cli pull`, but I need a way to identify that it's a platform token and not a regular user token.

Also changes how we display the tokens--we'll now hash the token when we store it and look it up by the hash. We'll only show you the token when you first create it.

There's a migration to allow older tokens to still function as they did before (but only newer tokens will be able to do `push` and `pull`).

Also includes the changes from https://github.com/instantdb/instant/pull/1206

### Migration plan
1. Merge into main
2. run migration
3. deploy code (there will be a short period between step 2 and 3 where people won't be able to generate tokens)

### DIFF
--- a/server/resources/migrations/67_personal_access_token_lookup.up.sql
+++ b/server/resources/migrations/67_personal_access_token_lookup.up.sql
@@ -1,0 +1,7 @@
+alter table instant_personal_access_tokens add column lookup_key bytea;
+
+update instant_personal_access_tokens set lookup_key = sha256(cast(cast(id as text) as bytea));
+
+alter table instant_personal_access_tokens alter column lookup_key set not null;
+
+create index on instant_personal_access_tokens (lookup_key);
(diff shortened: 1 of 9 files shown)


server/resources/migrations/67_personal_access_token_lookup.up.sql
```diff
@@ -0,0 +1,7 @@
+alter table instant_personal_access_tokens add column lookup_key bytea;
+
+update instant_personal_access_tokens set lookup_key = sha256(cast(cast(id as text) as bytea));
```
stopachka, line 3: 🔥

stopachka (approved):
Super cool!

PR #1206, opened 2025-05-16
Allow platform tokens for cli push/pull

dwwoelfel:
*Note to self: switch base branch back to main before merging*

Depends on https://github.com/instantdb/instant/pull/1205

Allows the `schema` and `rule` endpoints we use from the cli to be used with platform tokens.

This will allow us to do a `npx instant-cli pull` to bootstrap the schema and rules when selecting an existing Instant app on bolt.


stopachka (approved):
Ah, so nice!

dwwoelfel:
Going to merge this into #1205 and deploy both at the same time

PR #1208, opened 2025-05-19
Optionally include schema and perms in superadmin list apps

dwwoelfel:
Adds an `include` query param to `/superadmin/apps` that accepts a comma-separated list of extra data to include in the response. You can provide `schema`, and `perms`.

Initially, I tried to return attrs and rules from `app-model/list-by-creator-id`, but it would require returning the attrs as json from the database. That's harder than it seems because we join into the idents table so you can't just use `row_to_json`. But at least with the new `fetch-by-app-ids` methods we only generate 3 queries instead of 1 query to fetch apps + count(app_id) queries to fetch schema and rules.

### DIFF
--- a/server/src/instant/db/model/attr.clj
+++ b/server/src/instant/db/model/attr.clj
@@ -767,6 +767,33 @@
      ;; Don't cache if we're using a custom connection
      (get-by-app-id* conn app-id))))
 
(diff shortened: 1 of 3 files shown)


stopachka (approved):
Yess!

PR #1209, opened 2025-05-20
Adds a new `@instantdb/platform` package

dwwoelfel:
Adds an `@instantdb/platform` package. 

Right now it just has the `generateSchemaTypescriptFile` from `instant-cli` in it, which I'll use to generate the `instant.schema.ts` file for bolt.

I moved the `generateSchemaTypescriptFile` from `instant-cli` into this package and it gets imported from instant-cli. The plan is to move more and more code from the cli into the platform package, until the cli package only has cli-specific code (e.g. reading from local files, doing the cli auth, etc.).

I made a few minor changes to `generateSchemaTypescriptFile`:
 1. It's typed!
 2. The formatting before running prettier was a bit nicer. `prettier` would add too much to the bundle size, and I want people to be able to pull `@instantdb/platform` into their frontends.


stopachka (approved):
Niice!

PR #1212, opened 2025-05-21
Add `etype` and `label` to `attrs` table (part 1 of 2)

tonsky:
Currently almost all operations that include `attrs` have to join `idents` table to get an attr name or a namespace name.

This PR adds these columns directly to `attrs` table (`etype`, `label`, `reverse_etype`, `reverse_label`). It saves us an extra join at the cost of a little denormalization.

`idents` table is kept around for backwards compatibility, it might take some doing to get rid of it.

Not queries have been updated to use new fields. This is more to simplify writing future queries. But I did update some just to demonstrate how it feels.

Deploy plan:

1. Upgrade schema to 68
2. Deploy this PR #1212
3. Upgrade schema to 69
4. Deploy PR #1215

### DIFF
--- a/server/resources/migrations/68_attrs_etype_label_schema.up.sql
+++ b/server/resources/migrations/68_attrs_etype_label_schema.up.sql
@@ -1,0 +1,46 @@
+ALTER TABLE attrs ADD COLUMN etype text;
+ALTER TABLE attrs ADD COLUMN label text;
+ALTER TABLE attrs ADD COLUMN reverse_etype text;
+ALTER TABLE attrs ADD COLUMN reverse_label text;
+
+ALTER TABLE attrs
+ADD CONSTRAINT attrs_etype_label_unique UNIQUE (app_id, etype, label);
+
+ALTER TABLE attrs
+ADD CONSTRAINT attrs_reverse_etype_label_unique UNIQUE (app_id, reverse_etype, reverse_label);
+
+CREATE OR REPLACE FUNCTION check_attrs_unique_names ()
+  RETURNS TRIGGER
+  AS $$
+BEGIN
+  IF EXISTS (
+    SELECT 1
+      FROM attrs
+     WHERE (app_id, reverse_etype, reverse_label) = (NEW.app_id, NEW.etype, NEW.label)
+  ) THEN
+    RAISE EXCEPTION 'trigger violation trg_attrs_unique_names'
+       USING DETAIL = format('Key (app_id, reverse_etype, reverse_label)=(%s, %s, %s) already exists', NEW.app_id, NEW.etype, NEW.label),
+              TABLE = 'attrs',
+            ERRCODE = '23505';
+  END IF;
+
+  IF EXISTS (
+    SELECT 1
+      FROM attrs
+     WHERE (app_id, etype, label) = (NEW.app_id, NEW.reverse_etype, NEW.reverse_label)
+  ) THEN
+    RAISE EXCEPTION 'trigger violation trg_attrs_unique_names'
+       USING DETAIL = format('Key (app_id, etype, label)=(%s, %s, %s) already exists', NEW.app_id, NEW.reverse_etype, NEW.reverse_label),
+              TABLE = 'attrs',
+            ERRCODE = '23505';
+  END IF;
+
+  RETURN NEW;
+END;
+$$
+LANGUAGE plpgsql;
+
+CREATE TRIGGER trg_attrs_unique_names
+  BEFORE INSERT OR UPDATE
+  ON attrs
+  FOR EACH ROW EXECUTE FUNCTION check_attrs_unique_names ();
--- a/server/src/instant/db/model/attr.clj
+++ b/server/src/instant/db/model/attr.clj
@@ -190,27 +190,36 @@
   "Manual reflection of postgres attr table columns"
   [:id :app-id :value-type
    :cardinality :is-unique :is-indexed :is-required
-   :forward-ident :reverse-ident :on-delete :on-delete-reverse
+   :forward-ident :etype :label
+   :reverse-ident :reverse-etype :reverse-label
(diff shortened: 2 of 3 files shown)


stopachka (approved):
Niice!

server/resources/migrations/68_attrs_etype_label.up.sql
```diff
@@ -0,0 +1,56 @@
...
+CREATE OR REPLACE FUNCTION check_attrs_unique_names ()
+  RETURNS TRIGGER
+  AS $$
+BEGIN
+  IF EXISTS (
+    SELECT 1
+      FROM attrs
+     WHERE (app_id, reverse_etype, reverse_label) = (NEW.app_id, NEW.etype, NEW.label)
+  ) THEN
+    RAISE EXCEPTION 'Attr with (app_id, reverse_etype, reverse_label) (%, %, %) already exist', NEW.app_id, NEW.etype, NEW.label;
```
dwwoelfel: What does the error the developer sees in the console look like after this change?

tonsky: Same, really. I think `idents` constraint kicks in before this. When (if) we get rid of idents completely, then we might take a look at this.

```
ExceptionInfo: [instant-exception] `ttt` already exists on `xxx` {:instant.util.exception/type :instant.util.exception/record-not-unique, :instant.util.exception/message `ttt` already exists on `xxx`, :instant.util.exception/hint {:record-type :ident, :etype xxx, :label ttt}, :instant.util.exception/trace-id cf4b2e5abfc5bede654cc43b6f56a0f0}
  instant.util.exception/throw+                               exception.clj 79
  instant.util.exception/throw-record-not-unique!             exception.clj 263
  instant.util.exception/translate-and-throw-psql-exception!  exception.clj 482
  ... 34 common elements
Caused by: PSQLException: ERROR: duplicate key value violates unique constraint "app_ident_uq"
  Detail: Key (app_id, etype, label)=(15a25197-a82b-4109-aaca-bdfcb4c0caa4, xxx, ttt) already exists.

ExceptionInfo: [instant-exception] `xxx` already exists on `yyy` {:instant.util.exception/type :instant.util.exception/record-not-unique, :instant.util.exception/message `xxx` already exists on `yyy`, :instant.util.exception/hint {:record-type :ident, :etype yyy, :label xxx}, :instant.util.exception/trace-id 0e4b8c0f13bc5cb56876e7654ea4c200}
  instant.util.exception/throw+                               exception.clj 79
  instant.util.exception/throw-record-not-unique!             exception.clj 263
  instant.util.exception/translate-and-throw-psql-exception!  exception.clj 482
  ... 34 common elements
Caused by: PSQLException: ERROR: duplicate key value violates unique constraint "app_ident_uq"
  Detail: Key (app_id, etype, label)=(15a25197-a82b-4109-aaca-bdfcb4c0caa4, yyy, xxx) already exists.
```

server/src/instant/db/model/attr.clj
```diff
@@ -532,40 +534,39 @@
      conn
      (hsql/format
       {:with (concat
-              (if-let [attr-table-updates
-                       (seq (changes-that-require-attr-model-updates updates))]
```
dwwoelfel, line 536: Why don't we need to run this check any more?

server/resources/migrations/68_attrs_etype_label.up.sql
dwwoelfel: It will be easier for us to parse this later if it throws something in the same format as the duplicate constraint. I think something like this will do it (but I haven't tested it):

```
RAISE EXCEPTION 'duplicate key value violates unique constraint unique_attr_name',
  USING ERRCODE = '23505',
        DETAIL = 'Key (app_id, reverse_etype, reverse_label)=(%s, %s, %s) already exists.', NEW.app_id, NEW.etype, NEW.label,
        TABLE = 'attrs',
        CONSTRAINT = 'unique_attr_name';
```

server/src/instant/db/model/attr.clj
tonsky: Because all changes require attr-model-updates now. It was used before to separate changes that only touched `idents`, but now if you touch `idents` you touch attrs too

server/resources/migrations/68_attrs_etype_label.up.sql
tonsky: No, this might be confusing. This indicates that it comes from a constraint when in fact it comes from trigger. I’d prefer if it was explicit about what broke and where

tonsky: @dwwoelfel updated, please take a look

server/resources/migrations/68_attrs_etype_label_schema.up.sql
```diff
@@ -0,0 +1,46 @@
...
+ALTER TABLE attrs ADD COLUMN reverse_etype text;
+ALTER TABLE attrs ADD COLUMN reverse_label text;
+
+ALTER TABLE attrs
+ADD CONSTRAINT attrs_etype_label_unique UNIQUE (app_id, etype, label);
+
+ALTER TABLE attrs
+ADD CONSTRAINT attrs_reverse_etype_label_unique UNIQUE (app_id, reverse_etype, reverse_label);
+
+CREATE OR REPLACE FUNCTION check_attrs_unique_names ()
```
dwwoelfel, line 12: Should this part go in part 2? I think you'll hit errors with the null values.

tonsky: No, nulls don’t count in unique

dwwoelfel: Oh right, null won't equal null here.

dwwoelfel (approved):
LGTM!

PR #1214, opened 2025-05-21
Fix deleting the redirect url

dwwoelfel:
Fixes a bug where you couldn't click the delete button redirect because it would disappear when your mouse moved off of the url.

Changing from `display: hidden` to `visibility: hidden` keeps the button from disappearing.


stopachka (approved):
TIL group

PR #1216, opened 2025-05-22
Clean up stale ephemeral sessions (1 of 3)

tonsky:
I noticed it on tonsky.me, and #kosmik-js reported same recently. Sometimes ephemeral sessions get “stuck”, showing people present that are long gone, with no way to clean it up.

This PR adds periodic 30s heartbeat to every ws connection (new op `"heartbeat"`) and a periodic task on a server to clean up sessions that have not sent heartbeat in 2 to 3 intervals.

### DIFF
--- a/server/src/instant/util/hazelcast.clj
+++ b/server/src/instant/util/hazelcast.clj
@@ -1,6 +1,5 @@
 (ns instant.util.hazelcast
   (:require
-   [instant.util.coll :as ucoll]
    [instant.util.uuid :as uuid-util]
    [medley.core :refer [update-existing]]
    [taoensso.nippy :as nippy])
@@ -39,9 +38,7 @@
 (def room-broadcast-type-id 6)
 (def task-type-id 7)
 (def join-room-type-id 8)
-(def patch-type-id 9)
-
-(declare patch-assoc patch-assoc-in patch-merge-in patch-dissoc patch-dissoc-in)
+(def join-room-v3-type-id 9)
 
 ;; --------
 ;; Room key
@@ -160,6 +157,36 @@
   (make-serializer-config JoinRoomMergeV2
                           join-room-serializer))
 
+(defrecord JoinRoomMergeV3 [^UUID session-id ^String instance-id ^UUID user-id data]
+  BiFunction
+  (apply [_ room-data _]
+    (update room-data
+            session-id
+            (fn [existing]
+              (merge existing
+                     {:peer-id     session-id
+                      :instance-id instance-id
+                      :user        (when user-id
+                                     {:id user-id})}
+                     (if data
+                       {:data data}
+                       {:data (or (:data existing) {})}))))))
+
+(def ^ByteArraySerializer join-room-v3-serializer
+  (reify ByteArraySerializer
+    (getTypeId [_]
+      join-room-v3-type-id)
+    (write ^bytes [_ obj]
+      (let [{:keys [^UUID session-id ^String instance-id ^UUID user-id data]} obj]
+        (nippy/fast-freeze [session-id instance-id user-id data])))
+    (read [_ ^bytes in]
+      (let [[session-id instance-id user-id data] (nippy/fast-thaw in)]
+        (->JoinRoomMergeV3 session-id instance-id user-id data)))
+    (destroy [_])))
+
+(def join-room-v3-config
+  (make-serializer-config JoinRoomMergeV3
+                          join-room-v3-serializer))
 
 ;; ------------
 ;; Set presence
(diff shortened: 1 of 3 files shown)


server/src/instant/reactive/ephemeral.clj
```diff
@@ -221,6 +225,25 @@
                                              session-ids)))))
     (hazelcast/remove-session! (get-hz-rooms-map) room-key sess-id)))
 
+(defn clean-sessions []
+  (doseq [:let [now    (System/currentTimeMillis)
```
stopachka: Right now, the server also does a ping job for websocket sessions: 

https://github.com/instantdb/instant/blob/main/server/src/instant/lib/ring/websocket.clj#L93-L109

I had two questions when thinking through this PR: 

**1. Why was the server ping-job not enough to clean those sessions?** 

Maybe one reason is that if the server sends a ping, the client responds, but perhaps when the client is idle, it won't _initiate_ a ping. 

If this this case, I think your idea of sending a ping from the client is much better. 

**2. Should the ping job be in ephemeral.clj?** 

It's a bit surprising the the work is in `ephemeral.clj`, mainly because I would expect that we only need to clean up sessions in `session.clj`, and that should get reflected onto hazelcast. 

Would love thoughts!

tonsky: Oh nice, maybe I can reuse ping timings from there.

1. I’m not sure I know what are you talking about. If client doesn’t sent `pong`, whether idle or not, we should close the connection. 
2. So my thinking was: web socket lifecycle events might be unreliable, somehow ephemeral state might get stuck, and we should check and clean that periodically

tonsky: I don’t see anything wrong with your websocket ping impl, so we assume that works. At the same time, both me and kosmik observe “stuck” ephemeral state. So there might be some sort of race condition, or HZ sync bug, or smth. This PR addresses that, that’s why it’s in ephemeral.clj

dwwoelfel: My guess is that there's orphaned data in the hazelcast map from a session that was closed but wasn't removed from hazelcast or there is some race where the data gets added to hazelcast after the session is closed.

If that's the case, I'm not sure this would help because `rs/session` would return null.

One idea would be to encode the instance-id into the session id (or keep track of the instance-id somehow). Then we'd know that we need to remove the session from hazelcast if `rs/session` is null on the instance we're running from and we could remove anything for an instance that was shut down.

tonsky: Just realized we can’t just check sessions as they might be on another instance. So I think we should add to hz:

- list of instances with heartbeasts 
- instance field to session

tonsky: Do you guys think this could happen because of redeploy? New instances share cluster with old instances, right? So all the HZ state transfers over. Now imagine we shut HZ down before all of the websocket onClose callbacks fire. That will leave some sessions in there forever. @dwwoelfel @stopachka

server/src/instant/util/hazelcast.clj
```diff
@@ -28,15 +31,25 @@
       (.setTypeClass protocol)
       (.setImplementation serializer)))
 
+;; Must be unique within the project
+(def remove-session-type-id 1)
```
dwwoelfel, line 34: :heart:

server/src/instant/util/hazelcast.clj
```diff
@@ -228,13 +241,75 @@
   (make-serializer-config Task
                           task-serializer))
 
+;; -----------------
+;; Path serializer
+
+(defrecord Patch [edits]
```
dwwoelfel: I'd rather not create a single Patch method because it loses two properties of the previous one-record-per-type approach:

1. We transfer the minimal amount of data through hazelcast--it just needs to transfer the type-id and the serialized data. We've tried to keep the serialized data as small as possible. I want people to think about what they're serializing before they add new methods so that they can optimize it and don't do something silly like serialize a function.
2. It helps us to be careful about ensuring that we're not using new methods before they're available on every instance. I think it's going to be too easy to modify the patch method without realizing that the other machines don't have the same function body.

dwwoelfel:
> * Put heartbeats there, `{<instance-id> {:heartbeat <timestamp>}}`

Hazelcast has a map of the machines in the cluster (`(.getMembers (.getCluster (:hz @eph/hz)))`). If we use that then we don't have to maintain a separate map with heartbeats. It might already have the instance id in it, but if not you can set the instance id in the config:

```
modified   server/src/instant/reactive/ephemeral.clj
@@ -55,7 +55,11 @@
         tcp-ip-config        (.getTcpIpConfig join-config)
         aws-config           (.getAwsConfig join-config)
         serialization-config (.getSerializationConfig config)
-        metrics-config       (.getMetricsConfig config)]
+        metrics-config       (.getMetricsConfig config)
+        member-attribute-config (doto (com.hazelcast.config.MemberAttributeConfig.)
+                                  (.setAttribute "instance-id" (or @config/instance-id
+                                                                   "dev")))]
+    (.setMemberAttributeConfig config member-attribute-config)
     (.setInstanceName config instance-name)
     (.setEnabled (.getMulticastConfig join-config) false)
     (case env
```

server/src/instant/util/hazelcast.clj
tonsky: 1. From what I see currently it’s a lot of `nippy/fast-freeze` mostly, so not really optimized. The only really optimized one is `RoomKeyV1` and `RemoveSessionMergeV1`. Patch is `fast-freeze` too, so I was just following the pattern.
2. Well the idea was that you won’t need to modify patch. It kinda covers all use-cases: assoc, merge, dissoc. Everything else is probably a custom type anyway. We can put a note about not changing it.

Are you sure performance is that critical there? It feels a little bit un-clojurely, creating two classes, three functions and two-step deploy every time you need to assoc a field in a map. We can give fields and operations 1-letter names if that helps

dwwoelfel (approved):
LGTM!

PR #1221, opened 2025-05-25
Faster check to validated required attrs in the indexing job

dwwoelfel:
Patches an issue where setting required causes the query to time out while we verify that there are no entities that are missing the attribute.

If I'm reading it right, the old query checked for all entities that had any attribute with a matching etype to the required attr. The new query just looks for entities with the id attr. That should be equivalent.

The new query also looks up the id-attr from the `attrs` on the clojure side instead of doing it in the database.

I think we will eventually need an approach where we perform the check in batches, but this should hold us off for a little longer.

It also might help to require that the attribute is indexed to set it as required?

Also adds tests for "required" and "remove-required" jobs.


nezaj (approved):
Nice!

PR #1224, opened 2025-05-27
[Profiling] Add execution time to Sandbox

nezaj:
It would be nice to start giving users more observability for their queries and transactions. One simple win is to include approximate execution time in the Sandbox. This way users can at least get a sense for how long their slow queries are taking

CleanShot 2025-05-27 at 16 52 59@2x


dwwoelfel (approved):
Nice!

PR #1225, opened 2025-05-29
Schema push improvements for platform sdk

dwwoelfel:
A few small changes to make the platform SDK a bit nicer:

1. Allows the platform token to fetch a group of indexing jobs. I set the scope to `:apps/write`, since you'd only need this if you were modifying the schema.
2. The required job puts a list of failed entity ids into an error_data field on the job. I include those in the common `invalid-triples-sample` field when you fetch the job for the client.
3. I include the job-id in the steps we return from schema/push/apply so that it's easier to tie the job to the step. This lets me return a list of `steps` from platform sdk that includes information about the background job.

### DIFF
--- a/server/src/instant/db/indexing_jobs.clj
+++ b/server/src/instant/db/indexing_jobs.clj
@@ -132,7 +132,30 @@
 
               [:inline missing-required-error]
               {:select [[[:json_agg 't]]]
-               :from [[{:select :* :from 'triples :where :false} 't]]}] :invalid-triples-sample]]
+               :from [[{:select [:t.entity-id [nil :value] ["null" :json-type]]
+                        :from [[:triples :t]]
+                        :limit 10
+                        :where [:and
+                                [:= :app-id app-id]
+                                [:= :t.app_id :j.app_id]
+                                [:= :t.attr_id {:select :id
+                                                :from :attrs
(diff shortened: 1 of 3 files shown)


server/src/instant/db/indexing_jobs.clj
```diff
@@ -245,16 +268,19 @@
...
         indexed-attr (attr-model/seek-by-id attr_id attrs)
         _ (assert indexed-attr (str "no attr found with id " attr_id))]
     (if (not= (:value-type indexed-attr) :blob)
       ;; Just return false if it's not a blob
       [:= [:inline 1] [:inline 0]]
       [:and
        [:= :triples.app-id app_id]
-       [:= :triples.attr-id id-attr-id]
+       [:= :triples.attr-id (:id id-attr)]
+       (when (and (:unique? id-attr)
```
dwwoelfel, line 281: Applies the same optimization from the required check: https://github.com/instantdb/instant/pull/1221/commits/8e0e60f2e1362f9614f47628dccabb1ca74dcb62#diff-cdcec9fc9613966c0a6a5b3a6047e47ff655a36b1613164fcf49a78f385fb068R810

stopachka (approved):
Niice!

PR #1228, opened 2025-05-30
Put session count on left in overview, show sessions per machine

dwwoelfel:
Couple of small changes to the overview:

1. Puts the number of sessions for an app on the left so that we don't have to scroll
2. Shows the count of sessions per instance next to the total

It will look something like this:

image

### DIFF
--- a/client/www/pages/intern/overview.tsx
+++ b/client/www/pages/intern/overview.tsx
@@ -120,6 +120,14 @@ function flattenedSessionReports(machineToReport: any) {
   return items;
 }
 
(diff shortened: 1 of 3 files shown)


client/www/pages/intern/overview.tsx
```diff
@@ -120,6 +120,14 @@ function flattenedSessionReports(machineToReport: any) {
   return items;
 }
 
+function makeMachineSummary(machineToReport: any): Record<string, number> {
```
stopachka, line 123: niice!

PR #1232, opened 2025-05-30
Clean up stale ephemeral sessions (2 of 3) 

tonsky:
Continuing #1216, this actually adds `instance-id` to all ephemeral session data and schedules a job to clean orphaned sessions. Session is considered orphaned if:

- it has no `instance-id` (leftovers from previous deploy)
- it has `instance-id` that is no longer part of the hz cluster
- `instance-id` matches current machine and `rs/session` returns `nil`

Third PR will just clean up JoinRoomV2 message type

### DIFF
--- a/server/src/instant/reactive/ephemeral.clj
+++ b/server/src/instant/reactive/ephemeral.clj
@@ -1,6 +1,7 @@
 (ns instant.reactive.ephemeral
   "Handles our ephemeral data apis for a session (presence, cursors)"
   (:require
+   [chime.core :as chime-core]
    [datascript.core :as ds]
    [editscript.core :as editscript]
    [instant.config :as config]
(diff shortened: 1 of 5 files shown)


server/src/instant/reactive/ephemeral.clj
```diff
@@ -225,12 +236,36 @@
...
     (hazelcast/remove-session! (get-hz-rooms-map) room-key sess-id)))
 
+(defn clean-orphan-sessions [_time]
+  (let [instance-ids (-> (:hz @hz)
+                         HazelcastInstance/.getCluster
+                         Cluster/.getMembers
+                         (->>
+                          (map #(Member/.getAttribute % "instance-id"))
+                          (into #{})))]
+    (doseq [^Map$Entry entry (IMap/.entrySet (:hz-rooms-map @hz))
```
dwwoelfel: You could do `IMap/.localKeySet` to get only the keys that are on this machine. That way you wouldn't have to fetch the full set of objects on every machine.

The cluster as a whole should still process every entry.

server/src/instant/reactive/ephemeral.clj
```diff
@@ -225,12 +236,36 @@
                                              session-ids)))))
     (hazelcast/remove-session! (get-hz-rooms-map) room-key sess-id)))
 
+(defn clean-orphan-sessions [_time]
+  (let [instance-ids (-> (:hz @hz)
```
dwwoelfel: Could we put a tracer/with-span! around the whole thing? It would be interesting to see how long this process takes.

server/src/instant/reactive/ephemeral.clj
tonsky: You can for filtering out sessions that belong to dead instances. You can’t for sessions that belong to alive instances, have their connection closed but still hang around. Since there’s no guarantee that if session is open on current machine, its room will be in `localKeySet`

server/src/instant/reactive/ephemeral.clj
tonsky: I guess around 60-100 ms

Screenshot 2025-06-02 at 18 17 29

PR #1233, opened 2025-05-30
Add scopes to client-only oauth flow

dwwoelfel:
I accidentally hard-coded the scopes as `apps-write`. Now you have to pass in the scopes you want when you start the OAuth flow.

Also adds a basic playground at `/play/platform-demo`


stopachka (approved):
Niice!

PR #1242, opened 2025-06-02
Fix longest wait time

dwwoelfel:
This would always error when there was anything in the queue because the transducer wasn't given an initial value, so it would call `(min)`, which throws an `Wrong number of args (0) passed to: clojure.core/min` erorr.

Now we give it an initial value of `now`.


tonsky:
Ah damn, I thought it was like `reduce`, just combine first two arguments or use the first one if there’s only one

PR #1244, opened 2025-06-02
Disable hints for pagination also

dwwoelfel:
Saw some bad plans for queries that use pagination, so let's just ignore hints for that part of the query for now.

I also noticed we're getting a lot of the same queries in the test pipeline, so I added a cache to try to filter out queries we've already seen. It uses a ttl cache with a ttl of 1 minute--if we see the same query hash twice in that period, then we'll only run the query once (that's not exactly true, b/c we check when we put the query in the queue, but don't set the flag until the query executes, but it's close enough).

### DIFF
--- a/server/src/instant/db/instaql.clj
+++ b/server/src/instant/db/instaql.clj
@@ -3,7 +3,6 @@
    [clojure.set :as set :refer [map-invert]]
    [clojure.spec.alpha :as s]
    [clojure.string :as string]
-   [clojure.walk :as walk]
    [honey.sql :as hsql]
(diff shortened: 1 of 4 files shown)


server/src/instant/db/instaql.clj
```diff
@@ -1173,33 +1173,6 @@
      :forms forms
      :referenced-etypes referenced-etypes}))
 
-(defn clean-where-for-hash [where]
```
dwwoelfel, line 1176: This just moved into the `instaql-util` file so that I can import it from the hint plan testing ns.

PR #1245, opened 2025-06-02
Update pg_hint_plan installation instructions

dwwoelfel:
Puts the pg_hint_plan library in `/usr/local/lib` instead of the postgres application directory. Mac didn't like us messing around in a signed application bundle.

This method should also make it easier to keep pg_hint_plan working after postgres.app updates.


server/README.md
```diff
@@ -44,12 +44,13 @@ On a mac using postgres.app, it looks something like this:
 git clone https://github.com/ossc-db/pg_hint_plan.git
 git checkout PG16
```
stopachka, line 46:
```suggestion
cd pg_hint_plan
git checkout PG16
```

PR #1246, opened 2025-06-03
Don't send an error if we fail to handle the error op

dwwoelfel:
If we timeout while sending an error, then we'll send another error informing the session that we failed to send the error. We can get into an infinite loop of errors until the session eventually disconnects.

Easier to review with whitespace off: https://github.com/instantdb/instant/pull/1246/files?w=1


tonsky:
Looks good to me!

PR #1247, opened 2025-06-03
Fair grouped-queue

dwwoelfel:
The main idea here is to put an item at the end of the queue instead of the front of the queue when we get a new item for the same sub-queue.

With the existing code, we do a recur inside of the function that processes the item. That's equivalent to putting the item at the front of the queue. 

With this PR, we submit a new task to the executor when we find a new item in our queue. The executor is FIFO, so that's equivalent to putting the item at the back of the queue.

That does mean that there will be more `locking` on the queue. It would be nice if we could lock on the sub-queue's key, but I don't see a good way to accomplish that. I think the effect will be minimal because we probably don't recur that often.

There are a couple of other minor changes that will help things be a little more efficient:
1. We use `execute` instead of `submit`, which saves us an extra wrapper around the task. We don't care about the return value, so no need to use `submit`. More info on that [in this stack overflow post](https://stackoverflow.com/questions/18730290/what-is-the-difference-between-executorservice-submit-and-executorservice-execut)
2. We create a `ProcessTask` that implements `Runnable` so that we don't have to do a new allocation when we resubmit the item to the executor

### DIFF
--- a/server/src/instant/grouped_queue.clj
+++ b/server/src/instant/grouped_queue.clj
@@ -33,9 +33,9 @@
 
 (defn- clean-or-reschedule [process-task {:keys [executor groups] :as q} group key]
   (when (= ::reschedule (locking q
-                        (if (some? (Queue/.peek group))
-                          ::reschedule
-                          (Map/.remove groups key))))
+                          (if (some? (Queue/.peek group))
+                            ::reschedule
+                            (Map/.remove groups key))))
     (ExecutorService/.execute executor process-task)))
 
 (defn- process


server/src/instant/grouped_queue.clj
```diff
@@ -24,43 +24,57 @@
...
-                            (Map/.remove groups key))))
-          (recur)))))
+  (when @processing?
+    (when-some [item (poll group combine-fn)]
+      (try
+        (process-fn key item)
+        (catch Throwable t
+          (tracer/record-exception-span! t {:name "grouped-queue/process-error"})))
+      (AtomicInteger/.addAndGet num-items (- (::combined item 1))))
+    (when (= ::loop (locking q
```
tonsky: This probably should be called `::reschedule` or smth, now that we don’t loop

tonsky:
> That does mean that there will be more locking on the queue. It would be nice if we could lock on the sub-queue's key

What you can do is do `poll` and schedule next task with items extracted from it. `poll` doesn’t lock

dwwoelfel:
> What you can do is do `poll` and schedule next task with items extracted from it. `poll` doesn’t lock

I now do a `Queue/peek` and submit a task without locking if we found something in the queue. Similar idea, but this way I don't have to keep track of the items I would pull off the queue with a poll.

tonsky:
Nice! LGTM

PR #1249, opened 2025-06-04
[custom-emails] add domain verification instructions

stopachka:
A user let us know that protonmail was showing a suspicious-looking error on custom email magic code sends. 

This is because they were stricter about domain checks: specifically with the DKIM value. 

Well, When we create a sender signature, we actually also get the DKIM value that the user can add to their records. 

I went ahead and: 

1. Started to show the DKIM value
2. I also showed the confirmation status for the email.

image

**Some further work:** 
1. Right now it will take some time for Postmark to pick up the DKIM status. They do have an endpoint that lets us specifically check for DNS records -- it may be nice to push run that when someone says "Refresh status" 

@nezaj @dwwoelfel @tonsky

### DIFF
--- a/server/src/instant/dash/routes.clj
+++ b/server/src/instant/dash/routes.clj
@@ -944,12 +944,12 @@
 ;; Email templates
 
 (defn sender-verification-get [req]
-  (let [{{app-id :id} :app} (req->app-and-user! :admin req)
+  (let [{{app-id :id} :app} (req->app-and-user! :collaborator req)
         {postmark-id :postmark_id}
         (app-email-template-model/get-by-app-id-and-email-type
          {:app-id app-id :email-type "magic-code"})]
     (response/ok {:verification (when postmark-id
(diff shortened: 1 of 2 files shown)


server/src/instant/dash/routes.clj
```diff
@@ -943,6 +943,14 @@
 ;; ---------------
 ;; Email templates
 
+(defn sender-verification-get [req]
+  (let [{{app-id :id} :app} (req->app-and-user! :admin req)
+        {postmark-id :postmark_id}
+        (app-email-template-model/get-by-app-id-and-email-type
+         {:app-id app-id :email-type "magic-code"})]
+    (response/ok {:verification (when postmark-id
+                                  (:body (postmark/get-sender! {:id postmark-id})))})))
```
dwwoelfel: Might be worth doing a select-keys here in case there is stuff in the payload we don't want to expose to the user.

stopachka: Great idea!

PR #1250, opened 2025-06-04
implemented API for moving files

htinney:
Works kind of like `mv`. Allows you to change file paths without needing to re-upload files. Allows you to specify a directory and will recursively change all paths underneath it.


htinney:
Closed in favor of https://github.com/instantdb/instant/pull/1255

PR #1251, opened 2025-06-04
Convert InstantAPIPlatformSchema to InstantSchemaDef in platform package

dwwoelfel:
This PR hides the API format of the schema from the `@instantdb/platform` consumer.

Whenever we return the schema from the API, we always convert it to the `InstantSchemaDef` format (e.g. `i.schema({entities: {...}})`).

When you call any of the functions that would return a schema, like `api.getSchema(appId)`, we'll convert it before we return it, and you'll get a `InstantSchemaDef` back instead.

The code that generates the typescript `instant.schema.ts` file also takes a `InstantSchemaDef` now. I had to make one small change to `DataAttrDef` so that we could keep track of which attrs were inferred. It now has a `metadata` field that you can write to. I store whether the type was inferred there.

~~The code generation also generates `i.json()` instead of `i.any()` now. If that's a problem, I think I could just revert to the old way by always outputting `i.any()` when I see a DataAttrDef with type `json` to keep the behavior.~~ (updated to keep the old behavior)

Also fixes a bug where we weren't converting the app into the js format (e.g. the type said createdAt, but we returned created_at).

### DIFF
--- a/client/packages/cli/src/index.js
+++ b/client/packages/cli/src/index.js
@@ -2,6 +2,7 @@
 
 import {
   generatePermsTypescriptFile,
+  apiSchemaToInstantSchemaDef,
   generateSchemaTypescriptFile,
   translatePlanSteps,
 } from '@instantdb/platform';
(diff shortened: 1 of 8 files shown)


stopachka (approved):
Biyutiful!

PR #1252, opened 2025-06-05
Only remove hints from the single pagination cte

dwwoelfel:
In [a previous PR](https://github.com/instantdb/instant/pull/1244), I disabled a bit too much for the pagination hints. Instead of just disabling the hint for the cte that was doing pagination, I also disabled hints for any ctes before the pagination cte.

Now it only removes the hint for the single pagination cte.

I put it behind the testing flag so that we can see how it will affect queries before rolling it out to production.

### DIFF
--- a/server/src/instant/db/datalog.clj
+++ b/server/src/instant/db/datalog.clj
@@ -1480,7 +1480,9 @@
                           {:select [[[:exists has-previous-query]]]}
                           :not-materialized])
              :pg-hints (if true ;; disable hints for pagination until we can generate better plans
-                         []
+                         (if *testing-pg-hints*
+                           (:pg-hints (:query match-query))
+                           [])
                          (into (:pg-hints (:query match-query))
                                pg-hints))
              :select (kw last-row-table :.*)


stopachka (approved):
Let's do it!

dwwoelfel:
I'm going to close this and try something else. After looking at it a bit, trying to use the ea index whenever we pass `:ea` is going to make things worse in a lot of cases. In fact, checking for `triples.ea` is those cases is causing postgres to do unnecessary work.

Consider this query:

```clojure
[:ea ?tbl-0 :tbl/id _]
```

We generate a query that looks like:

```sql
SELECT * FROM triples WHERE (app_id = :app) AND ea AND (attr_id = :tbl/id)
```

The `ea_index` has `(app_id, entity_id, attr_id)`. We don't have the `entity_id` yet (that's what the query is trying to root out), so we're going to have to do a lot of searching on the `ea_index`. Instead we want something with an index like `(app_id, attr_id, maybe_some_other_stuff)`. The `triples_created_at` index has `(app_id, attr_id, created_at)`, so that's what postgres uses. A better query for us in this case would be:

```sql
SELECT * FROM triples WHERE (app_id = :app) AND (attr_id = :tbl/id)
```

We can drop the check for `ea` entirely. It's not helping us because we already know that every triple with `attr_id = :tbl/id` is going to have `ea`.

Where it gets complicated is when we have multiple CTEs that could be joined in different orders--in that case maybe the ea could help us out because we'll know the entity_id from one of the ctes further down the line.

PR #1255, opened 2025-06-05
[Storage] Allow updating file paths

nezaj:
This PR enables updates for `path` on `$files` triples. This also means we allow writing `update` rules in permissions for `$files` now.

This PR includes the following:

* Adds logic + tests for allowing updates on files
* Allows writing `update` rules for `$files`
* Includes some docs on using `update` for moving files

Tested this in the dashboard and the sandbox and all seems to work as expected!

**Update**

To protect against using invalid lookups we make `size` a required field. This is a bit of a hack to prevent creating new files.

We also discovered the validation for required attrs wasn't checking system attrs, so added that in too!

**Update 2**

We also discovered that `url` was marked as required in the system catalog but it shouldn't be. This caused test failures after updating our required attrs logic. Luckily the production attr for `url` is not marked as required so simply removing required from the system catalog should be good enough to get the tests to pass


htinney:
This looks great! I like how you can edit paths in the explorer as well!

Closed my other PR

client/www/pages/docs/storage.md
```diff
@@ -402,6 +402,35 @@ Use `db.storage.delete(path)` to delete a file.
...
+      path: file.path.replace(
+        'documents/my-video-project/',
+        'videos/my-video-project/',
+      ),
+    }),
+  ),
+);
+```
+
+`path` is a unique attribute so if the path already exists the transaction
```
dwwoelfel:
```suggestion
`path` is a unique attribute so if another file exists with that path, then the transaction
```

server/src/instant/db/transaction.clj
```diff
@@ -127,12 +127,16 @@
...
+                label (let [[_op _eid aid] t]
+                        (-> (attr-model/seek-by-id aid attrs)
+                            attr-model/fwd-label))]
+          :when (and (= etype "$files")
+                     (not (contains? #{"id" "path"} label)))]
     (ex/throw-validation-err!
      :tx-step
      [op t]
-     [{:message (format "update or merge is not allowed on $files in transact.")}])))
+     [{:message (format "update or merge is only allowed on `path` for $files in transact.")}])))
```
dwwoelfel: Do we need to do anything here to prevent users from updating the `id` field? Maybe if it's an `id`, we check that the `entity_id` and the `value` match?

dwwoelfel: What happens if you use a `lookup` to update the path?

dwwoelfel (approved):
Couple of comments, but LGTM!

server/src/instant/db/transaction.clj
nezaj: @dwwoelfel Updated PR to prevent changing ids and prevent updating files with non-existing lookups!

PR #1256, opened 2025-06-06
Don't override the required setting if it's not provided

dwwoelfel:
Sometimes we get only partial updates to an attr. In the cases where `required` wasn't provided as part of the update, we were overwriting the required field to false instead of falling back to the setting in the databse.

This is why setting links to required would fail (user report: https://discord.com/channels/1031957483243188235/1148284450992574535/1377662474438053969).

It would also remove required if you changed the attr name from the UI.

Also removed the `not_null` constraint from the database to match how is_index and is_unique work.


stopachka (approved):
Nice find!

server/src/instant/db/model/attr.clj
```diff
@@ -210,7 +210,7 @@
             [:cast (when cardinality (name cardinality)) :text]
             [:cast unique? :boolean]
             [:cast index? :boolean]
-            [:cast (or required? false) :boolean]
+            [:cast required? :boolean]
```
dwwoelfel, line 213: Don't give `required?` a value if none was provided in the update. This way it will default to the current value in the db.

server/src/instant/db/model/attr.clj
```diff
@@ -630,7 +630,7 @@
            :forward-identity [forward_ident etype label]
            :unique? is_unique
            :index? is_indexed
-           :required? is_required
+           :required? (or is_required false)
```
dwwoelfel, line 633: Make `required?` a boolean when we return it to the user--this is just to fix the tests which expect it to be exactly `false`.

PR #1257, opened 2025-06-06
Implement non-upsert mode for update/merge

tonsky:
Problem:

1. Client 1 deletes entity X
2. Client 2 updates entity X with name = “new_name”

If these updates arrive in 1-2 order to the server, entity X will be resurrected: update from client 2 will be treated as `create` operation. This seems to be against the intention.

Caveats:

- Can’t just change `update` to ignore updates to deleted entities. People might legitimately want to reuse entity ids they already deleted:
  - They’re using ids from an external system
  - ​​They generate the id by taking parts from two different entities to create a link between two entities
  - ​​They create links with a string other_namespace_id field instead of creating an actual link
- If entity was created/updated through lookup, there might be no way to prevent its resurrection. We have no id, it’ll just create new entity from lookup ref

Plan:

- [x] Add an option to `.update`/`.merge` called `upsert` with values `true | false | undefined`
- [x] `upsert: true` is the current behavior
- [x] `upsert: false` will throw if you try to update entity that doesn’t exist
- [x] `undefined` is the auto mode. It will check if entity exists in current query cache, if yes, it will set `upsert: false` (strict update), otherwise `upsert: true` (create)

Will look like

```
db.tx.todos[id].update({ completed: true }, { upsert: false })
```

Hope is that auto mode will work for the majority of the customers and they will never need to learn about `upsert` parameter. Later we’ll also add `create` method that will throw if entity already exists

### DIFF
--- a/server/src/instant/db/transaction.clj
+++ b/server/src/instant/db/transaction.clj
@@ -208,7 +208,9 @@
                                    (when (triple-model/eid-lookup-ref? e) e)))
                            (concat create-steps update-steps))]
     (when (or (seq eids) (seq lookups))
-      (let [query    "WITH eids AS (
+      (tracer/with-span! {:name "transaction/validate-mode"
+                          :attributes {:app-id app-id}}
+        (let [query    "WITH eids AS (
                         SELECT cast(elem AS uuid) AS id
                         FROM jsonb_array_elements_text(cast(?eids AS jsonb)) AS elem
                       ),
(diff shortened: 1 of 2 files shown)


Donnerstagnacht:
1. May I ask why you prefer a flag over a dedicated method like `.upsert` (if you consider create - which I would like to see)?
2. And if you choose a flag, why do you choose to set the default value be `{ upsert: undefined }`?

My reasoning would be that`{ upsert: false }` would make more sense as default behavior since it behaves like update and` { upsert: true }` behaves more like upsert and therefore the behavior would be more in line with the method name. 

Plus, the use cases mentioned by you are legitimate, but aren't they less common edge cases for most users? If a user would really like such a behavior, wouldn't they implement a soft delete or action log anyways?

tonsky:
@Donnerstagnacht

> May I ask why you prefer a flag over a dedicated method

We need to fix current `update` behavior, so it can’t be opt-in.

> { upsert: false } would make more sense as default behavior

Ideally, yes. But it won’t be backwards-compatible.

If we were starting from scratch, we would 100% just make `create`, `update` and `upsert`, with no options. But unfortunately we already have `update` that currently works like `upsert`, so we have to work around that.

Donnerstagnacht:
Makes sense! Probably you could tidy  that up one day with a 1.0 release or something which discards the backwards compatibility to keep the api clean in the long run.

server/src/instant/db/transaction.clj
```diff
@@ -177,6 +191,79 @@
...
+
+        ;; check update over missing entities
+        (when-some [missing (->> update-steps
+                                 (filter (fn [[_ e _ _ _]] (nil? (resolved e))))
+                                 not-empty)]
+          (ex/throw-validation-err!
+           :tx-step
+           missing
+           [{:message (str "Updating entities that don't exist: " (string/join ", " (map (fn [[_ e _ _ _]] e) missing)))}]))))))
+
```
stopachka, line 268: One thought I had: 

Is it possible for us to avoid the extra query? 

For example, maybe instead we could have this logic inside our `add-triple/deep-merge-triple` CTE: 

Since we always include the `id` triple in our transactions, perhaps we can have a special rule around it: 
1. When `create` mode, we throw if it exists
2. When `update` mode, we throw if it does not exist
3. When `upsert`, we no-op

I am not sure how expensive this is though; perhaps it's okay, just worry about including more queries in our transact path

client/packages/core/src/instaml.js
```diff
@@ -186,73 +187,134 @@ function expandUnlink(attrs, [etype, eidA, obj]) {
...
+          }
+        }
+      }
+    }
+  }
+  return false;
+}
+
+function convertOpts({ stores, attrs }, [etype, eid, obj_, opts]) {
+  checkEntityExists(stores, etype, eid);
```
stopachka: nit: this can be deleted?

stopachka (approved):
Super cool!

client/packages/core/src/instaml.js
tonsky: Good catch, thanks!

server/src/instant/db/transaction.clj
tonsky: I think it would make code way more complicated. These checks can only be run before any changes are done, and they need both `add` and `deep-merge` triples. We ran two separate queries for those.

As for performance, my measurements show that `validate-mode` on average costs ~2.5 ms (with round-trip and everything). For contrast, `hsql/format` inside insert-multi alone takes ~2 ms. Just going from honey data structure to sql string! The query itself takes ~25 ms.

So I’m not too convinced that one uber-query is all that beneficial. It also make code waaaaay harder to read, modify, measure, optimize, debug, split, reuse etc. It’s a direct opposite of all the reasonable best practices. And all that sacrificed only to make transactions somewhat faster? I’m not even sure we’re limited by transact speed.

stopachka: I hear you about the downsides. The way I see the tradeoff is like this: 

- **Uber query**
    - **Pros**
      - Timing model: Since everything is in one query, we can be more sure that the DB is seeing "one view" of the data.
      - Performance: When there are perf issues there's a lot of tuning we can do to Postgres.
      - Less IO back-and-forth
    - **Cons**
      - Hard to reason about
      - Hard to compose and re-compose: language of composition is SQL, which is bad at this
- **Split queries:** 
  - **Pros**
    - Easier to reason about: language of composition is clojure
  - **Cons**
    - Timing model: different queries can see different versions of the DB, so you have to be careful
    - IO cost: On average it can 2.5ms, but IO can sometimes surprise
  
I think on both sides we can look to improve things. For example, perhaps there's a way to make the uber-query easier to debug and introspect. Similarly, it's not the end of the world to have a few more queries, if it massively simplifies the uber query. 

--- 

If you think the cost is fine, **I'm down to just add a tracer around this and move forward**. If anything we can always optimize it.

stopachka (approved):
LGTM

PR #1258, opened 2025-06-06
Add MCP server

nezaj:
This adds a new package `@instantdb/mcp` which is a thin wrapper around our Platform SDK. With this folks should be able to integrate Instant w/ their editors.

I also added a `rules.txt`, which is the condensed version of the docs we've been using to provide extra context.

Next step would be to publish and verify this works as expected via `npx @instantdb/mcp --token ...`

After that will put up some docs!

### DIFF
--- a/client/packages/mcp/README.md
+++ b/client/packages/mcp/README.md
@@ -5,7 +5,7 @@ editor to enable creating, managing, and updating your InstantDB applications,
 
 ## Setup Instructions
 
-If you haven't already make sure to get a personal access token from your
+If you haven't already, make sure to get a personal access token from your
 [Instant dashboard](https://www.instantdb.com/dash?s=personal-access-tokens)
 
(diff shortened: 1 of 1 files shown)


client/packages/mcp/src/index.ts
```diff
@@ -0,0 +1,397 @@
...
+        };
+      } catch (error) {
+        return handleError(error);
+      }
+    },
+  );
+
+  server.tool(
+    'push-perms',
+    "Execute a permissions push. Use this after 'plan-perms-push' to apply changes",
```
dwwoelfel: There isn't a plan-perms-push, is there?

dwwoelfel (approved):
Cool!

client/packages/mcp/src/index.ts
nezaj: Ah yup! Had that in before but removed it, good catch!

PR #1259, opened 2025-06-06
@instantdb/platform helper to parse a schema file into an InstantSchemaDef

dwwoelfel:
Adds `schemaTypescriptFileToInstantSchema` to `@instantdb/platform`. 

You can use it to parse a schema file generated by the CLI into an `InstantSchemaDef` (e.g. `i.schema({...})`). 

It's not as robust as the parser we use in the `cli`. The `cli` uses a package called `unconfig`, which requires node APIs that aren't available in the browser.

The code to parse the schema file is pretty inflexible. It parses the typescript with babel, then walks the ast to find all of the `entities`, `links` and `rooms`. I think we could make it a lot nicer if we used `@babel/traverse`, but that doubles our bundle size. With just `@babel/parse`, our bundle is 88kb gzipped. But you'll only pay that bundle cost if you import the `schemaTypescriptFileToInstantSchema` helper. Otherwise it's closer to 13kb gzipped.

Example:

```ts
import { schemaTypescriptFileToInstantSchema } from '@instantdb/platform';

const code = `// Docs: https://www.instantdb.com/docs/modeling-data

import { i } from '@instantdb/core';

const _schema = i.schema({
  entities: {
    books: i.entity({
      title: i.string().unique(),
    }),
  },
});

// This helps Typescript display nicer intellisense
type _AppSchema = typeof _schema;
interface AppSchema extends _AppSchema {}
const schema: AppSchema = _schema;

export type { AppSchema };
export default schema;
`;

const schema = schemaTypescriptFileToInstantSchema(code);

// schema == i.schema({ entities: { books: i.entity({ title: i.string().unique() }) } })
```

### DIFF
--- a/client/packages/platform/src/typescript-schema.ts
+++ b/client/packages/platform/src/typescript-schema.ts
@@ -42,9 +42,8 @@ function astToJson(node: Node | null): JsonValue | null {
       const objNode = node as ObjectExpression;
       return objNode.properties.reduce(
         (obj: { [key: string]: JsonValue }, prop) => {
-          // Using prop.type directly for checks
           if (prop.type === 'ObjectProperty') {
-            const objectProperty = prop as ObjectProperty; // Cast to ObjectProperty
+            const objectProperty = prop as ObjectProperty;
             let key: string | number;
             if (objectProperty.key.type === 'Identifier') {
               key = (objectProperty.key as Identifier).name;
@@ -81,7 +80,6 @@ function astToJson(node: Node | null): JsonValue | null {
           if (element && element.type === 'SpreadElement') {
             return undefined;
           }
-          // element can be Expression, SpreadElement or null
           return astToJson(element as Expression | SpreadElement | null);
         })
         .filter((value) => value !== undefined) as JsonValue[];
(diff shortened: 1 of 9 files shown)


client/packages/platform/src/typescript-schema.ts
```diff
@@ -0,0 +1,455 @@
...
+
+    if (property.value.type !== 'ObjectExpression') {
+      throw new Error(
+        `Could not extract room \`${roomName}\`. Expected an object with topics and presence keys, got \`${property.value.type}\`.`,
+      );
+    }
+
+    rooms[roomName] = roomNodeToRoomDef(roomName, property.value);
+
+    // if (
```
nezaj: Did we want to leave this in for later?

nezaj (approved):
Nice! Curious, would we want to integrate this into the CLI's `pull-schema` command?

client/packages/platform/src/typescript-schema.ts
dwwoelfel: Whoops! Removed!

dwwoelfel:
> Curious, would we want to integrate this into the CLI's `pull-schema` command?

Not yet, because it's not very robust against changes to the format. For example, we would fail to parse if you did something like this:

```js
const users = i.entity({handle: i.string.unique()});

const schema = i.schema({entities: {users}});
```

PR #1261, opened 2025-06-06
[firefox] fix mouse-up that causes clicks when pointer-capture is involved

stopachka:
# Problem

In Firefox on mac, if you:
1. When to the sandbox
2. Started to drag your mouse from the monaco editor
3. And released your mouse on a nav item
4. The _nav item_ would trigger and you'd be redirected! 

https://github.com/user-attachments/assets/c8900a14-e514-42ba-ad85-315839eb5a11

# Root Cause

This is a bug in Firefox on Mac. 

Some relevant links: 
 * Bug report in [Monaco](https://github.com/microsoft/monaco-editor/issues/4379)
 * Bug report in [Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1556240)
 * W3C [discussion](https://github.com/w3c/pointerevents/issues/356) 
 * Cool [repro](https://output.jsbin.com/zuwiwep)  

What happens is: 
1. Monaco uses `setPointerCapture`
2. When it releases the setPointerCapture in mouseup, Firefox has to decide where the click happened: Is it on the origin target, or where the mouse up is? 
3. Most browsers decide it's the origin. Firefox decides it's where the mouse up is (the menu item!) 

# Solution

I wrote a custom patch for Firefox. This saves the target on mouseUp. If it's the _same_ as the target on click, we continue on as normal. 

@dwwoelfel @nezaj @tonsky

### DIFF
--- a/client/www/lib/patchFirefoxClicks.ts
+++ b/client/www/lib/patchFirefoxClicks.ts
@@ -1,0 +1,70 @@
+/**
+ * There's a bug in Firefox, where click events are dispatched to the wrong target element.
+ *
+ * This happens around our Monaco Editor.
+ *
+ * Here's the repro scenario:
+ *
+ * 1. Click inside our Monaco Editor, and start dragging your cursor
+ * 2. Release your cursor over a Nav item
+ * 3. Firefox will trigger a "click" event on the Nav item!
+ *
+ *
+ * Root cause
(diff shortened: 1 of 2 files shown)


dwwoelfel (approved):
Nice!

PR #1262, opened 2025-06-06
Update a couple of client repos with new 

dwwoelfel:
Updates the client readme to include the platform and mcp packages.

Updates the mcp readme to include the common header.


nezaj (approved):
Heck yeah!

PR #1264, opened 2025-06-07
[MCP] Add required and onDelete to schema additions

nezaj:
Adds support for `required` and `onDelete` for links when creating/modifying schema


stopachka (approved):
Lesgo!

PR #1266, opened 2025-06-07
Change `uploadFile` typing to allow uploading `Blob`s

matubu:
changed the typing of `uploadFile` to allow uploading a `Blob`s bcz its already supported and its common to have `Blob`s in JS

also minor change in client README to add syntax highlighting

btw great work. thanks!


dwwoelfel:
Thanks @matubu!

PR #1269, opened 2025-06-10
[admin] magic codes on the backend

stopachka:
A user requested that we also support sending and verifying magic codes on the backend. This can be useful if you want to do some extra work after creating the user token. 

I went ahead and abstracted most of the work into a `magic-code-auth` namespace, and replaced the logic in smaller wrappers around the admin api and the runtime api. 

@dwwoelfel @nezaj @tonsky


dwwoelfel (approved):
LGTM!

PR #1274, opened 2025-06-10
RFC: Set `required` on reverse attr for one-to-one required links

dwwoelfel:
It seems right to set the `required` flag if the link is one-to-one and it's required on the forward attribute. This makes the types nicer, e.g.:

```js
const authors = await db.query({ authors: { profile: {} }});
const author = authors[0];
if (!author) { throw('No author') }
// no errors
const name = author.profile.name;
```

Opening as an RFC in case I'm overlooking something.

### DIFF
--- a/client/packages/platform/src/api.ts
+++ b/client/packages/platform/src/api.ts
@@ -352,7 +352,7 @@ function apiSchemaAttrToLinkDef(attr: InstantDBAttr) {
       has: rhas,
       label: rlabel,
       required:
-        (attr['required?'] && fhas === 'one' && rhas === 'one') || undefined,      
+        (attr['required?'] && fhas === 'one' && rhas === 'one') || undefined,
       onDelete:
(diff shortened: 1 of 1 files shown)


dwwoelfel:
Actually didn't seem to fix the type--will see if it's easier to fix the type directly instead of changing the schema.ts

dwwoelfel:
I'm going to put this aside to avoid getting distracted, but I think what we'd need to do is add Required to LinkAttrDef (https://github.com/instantdb/instant/blob/main/client/packages/core/src/schemaTypes.ts#L67), and then do something different with the `| undefined` here: https://github.com/instantdb/instant/blob/main/client/packages/core/src/queryTypes.ts#L190

stopachka:
Yes! We may also need to address this issue:

https://github.com/instantdb/instant/pull/1148#discussion_r2067592941

(Even with a required link, right now I think you can still get a partial result in InstaQL)

PR #1275, opened 2025-06-10
[Docs] Update perms example

nezaj:
One of our users noticed a slight security flaw in our example. Before the example wouldn't prevent someone from creating a task and then linking it over to another user (by using their user ID). This change would close that hole


client/www/pages/docs/permissions.md
```diff
@@ -7,24 +7,29 @@ To secure user data, you can use Instant’s Rule Language. Our rule language
 takes inspiration from Rails’ ActiveRecord, Google’s CEL, and JSON.
 Here’s an example ruleset below
 
-```typescript {% showCopy=true %}
+<!-- prettier-ignore-start -->
```
nezaj, line 10: Adding this so prettier doesn't try to change the formatting of the code block.

The previous version wasn't copy/pasteable into our dashboard permissions

PR #1276, opened 2025-06-10
Add a route to get token info

dwwoelfel:
Adds a route to check if the platform OAuth token is still good.

This is useful to check if the token needs to be refreshed.

Returns JSON with `expires_in`, `scopes`, and `token_type`.


stopachka (approved):
SGTM!

PR #1277, opened 2025-06-11
Self Host: Coolify

bioshazard:
**BLUF:** Enable self-hosted InstantDB on Coolify with Postgres + Minio (S3-compatible), all opt-in via new ENV VARs and Docker Compose.

**Background:**

* InstantDB offers an offline-first sync layer backed by Postgres and S3/Hazelcast.
* Goal: self-host on homelab (Coolify) using Minio, without disrupting existing setups.

**What’s included:**

* **ENV VARs** for Minio/S3 integration (no default changes to current deployments).
* **GitHub-hosted Docker image** for manual builds.
* **Docker Compose** manifest for zero-friction Coolify deployment.
* **`./selfhost` directory** with one-shot Coolify + psql-admin example.
* **Docs** on mounting an `override.edn` at runtime (avoids rebuilding the JAR).

*Open to feedback or a live walkthrough of my use case. Thanks!*


bioshazard:
ok added a few more commits to also enable Keycloak as OIDC provider. Its AWESOME that InstantDB delivers arbitrary OIDC. How are people not freaking out over this project that does everything for you in a 2-way sync layer!

stopachka (commented):
Het @bioshazard , we don't have the bandwidth to maintain the docker bits at the moment. However, to unblock we'd love to add in the parts where localhost doesn't seem to make sense. If you like please pop in a PR with those changes, and we'd love to land that.

PR #1279, opened 2025-06-11
Fix flaky instaql-test/fields-with-rules

tonsky:
This test is counting number of queries executed, and sometimes attr model gets evicted and re-queried, that throws off the counter

### DIFF
--- a/server/test/instant/db/instaql_test.clj
+++ b/server/test/instant/db/instaql_test.clj
@@ -4236,26 +4236,21 @@
 (deftest fields-with-rules
   (with-zeneca-app
     (fn [app r]
-      (let [make-ctx (fn []
-                       (let [attrs (attr-model/get-by-app-id (:id app))]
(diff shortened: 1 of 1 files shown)


server/test/instant/db/instaql_test.clj
```diff
@@ -4253,6 +4253,9 @@
                                  :bookshelves {:allow {:view "data.name == 'Nonfiction'"}}
                                  :books {:allow {:view "data.isbn13 == '9780316486668'"}}}})
 
+        ;; prevent attrs lookup from messing with query count
+        (attr-model/get-by-app-id (:id app))
```
dwwoelfel: It might work a little better to put the `(make-ctx)` in a `let` that wraps the `(binding [sql/*in-progress-stmts* query-tracker]` instead.

dwwoelfel (approved):
One suggestion, but LGTM!

PR #1280, opened 2025-06-11
Set up platform API to refresh tokens

dwwoelfel:
This enables automatic token refresh for the platform API.

If you provide an `auth` object with `{clientId, clientSecret, accessToken, refreshToken}`, then we'll notice when the token is expired and we'll refresh it and try again.

The way we determine that it's expired is to catch exceptions that return 401 and refresh. It might be nicer to check the expiredAt first, but I didn't want to make callers keep track of the `expiresAt` for their token. This seemed like a nicer interface.

If you want to know when the token refreshed so that you can save the new accessToken, you can provide an `onRefresh` callback in the `auth` object.

If you want to manage refreshing tokens yourself, just keep passing `{accessToken: token}` as the `auth` arg when you create the `PlatformAPI`.

### DIFF
--- a/client/packages/platform/src/api.ts
+++ b/client/packages/platform/src/api.ts
@@ -358,8 +358,6 @@ function apiSchemaAttrToLinkDef(attr: InstantDBAttr) {
       on: re,
       has: rhas,
       label: rlabel,
-      required:
-        (attr['required?'] && fhas === 'one' && rhas === 'one') || undefined,
       onDelete:
         attr['on-delete-reverse'] === 'cascade'
           ? ('cascade' as 'cascade')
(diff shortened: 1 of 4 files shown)


client/packages/platform/src/api.ts
```diff
@@ -351,6 +358,8 @@ function apiSchemaAttrToLinkDef(attr: InstantDBAttr) {
       on: re,
       has: rhas,
       label: rlabel,
+      required:
+        (attr['required?'] && fhas === 'one' && rhas === 'one') || undefined,
```
stopachka: nit: was this meant to be included?

stopachka (approved):
Cool!

client/packages/platform/src/api.ts
dwwoelfel: Whoops! Will remove!

PR #1281, opened 2025-06-11
Remote mcp

dwwoelfel:
Adds support for a remote mcp that we can deploy at `mcp.instantdb.com`.

It uses streamable http with OAuth.

The way the MCP server does OAuth is a little odd--it really wants you to implement "dynamic OAuth client registration". We don't really want to do that on the Instant side, so we have the mcp server itself manage generating OAuth clients. The flow looks something like this:

1. The mcp client asks for a new oauth client
2. We generate a new client for them and put it in the mcp app's store (which is an InstantDB app!)
3. The mcp client starts the oauth login with their new client by redirecting to the `/authorize` endpoint that the mcp server manages
4. We store some info about the oauth request, then we redirect them to `api.instantdb.com/platform/oauth/start` with an OAuth client I created for the MCP server. The users grants access.
5. We store the instantdb code in the mcp app's store, then create a new code that we give to the mcp client
6. The mcp client exchanges that code for a token by hitting the `/token` endpoint that the mcp server manages. At that point, we exchange the instantdb code for a platform token, generate a new token for the mcp server to return back to the client, and associate the mcp server's token with the instantdb token.
7. The mcp client uses its new token to make a request to `/mcp`. We look up the instant token, instantiate a new `PlatformApi` with it, and register the tools. Now they can call all of the methods.

Still TODO:
- [x] Instructions in README
- [x] Add a consent screen to oauth/start
- [x] Add sse as a transport at /sse
- [x] Push up schema and perms to the prod instantdb app


client/packages/mcp/src/db/instant.perms.ts
```diff
@@ -0,0 +1,13 @@
+// Docs: https://www.instantdb.com/docs/permissions
```
stopachka, line 1: so cool!

stopachka (approved):
Suuper cool!

PR #1283, opened 2025-06-11
[dash] ensure controlled component

stopachka:
A user reported they saw a warning that said: 

client.ts:59 A component is changing from uncontrolled to controlled. 

I repro'd this by loading /dash without an pre-existing app set. 

What is happening is that `Combobox` was receiving `undefined`, which caused it to believe that it was an uncontrolled component. 

I went ahead and used `null` instead. 

@dwwoelfel @nezaj @tonsky


dwwoelfel (approved):
Nice!

PR #1284, opened 2025-06-12
Upgrade turborepo

dwwoelfel:
Upgrades turborep from v1 to v2. 

They have some fixes for `turbo prune` that were blocking me from generating a dockerfile to run the remote mcp server. Upgrading turbo allowed me to get further (still not 100% of the way there).


stopachka (approved):
Let's do it! 🔥

PR #1286, opened 2025-06-12
Make deny-redirect throw if there is no redirect

dwwoelfel:
Fixes the error stopa ran into when clicking deny (I think the request had previously been approved, maybe?). Right now it just returns JSON. It would be nicer if we returned an HTML page, but I think that's a bigger project.


stopachka (approved):
Nice!

PR #1287, opened 2025-06-12
Show the logged in user in the oauth confirmation page

dwwoelfel:
Allows the user to see which Instant account they're giving access to.

Looks like this:

image


nezaj (approved):
Nice!

PR #1289, opened 2025-06-13
Support $not on reverse links

dwwoelfel:
Fixes a bug where we would throw an error if you used `$not` on a reverse link. Reported on Discord [here](https://discord.com/channels/1031957483243188235/1382553860874174526).

For example, in the zeneca app where users -> bookshelves:

The forward link worked:
```clojure
{:users {:$ {:where {:bookshelves {:$not bookshelf-id}}}}
```

But the backward link didn't:
```clojure
{:bookshelves {:$ {:where {:users {:$not user-id}}}}
```

That's because we have to reverse the e and v in the pattern when we're comparing on the reverse link. We didn't support that for `$not`--the spec didn't allow `$not` in the `e` position. Now we'll handle that case.

### DIFF
--- a/server/src/instant/db/datalog.clj
+++ b/server/src/instant/db/datalog.clj
@@ -129,9 +129,9 @@
    (map-indexed
     (fn [i c]
       (if (or
-           (and (or (= i 0))
+           (and (= i 0)
                 (map? c)
-                (or (contains? c :$not)))
+                (contains? c :$not))
            ;; Don't override function clauses
            (and (= i 2)
                 (map? c)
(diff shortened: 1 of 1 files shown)


server/src/instant/db/datalog.clj
```diff
@@ -55,6 +55,8 @@
         :variable symbol?))
 
 (s/def ::$not ::triple-model/value)
+;; $not for the entity_id should always be a uuid
+(s/def :datalog-entity-id/$not uuid?)
```
stopachka, line 59: til you can have a slash here!

PR #1300, opened 2025-06-23
Fix slow cascade delete

tonsky:
Report:

https://discord.com/channels/1031957483243188235/1384212875584737321

Cause:

Suboptimal query: we were filtering attrs by app_id just in case, it was completely ruining PostgreSQL optimizer. Feels safe to remove, since attrs are joined from triples, triples are already pre-filtered by app_id, so attrs will have to be from the same app_id.

I also switched away from honeysql for performance reasons, so it’s hard to see, but the only change is the removal of

```
[:or
  [:= 'attrs/app-id app-id]
  [:= 'attrs/app-id system-catalog-app-id]]
```

from both `forward-attrs` and `reverse-attrs` CTEs,

and addition of

```
entity_id_jsonb
```

for even more perf.

### DIFF
--- a/server/src/instant/db/model/triple.clj
+++ b/server/src/instant/db/model/triple.clj
@@ -684,46 +684,59 @@
    2. Deletes all reference triples where this entity is the value:
       [_ _ id]"
   [conn app-id id+etypes]
-  (let [id-etypes
-        (sql/tupleset id+etypes
-                      [{:as 'entity_id :type :uuid}
-                       {:as 'etype :type :text}])
+  (let [query (sql/format
+               "WITH
 
-        forward-attrs
-        {:select 'triples/ctid
-         :from   'triples
-         :join   ['id-etypes [:= 'triples/entity_id 'id-etypes/entity_id]
-                  'attrs     [:= 'triples/attr_id 'attrs/id]]
-         :where  [:and
-                  [:= 'triples/app-id app-id]
-                  [:= 'attrs/etype 'id-etypes/etype]
-                  [:or
-                   [:= 'attrs/app-id app-id]
-                   [:= 'attrs/app-id system-catalog-app-id]]]}
+                id_etypes AS (
+                  SELECT
+                    cast(elem ->> 0 AS uuid) AS entity_id,
+                    cast(elem ->> 1 AS text) AS etype,
+                    TO_JSONB(cast(elem ->> 0 AS uuid)) AS entity_id_jsonb
+                  FROM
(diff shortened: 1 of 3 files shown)


server/src/instant/db/model/triple.clj
```diff
@@ -684,46 +684,59 @@
...
+                ),
+
+                forward_attrs AS (
+                  SELECT
+                    triples.ctid
+                  FROM
+                    triples
+                    JOIN id_etypes ON triples.entity_id = id_etypes.entity_id
+                    JOIN attrs ON triples.attr_id = attrs.id
+                  WHERE triples.app_id = ?app-id
```
stopachka, line 706: What happens in the case where we have a cascade delete on a system-catalog triple? 

(My assumption is that here it would need to be an OR with the system-catalog-app-id, or no?)

stopachka: Dohp, nevermind. Regardless if it's a system-catalog-app, the app-id will be the original one on triples! Carry on

tonsky: No, we only filter triples by app_id and triples always have your app_id. If we join triples to attrs then triple.attrs.app_id could be system catalogue, but we don’t filter by that because we already filtered triples by your app_id

server/src/instant/db/model/triple.clj
```diff
@@ -684,46 +684,59 @@
...
-                :returning   ['entity_id 'attr_id 'value 'created-at]}]
-    (sql/execute! ::delete-entity-multi! conn (hsql/format query))))
+  (let [query (sql/format
+               "WITH
+
+                id_etypes AS (
+                  SELECT
+                    cast(elem ->> 0 AS uuid) AS entity_id,
+                    cast(elem ->> 1 AS text) AS etype,
+                    TO_JSONB(cast(elem ->> 0 AS uuid)) AS entity_id_jsonb
```
stopachka, line 694: This is soo nice. I wonder if we do something like this in `insert-multi` et al, if we can _massively_ increase the number of chunks we can handle in one transaction. Some users reported we get a "[too many parameters](https://discord.com/channels/1031957483243188235/1385029412117811342/1385113920976584759)" error from postgres

tonsky: We _should_ do this in insert multi

tonsky: "[too many parameters](https://discord.com/channels/1031957483243188235/1385029412117811342/1385113920976584759)" is kinda good enough reason for me to redo insert-multi! I feel like. It’s user-visible

PR #1306, opened 2025-06-24
Better index selection for pg_hint_plan

dwwoelfel:
Analyzes the datalog pattern to determine which index we should use based on which columns are known.

Consider this datalog query:

```clojure
[:ea ?tbl-0 :tbl/id _]
```

We generate a query that looks like:

```sql
SELECT * FROM triples WHERE (app_id = :app) AND ea AND (attr_id = :tbl/id)
```

We know that `a` is a constant, but `e` and `v` are unknown, so we should use an index that starts with `a`. The only index like that in the `triples` table is the `triples_created_at_idx`, so we use that. 

Before we would try to use the `ea` index, which doesn't help us at all because it has `e` before the `a`.

I'll run this in testing mode to see if there are any queries that it makes worse, but I didn't find any when testing the usual suspects locally.

### DIFF
--- a/server/src/instant/db/datalog.clj
+++ b/server/src/instant/db/datalog.clj
@@ -172,6 +172,12 @@
       :map (:idx-key v)
       :keyword v)))
 
+(defn idx-data-type [idx]
+  (let [[tag v] idx]
+    (case tag
+      :map (:data-type v)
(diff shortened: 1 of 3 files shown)


stopachka (approved):
Let's do this!

PR #1310, opened 2025-06-25
Get all of the logs

dwwoelfel:
We were losing some logs on shutdown. This makes two changes:

1. We upload the logs through a docker logger instead of letting cloudwatch load the file. I have a feeling the file isn't getting uploaded in time when weird things happen to the instance.
2. We shutdown the tracer so that it will flush its logs before the process exits.


nezaj (approved):
Nice!

PR #1311, opened 2025-06-25
Allow app admin tokens in the platform SDK

dwwoelfel:
Allows the admin token for an app to be used with the platform SDK for operations that take an appId.

Example:

```
const api = new PlatformApi({auth: {token: 'YOUR_APP_ADMIN_TOKEN'}});
await api.getSchema('YOUR_APP_ID');
```


stopachka (approved):
Niice!

PR #1313, opened 2025-06-26
Show users the oauth apps they've authorized

dwwoelfel:
Adds a user settings page that will show users the OAuth apps that they've authorized and allow them to revoke access.

Here's what it looks like:

https://github.com/user-attachments/assets/05b76b1d-0f22-4825-855e-f78bb758493b


dwwoelfel:
> Should they just be in Account Settings only?

Fixed the tabs!

PR #1314, opened 2025-06-27
[CLI] Support env.local etc. files

nezaj:
Replaces `dotenv` with `dotenv-flow` to automatically load environment files in the standard hierarchy (`.env`, `.env.local`, `.env.development`, etc.)

Previously, the CLI only read from `.env`. Developers often use `.env.local` for local overrides that shouldn't be committed to version control. This aligns with common patterns used by next and other frameworks.


stopachka (approved):
SGTM!

PR #1317, opened 2025-06-27
[CLI] Add create-app

nezaj:
While using the CLI I noticed it would be nice to have a separate command to generate new app ids but not necessarily update my schema and perms -- this is especially helpful when trying scaffold a new app with schema from an existing app.

If this looks good will follow-up with some docs on `create` for the cli

### DIFF
--- a/client/packages/cli/src/index.js
+++ b/client/packages/cli/src/index.js
@@ -326,7 +326,7 @@ program
 
 program
   .command('create-app')
-  .description('Generate an ')
+  .description('Generate an app ID and admin token for a new app.')
   .action(async () => {
     await checkVersion();
     const authToken = await readAuthTokenOrLoginWithErrorLogging();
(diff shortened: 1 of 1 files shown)


client/packages/cli/src/index.js
```diff
@@ -324,6 +324,24 @@ program
...
+    const authToken = await readAuthTokenOrLoginWithErrorLogging();
+    if (!authToken) return process.exit(1);
+
+    const result = await promptCreateApp();
+    if (!result.ok) return process.exit(1);
+
+    const { appId, appTitle, appToken } = result;
+    console.log(`\n✅ Created app "${chalk.green(appTitle)}"\n`);
+    console.log(`App ID: ${chalk.cyan(appId)}`);
+    console.log(`App Admin Token: ${chalk.cyan(appToken)}`);
```
dwwoelfel: I don't know if it's a good idea to print out the admin token, since some people might call `create-app` while live-streaming. Is there some way we could obfuscate it?

nezaj: @dwwoelfel pushed up a tweak to show the first 4 digits only but you can still double click to select/copy

stopachka (approved):
Cool!

PR #1318, opened 2025-06-27
[CLI] Include admin token on init

nezaj:
The create app endpoint returns the admin token and I thought it would be nicer for us to include that in the `.env` file we generate on `init` to make it slightly easier for folks who want to get started with Instant on the backend


stopachka (approved):
SGTM! 

I had a thought about if this could accidentally end up in a bundle. But I don't think that's the case: Expo, Vite, Next all use _public prefixes

PR #1319, opened 2025-06-27
[DOCS] Add docs for cli create-app

nezaj:
What it says on the tin! We've seen focus sometimes be surprised that they need to run `login` first before `init` so thought it would be nice to include a blurb about that too


dwwoelfel (approved):
LGTM!

PR #1322, opened 2025-06-30
Do not send attrs on every refresh-ok

tonsky:
Attrs dominate every payload. For our todo example, we send ~1kb of data and 13kb of attrs on every refresh-ok. There’s no reason to send attrs if they don't change. This PR fixes that. Unfortunately requires client upgrade, so adoption will be gradual


server/src/instant/reactive/session.clj
```diff
@@ -214,7 +225,16 @@
...
+                      :tx-latency-ms (e2e-tracer/tx-latency-ms (:tx-created-at event))}
+        {prev-attrs :session/attrs
+         version :session/versions} (rs/session store sess-id)
+        parsed-version  (some-> version (get core-version-key) (semver/parse))
+        can-skip-attrs? (and (flags/refresh-skip-attrs? app-id)
+                             parsed-version
+                             (pos? (semver/compare-semver parsed-version refresh-skip-attrs-min-version)))
+        attrs-changed?  (not= prev-attrs attrs)]
+    (when (and can-skip-attrs? attrs-changed?)
+      (rs/assoc-session! store sess-id :session/attrs attrs))
```
dwwoelfel: Can we just store a hash of the attrs instead of storing the full set of attrs in the session?

We already have a thing that hashes in `instaql-query-reactive!`. We could either store two hashes there or store a single hash in the session and stop inluding the attrs in the hash in `instaql-query-reactive!`.

server/src/instant/db/model/attr.clj
```diff
@@ -681,59 +656,82 @@
...
       this))
 
+  clojure.lang.IHashEq
+  (hasheq [_this]
+    (clojure.lang.Murmur3/hashUnordered elements))
+
+  java.lang.Object
+  (equals [_this o]
+    (and (instance? Attrs o)
+         (= (set elements)
```
dwwoelfel: If we added `order by id` to the `get-by-app-id` function, then we could skip putting the elements into a `set` in the `equals` function.

server/src/instant/db/model/attr.clj
```diff
@@ -681,59 +656,82 @@
...
   (cons [_this o]
-    (let [new-elements (cons o elements)]
-      (Attrs. new-elements (delay (index-attrs new-elements)))))
+    (wrap-attrs (cons o elements)))
   (seq [this]
     (if (empty? elements)
       nil
       this))
 
+  clojure.lang.IHashEq
```
dwwoelfel, line 684: Oops, I thought iseq handled the hash for me--we've been sending unnecessary spam whenever the attrs got pushed out of the cache all this time!

server/src/instant/db/model/attr.clj
```diff
@@ -681,59 +656,82 @@
...
+    (get @ea-ids-by-etype-cache etype #{}))
+  (unwrap [_this]
+    elements))
 
 (defn wrap-attrs [attrs]
-  (Attrs. attrs (delay (index-attrs attrs))))
+  (Attrs.
+   attrs
+   ;; by-id-cache
+   (delay
```
dwwoelfel, line 699: My thinking was that we'd save cpu cycles in the long run by building all of the caches we'd use in one scan of the attrs, even if we populated caches that we didn't use. Are you sure multiple traversals over the attrs isn't going to perform worse?

dwwoelfel (approved):
Couple of comments, but LGTM!

server/src/instant/db/model/attr.clj
tonsky: Done

server/src/instant/db/model/attr.clj
tonsky: Yes. The time is dominated by assoc anyways. We are actually faster after my changes, 35 µs → 30 µs to deref everything

PR #1324, opened 2025-06-30
Use ave index for string operations

dwwoelfel:
We use a gist index for strings so that `like` will be efficient, but it's not very good for exact matches, and it's not very good for ordering because it doesn't produce a sorted output ([according to the documentation](https://www.postgresql.org/docs/current/indexes-ordering.html), only B-Tree indexes produce sorted output).

We already have everything in the `ave` index and it sorts in the same order as strings do. We can use it for any index comparisons that don't use `like` or `ilike`.

This PR uses the `ave` index when we're sorting by strings, but still uses the gist index when comparing with `like` or `ilike`.

### DIFF
--- a/server/src/instant/db/datalog.clj
+++ b/server/src/instant/db/datalog.clj
@@ -585,7 +585,7 @@
          [:json_null_to_null :value]
          :value)
        (value->jsonb val)]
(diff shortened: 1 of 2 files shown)


stopachka (approved):
Exciting!

PR #1325, opened 2025-07-01
Fix create not adding new attrs

tonsky:
I notices that implicit attrs creation doesn't work with `create` op


stopachka (approved):
SGTM!

PR #1327, opened 2025-07-02
Fix deleting entity that shares ids with other namespaces

tonsky:
Report: https://discord.com/channels/1031957483243188235/1258878823529844766/1390020531407028365

Problem:

> It looks like we might have a bug in the required fields validation when an id is shared between multiple entities.


dwwoelfel (approved):
LGTM!

PR #1329, opened 2025-07-02
Use github as the source of truth for prod migrations

dwwoelfel:
This change will make gomigrate fetch the migrations from the main branch of GitHub instead of looking at your local filesystem.

This will prevent any incidents where someone was working on a migration locally and forgot to checkout main and stash their changes before running the migrations.

Only affects `make prod-up`, `make prod-down` and `make prod-version`.

gomigrate docs: https://github.com/golang-migrate/migrate/tree/master/source/github

Also adds a `make create-migration` that will generate the two blank up/down files for you.


nezaj (approved):
LGTM!

PR #1326, opened 2025-07-02
Put uuids in the vae and eav index

dwwoelfel:
We only put refs in the vae and eav indexes so we know that the value will be a uuid. This PR stores those uuids as actual uuids in the index itself. 

This makes the indexes take up less space (they're about 30% smaller).

The big change is that now we can use `vae` instead of `eav` when we do a join on a ref, which saves us doing an expensive json->uuid conversion in the join filter. Here's a simple example:

```clojure
{:books {:where {:category "western" :author :eid/elmer-kelton}}}
```

That generates two ctes:

```
with m_0 as (
  select entity_id as m_0_eid from triples t0 where attr_id = :category-eid and value = "western"
),
m_1 as (
  select m_0.*, entity_id as m_1_eid from triples t1 where value = :eid/elmer-kelton and m_0.m_0_eid = json_to_uuid(value)
)
```

If we use the `eav` index for m_1, the `v` is at the end of the index. Postgres does a nested loop and filters the join on `entity_id = json_to_uuid(value)`. It looks something like this:

```
    ->  Nested Loop
          Join Filter: (t0.entity_id = ((t1.value ->> 0))::uuid)
          Rows Removed by Join Filter: N
          ->  Index Scan using ave_index on triples t0
                Index Cond: (attr_id = :category-eid) AND (value = '"western"'::jsonb)
          ->  Index Scan using eav_uuid_index on triples t1
                Index Cond: ((entity_id = :eid) AND (attr_id = :author-eid))
```

If we use the `vae` index for m_1, then the `v` is at the start of the index and postgres will use that in the index cond instead of the less efficient join filter. It looks something like this:

```
    ->  Nested Loop 
          ->  Index Scan using ave_index on triples t0 
                Index Cond: (attr_id = :category-eid) AND (value = '"western"'::jsonb)
          ->  Index Scan using vae_uuid_index on triples t1
                Index Cond: ((((value ->> 0))::uuid = t0.entity_id) AND (attr_id = :author-eid) AND (entity_id =:eid))
```

`((value ->> 0))::uuid` is already in the index, so we never have to do any json->uuid conversion at all in the query itself. We avoid the expensive join filter, which in my backtesting makes queries much faster.

We also don't have to carry around the `match_x_x_value_uuid` any more. Instead of `value_uuid`, we track `is_ref_value` so that we know to convert it from a string to a uuid in `sql-row->triple`.

<details>
<summary>Results from the backtest:</summary>

[...]

PR #1330, opened 2025-07-02
Remove old vae and eav indexes

dwwoelfel:
Removes the indexes that were replaced in https://github.com/instantdb/instant/pull/1326

I'll delay merging and check the autoexplain to be sure we don't still have queries relying on these indexes.

## Deployment plan

- [ ] Merge to main
- [ ] Drop the indexes concurrently
```sql
drop index concurrently if exists vae_index;

drop index concurrently if exists eav_index;
```
- [ ] Run the migrations


stopachka (approved):
Heeck yeah!

PR #1333, opened 2025-07-03
Set default scopes for the MCP server if they're not provided

dwwoelfel:
Fixes a bug with VSCode, where it will generate the client without any scopes, then fail to go through the OAuth flow because the client doesn't have the requested scope.


stopachka (approved):
Niice!

PR #1334, opened 2025-07-03
Stop using the vae_index

dwwoelfel:
Follow-up to https://github.com/instantdb/instant/pull/1326 and prerequisite for https://github.com/instantdb/instant/pull/1330

Modifies some of the datalog queries to replace usage of the `vae_index` with the new `vae_uuid_index`. Once we're no longer using the `vae_index` at all, we can drop it and save ourselves some disk space and cpu on the db instance.

Three main changes:

1. Prefer the `:vae` index over the `:av` index in attr-pat when both are applicable (i.e. querying a unique ref attr). This saves us converting an entity_id to JSON in a lot of cases--we can just directly compare the entity_id to `json_uuid_to_uuid(value)`, which requires no calculation because it's already in the `vae_uuid_index`.
1. Update the join-cols to use the `json_uuid_to_uuid(value) = entity_id` instead of `value = to_jsonb(entity_id)` for better use of the vae index.
2. Update `in-or-eq` to use `[:json_uuid_to_uuid :value]` when it encounters the `vae` index, instead of using `[:to_jsonb :value]`

The structure of the symbol map changed. Instead of being a two-element array, it's now a map with some extra details. `[0 1]` => `{:pattern-idx 0 :triples-idx 1, :ctype :a}`. I did that so we could pass down whether the value was a ref into the join cols. It also has the benefit of making it easier to write the `join-cond-for-or`. The `or`s need to fully qualify the 

Minor changes:
1. Renamed :value-blob to just :value. It's no longer needed now that we stopped tracking the `:value-uuid` and makes the code a bit cleaner.
2. Replaced the `:match-0-` prefix with just `:m`. It makes the queries easier to read and it's less data to and from the db.

Did a backtest and didn't find any queries that got faster or slower.

### DIFF
--- a/server/src/instant/db/datalog.clj
+++ b/server/src/instant/db/datalog.clj
@@ -830,8 +830,6 @@
 (def ^:private component-type->col-name
   {:e :entity-id :a :attr-id :v :value :created-at :created-at})
 
-(s/def ::join-col-ctype #{:e :a :v :created-at :v-ref-value})
-
 (defn qualify-col [prefix binding-path col]
   (kw prefix (:pattern-idx binding-path) "-" col))
 


stopachka (approved):
Biyutiful!

server/test/instant/db/datalog_test.clj
```diff
@@ -80,30 +80,21 @@
...
   (testing "join conditions match values"
-    (is (= '(([:= :value :match-0-value-blob]))
+    (is (= '(([:= :value :match-0-value]))
            (raw-pats->join-conds
             '[[:eav _ _ ?a]
               [:av _ _ ?a]]))))
   (testing "join conditions matches values to coerced entities"
-    (is (= '(([:= :value [:to_jsonb :match-0-entity-id]]))
-           (raw-pats->join-conds '[[:ea ?a] [:eav _ _ ?a]]))))
-  (testing "join conditions matches values to coerced attrs"
```
dwwoelfel, line 98: I removed the test cases where we compare on attr ids because we never actually compare attr ids to anything.

PR #1335, opened 2025-07-04
feat(types): add type safety for order fields

lmssiehdev:
Recording 2025-07-04 at 19 21 11

### DIFF
--- a/client/packages/core/src/queryTypes.ts
+++ b/client/packages/core/src/queryTypes.ts
@@ -68,12 +68,13 @@ type Cursor = [string, string, any, number];
 
 type Direction = 'asc' | 'desc';
 
-type Order = { [key: string]: Direction };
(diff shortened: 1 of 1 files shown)


dwwoelfel (changes requested):
Thanks for submitting this PR!

There are two issues to address:

1. There is a special `serverCreatedAt` field that will sort by when the entity was created. It's not in the list of fields, so anyone sorting by `serverCreatedAt` would get a type error. That should be the easier one to fix--I think we'd just need to hardcode that value in the `Order` type.

2. We can only sort by fields that are typed and indexed. This one is a little harder to implement. First, we'd have to parameterize `$Option` by an array of `DataAttrDef`s instead of an array of strings. We'd also have to add whether the field is indexed to the `DataAttrDef`. Right now we only track `ValueType` and `IsRequired` in the `DataAttrDef`, but we could add a 3rd `IsIndexed` field.

PR #1336, opened 2025-07-04
Upgrade uuid dependency

nezaj:
Looks like we were on an older version of the `uuid` package which was causing us to break in Nuxt. [Discord](https://discord.com/channels/1031957483243188235/1390428798255628339)

Upgrading to latest fixes the issue with Nuxt. 

Did the following smoke tests to see if this causes any unexpected break-age

* Verified our sandbox apps work / admin tests work
* Published experimental package and verified things work w/ react


stopachka (approved):
SGTM!

PR #1339, opened 2025-07-07
[CLI] Include etype in cli

nezaj:
Per user request! Cleaned up the logic a little bit too to make it easier to get the fences to line up

**Before**
CleanShot 2025-07-07 at 09 59 46@2x

**After**

CleanShot 2025-07-07 at 09 55 08@2x

### DIFF
--- a/client/packages/cli/src/index.js
+++ b/client/packages/cli/src/index.js
@@ -966,7 +966,7 @@ function indexingJobCompletedMessage(job) {
       );
 
       const columns = [
-        { header: 'etype', width: 15, getValue: () => etype },
+        { header: 'namespace', width: 15, getValue: () => etype },
         {
           header: 'id',
           width: 37,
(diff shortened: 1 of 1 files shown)


client/packages/cli/src/index.js
```diff
@@ -943,36 +962,44 @@ function indexingJobCompletedMessage(job) {
       const samples = formatSamples(job.invalid_triples_sample);
       const longestValue = samples.reduce(
         (acc, { value }) => Math.max(acc, value.length),
-        // Start with length of label
         label.length,
       );
 
+      const columns = [
+        { header: 'etype', width: 15, getValue: () => etype },
```
dwwoelfel:
```suggestion
        { header: 'namespace', width: 15, getValue: () => etype },
```

I think we typically call this a `namespace` (or sometimes `table`) in user-facing areas.

PR #1343, opened 2025-07-07
Add InstantAuthContext and useCurrentUser

nezaj:
This adds `InstantAuthContext` and `useCurrentUser` to the react and react-native packages

## Problem

`db.useAuth` requires type assertions even when we already know for sure we are authed

```tsx
// Imagine we already authed earlier
function MyComponent() {
  const { user } = db.useAuth();
  // Need user! type assertions since TypeScript doesn't know user is guaranteed
  return <div>Hello {user!.email}</div>;
}
```

## Solution

Added `InstantAuthContext` and `useCurrentUser` hook to eliminate loading states and null checks in child components:

```tsx
iimport { init, InstantAuthContext, useCurrentUser } from '@instantdb/react';
import { schema } from '../instant.schema.ts';
                                                                             
const db = init({ appId: 'your-app-id', schema });
                                                                             
function App() {
  const { user, isLoading, error } = db.useAuth();
  if (isLoading) return null; // or a loading spinner
  if (error) return <div>Error: {error.message}</div>;
                                                                             
                                                                             
  return (
    <InstantAuthContext.Provider value={user}>
      <MyComponent />
    </InstantAuthContext.Provider>
  );
}
                                                                             
function MyComponent() {
  const user = useLoggedInUser(); // Never null when user is authenticated
  return <div>Hello {user.email}</div>;
}

```

## Next Steps

To make this pattern more ergonomic we can add an `InstantAuthProvider` component with some defaults for loading, error, and login states. We can also allow users to supply their own logic for those. 

Wanted to put this up though to get thoughts. If we like this can go ahead and add `InstantAuthProvider` and then some docs.

### DIFF
--- a/client/packages/react/src/InstantReactAbstractDatabase.ts
+++ b/client/packages/react/src/InstantReactAbstractDatabase.ts
@@ -365,17 +365,17 @@ export default abstract class InstantReactAbstractDatabase<
    * }
    *
    * function MyComponent() {
-   *   const user = db.useCurrentUser(); // Never null!
+   *   const user = db.useLoggedInUser(); // Never null!
    *   return <div>Hello {user.email}</div>;
    * }
    * ```
    */
-  useCurrentUser = (): User => {
+  useLoggedInUser = (): User => {
     const user = useContext(InstantAuthContext);
 
     if (user === null || user === undefined) {
       throw new Error(
-        'useCurrentUser must be used within an InstantAuthContext.Provider with an authenticated user. ' +
+        'useLoggedInUser must be used within an InstantAuthContext.Provider with an authenticated user. ' +
           'Make sure you have wrapped your component with InstantAuthContext.Provider and passed a valid user.',
       );
     }
--- a/client/sandbox/react-nextjs/pages/play/auth-provider-test.tsx
+++ b/client/sandbox/react-nextjs/pages/play/auth-provider-test.tsx
@@ -5,9 +5,9 @@ import Login from '../../components/Login';
 // Initialize the database
 const db = init(config);
 
-// Test component that uses useCurrentUser
+// Test component that uses useLoggedInUser
 function AuthenticatedApp() {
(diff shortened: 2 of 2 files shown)


client/packages/react-native/src/index.ts
```diff
@@ -7,6 +7,8 @@ import version from './version';
 import {
   // react
   InstantReactAbstractDatabase,
+  InstantAuthContext,
+  useCurrentUser,
```
stopachka: This would be the first hook that we expose outside of `db`. 

Maybe we should push it back in there?

client/packages/react-native/src/index.ts
```diff
@@ -7,6 +7,8 @@ import version from './version';
 import {
   // react
   InstantReactAbstractDatabase,
+  InstantAuthContext,
```
stopachka, line 10: i.m.o we should ship this PR with the `InstantAuthProvider`.

client/packages/react/src/useCurrentUser.ts
```diff
@@ -0,0 +1,49 @@
...
+ *   );
+ * }
+ *
+ * function MyComponent() {
+ *   const user = useCurrentUser(); // Never null when user is authenticated
+ *   return <div>Hello {user.email}</div>;
+ * }
+ * ```
+ */
+export function useCurrentUser(): User {
```
stopachka: maybe we can have this called like `useLoggedInUser`. This way it's a bit clearer?

client/packages/react-native/src/index.ts
nezaj: Good point! Updated!

client/packages/react/src/useCurrentUser.ts
nezaj: Updated!

nezaj:
Will close this for now and come back to it later with a proposal with how a Provider pattern

PR #1344, opened 2025-07-08
Add hints to JS errors

tonsky:
While debugging #1342, I was struggling with understanding what the problem was. The error I was seeing was too generic:

Screenshot 2025-07-08 at 17 32 40 copy

Turned out we actually have more detailed error, but it’s hidden inside hints. This PR adds it to the main message:

Screenshot 2025-07-08 at 17 32 40


stopachka (approved):
Nice!

client/packages/core/src/utils/fetch.ts
```diff
@@ -83,7 +83,32 @@ export class InstantAPIError extends Error {
...
-    const message = error.body?.message || `API Error (${error.status})`;
+    var message = `API Error (${error.status})`;
+
+    let body = error.body;
+
+    if (body?.message) {
+      message = body?.message;
+    }
+
+    if (body?.type === 'validation-failed' && body.hint?.errors) {
```
stopachka: [We have similar kinds of error massaging in www too: ](https://github.com/instantdb/instant/blob/2c2e6f1518fe086cd4ecc0dfc70f1f511882b801/client/www/lib/errors.ts#L14)

This made me think: maybe down the road we should actually do this kind of massaging in the server. This way all clients would get better errors by default.

tonsky:
Done, please take a look @dwwoelfel @stopachka

server/dev/user.clj
```diff
@@ -9,15 +9,18 @@
 
 (.doReset #'*warn-on-reflection* true)
 
-(clojure+.error/install!)
+(clojure+.error/install!
+ {:trace-transform
+  (fn [trace]
+    (take-while #(not (#{"Compiler" "clj-reload" "clojure-sublimed"} (:ns %))) trace))})
```
dwwoelfel, line 15: What does this do?

tonsky: This is for dev mode only, cuts down parts of stacktraces that you usually don’t care about (deep in Clojure compiler or clj-reload)

PR #1346, opened 2025-07-08
typed where clause

drew-harris:
Creates new types to more strictly enforce values in the ${ where: {}} clause based on the schema. 
Mostly backwards compatible with queries that don't involve a schema except for 2 regressions. 
```ts
  const r9 = dummyQuery({ users: { $: { where: { foo: {} } } } });
```
Foo should error because we dont want to allow empty `WhereClauseValueComplex`s but that breaks the detection for `$isNull`

```ts
  const s5 = dummyQuery({
    users: { $: { where: { val: { $gt: { val: 'a' } } } } },
  });
 ```
` BaseWhereClauseValueComplex<V>` is too permissive but the way that V is established should already narrow it down but unsure where that happens. 

# Exactly 
From queryTypes.ts
>   Typescript can permit extra keys when a generic extends a type.
> For some reason, it makes it possible to write a query like so:
> ```ts
> dummyQuery({
>  users: {
>    $: { where: { "foo": 1 } },
>    posts: {
>      $: { "far": {} }
>    }
>  }
>  ```
> 
>  The problem: $: { "far": {} }
> 
>  This passes, when it should in reality fail. I don't know why adding `Exactly` fixes this, but it does.

For this reason, `Exactly` was added to the query methods in the core package and the react package. 
Typescript allows extra values but what Exactly does is check each key in the object to see if it is part of the specified type and if not it will reject it as `never`

### CODE REVIEW

File: client/packages/core/__tests__/src/schema.test.ts [hunk 1]
```diff
@@ -106,4 +106,21 @@ test('runs without exception', () => {
   type DemoQueryResult = InstaQLResult<Schema, typeof demoQuery>;
   queryResult?.users[0].friends[0]._friends[0].bio;
   queryResult?.users[0].posts[0].author?.junk;
+
+  // Test dot notation in where clauses
```
stopachka: You might want to add a few examples that you expect to fail too

client/packages/core/src/queryTypes.ts
```diff
@@ -70,9 +180,36 @@ type Direction = 'asc' | 'desc';
...
+    first?: number;
+    offset?: number;
+    after?: Cursor;
+    before?: Cursor;
+    fields?: Fields;
+  };
+};
+
+// Typed version that supports dot notation for nested queries
+type $OptionWNest<
```
stopachka: Q: What does WNest stand for? 

Maybe we can think of a better name here.

client/packages/core/src/queryTypes.ts
```diff
@@ -70,9 +180,36 @@ type Direction = 'asc' | 'desc';
 
 type Order = { [key: string]: Direction };
 
-type $Option<Fields extends string[]> = {
+type $Option<
```
stopachka: [to-discuss] I wonder if we need $Option. 

Afaik we always will have an `S` passed into the query now.

client/packages/core/src/queryTypes.ts
```diff
@@ -70,9 +180,36 @@ type Direction = 'asc' | 'desc';
...
+    before?: Cursor;
+    fields?: Fields;
+  };
+};
+
+// Typed version that supports dot notation for nested queries
+type $OptionWNest<
+  S extends IContainEntitiesAndLinks<any, any>,
+  K extends keyof S['entities'],
+  Fields extends string[],
```
stopachka: nit: but I wonder if we need to actually pass these in as third and forth args. They can both be derived from `S` and `K`. It might make this type simpler to read

client/packages/core/src/queryTypes.ts
```diff
@@ -70,9 +180,36 @@ type Direction = 'asc' | 'desc';
...
+  WhereFieldTypes extends {
+    [key: string]: any;
+  } = Record<any, BSUnknown>,
+> = {
   $?: {
-    where?: WhereClause;
+    where?: WhereClause<WhereFieldTypes> & TypedFreeWhereClause<S, K>;
```
stopachka: [to-discuss] Why do we need both `WhereClause` and `TypedFreeWhereClause`? Maybe we can get away with just one?

client/packages/core/src/queryTypes.ts
```diff
@@ -42,20 +45,127 @@ type WhereArgs = {
...
+                  S,
+                  LinkedEntityName,
+                  [never, 0, 1, 2, 3][Depth]
+                >}`
+              : never
+            : never;
+        }[keyof S['entities'][K]['links']];
+
+type WhereClauseWithCombination<T extends Record<any, unknown>> = {
+  // do infer here
```
stopachka: did you intend to do an infer here?

client/packages/core/src/queryTypes.ts
```diff
@@ -42,20 +45,127 @@ type WhereArgs = {
...
-  or?: WhereClause[] | WhereClauseValue;
-  and?: WhereClause[] | WhereClauseValue;
+// Helper type to infer the value type from a nested path
+type InferNestedValueType<
+  S extends IContainEntitiesAndLinks<any, any>,
+  K extends keyof S['entities'],
+  Path extends string,
+  Depth extends number = 3,
+> = Depth extends 0
+  ? BSUnknown
```
stopachka: I assume Depth is counting down? I was a bit surprised, as I would have expected the number to count up

client/packages/core/src/queryTypes.ts
```diff
@@ -324,18 +461,47 @@ type InstaQLQuerySubqueryParams<
...
+        S,
+        K,
+        InstaQLFields<S, K>,
+        MakeIdOptional<InstaQLEntity<S, K>>
+      >
+    | ($OptionWNest<
+        S,
+        K,
+        InstaQLFields<S, K>,
+        MakeIdOptional<InstaQLEntity<S, K>>
```
stopachka: Why do we need to make id optional?

stopachka (commented):
So cool!

client/packages/core/__tests__/src/schema.test.ts
```diff
@@ -106,4 +112,527 @@ test('runs without exception', () => {
...
+  });
+
+  // Numeric comparisons
+  dummyQuery({
+    posts: {
+      $: {
+        where: {
+          // Using any field for numeric comparisons since we don't have numeric fields
+          // This tests the BSUnknown typing
+          'author.stuff.custom': { $gt: 'a', $lt: 'z' },
```
dwwoelfel: We don't throw on this, but it's also not supported to provide more than one property in the query right now. We'll silently ignore either the `$gt` or the `$lt`. 

To get the query you want, you'd have to do:

```js
{where: {and: [{'author.stuff.custom': {$gt: 'a'}},
               {'author.stuff.custom': {$lt: 'z'}}]}}
```

But it could be a nice project to make multiple filters work like this.

client/packages/core/src/sanityCheckQueries.ts
```diff
@@ -130,7 +120,7 @@ const sanityCheckQueries = () => {
   // Bad $ clauses fail
   // @ts-expect-error
   const r8 = dummyQuery({ users: { $: { where: 'foo' } } });
-  // @ts-expect-error
+  // REMOVED BY DREW
   const r9 = dummyQuery({ users: { $: { where: { foo: {} } } } });
```
stopachka, line 124: I would add a note here explaining the issue

client/packages/core/src/sanityCheckQueries.ts
```diff
@@ -149,8 +139,7 @@ const sanityCheckQueries = () => {
   });
 
   const s5 = dummyQuery({
-    // @ts-expect-error
-    users: { $: { where: { val: { $not: { val: 'a' } } } } },
```
stopachka, line 153: should this be undone?

client/packages/core/__tests__/src/schema.test.ts
```diff
@@ -106,4 +112,527 @@ test('runs without exception', () => {
...
+          ],
+        },
+      },
+    },
+  });
+
+  // ===== COMPREHENSIVE TEST QUERIES =====
+
+  // Basic filtering with different operators
+  dummyQuery({
```
stopachka: Nit: it's not 100% clear what we should be expecting here. 

I would make some of this a bit more descriptive. 

For example, instead of "comprehensive test queries"

Maybe this section is really about testing the $ options 

Then for each one in this were, I would add a comment explaining that what you expect that to work. 

i.e `name` works cuz it's string
email cuz it's indexed, bio cuz it's string or something like that.

Then we can add a specific example where we expect this to fail. i.e name is number, another one where email is not indexed etc

client/packages/core/__tests__/src/schema.test.ts
```diff
@@ -106,4 +112,527 @@ test('runs without exception', () => {
...
+              email: { $like: '%@company.com' },
+            },
+          },
+        },
+      },
+    },
+  });
+
+  // Complex nested filtering with dot notation
+  const resulsjdlf = dummyQuery({
```
stopachka: Maybe each of these examples should have a result? Otherwise it's going to be hard to hover over the stuff later to figure out what the output type is.

client/packages/core/src/queryTypes.ts
```diff
@@ -84,9 +139,34 @@ type $Option<Fields extends string[]> = {
...
+    fields?: HasIndexSignature<InstaQLEntity<S, K>> extends true
+      ? string[]
+      : InstaQLFields<S, K>;
+  };
+};
+
 type Subquery = { [namespace: string]: NamespaceVal };
 
-type NamespaceVal = $Option<string[]> | ($Option<string[]> & Subquery);
+type NamespaceVal =
```
stopachka, line 154: Nit: maybe we can undo this? afaik we don't really use this anywhere

client/packages/core/src/queryTypes.ts
```diff
@@ -84,9 +139,34 @@ type $Option<Fields extends string[]> = {
...
+  $?: {
+    where?: WhereClause<S, K>;
+    order?: Order;
+    limit?: number;
+    last?: number;
+    first?: number;
+    offset?: number;
+    after?: Cursor;
+    before?: Cursor;
+    fields?: HasIndexSignature<InstaQLEntity<S, K>> extends true
```
stopachka: Can we try to remove this check? 

imo if the index signature exists here, wouldn't InstaQLFields end up basically returning string[]? 

The idea being we should do our best to not special-case against a specific kind of schema

client/packages/core/src/queryTypes.ts
```diff
@@ -42,20 +45,67 @@ type WhereArgs = {
...
-  and?: WhereClause[] | WhereClauseValue;
+type WhereClauseWithCombination<T extends Record<any, unknown>> = {
+  or?: BaseAttrWhereClause<T>[] | WhereClauseValue<BSUnknown>;
+  and?: BaseAttrWhereClause<T>[] | WhereClauseValue<BSUnknown>;
 };
 
-type WhereClause =
-  | WhereClauseWithCombination
-  | (WhereClauseWithCombination & BaseWhereClause);
+type BaseAttrWhereClause<T extends Record<any, any>> =
```
stopachka: Maybe we can all this SchemaWhereClause? 

Idea being it's similar to what we are now trying to do with SchemaOption.

stopachka: Maybe this can take an S, K as well

client/packages/core/src/queryTypes.ts
```diff
@@ -42,20 +45,67 @@ type WhereArgs = {
...
+  K extends keyof S['entities'],
+> = BaseAttrWhereClause<InstaQLEntity<S, K>> & {
+  [key: `${string}.${string}`]: WhereClauseValue<BSUnknown>;
 };
 
-type WhereClauseWithCombination = {
-  or?: WhereClause[] | WhereClauseValue;
-  and?: WhereClause[] | WhereClauseValue;
+type WhereClauseWithCombination<T extends Record<any, unknown>> = {
+  or?: BaseAttrWhereClause<T>[] | WhereClauseValue<BSUnknown>;
```
stopachka: Do we need BSUnknown here?

Maybe what we can do is give a type of [string]: string | number | boolean

stopachka: In which case maybe WhereClauseValue can just return V | undefined

client/packages/core/src/queryTypes.ts
```diff
@@ -28,7 +28,10 @@ type NonEmpty<T> = {
   [K in keyof T]-?: Required<Pick<T, K>>;
 }[keyof T];
 
-type WhereArgs = {
+// // Branded Schemaless unknown
+// type BSUnknown = { _unknown: '_unknown' };
```
stopachka: Remove

client/packages/core/src/queryTypes.ts
```diff
@@ -84,9 +152,34 @@ type $Option<Fields extends string[]> = {
   };
 };
 
+// Helper type to check if an entity has index signature (unknown schema)
+type HasIndexSignature<T> = string extends keyof T ? true : false;
```
stopachka: Rm! : D

client/packages/core/src/queryTypes.ts
```diff
@@ -70,23 +102,27 @@ type Direction = 'asc' | 'desc';
 
 type Order = { [key: string]: Direction };
 
-type $Option<Fields extends string[]> = {
+// Typed version that supports dot notation for nested queries
```
stopachka: rm

client/packages/core/src/sanityCheckQueries.ts
```diff
@@ -191,18 +187,7 @@ const sanityCheckSchemalessResponses = () => {
   // @ts-expect-error
   r3.$;
 };
+
 function sanityCheckSchemadResponses() {
```
stopachka: Let's just delete the function then?

drew-harris:
Note: the Exactly type was removed which significantly reduces type safety in some cases but the test's have been kept, however the `@ts-expect-error` has been replaced with `#ts-expect-error` for the affected cases so that when we are ready to fix those, search and replace can be used.

stopachka:
cc @drew-harris could we push up another experimental version to confirm with Cam?

client/packages/core/src/queryTypes.ts
```diff
@@ -30,34 +32,74 @@ type NonEmpty<T> = {
...
+            }
+          : {}));
+
+// Make type display better
+type WhereClauseValue<
+  D extends DataAttrDef<string | number | boolean, boolean, boolean>,
+> =
+  D extends DataAttrDef<infer V, infer R, infer I>
+    ?
+        | (IsAny<V> extends true ? string | number | boolean : V)
```
stopachka, line 77: This should not be necessary

client/packages/core/src/queryTypes.ts
```diff
@@ -30,34 +32,74 @@ type NonEmpty<T> = {
...
+  $in?: V[];
+  $not?: V;
+  $gt?: V;
+  $lt?: V;
+  $gte?: V;
+  $lte?: V;
 };
 
-type WhereClauseValue = string | number | boolean | NonEmpty<WhereArgs>;
+type IsAny<T> = boolean extends (T extends never ? true : false) ? true : false;
```
stopachka, line 46: We shouldn't need IsAny

PR #1356, opened 2025-07-10
apply button to modify constraints for attrs all at once

drew-harris:


client/www/lib/indexingJobs.ts
```diff
@@ -61,14 +61,15 @@ export function jobFetchLoop(appId: string, jobId: string, token: string) {
           { headers: { authorization: `Bearer ${token}` } },
         );
         const body = await res.text();
-        const json = JSON.parse(body);
+        const json = await JSON.parse(body);
```
dwwoelfel: You don't need an await here, do you?

client/www/lib/hooks/useEditBlobConstraints.ts
```diff
@@ -0,0 +1,204 @@
...
+    );
+    if (isRunning) {
+      return;
+    }
+
+    // Pending requirement job
+    if (isRequired === attr.isRequired) {
+      setPendingJobs((p) => ({ ...p, require: undefined }));
+    } else if (isRequired) {
+      setPendingJobs((p) => ({ ...p, require: { jobType: 'required' } }));
```
stopachka, line 53: This made me think: is it possible to get in some kind of race condition where we are running a `required` job, but immediately try to run a `remove-required` job. 

Can something like that happen?

client/www/lib/hooks/useEditBlobConstraints.ts
```diff
@@ -0,0 +1,204 @@
...
+          ...p,
+          type: { jobType: 'check-data-type', checkedDataType },
+        }));
+      }
+    }
+  }, [isRequired, isIndexed, isUnique, checkedDataType, attr, runningjobs]);
+
+  const apply = async () => {
+    // Clean up previous errors
+    setRunningJobs({});
```
stopachka, line 109: Should we make this no-op if someone clicks apply twice?

client/www/components/ui.tsx
```diff
@@ -1077,6 +1077,48 @@ export const InfoTip = ({ children }: PropsWithChildren) => {
...
+  ...props
+}: PropsWithChildren<{
+  percentage?: number;
+  loading?: boolean;
+  className?: string;
+  variant?: 'primary' | 'secondary' | 'subtle' | 'destructive' | 'cta';
+}> &
+  Parameters<typeof Button>[0]) {
+  const progressFillStyle = {
+    width: loading ? `${Math.max(0, Math.min(100, percentage))}%` : '0%',
```
stopachka: If I remember correctly `width` can cause layout reflows in the browser. Maybe transform: scaleX will make this faster?

client/www/lib/hooks/useEditBlobConstraints.ts
```diff
@@ -0,0 +1,204 @@
...
+          setProgress((prev) => ({ ...prev, [jobType]: percent }));
+
+          setRunningJobs((prev) => ({
+            ...prev,
+            [jobType]: updatedJob,
+          }));
+        }
+      });
+    }
+  }, [runningjobs, appId, token]);
```
stopachka: One thought I had: 

Is this useEffect necc? 

It _feels_ like we should be able to kick all of this off inside `apply`?

client/www/components/dash/explorer/EditNamespaceDialog.tsx
```diff
@@ -1016,371 +1033,340 @@ async function updateRequired({
...
 }
 
-function EditRequired({
-  appId,
-  attr,
-  isSystemCatalogNs,
-  pushNavStack,
-}: {
-  appId: string;
+type BlobConstraintControlComponent<V> = (props: {
```
stopachka, line 1036: nice idea!

drew-harris:
Converted to draft: clicking label does not toggle checkbox on chrome

stopachka (approved):
Left some thoughts! Overall LGTM!

client/www/lib/indexingJobs.ts
drew-harris: I need to parse the body so I can check if the job is finished to stop polling

dwwoelfel: But isn't JSON.parse synchronous already?

drew-harris: Oh wait you are right. I was thinking it was doing Response.json()

PR #1357, opened 2025-07-11
Persist attr changes to pendingMutations

dwwoelfel:
This fixes a bug where we hold on to old mutations that have `add-attr` and then overwrite the attrs from the server with our old `add-attr` when we call `optimisticAttrs`.

You can sometimes hit the bug if you add a new attr to a namespace from the explorer and then try to modify the attr by e.g. indexing it. If we keep the `add-attr` mutation around for a query that hasn't seen a `refresh-ok`, then we'll apply the `add-attr` over the actual attr in `optimisticAttrs` and it will look like the attr hasn't been indexed.

The fix implemented in the PR does two things:

1. It filters out `add-attr/update-attr/delete-attr` that we have a newer versions of in `rewriteMutations`, if you provide a `processedTxId` to the function
2. It persists the result of `rewriteMutations` to `pendingMutations` when we call it in `refresh-ok`. This seems like a good idea in general, because then we don't need to spend effort rewriting the mutations every time. But it's also necessary here because our fix would break when you refreshed the page otherwise.

As a side effect, it also does a little less work in `rewriteMutations`. While there are no changes to the attrs, we don't bother calling `instaml.rewriteStep`. Then we can use `===` to see if we need to save the new mutations in the `refresh-ok` handler.

### DIFF
--- a/client/packages/core/src/Reactor.js
+++ b/client/packages/core/src/Reactor.js
@@ -451,18 +451,27 @@ export default class Reactor {
         break;
       case 'refresh-ok':
         const { computations, attrs } = msg;
+        const processedTxId = msg['processed-tx-id'];
         if (attrs) {
           this._setAttrs(attrs);
         }
 
         this._cleanupPendingMutationsTimeout();
 
-        const mutations = this._rewriteMutationsSorted(
+        const rewrittenMutations = this._rewriteMutations(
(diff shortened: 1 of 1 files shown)


stopachka (approved):
Nice sleuthing @dwwoelfel !

PR #1359, opened 2025-07-11
[WWW] Add intern index page

nezaj:
Thought it would be nice to have an index page for all our current internal tools!

This also adds a `dash` endpoint to check whether we are an admin. 

CleanShot 2025-07-11 at 09 28 03@2x


tonsky:
That’s a fantastic idea! LGTM

PR #1360, opened 2025-07-11
Speed up the intern/paid query

dwwoelfel:
The paid page is timing out because it takes too long to figure out how big the databases are for all of the paid apps.

We should build a faster way of getting the db size (best option is probably a trigger on triples insert/update/delete that writes a new row to a usage table, then we periodically delete everything in the usage table, get the sum, and add it to a total on the app). For now this just samples the first 100,000 triples for the app and then multiples by the total triples over the sample size to get an estimate.

It's fairly accurate, the worst estimate for the current apps is off by 8%.

As a bonus, we get the number of triples in the app on the paid page.

### DIFF
--- a/client/www/pages/intern/paid.tsx
+++ b/client/www/pages/intern/paid.tsx
@@ -78,33 +78,86 @@ function humanBytes(bytes: number) {
 }
 
 function PaidTable({ data }: { data: any }) {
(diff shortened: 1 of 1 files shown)


nezaj (approved):
Neat workaround!

PR #1361, opened 2025-07-11
Destructure error in the docs

dwwoelfel:
Updates the docs in a few places to destructure `error` when using `db.useQuery`. 

I noticed that claude didn't destructure the error and it made it harder to debug the failing query. Hopefully giving it more examples will clue it in that it needs to handle query errors.


nezaj (approved):
Nice!

PR #1364, opened 2025-07-11
[core] Add configurable offline query cache size (#1362)

nezaj:
Adds a new option to `init` to allow users to specify the number of queries they want to cache. The default it still 10, but now you can do ``init({ appId ..., schema, queryCacheLimit: 50 })` to persist the last 50 queries.

You can test this out in the sandbox app at `http://localhost:4000/play/query-cache-test`

https://github.com/user-attachments/assets/e3748b81-6aa0-46ce-85c5-21598d391d01

If this looks good can add docs in a follow-up PR!

--

In making the sandbox app I discovered two bugs

### Problem
- When switching between different todos (changing the query), the component would briefly show data from the previously selected todo
- This created a confusing UX where clicking Todo B, which was not cached, showed Todo A data

Looking at the `useQueryInternal` function in `/packages/react/src/useQuery.ts`, there was an issue in how `resultCacheRef` was being managed:

```typescript
// Line 56-58: This ref is only initialized ONCE on mount
const resultCacheRef = useRef<InstaQLLifecycleState<Schema, Q>>(
  stateForResult(_core._reactor.getPreviousResult(query)),
);
```
When the query changes (e.g., selecting a different todo), the `queryHash` changes and a new `subscribe` function is created, BUT the `resultCacheRef` still contains the data from the **previous query**.

### Fix
Updated the `useQuery.ts` file to properly handle query changes by tracking when the query hash changes and updating the result cache accordingly.

```typescript
// Added after resultCacheRef initialization:
// Update the ref when the query changes to avoid showing stale data
const previousQueryHash = useRef(queryHash);
if (previousQueryHash.current !== queryHash) {
  resultCacheRef.current = stateForResult(
    _core._reactor.getPreviousResult(query),
  );
  previousQueryHash.current = queryHash;
}
```

How it works now

1.`previousQueryHash` is now responsible for detecting when the query actually changes
2. When the query does changes, we update `resultCacheRef.current` to show previous data we had for the new query (if there was any)
This ensures you don't see Todo A's data when you click on Todo B

### Problem
In the `_onMergeQuerySubs` method, when deciding which stored queries to restore from IndexedDB, the selection was based on:

```typescript
const storageKsToAdd = Object.keys(storageSubs)
  .filter((k) => !inMemorySubs[k])
  .slice(0, this.queryCacheLimit);
```

This approach had no consideration for which queries were most recently or frequently used.

### Fix
Added a timestamp so we can keep the most recently accessed queries:

1. **Track Access Time**: Added `lastAccessed` timestamp when creating/accessing query subscriptions in `_startQuerySub()`:

   ```typescript
   prev[hash].lastAccessed = Date.now();
   ```

2. **Sort by Recency**: Updated the selection logic to sort by `lastAccessed` before taking the first N queries:
   ```typescript
   const storageKsToAdd = Object.keys(storageSubs)
     .filter((k) => !inMemorySubs[k])
     .sort((a, b) => {
       // Sort by lastAccessed, newest first
       const aTime = storageSubs[a]?.lastAccessed || 0;
       const bTime = storageSubs[b]?.lastAccessed || 0;
       return bTime - aTime;
     })
     .slice(0, this.queryCacheLimit);
   ```

### DIFF
--- a/client/packages/core/src/Reactor.js
+++ b/client/packages/core/src/Reactor.js
@@ -173,6 +173,7 @@ export default class Reactor {
     versions,
   ) {
     this.config = { ...defaultConfig, ...config };
+    this.queryCacheLimit = this.config.queryCacheLimit ?? 10;
 
     this._log = createLogger(
       config.verbose || flags.devBackend || flags.instantLogs,
@@ -346,7 +347,13 @@ export default class Reactor {
     // we can keep usage information about which queries are popular.
     const storageKsToAdd = Object.keys(storageSubs)
       .filter((k) => !inMemorySubs[k])
-      .slice(0, 10);
+      .sort((a, b) => {
+        // Sort by lastAccessed, newest first
+        const aTime = storageSubs[a]?.lastAccessed || 0;
+        const bTime = storageSubs[b]?.lastAccessed || 0;
+        return bTime - aTime;
+      })
+      .slice(0, this.queryCacheLimit);
 
     storageKsToAdd.forEach((k) => {
       ret[k] = storageSubs[k];
@@ -702,6 +709,7 @@ export default class Reactor {
     const eventId = uuid();
     this.querySubs.set((prev) => {
       prev[hash] = prev[hash] || { q, result: null, eventId };
+      prev[hash].lastAccessed = Date.now();
       return prev;
     });
     this._trySendAuthed(eventId, { op: 'add-query', q });
--- a/client/packages/react/src/useQuery.ts
+++ b/client/packages/react/src/useQuery.ts
@@ -56,6 +56,15 @@ export function useQueryInternal<
     stateForResult(_core._reactor.getPreviousResult(query)),
   );
 
+  // Update the ref when the query changes to avoid showing stale data
+  const previousQueryHash = useRef(queryHash);
+  if (previousQueryHash.current !== queryHash) {
+    resultCacheRef.current = stateForResult(
+      _core._reactor.getPreviousResult(query),
+    );
+    previousQueryHash.current = queryHash;
+  }
+
   // Similar to `resultCacheRef`, `useSyncExternalStore` will unsubscribe
   // if `subscribe` changes, so we use `useCallback` to memoize the function.
   const subscribe = useCallback(
--- a/client/sandbox/react-nextjs/components/EphemeralAppPage.tsx
+++ b/client/sandbox/react-nextjs/components/EphemeralAppPage.tsx
@@ -6,6 +6,7 @@ import {
   InstantReactAbstractDatabase,
   InstantSchemaDef,
   LinksDef,
+  Config,
 } from '@instantdb/react';
 import { useEffect, useState } from 'react';
 import { RoomsDef, TransactionChunk } from '../../../packages/core/dist/esm';
@@ -67,6 +68,7 @@ function AppPage<
   perms,
   onCreateApp,
   Component,
+  addConfig,
 }: {
   urlAppId: string | undefined;
   schema?: InstantSchemaDef<Entities, Links, Rooms>;
@@ -82,9 +84,10 @@ function AppPage<
     db: InstantReactAbstractDatabase<InstantSchemaDef<Entities, Links, Rooms>>;
     appId: string;
   }>;
+  addConfig?: Partial<Omit<Config, 'appId' | 'schema'>>;
 }) {
   const router = useRouter();
-  const [appId, setAppId] = useState();
+  const [appId, setAppId] = useState<string | undefined>();
   const [error, setError] = useState<string | null>(null);
 
   const provisionApp = async () => {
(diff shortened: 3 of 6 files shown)


client/sandbox/react-nextjs/components/EphemeralAppPage.tsx
```diff
@@ -109,13 +112,19 @@ function AppPage<
...
-      });
+      verifyEphemeralApp({ appId: urlAppId })
+        .then((res): any => {
+          if (res.app) {
+            setAppId(res.app.id);
+          } else {
+            provisionApp();
+          }
+        })
+        .catch(() => {
```
dwwoelfel: You should be able to tell the difference between being offline and the app not existing. 

If you get a 400 back and the body looks like this, then you know that the app is invalid:

```json
{
    "type": "record-not-found",
    "message": "Record not found: app",
    "hint": {
        "args": [
            {
                "id": "db8b51c1-b055-41a6-b0ac-891033908e7a"
            }
        ],
        "record-type": "app"
    },
    "trace-id": "d92b9c84bcbf8d7f86e5f84621351be9"
}
```

client/sandbox/react-nextjs/components/EphemeralAppPage.tsx
```diff
@@ -82,9 +84,10 @@ function AppPage<
     db: InstantReactAbstractDatabase<InstantSchemaDef<Entities, Links, Rooms>>;
     appId: string;
   }>;
+  addConfig?: Partial<Omit<Config, 'appId' | 'schema'>>;
```
dwwoelfel: Maybe we should call this `initConfig`?

nezaj: @dwwoelfel curious why `initConfig`? For context, I intended this variable to allow folks to override the config (I needed this for this case where I'm adjust the cache limit)

client/sandbox/react-nextjs/components/EphemeralAppPage.tsx
nezaj: @dwwoelfel updated! Beforehand the logic handled the record-not-found case in the then statement

```typescript
if (res.app) {
  setAppId(res.app.id);
} else {
  provisionApp();
}
```

I wasn't sure if it was written like that on purpose so didn't change it initially. I do think moving all the error handling in the catch is nicer so I updated `verifyEphemeralApp` to throw if we don't get a 200

```typescript
async function verifyEphemeralApp({ appId }: { appId: string }) {
  const r = await fetch(`${config.apiURI}/dash/apps/ephemeral/${appId}`, {
    headers: {
      'Content-Type': 'application/json',
    },
  });

  const data = await r.json();

  if (!r.ok) {
    throw data;
  }
  return data;
}
```

Now I'll differentiate between being offline / the app not existing or being malformed

```typescript
.catch((err) => {
  if (
    err.type === 'record-not-found' ||
    err.type === 'param-malformed'
  ) {
    // App ID is not valid, provision a new one
    console.error('Error verifying ephemeral app:', err);
    console.log('Provisioning new ephemeral app');
    provisionApp();
    return;
  }
  if (!err.type) {
    // Some other error, maybe we're offline - let's just trust the
    // app ID so we can test offline
    setAppId(urlAppId as any);
    return;
  }
```

client/sandbox/react-nextjs/components/EphemeralAppPage.tsx
stopachka: i.m.o `addConfig` can be a bit confusing because it may mean that 

(a) there is some function called add which we are configuring
(b) there was no config before, but we are adding it 

--- 

Maybe `extraConfig`, since we don't allow app and schema?

client/packages/core/src/Reactor.js
```diff
@@ -346,7 +347,13 @@ export default class Reactor {
     // we can keep usage information about which queries are popular.
     const storageKsToAdd = Object.keys(storageSubs)
       .filter((k) => !inMemorySubs[k])
-      .slice(0, 10);
+      .sort((a, b) => {
+        // Sort by lastAccessed, newest first
+        const aTime = storageSubs[a]?.lastAccessed || 0;
```
stopachka, line 352: nice!

client/packages/react/src/useQuery.ts
```diff
@@ -56,6 +56,15 @@ export function useQueryInternal<
     stateForResult(_core._reactor.getPreviousResult(query)),
   );
 
+  // Update the ref when the query changes to avoid showing stale data
+  const previousQueryHash = useRef(queryHash);
```
stopachka: Maybe we move

```
    resultCacheRef.current = stateForResult(
      _core._reactor.getPreviousResult(query),
    );
```

Inside the useCallback below? This would remove the need for the extra check here. It also makes it so we can keep the two changes that rely on queryHash together: 
 * To update resultCacheRef
 * To update the subscribe callback

stopachka (approved):
Added one comment but SGTM!

client/packages/react/src/useQuery.ts
nezaj: Updated!

PR #1367, opened 2025-07-14
auto install needed react native libs on init command

drew-harris:
When user selects `@instantdb/react-native` as the package they want to install during the cli init command, also install the needed dependencies react-native-get-random-values and react-native-async-storage

### DIFF
--- a/client/packages/cli/src/index.js
+++ b/client/packages/cli/src/index.js
@@ -648,17 +648,31 @@ async function getOrInstallInstantModuleWithErrorLogging(pkgDir) {
   });
 
   const packageManager = await detectPackageManager(pkgDir);
-  const installCommand = getInstallCommand(packageManager, moduleName);
+
+  const packagesToInstall = [moduleName];
+  if (moduleName === '@instantdb/react-native') {
+    packagesToInstall.push(
+      'react-native-get-random-values',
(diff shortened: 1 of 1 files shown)


client/packages/cli/src/index.js
```diff
@@ -647,17 +647,33 @@ async function getOrInstallInstantModuleWithErrorLogging(pkgDir) {
...
   try {
     await execAsync(installCommand, pkgDir);
-    spinner.succeed(`Installed ${moduleName} using ${packageManager}.`);
+    spinner.succeed(
+      `Installed ${packagesToInstall.join(', ')} using ${packageManager}.`,
+    );
   } catch (e) {
-    spinner.fail(`Failed to install ${moduleName} using ${packageManager}.`);
+    spinner.fail(
+      `Failed to install ${packagesToInstall.join(', ')} using ${packageManager}.`,
```
nezaj: It would be great if we could indicate which package specifically failed but perhaps that's too tough

drew-harris: Should I change it to say "Error running {the install command here} that way they can copy paste to run it themselves and see the error?

nezaj: I like that!

PR #1373, opened 2025-07-14
add necessary cli options to skip prompts when using --yes

drew-harris:
When using any of the commands with --yes to skip prompts, require either an -app flag to specify the app or a --title flag to create one.


client/packages/cli/src/index.js
```diff
@@ -386,6 +396,11 @@ program
     '--skip-check-types',
     "Don't check types on the server when pushing schema",
   )
+  .option('-t --title', 'Title for the created app')
```
nezaj, line 400: Curious, why do push and pull need title?

drew-harris: If you run the push/pull command without an app linked, it will give you the option to create one, which you would need to input a title. I do agree that it feels like a weird arg to have but its technically necessary if we want --yes to be able to avoid all prompts. 
Not sure what call to make here

nezaj: cc @stopachka curious your thoughts here.

Does it makes sense to have a flow for creating an app on `push` or `pull` if we don't have an app linked?

stopachka: I originally allowed for this behavior, to make it easier for hackers. But it is annoying if you _know_ what you're doing. 

Maybe we start throwing an error if an app doesn't exist with pull and push?

cc @dwwoelfel @tonsky if you have thoughts

dwwoelfel: I would keep the flow for creating/selecting an app in `push/pull`. It's much nicer to go through the create/select app flow instead of being redirected to init.

nezaj (approved):
Nice!!

PR #1374, opened 2025-07-14
cli: check for src/ folder for initial schema placement

drew-harris:
When pulling the schema and perms for the first time, if there is a `src/` folder the schema and perms will be placed there.


nezaj (approved):
LGTM!

PR #1377, opened 2025-07-15
Use react-markdown for essays

dwwoelfel:
Switches from `marked` to react markdown for essays to make it easier to include dynamic code.

The only difference I can see is that the footnotes links list at the end have a `↩` link instead of linking the number at the start of the footnote.


stopachka (approved):
So cool, thank you @dwwoelfel !

PR #1379, opened 2025-07-15
Increase cors cache to 1 day

dwwoelfel:
The cors cache is working in prod, but 5 minutes is pretty short. This PR increases it to a day.


stopachka (approved):
Woohoo!

PR #1381, opened 2025-07-15
[Docs][Easy] Fuller init example

nezaj:
Slight tweak to the `init` docs to use `schema` and be in typescript

### DIFF
--- a/client/www/pages/docs/init.md
+++ b/client/www/pages/docs/init.md
@@ -63,6 +63,7 @@ times without worrying about creating multiple connections or
 performance overhead. However we do recommend the pattern of exporting a
 reference from a utility file like so:
 
+<!-- prettier-ignore-start -->
 ```typescript
 // lib/db.ts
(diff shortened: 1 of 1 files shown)


client/www/pages/docs/init.md
```diff
@@ -78,3 +54,33 @@ const db = init({ appId: APP_ID, schema });
...
+export const db = init({ appId: APP_ID, schema });
+
+// app/page.tsx
+"use client";
+
+import { db } from '../lib/db';
+
+function App() {
+  // do some instant magic 🪄
+  db.useQuery({ ... });
```
stopachka: Thought: perhaps we should fill this in with `{ posts: { } } `

Or something like that?

stopachka (approved):
Cool!

client/www/pages/docs/init.md
nezaj:
```suggestion
  db.useQuery({ todos: {} });
```

PR #1383, opened 2025-07-15
[Pricing] Add platform to pricing

nezaj:
Thought it would be nice to update the pricing page with the launch

CleanShot 2025-07-15 at 09 47 23@2x


stopachka (approved):
Nice!

PR #1386, opened 2025-07-15
Refactor permissioned-transact

tonsky:
Right now permissioned-transact does a lot of things in one place, which makes code harder to read, understand, modify. This PR refactors the code into small independent steps. Each step does one thing and works on the same simple abstraction: tx-step-maps

Changes:

- Removed `etype` checks for delete-entity. We confirmed with prod metrics that all delete-entities have etype now (#1394 #1398)
- Internal `tx-step` representation is replaced with a map (`[op eid aid value]` → `{:op :eid :aid :value}`). This allowed me to calculate and add `:etype` to steps once and use it in all subsequent operations. It’s a big change but I belive it’s worth it in the long run. Eventually it’ll be nice to migrate all code to tx-step-map and get rid of vector form altogether (it’s not client-facing anyway)
- A lot of code got simpler with tx-step-map. Logic like `if op = add-triple then eid else etype` was scattered all around transaction/permissioned-transaction
- `preload-triples` (renamed to `load-entities-map`) now operate on eid + etype (was eid + etype + action before)
- Most of permissioned-transaction is now a streamlined pipeline that operates on a list of `tx-step-maps`. Before there were several data structures, sometimes maps, in which keys held additional info (e.g. `group-object-tx-steps` was calculating etype and action and adding it to _keys_)
- Operating on single data structure (list of tx-step-maps) makes it easier to read and modify code. Need new check? Easy! Add new function that reads well-defined data structure and returns the same data structure. Pipelines!
- Understanding code is easier because most pipeline functions do only 1 thing. No more “transform steps + calculate etype + generate rules + validate uuid values”.
- Generating checks is now 1-step process (was 2-step before, object-checks + check-program)
- Untangled `rule-params-to-copy`
- Structured code reuse between transaction and permissioned-transaction a little bit better
- Include `:update-attr` and `:delete-attr` in optimistic-attr calculations
- Docstrings on most functions
- Overall I was able to shove off almost 200 lines. Should be a good sign!

Deployment plan:

- push flags config `cd server/flags-config && pnpm i && pnpm run push`

### DIFF
--- a/server/test/instant/db/transaction_test.clj
+++ b/server/test/instant/db/transaction_test.clj
@@ -61,6 +61,7 @@
                     (coll/update-in-when [:add-triple] set)
                     (coll/update-in-when [:retract-triple] set)
                     (coll/update-in-when [:delete-entity] set)
+                    (coll/update-in-when [:add-attr :attrs] set)
                     (coll/update-in-when [:add-attr :idents] set)))))))
 
 (defn- normalize-results


tonsky:
I’m sorry, I got carried away and when I realized what is going on it was too late to turn back. This is more of a rewrite than refactoring, which is hard to review. I’ll try not to do this in the future, but right now this refactoring exists and I believe is net beneficial to the codebase.

Anyways. This is what is going on here:

- Removed `etype` checks for delete-entity. We confirmed with prod metrics that all delete-entities have etype now (#1394 #1398)
- Internal `tx-step` representation is replaced with a map (`[op eid aid value]` → `{:op :eid :aid :value}`). This allowed me to calculate and add `:etype` to steps once and use it in all subsequent operations. It’s a big change but I belive it’s worth it in the long run. Eventually it’ll be nice to migrate all code to tx-step-map and get rid of vector form altogether (it’s not client-facing anyway)
- A lot of code got simpler with tx-step-map. Logic like `if op = add-triple then eid else etype` was scattered all around transaction/permissioned-transaction
- `preload-triples` (renamed to `load-entities-map`) now operate on eid + etype (was eid + etype + action before)
- Most of permissioned-transaction is now a streamlined pipeline that operates on a list of `tx-step-maps`. Before there were several data structures, sometimes maps, in which keys held additional info (e.g. `group-object-tx-steps` was calculating etype and action and adding it to _keys_)
- Operating on single data structure (list of tx-step-maps) makes it easier to read and modify code. Need new check? Easy! Add new function that reads well-defined data structure and returns the same data structure. Pipelines!
- Understanding code is easier because most pipeline functions do only 1 thing. No more “transform steps + calculate etype + generate rules + validate uuid values”.
- Generating checks is now 1-step process (was 2-step before, object-checks + check-program)
- Untangled `rule-params-to-copy`
- Structured code reuse between transaction and permissioned-transaction a little bit better
- Docstrings on most functions
- Overall I was able to shove off almost 200 lines. Should be a good sign!

Since this is such a huge change to such an important part of our system, deploying won’t be easy. I’ll try tomorrow to add old impl back and add a flag that let us switch back and forth just in case. But I want to get the ball rolling already, so please take a look!

tonsky:
Okay, I restored old permissioned-transact & added a flag to enable new impl. Should be safe to deploy now (will default to old impl by default) and then gradually enable e.g. per app

server/src/instant/db/cel.clj
```diff
@@ -454,102 +454,109 @@
...
-                        (str "Tried to evaluate a cel program that used unknown variables. "
-                             "Eval returned unknown, but there are no unknown refs.")))))
-
-            :else
-            result))
-    (catch CelEvaluationException e
-      (ex/throw-permission-evaluation-failed!
-       etype action e (:show-cel-errors? ctx)))))
+  (if (contains? program :result)
+    (:result program)
```
dwwoelfel, line 460: Why do we need this?

server/src/instant/flags.clj
```diff
@@ -99,6 +100,24 @@
                                 :default-value default-value
                                 :disabled? disabled?}))
 
+        new-permissioned-transact
+        (when-let [hz-flag (-> (get result "new-permissioned-transact")
```
dwwoelfel, line 104: Can you add this new entity to https://github.com/instantdb/instant/blob/main/server/flags-config/instant.schema.ts, then add a deployment plan in the PR body that includes pushing up the flag?

dwwoelfel:
> Removed etype checks for delete-entity. We confirmed with prod metrics that all delete-entities have etype now (https://github.com/instantdb/instant/pull/1394 https://github.com/instantdb/instant/pull/1398)

How integral to this PR is removing the etype check? I'm seeing logs from apps that are sending transactions without the etype:   [honeycomb logs](https://ui.honeycomb.io/instantdb/environments/prod/datasets/instant-server/result/qGTC8wYfRu5?vs=hideCompare) [cloudwatch logs](https://us-east-1.console.aws.amazon.com/cloudwatch/home?region=us-east-1#logsV2:logs-insights$3FqueryDetail$3D~(end~0~start~-604800~timeType~'RELATIVE~tz~'UTC~unit~'seconds~editorString~'fields*20*40timestamp*2c*20*40message*2c*20*40logStream*2c*20*40log*0a*7c*20filter*20*40message*20like*20*27missing-etype*27*0a*7c*20sort*20*40timestamp*20desc*0a*7c*20limit*2010000~queryId~'c5a5e79e-6acf-4acd-807f-74580ae72a36~source~(~'*2faws*2felasticbeanstalk*2fInstant-docker-prod-env-2*2fvar*2flog*2feb-docker*2fcontainers*2feb-current-app*2fstdouterr.log)~lang~'CWLI))

tonsky:
So there’s only one guy (app_id d7dd3078-9350-4e3e-8c73-d66416b6ca46, app Notes, created 2024-08-29) who is genuinely using delete-entity without etype. I guess we will have to restore that path (but it’s a simple pre-processing step)

The rest are add-triple with unknown attrs — corresponding transaction fails anyways, but the step is reported anyway as “missing etype”

tonsky:
@dwwoelfel restored delete without etype support

server/src/instant/db/permissioned_transaction_new.clj
```diff
@@ -0,0 +1,408 @@
...
+   5. Transact tx-steps
+   6. Fetch new lookup-ref data for newly inserted data
+   7. Run create checks
+
+   We run `create` checks _after_ the transaction, so we can query off of the
+   object. For example, if we created a new `post`, we may want a check that says:
+   `auth.id in data.ref('creator.id')`"
+  [{:keys [db app-id admin? admin-check? admin-dry-run? attrs] :as ctx} tx-step-vecs]
+  (tracer/with-span! {:name "permissioned-transaction/transact!"
+                      :attributes {:app-id app-id}}
```
dwwoelfel: Can you add an attr like `new-transact: true` here so that we can track it in honeycomb?

server/src/instant/db/permissioned_transaction_new.clj
```diff
@@ -0,0 +1,408 @@
...
+            ctx                  (assoc ctx
+                                        :db {:conn-pool tx-conn}
+                                        :attrs optimistic-attrs)]
+        (validate-reserved-names! ctx tx-step-maps)
+        (if admin?
+          (tx/transact-without-tx-conn-impl! tx-conn optimistic-attrs app-id tx-step-maps {})
+          (let [;; pre-processing tx
+                tx-step-maps         (->> tx-step-maps
+                                          (coerce-value-uuids ctx)
+                                          (validate-lookup-etypes ctx))
```
dwwoelfel: We never update the etypes in tx-step-maps with optimistic attrs, so this will fail if we're updating a newly created attr with a lookup.

server/src/instant/db/transaction.clj
```diff
@@ -264,87 +317,75 @@
...
+       [{:message "update or merge is only allowed on `path` for $files in transact."}]))))
+
+(defn resolve-lookups-for-delete-entity [conn app-id tx-step-maps]
+  (let [[lookup-ref-deletes rest] (coll/split-by
+                                   #(and (= :delete-entity (:op %))
+                                         (triple-model/eid-lookup-ref? (:eid %)))
+                                   tx-step-maps)
+        lookup-refs               (map :eid lookup-ref-deletes)
+        resolved                  (resolve-lookups conn app-id lookup-refs)]
+    (concat
```
dwwoelfel, line 365: This changes the order of the tx-steps, which will produce different outcomes in the result. For example:

```
db.transact([
  db.tx.items["my-id"].delete({}), 
  db.tx.items["my-id"].update({a: 1}})
])
```

If you move the deletes to the end, then there will be no item in the db at the end.

server/src/instant/db/cel.clj
tonsky: Because sometimes we already know the result and can skip the program, e.g. here:

https://github.com/instantdb/instant/blob/71693cf121ff4971a68245322c475448d99b2b36/server/src/instant/db/permissioned_transaction_new.clj#L194

It was like this before, but the filtering was rather complicated on the permssioned_transaction side:

https://github.com/instantdb/instant/blob/f61b80a33a662abcc955be94f3155111cd2df957/server/src/instant/db/permissioned_transaction.clj#L354-L368

This way it’s much cleaner, and also :result key is not arbitrary: the output of eval-programs is list of the same programs with :result set on them

server/src/instant/db/transaction.clj
tonsky: I don’t think we ever maintained the order of tx-steps. In current implementation, we group them all in the map, then concat all the map values, so the order is already messed up. Do you think we should start guaranteeing it?

tonsky: No, I don’t think we can, actually. There’s a reason we group e.g. all deletes together—so that they all executed as one query. If we start respecting the order, we’ll need to run one query per consecutive group of similar operations. I don’t think it’s worth it

server/src/instant/db/permissioned_transaction_new.clj
tonsky: Right! Mapify should be done on optimistic attrs, probably. Also will need a test for this case

server/src/instant/db/transaction.clj
dwwoelfel: I saw different behavior in the deletes when running this PR locally vs. production. I assumed it was this, but maybe it's something else.

tonsky: Wow wow wow, what do you mean? Maybe we should look into this? What was the difference? Any details?

server/src/instant/db/permissioned_transaction_new.clj
tonsky: Added `transact-version = 2`

tonsky:
@dwwoelfel I think I addressed all notes, please take a look:

- f3c75247550850c7d5627a363ed348229482ae30
- bfff8fe510e747ec00d590a48a34aabc5c792e61

server/src/instant/db/transaction.clj
dwwoelfel: In production, if I do:

```
db.transact([
  db.tx.items["my-id"].delete({}), 
  db.tx.items["my-id"].update({a: 1}})
])
```

Then at the end of the transaction, there is an item in the database.

When I tested this PR locally, there was no item in the database at the end.

dwwoelfel: I added a test for it here: https://github.com/instantdb/instant/pull/1459

tonsky: I can add behavior that mimics group-by logic (whatever key appeared first that group will be executed first). But I really think this is out of scope to introduce full step-by-step transactions (and I’m not even sure we want it)

tonsky: Done [740d355](https://github.com/instantdb/instant/pull/1386/commits/740d355782766395a315e1193e59686f7f6f018d). Also incorporated your test in this PR

server/src/instant/db/transaction.clj
```diff
@@ -518,18 +591,48 @@
...
+   in order they first appeared in tx-steps"
+  [tx-step-vecs]
+  (distinct (map first tx-step-vecs)))
+
+(defn reorder-tx-steps
+  [ops-order tx-step-maps]
+  (for [op ops-order
+        tx-step-map tx-step-maps
+        :when (= op (:op tx-step-map))]
+    tx-step-map))
```
dwwoelfel:
```suggestion
  (let [groups (group-by :op tx-step-maps)]
    (mapcat #(get groups %) ops-order))
```

server/src/instant/db/transaction.clj
```diff
@@ -518,18 +591,48 @@
...
+
+(defn optimistic-attrs [attrs tx-step-vecs]
+  (->> tx-step-vecs
+       (reduce
+        (fn [acc [op value]]
+          (case op
+            :add-attr    (assoc! acc (:id value) value)
+            :update-attr (coll/update! acc (:id value) merge value)
```
dwwoelfel: Probably safer to just replace the attr instead of merging it, because it looks like that's what we do when we update the attr in the database.

Also might be worth calling out in the PR description that we'll now include attr deletes and updates in optimistic attrs (previously was just add-attr). This change seems more correct, but it would be good to call it out in case there are some unexpected effects.

server/test/instant/db/transaction_test.clj
```diff
@@ -1323,7 +1388,7 @@
          (aurora/conn-pool :write)
          (attr-model/get-by-app-id app-id)
          app-id
-         [[:delete-entity billy-eid]])
+         [[:delete-entity billy-eid "users"]])
```
dwwoelfel, line 1486: Can we revert the changes that add etype to the tests for :delete-entity?

server/src/instant/db/permissioned_transaction_new.clj
```diff
@@ -0,0 +1,410 @@
...
+                                           different from the namespace of the attribute that is
+                                           being set")}])))
+  tx-steps)
+
+(defn load-entities-map
+  "Takes tx-steps, fetches from DB map of
+
+     {{:eid :etype} -> <entity map>}"
+  [{:keys [datalog-query-fn attrs] :as ctx} tx-steps]
+  (let [eids+etypes (for [{:keys [eid etype value rev-etype]} tx-steps
```
dwwoelfel, line 81: We can skip running the datalog query if eids+etypes is empty.

server/src/instant/db/permissioned_transaction_new.clj
```diff
@@ -0,0 +1,410 @@
...
+         (-> (entity-model/triples->map ctx triples)
+             (assoc "id" (ffirst triples))))))))
+
+(defn resolve-lookups-tx-steps
+  "Given known entities-map, resolves as much lookup-refs as possible"
+  [_ctx entities-map tx-steps]
+  (for [{:keys [eid etype value rev-etype] :as tx-step} tx-steps]
+    (cond-> tx-step
+      (lookup-ref? eid)
+      (update :eid #(-> entities-map (get {:eid % :etype etype}) (get "id") (some-> uuid-util/coerce) (or %)))
```
dwwoelfel: Can we get some newlines in between the statements? It's hard to see where each function in the thread macro starts.

server/src/instant/db/permissioned_transaction_new.clj
```diff
@@ -0,0 +1,410 @@
...
+                tx-step-maps         (resolve-lookups-tx-steps ctx entities-map tx-step-maps)
+                entities-map         (resolve-lookups-entities-map ctx entities-map)
+                updated-entities-map (update-entities-map ctx entities-map tx-step-maps)
+                rule-params-map      (into {}
+                                           (for [{:keys [op eid etype value]} tx-step-maps
+                                                 :when (= :rule-params op)]
+                                             [{:eid eid :etype etype} value]))
+
+                ;; pre checks
+                pre-check-results    (run-checks! ctx (pre-checks ctx entities-map updated-entities-map rule-params-map tx-step-maps))
```
dwwoelfel: Can you add back the io/warn-io when running the checks? The idea is that we should only be touching the database when we preload data in batches.

An improvement on how we were doing it would be to wrap the whole `transact!` in `io/warn-io` and then add an `io/allow-io` around the places where we expect to do io.

server/src/instant/db/permissioned_transaction_new.clj
```diff
@@ -0,0 +1,410 @@
...
+       (do
+         (when-not (:admin-check? ctx)
+           (ex/assert-permitted! :perms-pass? [etype scope] result))
+         (-> check
+             (dissoc :result)
+             (assoc
+              :check-result result
+              :check-pass?  (boolean result))))))))
+
+(defn lock-tx-on! [tx-conn big-int]
```
dwwoelfel: We can just remove this since it's not used any more.

server/src/instant/db/permissioned_transaction_new.clj
```diff
@@ -0,0 +1,410 @@
...
+                       :let [{:keys [eid]} check]
+                       :when (lookup-ref? eid)]
+                 (ex/throw-validation-err! :lookup eid [{:message "Could not find the entity for this lookup"}]))
+        results (cel/eval-programs! ctx checks)]
+    (doall
+     (for [{:keys [scope etype result] :as check} results]
+       (do
+         (when-not (:admin-check? ctx)
+           (ex/assert-permitted! :perms-pass? [etype scope] result))
+         (-> check
```
dwwoelfel, line 332: Before we had a `data` key with `original` and `updated` in the result. Now we return `:bindings` with `:data` and `:new-data`. The sandbox on the dashboard won't display the results properly if it doesn't get the data in that shape.

server/test/instant/db/transaction_test.clj
```diff
@@ -26,6 +26,14 @@
   (:import
    (java.util UUID)))
 
+(test/use-fixtures :each
```
dwwoelfel, line 104: Can we add a with-redefs that captures the outputs of `transact!` from each run and then fail the test if there are any differences? Might have to write something to replace uuids and tx-id, but everything else should be the same?

The data gets sent to the client when they run queries from the dashboard's sandbox, so the output format is kind of important.

server/test/instant/db/transaction_test.clj
tonsky: I don’t think we should. There’s now just 1 guy who does like 1 transaction/week that uses old delete. I think it’s better if we test more for majority of users. There’s a test for delete without etype, I added one more, I think the rest should test our main use-case, not some old weird case

server/src/instant/db/permissioned_transaction_new.clj
tonsky: Fixed the sandbox. New format uses the same names that actual CEL programs use (does not invent new concepts), so I’d like to keep it

server/src/instant/data/bootstrap.clj
```diff
@@ -27,7 +25,7 @@
                                    (not v)))))
         eid->uuid (->> triples
                        (map (fn [[id _ _]]
-                              [id (java.util.UUID/randomUUID)]))
+                              [id (random-uuid)]))
```
dwwoelfel, line 28: ❤️

PR #1387, opened 2025-07-15
Single query for create-app

dwwoelfel:
This makes creating an app a little bit faster because we don't have to do two round trips to the database. 

Inserts both the app and the admin token with a CTE instead of two queries inside of a transaction.


nezaj (approved):
Nice!

PR #1389, opened 2025-07-15
fix core recursion error

drew-harris:
fixes an error where deleting an object with a link to an entity with the same id recurses infinitely


stopachka (approved):
Would test, but LGTM! Let's ship

PR #1390, opened 2025-07-15
Fix inflection on schema operations

dwwoelfel:
Fixes a bug where it used the wrong inflection, now it will say "Finished 1 schema operation" or "Finished 10 schema operations".


stopachka (approved):
Woohoo!

PR #1393, opened 2025-07-16
Fix cors

dwwoelfel:
We weren't varying the response based on the headers we return in `Access-Control-Allow-Headers`, which was causing signup to fail. Now we add `Access-Control-Request-Headers` to the `vary` header, which will force the cdn to send the request back to the origin when it needs new headers.

Also puts the max-age in a new `flags` field in `instant-config`.

### DIFF
--- a/server/flags-config/instant.schema.ts
+++ b/server/flags-config/instant.schema.ts
@@ -3,13 +3,13 @@
 import { i } from "@instantdb/core";
 
 const _schema = i.schema({
-  // We inferred 7 attributes!
+  // We inferred 8 attributes!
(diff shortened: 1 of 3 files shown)


server/flags-config/instant.schema.ts
```diff
@@ -74,10 +84,6 @@ const _schema = i.schema({
       "dualWrite?": i.boolean().optional(),
       "useLocationId?": i.boolean(),
     }),
-    "log-sampled-apps": i.entity({
```
dwwoelfel, line 77: This was included twice

stopachka (approved):
Beautiful!

PR #1394, opened 2025-07-16
Trace tx-steps without etype

tonsky:
The code of permissioned-transact is filled with checks for nil etype. My suspicion is that it was only relevant for delete in the past, but since: a) clients have updated, and b) we have a transact step that fills delete etype for us in advance.

So I want to get rid of the checks and greatly simplify the code, but it’s scary. That’s why I want to be double sure I’m not missing something. Thus the tracing


server/src/instant/db/permissioned_transaction.clj
```diff
@@ -268,6 +268,9 @@
                   etype (if (sequential? eid)
                           (extract-lookup-etype! ctx eid aid-etype tx-step)
                           aid-etype)
+                  _ (when (nil? etype)
```
dwwoelfel, line 272: You probably want to add a whole new top-level span here to make it less likely that your trace is going to get sampled.

PR #1399, opened 2025-07-17
Make i.date() return dates

drew-harris:
In the init function of the client libraries, you can pass `useDateObjects` to make any i.date() column return date objects instead of a `string | number`. You are still able to pass in `string | number | Date` for update functions.


client/packages/core/src/schemaTypes.ts
```diff
@@ -277,19 +277,28 @@ type OptionalKeys<Attrs extends AttrsDefs> = {
  *   - Required keys => `key: ValueType`
  *   - Optional keys => `key?: ValueType`
  */
-type MappedAttrs<Attrs extends AttrsDefs> = {
+type MappedAttrs<Attrs extends AttrsDefs, UseDates extends boolean> = {
   [K in RequiredKeys<Attrs>]: Attrs[K] extends DataAttrDef<infer V, any, any>
-    ? V
+    ? V extends Date
```
dwwoelfel, line 283: What will happen here if the user specified Date in their asType, but they don't pass `useDateObjects` e.g.:

```js
i.schema({
  someEntity: i.entity({
    date: i.date()
  }).asType({date: Date})
})
```

dwwoelfel (commented):
Can you add some tests that show that transact will store dates in the store when `useDateObjects` is true and will store whatever it was provided if `useDateObjects` is false?

It would also be nice to have a playground in `sandbox/react-nextjs/` that lets you toggle `useDateObjects` and insert new objects with dates. The missing-attrs playground is a good template to follow: http://localhost:4000/play/missing-attrs

client/packages/core/src/schemaTypes.ts
drew-harris: They will just get a string or number but typed as a date. What would the expected behavior be here?

client/packages/core/__tests__/src/store.test.js
```diff
@@ -610,3 +610,73 @@ test('recursive links w same id', () => {
...
+          title: i.string().optional(),
+        }),
+      },
+    });
+    const sameId = id();
+    const ops = [
+      instatx.tx.todos[sameId].update({
+        title: 'todo',
+        completed: false,
+        createdAt: new Date(),
```
dwwoelfel, line 630: Since you can insert a date, string, or number, can we test that each of those produces a date object?

client/packages/core/__tests__/src/store.test.js
```diff
@@ -610,3 +610,73 @@ test('recursive links w same id', () => {
...
+      }),
+    ];
+
+    const steps = instaml.transform({ attrs: {}, schema }, ops);
+    const store = createStore({}, [], true, createLinkIndex(schema), true);
```
dwwoelfel, line 635: A comment that we're testing `useDateObjects=true` would be handy here, especially in the future when we forget what we're testing. It's not very easy to see from the argument to `createStore`.

client/packages/core/src/store.js
```diff
@@ -47,18 +47,44 @@ function setInMap(m, path, value) {
   setInMap(nextM, tail, value);
 }
 
-function createTripleIndexes(attrs, triples) {
+function isDateAttr(attr) {
+  return attr['checked-data-type'] === 'date';
+}
+
+function coerceToDate(value) {
```
dwwoelfel: There are some date formats that we support that will return an invalid date from `new Date(value)`: https://github.com/instantdb/instant/blob/main/server/test/instant/db/model/triple_test.clj#L11

The formats that fail are:

- `2025-01-02T00:00:00-08`
- `"2025-01-02T00:00:00-08"`
- `"2025-01-15 20:53:08"`

I see two possible solutions:
1. Write a custom date resolver in js that can convert those formats
2. Send the flag to the backend and have it return dates as something that `Date` can parse (we would use the postgres function we've defined called `triples_extract_date_value`)

I kind of prefer option 2, because then you're getting the same value we stored in the index. Anything that `transact` accepts would work for `query`. If we only do option 1, then the client will be broken until we can push an update to accept the new format.

But we would probably also need to implement option 1 so that flipping `useDateObjects` from `true` to `false` won't break when the data in the store used the opposite flag.

client/packages/core/src/store.js
```diff
@@ -257,6 +287,14 @@ function addTriple(store, rawTriple) {
     return;
   }
 
+  if (
+    attr['checked-data-type'] &&
```
dwwoelfel:
```suggestion
```

client/sandbox/react-nextjs/pages/play/date-conversion.tsx
```diff
@@ -0,0 +1,93 @@
...
+      title: i.string().unique().indexed(),
+      completed: i.boolean(),
+      createdAt: i.date(),
+    }),
+  },
+});
+
+function Example({ appId }: { appId: string }) {
+  const myConfig = { ...config, appId };
+  const db = init({ ...myConfig, schema, useDateObjects: true });
```
dwwoelfel: In http://localhost:4000/play/missing-attrs, there's a query param you can add to change whether we pass the schema. Could you add a query param to make it easy to switch back and forth from `useDateObjects=true` to `useDateObjects=false`?

drew-harris:
Playground URL:
http://localhost:4000/play/date-conversion?useDateObjects=true

client/packages/core/src/utils/dates.ts
```diff
@@ -0,0 +1,70 @@
+export function coerceToDate(value: unknown): Date {
```
dwwoelfel: We have a bunch of date parsers in [triple.clj](https://github.com/instantdb/instant/blob/main/server/src/instant/db/model/triple.clj#L930-L1008). If we could match the same structure here, then it would be easier for us to handle new date formats in the future.

It could also use a test on the same dates we test in [triples_test](https://github.com/instantdb/instant/blob/main/server/test/instant/db/model/triple_test.clj#L11)

client/packages/core/src/utils/dates.ts
```diff
@@ -0,0 +1,142 @@
+// Main parse function that handles strings and numbers
+export function coerceToDate(x: unknown): Date {
```
dwwoelfel: We generally follow the no-use-before-define rule on the javascript side since it's required on the clojure side.

It's not enforced by a lint rule or anything, but it would be nice if we all followed that convention since it's already been established.

client/packages/core/src/utils/dates.ts
```diff
@@ -0,0 +1,142 @@
...
+  if (typeof x === 'string') {
+    const result = dateStrToInstant(x) || jsonStrToInstant(x);
+    if (!result) {
+      throw new Error(`Unable to parse date string ${x}`);
+    }
+    return result;
+  } else if (typeof x === 'number') {
+    return new Date(x);
+  }
+  throw new Error(`Invalid date value type: ${typeof x}`);
```
dwwoelfel: Can you include the date value in here also for easier debugging?

client/packages/core/__tests__/src/utils/dates.test.ts
```diff
@@ -0,0 +1,91 @@
...
+            expect(result.toISOString()).toBe('2025-01-15T20:53:08.000Z');
+            break;
+          case 'Wed Jul 09 2025':
+            expect(result.getUTCFullYear()).toBe(2025);
+            expect(result.getUTCMonth()).toBe(6); // July is month 6 (0-indexed)
+            expect(result.getUTCDate()).toBe(9);
+            expect(result.getUTCHours()).toBe(0);
+            expect(result.getUTCMinutes()).toBe(0);
+            expect(result.getUTCSeconds()).toBe(0);
+            break;
```
dwwoelfel, line 55: Can you add a default case that throws

client/sandbox/react-nextjs/pages/play/date-conversion.tsx
```diff
@@ -0,0 +1,141 @@
...
+const getRandomTodo = () =>
+  exampleTodos[Math.floor(Math.random() * exampleTodos.length)];
+
+const dateTypes = {
+  number: new Date().getTime(),
+  string: new Date().toISOString(),
+  jsonString: new Date().toJSON(),
+  invalidJsonString: JSON.stringify(new Date()),
+  invalidString: 'invalid',
+  backendCanParseOnly: '2025-01-02T00:00:00-08',
```
dwwoelfel: What does it mean to be `backendCanParseOnly`? Isn't this an example of something the client can coerce?

drew-harris: Previously this was because only the backend was capable of parsing the string, i've updated it to "requiresSpecialParsing" now that we can handle it on the frontend too.

client/packages/core/src/package.json
```diff
@@ -0,0 +1,3 @@
+{
```
dwwoelfel: What is this for?

client/packages/admin/src/index.ts
```diff
@@ -696,16 +711,19 @@ type AdminQueryOpts = {
...
-  config: InstantConfigFilled<Schema>;
+class InstantAdminDatabase<
+  Schema extends InstantSchemaDef<any, any, any>,
+  Config extends InstantConfig<Schema, boolean> = InstantConfig<Schema, false>,
+> {
+  config: InstantConfigFilled<Schema, Config['useDateObjects']>;
   auth: Auth;
   storage: Storage;
-  rooms: Rooms<Schema>;
+  rooms: Rooms<Config['schema']>;
```
dwwoelfel: Why do we pass Config into `Rooms`? Doesn't it take a Schema?

dwwoelfel: Oh nvm, I see it's taking the schema from config

client/packages/core/src/utils/dates.ts
```diff
@@ -0,0 +1,142 @@
...
+
+// Main parse function that handles strings and numbers
+export function coerceToDate(x: unknown): Date {
+  if (x instanceof Date) {
+    return x;
+  }
+  if (typeof x === 'string') {
+    const result = dateStrToInstant(x) || jsonStrToInstant(x);
+    if (!result) {
+      throw new Error(`Unable to parse date string ${x}`);
```
dwwoelfel:
```suggestion
      throw new Error(`Unable to parse \`${s}\` as a date.`);
```

client/packages/core/src/utils/dates.ts
```diff
@@ -0,0 +1,142 @@
...
+    }
+    return result;
+  } else if (typeof x === 'number') {
+    return new Date(x);
+  }
+  throw new Error(`Invalid date value type: ${typeof x}, x`);
```
dwwoelfel:
```suggestion
  throw new Error(`Invalid date value \`${x}\`. Expected a date, number, or string, got type ${typeof x}.`);
```

PR #1400, opened 2025-07-17
Make it possible to override config

stopachka:
Added a flag so we can set custom handle receive overrides for specific apps. 

### DIFF
--- a/server/src/instant/reactive/session.clj
+++ b/server/src/instant/reactive/session.clj
@@ -561,7 +561,8 @@
                              :in-progress-stmts in-progress-stmts
                              :silence-exceptions silence-exceptions}]
         (swap! pending-handlers conj pending-handler)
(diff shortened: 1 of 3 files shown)


server/src/instant/reactive/session.clj
```diff
@@ -542,6 +542,10 @@
       (let [pending-handlers (:pending-handlers (:session/socket session))
             in-progress-stmts (sql/make-statement-tracker)
             debug-info (atom nil)
+            app-id (-> session :session/auth :app :id)
+            timeout-ms (or (when app-id
```
dwwoelfel, line 546: Could we also add the timeout-ms to the `:attributes` for the `handle-receive` span? Might be useful for debugging later.

dwwoelfel (approved):
LGTM!

PR #1402, opened 2025-07-17
[Sandbox] Prevent double running

nezaj:
A user reported an edge case where we could accidentally double-run in the sandbox. This makes it so we can only run one exec at a time


client/www/components/dash/Sandbox.tsx
```diff
@@ -86,6 +87,9 @@ export function Sandbox({
   }, [output]);
 
   const exec = async () => {
+    if (isExecuting) return;
```
stopachka, line 90: I would add this in the `disabled` option on the button too.

nezaj: I was thinking it's nicer to wait a little before changing the UI so there isn't a flicker

PR #1404, opened 2025-07-18
Add created date to feedback app

nezaj:
Main change is adding `createdAt` to the feedback app so we can show when someone gave us feedback about our docs.

CleanShot 2025-07-17 at 17 33 58@2x

Also changed the following:

* Added deep-linking so when you refresh you don't go back to the index
* Updated docs / removed the makefile command for updating the feedback app

I had trouble with the Makefile command not being able to pick up the schema/perms, it always wanted to create a new app. Figured instead of having those commands would be nicer to just be explicit on how to update the app


dwwoelfel (approved):
LGTM, but you might want to check with @stopachka if removing the Makefile commands breaks his workflow

PR #1406, opened 2025-07-18
Faster add-attr when not indexed

dwwoelfel:
Makes add-attr fast when you're adding a new non-indexed attr on a large table.

If the attr is indexed, we have to add nulls to every entity in the triples table. Postgres was getting confused and trying to check which entities needed nulls even when you added an unindexed attr. 

Removing the joins to the idents table allowed postgres to build a better query plan that skipped the whole check when you weren't adding an indexed attr.

Adding indexed attrs to an existing table is still slow. I'll follow-up with a PR that does that in two steps: 1. adds the attr without indexing it, 2. creates a background job to index it.

### DIFF
--- a/server/src/instant/db/model/attr.clj
+++ b/server/src/instant/db/model/attr.clj
@@ -406,19 +406,11 @@
                            :where [:and
                                    [:= :attr-inserts.value-type [:inline "blob"]]
                                    :attr-inserts.is-indexed]
-                           :join [[:attr-idents :ident]
-                                  [:= :attr-inserts.forward-ident :ident.id]
-
-                                  [:idents :id-ident]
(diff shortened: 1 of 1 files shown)


server/src/instant/db/model/attr.clj
```diff
@@ -429,7 +421,6 @@
                                    [:not [:exists {:select :1
                                                    :from :triples
                                                    :where [:and
-                                                           :triples.ave
```
dwwoelfel, line 432: The `triples.ave` was no help here, it just added extra work for the postgres. It would have to look at the row instead of doing an index-only scan on the triples_pkey (which has app-id, attr-id, and entity-id)

PR #1407, opened 2025-07-18
Fix standalone packaging for core

nezaj:
@instantdb/core's standalone files were sometimes missing from unpkg. Seemed like an intermittement issue

I think what was happening was when publishing, npm would sometimes pack the package before vite finished writing the standalone files to disk.

Talking it through with claude it suggested it only affected the core package because it publishes immediately after building, while react and platform packages have a natural delay (they build after core, giving core's files time to fully write).

The fix here adds a `prepublishOnly` script to ensure the standalone build completes right before npm creates the package:

This should ensure the dist/standalone directory exists when npm packs the files

I was able to get this working 3x in a row w/ `pnpm run dev` running as well

* [v.20.9-experimental.j3](https://www.unpkg.com/@instantdb/core@0.20.9-experimental.j3/dist/standalone/index.umd.cjs), 
* [v.20.9-experimental.j4](https://www.unpkg.com/@instantdb/core@0.20.9-experimental.j4/dist/standalone/index.umd.cjs)
* [v.20.9-experimental.j5](https://www.unpkg.com/@instantdb/core@0.20.9-experimental.j5/dist/standalone/index.umd.cjs)

### DIFF
--- a/client/packages/core/package.json
+++ b/client/packages/core/package.json
@@ -41,6 +41,7 @@
     "build": "rm -rf dist; npm run build:tshy && npm run build:standalone && npm run check-exports",
     "build:tshy": "tshy",
     "build:standalone": "vite build",
+    "prepublishOnly": "npm run build:standalone",
     "publish-package": "pnpm publish --access public --no-git-checks"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


nezaj:
Closing in favor of https://github.com/instantdb/instant/pull/1410

PR #1408, opened 2025-07-18
[Docs] Add pattern for local vs prod app

nezaj:
What it says on the tin!

CleanShot 2025-07-18 at 14 38 22@2x

### DIFF
--- a/client/www/pages/docs/patterns.md
+++ b/client/www/pages/docs/patterns.md
@@ -401,8 +401,8 @@ import { init } from '@instantdb/react';
 // Next.js example
 const APP_ID =
   process.env.NODE_ENV === 'production'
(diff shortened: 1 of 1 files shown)


client/www/pages/docs/patterns.md
```diff
@@ -387,6 +387,55 @@ function Login() {
...
+for production. You can then use an environment variables to switch between the two
+apps.
+
+```javascript
+// lib/db.ts
+import { init } from '@instantdb/react';
+
+// Next.js example
+const APP_ID =
+  process.env.NODE_ENV === 'production'
```
dwwoelfel, line 403: I think the more common thing for people to do is to just override `NEXT_PUBLIC_INSTANT_APP_ID` in .env.local

That way you don't have to have separate env var names for the app id.

PR #1409, opened 2025-07-18
Use the eav index when it's better

dwwoelfel:
Sometimes it's better to use the eav index, for example in the zeneca query:

### Examples
```clj
{:books {:$ {:where {:bookshelves.users.handle "stopa"}}
```

We send patterns to datalog that look like:

```clj
[[:av ?users-2 :user-aid "stopa"]
 [:vae ?users-2 :bookshelves-aid ?bookshelves-1]
 [:vae ?bookshelves-1 :books-aid ?books-0]]
```

Since we've resolved `?users-2` as an entity-id in the first pattern and it's in the e position in the second pattern, using the eav index will hit the `e` and the `a`, which will usually make a good plan for postgres.

On the other hand, if we went the other way through the links:

```clj
{:users {:$ {:where {:bookshelves.books.title "The Count of Monte Cristo"}}}}
```

We'll create patterns:

```clj
[[:ave ?books-2 :title-eid "The Count of Monte Cristo"]
 [:vae ?bookshelves-1 :book-eid ?books-2]
 [:vae ?users-0 :bookshelf-eid ?bookshelves-1]]
 ```
 
Since we've resolved `?books-2` as an entity-id in the first pattern and it's in the v position in the second pattern, using the vae index will hit the v and the a.

### Explanation

You may be wondering why we put `vae` as the index in the first example, when it should obviously be `eav`. That's because the patterns were reversed in `optimize-attr-pats` after we chose the index to use. It was the right index when we chose it.

That's why we rewrite the `named-pattern` in the datalog.clj. It's just the easiest place to figure out what we're joininig to. In the future, attr-pat shouldn't have any say in the index we use, it should just give us the properties of the attr. Then we can decide which index in datalog.clj based on the full query.

### DIFF
--- a/server/src/instant/db/datalog.clj
+++ b/server/src/instant/db/datalog.clj
@@ -622,6 +622,7 @@
                     ;; Make sure av uses the av_index
                     :av [:json_null_to_null :value]
                     ;; Make sure vae uses the vae_uuid_index
+                    ;; and eav uses the eav_uuid_index
                     (:eav :vae) [:json_uuid_to_uuid :value]
 
                     :value)
--- a/server/src/instant/db/datalog.clj
+++ b/server/src/instant/db/datalog.clj
@@ -1034,10 +1034,12 @@
 ;; ---
 ;; match-query
 
-(defn flatten-symbol-map-values [values]
-  (if (set? values)
-    (mapcat flatten-symbol-map-values values)
-    values))
+(defn flatten-symbol-map-values [vs]
+  (mapcat (fn [v]
+            (if (set? v)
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: server/src/instant/db/datalog.clj [hunk 1]
```diff
@@ -622,10 +622,11 @@
                     ;; Make sure av uses the av_index
                     :av [:json_null_to_null :value]
                     ;; Make sure vae uses the vae_uuid_index
```
nezaj, line 624: Nit. Should we update this comment?

stopachka (approved):
Heck yeah @dwwoelfel!

PR #1410, opened 2025-07-21
Publish packages on ci

dwwoelfel:
Publishes our packages to npm through github actions.

On each push, it will check to see if the packages have changed. If they have, then it will publish an experimental version of the package to npm. Here's an example from this branch: https://www.npmjs.com/package/@instantdb/react/v/0.20.9-experimental.publish-on-ci.16426426678

When you're ready to publish the package for real:

1. Create a new branch and update the version.md with your new version.
2. Open a PR with that branch.
3. Wait for the GitHub actions to publish the packages and add a commit to the branch
4. Merge your branch.

If you forget how it works, just call `make publish` and it will print those instructions. If you really need to publish manually, call `make publish-ci`.


client/turbo.json
```diff
@@ -1,5 +1,58 @@
 {
   "$schema": "https://turborepo.org/schema.json",
+  "globalPassThroughEnv": [
```
dwwoelfel, line 3: The first two are added to allow us to publish from turbo (we pass the auth token as an env var, so we need to tell turbo to let those through).

The rest of the env vars are needed to enable provenance, which lets npm users know how we built the package. It gives us a nice stamp next to the version number and a provenance section at the bottom that shows how it was built: https://www.npmjs.com/package/@instantdb/react/v/0.20.9-experimental.publish-on-ci.16426426678#provenance

Some of the env vars are probably not necessary, but I couldn't find a list of which are needed and which aren't and it's a hassle to try each one.

client/Makefile
```diff
@@ -20,6 +20,14 @@ index-docs-manual:
 	pnpm exec tsx ./www/scripts/index-docs.ts
 
 publish:
+	@echo "There is a new way to publish packages!"
```
nezaj, line 23: Nice touch!

.github/workflows/js.yml
```diff
@@ -136,12 +137,36 @@ jobs:
...
+          fi
+        done
+    - name: Detect version.md change
+      id: detect-version-change
+      run: |
+        changeArr=($(echo ${{ steps.changed-files.outputs.changed_files }}))
+        echo "version_did_change=False" >> $GITHUB_OUTPUT
+        for file in ${changeArr[*]}
+        do
+          if [[ $file =~ version.md ]]; then
```
nezaj: Nit. Should we do `$file == version.md` to match exactly?

.github/workflows/js.yml
```diff
@@ -188,6 +213,138 @@ jobs:
...
+          sudo mv bb /usr/local/bin/
+          bb --version
+
+      - name: Set branch name
+        run: |
+          if [ "${{ github.event_name }}" = "pull_request" ]; then
+            echo "BRANCH_NAME=${{ github.head_ref }}" >> $GITHUB_ENV
+          else
+            echo "BRANCH_NAME=${{ github.ref_name }}" >> $GITHUB_ENV
+          fi
```
nezaj: I asked Claude about this and said there is potential someone creates a valid branch name (e.g. `feature/add-auth`) that would be invalid for npm

```suggestion
      - name: Set branch name
        run: |
          if [ "${{ github.event_name }}" = "pull_request" ]; then
            BRANCH_NAME="${{ github.head_ref }}"
          else
            BRANCH_NAME="${{ github.ref_name }}"
          fi
          # Sanitize branch name for npm version compatibility
          # Replace / and other invalid characters with -
          SAFE_BRANCH_NAME=$(echo "$BRANCH_NAME" | sed 's/[^a-zA-Z0-9-]/-/g')
          echo "BRANCH_NAME=$SAFE_BRANCH_NAME" >> $GITHUB_ENV
```

nezaj (approved):
LGTM w/ minor nits!

.github/workflows/js.yml
dwwoelfel: Good catch!

.github/workflows/js.yml
dwwoelfel: Updated to check for an exact match.

PR #1411, opened 2025-07-21
[Metrics] Add signups to overview

nezaj:
Feels like we've been getting a burst of signups lately and wanted to add some metrics for us to see that.

### DIFF
--- a/server/src/instant/intern/metrics.clj
+++ b/server/src/instant/intern/metrics.clj
@@ -212,7 +212,8 @@
 (defn rolling-avg-signups
   "Get rolling 7-day average signup counts for the last n weeks"
   ([conn weeks]
-   (let [end-date (LocalDate/now)
+   (let [;; Exclude today since it has incomplete data
+         end-date (.minusDays (LocalDate/now) 1)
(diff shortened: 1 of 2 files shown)


server/src/instant/intern/metrics.clj
```diff
@@ -201,6 +201,76 @@
...
+    (.format date-val (java.time.format.DateTimeFormatter/ofPattern "MMM d"))
+    (let [sql-date (if (instance? java.sql.Timestamp date-val)
+                     (.toLocalDate (.toLocalDateTime date-val))
+                     (.toLocalDate date-val))]
+      (.format sql-date (java.time.format.DateTimeFormatter/ofPattern "MMM d")))))
+
+(defn rolling-avg-signups
+  "Get rolling 7-day average signup counts for the last n weeks"
+  ([conn weeks]
+   (let [end-date (LocalDate/now)
```
dwwoelfel: If you use `LocalDate/now` in prod, I think that's going to give the date in UTC. Should we explicitly set the timezone to pacific?

PR #1413, opened 2025-07-21
better typed presence

drew-harris:
This PR re-exports the utility types from core in the react library and adds an Expand inside RoomsForDef so that hovers look better and are easier to reconstruct. 

Example before:
image
Example after:
image


client/packages/react/src/index.ts
```diff
@@ -36,6 +36,10 @@ import {
   type InstaQLEntity,
   type InstaQLFields,
   type InstaQLResult,
+  type RoomsOf,
```
dwwoelfel, line 39: Do we need to export these from the react-native package and the admin package?

drew-harris: Good point, doing that now

client/sandbox/react-nextjs/pages/play/typed-presence.tsx
```diff
@@ -0,0 +1,104 @@
+import React, { useEffect, useRef, useState } from 'react';
+import Head from 'next/head';
+import { i, init, RoomHandle } from '@instantdb/core';
```
dwwoelfel, line 3: We use `RoomHandle` here, do we also need to export it from `@instantdb/react` and friends, or are we not expecting users to use `RoomHandle`?

PR #1414, opened 2025-07-22
[Metrics[ Split discord ping into two messages

nezaj:
Thought it would be nicer to ping the actives and signups metrics in two messages so you can easily see the charts w/o needing to click into them


dwwoelfel (approved):
LGTM!

PR #1416, opened 2025-07-22
Fix missing pageInfo

nezaj:
I noticed we can get into a state where `pageInfo` returns undefined.

## Repro steps below.

1. Set up a paginated query
2. Create enough items to go to page 2
3. Delete the items from page 2
4. See that `pageInfo` becomes undefined

This happens for both offset and cursor pagination. The main fix was to check whether we currently have an `offset` > 0 or an `after-cursor`, if we do then we set `hasPreviousPage` to true.

## Additional things:

* Noticed the `pagination` example in the sandbox was broken due to required attributes -- fixed this up
* Added a `paginated-messages` example to demonstrate this works as expected

## Question:

I notice we don't export `Cursor` -- should we? For the `paginated-messages` example I manually type the before/after cursor as `[string, string, any, number]`

## Demo

**Before**

https://github.com/user-attachments/assets/d1eea951-4938-457f-9b7a-e1c078db9008

**After**

https://github.com/user-attachments/assets/64404b9a-1ab5-4d02-ac93-9f9afaedd03f

### DIFF
--- a/server/test/instant/db/instaql_test.clj
+++ b/server/test/instant/db/instaql_test.clj
@@ -2454,7 +2454,6 @@
             (is (= #{"null" "undefined"} (run-query {:etype {:$ {:where {:boolean {:$isNull true}}}}})))
             (is (= #{"a" "b" "c"} (run-query {:etype {:$ {:where {:boolean {:$isNull false}}}}})))))
 
-
         (testing "$not"
           (testing "string"
             (is (= #{"null" "undefined" "a" "c"}  (run-query {:etype {:$ {:where {:string {:$not "1"}}}}}))))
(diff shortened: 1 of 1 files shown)


server/src/instant/db/datalog.clj
```diff
@@ -2140,10 +2140,18 @@
                                                                first
                                                                (get "exists")))
                               (:page-info group) (assoc-in [:page-info :has-previous-page?]
-                                                           (-> sql-res
-                                                               (get (name (has-prev-tbl table)))
-                                                               first
-                                                               (get "exists"))))
+                                                           (let [has-prev (-> sql-res
```
dwwoelfel, line 2143: If you have a namespace with 0 items and you query `{:ns {:$ {:offset 10}}}` will this return `hasPreviousPage? true`?

nezaj: Indeed it will.

For posterity -- we did a lot of digging and settled on sticking with approach. The alternative we explored would be much more complex to support both cursor and offset pagination

PR #1419, opened 2025-07-23
detect content-type from filename if not provided

drew-harris:


server/deps.edn
```diff
@@ -72,6 +72,7 @@
         com.stripe/stripe-java {:mvn/version "25.1.0"}
         inflections/inflections {:mvn/version "0.14.1" :exclusions [noencore/noencore]}
         noencore/noencore {:mvn/version "0.3.8"}
+        org.apache.tika/tika-core {:mvn/version "2.9.1"}
```
nezaj: Nit. Should we use the latest version 3.2.1?

https://mvnrepository.com/artifact/org.apache.tika/tika-core/3.2.1

server/src/instant/storage/s3.clj
```diff
@@ -90,11 +92,19 @@
...
 
-(defn upload-file-to-s3 [{:keys [app-id location-id] :as ctx} file]
+(defn upload-file-to-s3 [{:keys [app-id location-id content-type] :as ctx} file]
   (when (not (instance? java.io.InputStream file))
     (throw (Exception. "Unsupported file format")))
-  (let [ctx* (assoc ctx :object-key (->object-key app-id location-id))]
-    (s3-util/upload-stream-to-s3 (s3-async-client) bucket-name ctx* file)))
+  (let [tika-stream (TikaInputStream/get file)
+        detected-mime-type (when-not content-type
+                             (.detect (Tika.) tika-stream))
```
dwwoelfel: I'd put this in a def, e.g.

`(def detector (Tika.))`

And then

`(.detect detector tika-tika-stream)`

server/src/instant/storage/s3.clj
```diff
@@ -90,10 +91,13 @@
 ;; Instant <> S3 integration
 ;; ----------------------
 
-(defn upload-file-to-s3 [{:keys [app-id location-id] :as ctx} file]
+(defn upload-file-to-s3 [{:keys [app-id location-id content-type] :as ctx} file]
   (when (not (instance? java.io.InputStream file))
     (throw (Exception. "Unsupported file format")))
-  (let [ctx* (assoc ctx :object-key (->object-key app-id location-id))]
+  (let [detected-mime-type (when-not content-type (.detect (Tika.) ^String (:path ctx)))
```
nezaj: Detecting based on the path won't help for use cases where folks don't include the extension in the path (this is common for example for setting a profile picture).

Let's use `TikaInputStream` and put the detection behind a feature flag -- we can post in our storage-feedback channel that we're testing this new change and allow folks to ping us their app ids to enable it

server/src/instant/storage/s3.clj
```diff
@@ -90,11 +93,32 @@
...
+            tika (Tika.)
+            mime-type (.detect tika ^TikaInputStream tika-stream)
+            ctx* (assoc ctx
+                        :object-key (->object-key app-id location-id)
+                        :content-type mime-type)]
+        (try
+          (s3-util/upload-stream-to-s3 (s3-async-client) bucket-name ctx* tika-stream)
+          (finally
+            (.close tika-stream))))
+      (let [detected-mime-type (.detect (Tika.) ^String (:path ctx))
```
nezaj: I think we should cut this. Adding this is in now is new behavior and it would be nice to keep things as-is while trying out this new dependency

server/src/instant/storage/s3.clj
```diff
@@ -90,11 +93,32 @@
...
 ;; ----------------------
 
-(defn upload-file-to-s3 [{:keys [app-id location-id] :as ctx} file]
+(defn upload-file-to-s3 [{:keys [app-id location-id content-type] :as ctx} file]
   (when (not (instance? java.io.InputStream file))
     (throw (Exception. "Unsupported file format")))
-  (let [ctx* (assoc ctx :object-key (->object-key app-id location-id))]
-    (s3-util/upload-stream-to-s3 (s3-async-client) bucket-name ctx* file)))
+  (if content-type
+    ;; Content-type provided
```
nezaj: Nit. I would remove this comment, we can already tell from the if check

server/src/instant/storage/s3.clj
```diff
@@ -90,11 +93,32 @@
...
+    (let [ctx* (assoc ctx
+                      :object-key (->object-key app-id location-id)
+                      :content-type content-type)]
+      (s3-util/upload-stream-to-s3 (s3-async-client) bucket-name ctx* file))
+    ;; No content-type
```
nezaj: Similar would remove this comment too

server/src/instant/storage/s3.clj
```diff
@@ -90,11 +93,32 @@
...
+(defn upload-file-to-s3 [{:keys [app-id location-id content-type] :as ctx} file]
   (when (not (instance? java.io.InputStream file))
     (throw (Exception. "Unsupported file format")))
-  (let [ctx* (assoc ctx :object-key (->object-key app-id location-id))]
-    (s3-util/upload-stream-to-s3 (s3-async-client) bucket-name ctx* file)))
+  (if content-type
+    ;; Content-type provided
+    (let [ctx* (assoc ctx
+                      :object-key (->object-key app-id location-id)
+                      :content-type content-type)]
```
nezaj: `content-type` will already be in `ctx` so no need to assoc it again

nezaj (commented):
Instead of these if checks I think we can do the following

* Determine a final content-type (use the one provided or tika if someone opted in)
* Assoc `:content-type` onto `ctx*`with this finalized content-type
* Call upload

This should be a bit more streamlined to read

server/src/instant/flags.clj
```diff
@@ -147,10 +147,14 @@
...
-                        (assoc acc (keyword setting) value))
-                      {}
-                      (get result "flags"))
+        flags (->
+               (reduce (fn [acc {:strs [setting value]}]
+                         (assoc acc (keyword setting) value))
+                       {}
+                       (get result "flags"))
+               (update :tika-enabled-apps (fn [vs]
+                                            (set (map parse-uuid vs)))))
```
dwwoelfel:
```suggestion
        flags (-> (reduce (fn [acc {:strs [setting value]}]
                            (assoc acc (keyword setting) value))
                          {}
                          (get result "flags"))
                  (update :tika-enabled-apps (fn [vs]
                                               (set (map parse-uuid vs)))))
```

server/src/instant/storage/s3.clj
```diff
@@ -90,11 +93,32 @@
...
+      (s3-util/upload-stream-to-s3 (s3-async-client) bucket-name ctx* file))
+    ;; No content-type
+    (if (contains? (flags/flag :tika-enabled-apps) app-id)
+      (let [tika-stream (TikaInputStream/get file)
+            tika (Tika.)
```
dwwoelfel: Do we need a new `Tika` for every invocation? I think it's safe to do `(def ^Tika detector (Tika.))` higher up.

client/packages/admin/src/index.ts
```diff
@@ -573,11 +573,17 @@ class Storage {
...
       path,
-      'content-type': metadata.contentType || 'application/octet-stream',
     };
     if (metadata.contentDisposition) {
       headers['content-disposition'] = metadata.contentDisposition;
     }
+    if (!metadata.contentType) {
+      // headers.content-type will become "undefined" (string)
+      // if not removed from the object
+      delete headers['content-type'];
```
stopachka: nit: this is a bit confusing. 

If there _is_ a metadata.contentType
- We overwrite headers['content-type']
But if there _isn't_ a metadata.contentType 
- We _delete_ headers['content-type']

In that case, could we move `delete headers['content-type']` of the if statements?

drew-harris: Flipped the logic to 
```ts
delete headers['content-type'];

if (metadata.contentType) {
  headers['content-type'] = metadata.contentType;
}
```

server/src/instant/storage/s3.clj
```diff
@@ -90,11 +94,32 @@
...
+  (and
+   (not content-type)
+   (contains? (flags/flag :tika-enabled-apps) app-id)))
+
+(defn deduce-content-type [{:keys [content-type] :as ctx} file]
+  (if (requires-tika? ctx)
+    (.detect tika ^TikaInputStream file)
+    content-type))
+
+(defn get-file-stream [ctx file]
```
dwwoelfel:
```suggestion
(defn get-file-stream ^InputStrem [ctx ^InputStream file]
```

With this change, the type hints act as extra documentation. Then you can get rid of the inline type hint on line 111 and the type hint on line 117.

server/src/instant/storage/s3.clj
```diff
@@ -90,11 +94,32 @@
...
+  (if (requires-tika? ctx)
+    (TikaInputStream/get ^InputStream file)
+    file))
+
+(defn upload-file-to-s3 [{:keys [app-id location-id] :as ctx} _file]
```
dwwoelfel: If you put an underscore in front of `file`, then that tells the linter that it shouldn't report an error if you don't use the argument. I'd put it at the end (i.e. `file_`) so that the linter will complain if we forget to use it.

stopachka (commented):
Getting suuper close! Two final things: 

1. The PR description is no longer related. Let's update it
2. Since this adds a flag, it would be good to include a deploy plan in the PR

PR #1422, opened 2025-07-24
Add response cache control

nezaj:
Following up from #977 , let's set a response cache header to speed up subsequent loads of a presigned url.

**Before** (on reload)

CleanShot 2025-07-24 at 11 47 24@2x

**After** (on reload)

CleanShot 2025-07-24 at 11 48 24@2x


dwwoelfel (approved):
Cool!

PR #1425, opened 2025-07-25
Enable auth.ref for storage rules

nezaj:
Folks weren't able to use `auth.ref` for `create` and `delete` permissions in storage (also for `view` for those who go through the deprecated `create-download-url`). This was because we missing some params for context to enable traversing refs.

Added some tests for this too!


dwwoelfel (approved):
LGTM!

PR #1426, opened 2025-07-25
Better pg_hint_plan with counts

dwwoelfel:
Uses counts from the database to choose better hints.

This is an experiment to see how good we can be at generating query plans if we have count information from the database. Right now it does an actual query like `count * from triples where attr_id = :attr-id and value = :value`. But if this works, then we should be able to replace that database query with a lookup in a [count-min sketch](https://en.wikipedia.org/wiki/Count%E2%80%93min_sketch) that we maintain for each attr.

**How it works**

If the *testing-pg-hints* flag is on, we'll walk the `nested-named-patterns` to determine the best index to use for each pattern and update each pattern with the index to use. We walk it in two phases--in phase 1, we collect all of the count queries we will need to make. Then we fetch all of the counts in one go. In phase 2, we make index decisions based on the counts.

As we go through the patterns, we keep track of the information we've gathered in each previous pattern to get the best hints. For example, if we know the `e` from a previous pattern returns a very small number, then we might look at indexes that have the e first.

To keep the count queries from getting too out of control, we do a `select count(*) from (select * from x limit 10000))`.

### DIFF
--- a/server/src/instant/db/datalog.clj
+++ b/server/src/instant/db/datalog.clj
@@ -1049,7 +1049,10 @@
         10000)
       (get-in ctx [:counts count-info]))))
 
-(defn index-compare [a b]
+(defn index-compare
+  "Compares the indexes pairwise to try to pick the best one.
+   A scoring system might be better, but this is a easier to debug."
+  [a b]
   (cond (and (empty? (:known-remaining (:index-costs a)))
              (seq (:known-remaining (:index-costs b))))
         -1
@@ -1116,6 +1119,7 @@
    defined at this point in the query."
   [ctx named-p symbol-map]
   (let [index-candidates
+        ;; Filters the indexes that can't work for this CTE (e.g. ea index on a ref)
         (remove (fn [idx-config]
                   (or (and (:idx-key idx-config)
                            (not= (:idx-key idx-config)
@@ -1137,6 +1141,8 @@
                                                             :op)))))))))
                 index-configs)
 
+        ;; Gets the components that we know (either constants or defined
+        ;; in the symbol map by a previous CTE)
         known-components
         (reduce (fn [acc c]
                   (let [[tag value] (get named-p c)]
@@ -1166,6 +1172,8 @@
                 {}
                 [:e :a :v :created-at])
 
+        ;; Gets the components that we're joining on from a previous
+        ;; CTE
         join-components
         (reduce (fn [acc c]
                   (let [[tag value] (get named-p c)]
@@ -1176,6 +1184,8 @@
                 {}
                 [:e :a :v :created-at])
 
+        ;; Gets the variable components that we don't have in the
+        ;; symbol map
         needed-components (reduce (fn [acc c]
                                     (let [[tag value] (get named-p c)]
                                       (if (and (= tag :variable)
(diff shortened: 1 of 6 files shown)


server/src/instant/db/datalog.clj
```diff
@@ -677,7 +678,7 @@
     (uuid-util/coerce s)
     (uuid-util/coerce (str s (subs all-fs-uuid (count s))))))
 
-(defn- value-function-clauses [app-id idx [v-tag v-value]]
+(defn- value-function-clauses [app-id triples-alias idx [v-tag v-value]]
```
dwwoelfel, line 681: This gives the subqueries a better alias name. Before it was always just "t", now it will be "t1_subquery" or "t2_subquery". 

This will let us add hints to the subqueries (not something we do yet) and makes it easier to distinguish each query when we call `determine-indexes-used`.

server/src/instant/db/datalog.clj
```diff
@@ -808,17 +820,23 @@
...
+  (let [attr-ids-where (when-let [a (:a named-pattern)]
+                         (when (named-constant? a)
+                           (constant->where-part idx app-id :a a)))
+        rest-wheres (concat (->> (dissoc named-pattern :a)
+                                 constant-components
+                                 (map (fn [[component-type v]]
+                                        (constant->where-part idx app-id component-type v))))
+                            (function-clauses app-id triples-alias named-pattern)
+                            (patch-values-for-av-index (idx-key idx) additional-clauses))]
+    (list*
```
dwwoelfel, line 833: This change causes some problems, so I'm going to update this to only leave out the index if we're doing the count query.

server/src/instant/db/hint_testing.clj
```diff
@@ -0,0 +1,194 @@
+(ns instant.db.hint-testing
```
dwwoelfel, line 1: This replaces the rule-where-testing ns that we hijacked for hint testing, now it's a little better for testing hints:

It runs the query once to collect the datalog queries, then it compares running with and without pg_hint_plan. This helps us separate out the initial query vs the query to load permission data.

server/src/instant/db/datalog.clj
```diff
@@ -808,17 +820,28 @@
...
-   :and
-   [:= :app-id app-id]
-   [:= (idx-key idx) :true]
-   (concat (->> named-pattern
-                constant-components
-                (map (fn [[component-type v]]
-                       (constant->where-part idx app-id component-type v))))
-           (function-clauses app-id named-pattern)
-           (patch-values-for-av-index (idx-key idx) additional-clauses))))
+  [{:keys [app-id
```
dwwoelfel, line 823: Small change here to remove the index from the query (e.g. `:ea = :true`) when we're doing the counts. Doesn't affect normal queries.

server/src/instant/db/datalog.clj
```diff
@@ -978,60 +1020,386 @@
...
+
+        (> (count (:path (:index-costs a)))
+           (count (:path (:index-costs b))))
+        -1
+
+        (> (count (:path (:index-costs b)))
+           (count (:path (:index-costs a))))
+        1
+
+        (and (< 0 (count (:path (:index-costs a))))
```
nezaj, line 1072: We should probably check count of `b` to prevent a divide by zero risk

PR #1427, opened 2025-07-25
Better changed files action

dwwoelfel:
Better changed-files action that works even when you do a force push.

It uses the github-script action to query the API for changed files instead of doing a git diff. 

It's faster than the old step because it doesn't need to check out the repo.

Tested that it works on a fresh PR here: https://github.com/instantdb/instant/pull/1431

### DIFF
--- a/.github/workflows/clojure.yml
+++ b/.github/workflows/clojure.yml
@@ -42,7 +42,7 @@ jobs:
           for (const file of files) {
             const changes = {
               server: file.startsWith('server/'),
-              refinery: file.startsWith('server/refinery/')
+              refinery: file.startsWith('server/refinery/'),
               workflowFile: file === '.github/workflows/clojure.yml'
             };
             console.log('file', file);


client/packages/mcp/src/oauth-service-provider.ts
```diff
@@ -106,8 +106,6 @@ export function makeApiAuth(
   };
 }
 
-// https://github.com/modelcontextprotocol/modelcontextprotocol/issues/653
-// Anthropic says it's fixed, but it doesn't seem like it
```
dwwoelfel, line 110: This comment is duplicated below (made the change here to test the detection was working)

client/www/pages/docs/instaml.md
```diff
@@ -14,7 +14,7 @@ import { init, id } from '@instantdb/react';
 
 // Instant app
 const APP_ID = '__APP_ID__';
-const db = init({ appID: APP_ID });
+const db = init({ appId: APP_ID });
```
nezaj, line 17: Nice!

.github/workflows/clojure.yml
```diff
@@ -5,44 +5,68 @@ on:
...
+            refinery_did_change: false,
+          };
+
+          for (const file of files) {
+            const changes = {
+              server: file.startsWith('server/'),
+              refinery: file.startsWith('server/refinery/'),
+              workflowFile: file === '.github/workflows/clojure.yml'
+            };
+            console.log('file', file);
```
nezaj, line 48: Curious, did you intended to leave the logs in?

nezaj (approved):
LGTM w/ one question!

.github/workflows/clojure.yml
dwwoelfel: Yes, it makes it much easier to debug why something happened or didn't. Plus there's no extra charge for the logs!

PR #1428, opened 2025-07-25
[Docs] Update CDN url

nezaj:
We have our standalone packages again thanks to @dww! Updating docs here to include the right link


dwwoelfel (approved):
LGTM!

PR #1432, opened 2025-07-26
Skip child order

dwwoelfel:
This ignores ordering for child forms on the postgres side. We'll still give the user the data in the right order when they get it from the client or the admin query. 

We don't support pagination on child forms and we don't rely on the order of the triples from postgres to apply the order-by on the client (or in admin query). So any ordering we do in postgres for child forms is just wasted cycles.

In this query:

```clj
{:users {:$ {:where {:email {:$ilike "%org.com"}, :order {createdAt "desc"}
             :comments: {:$ {:order {:createdAt "desc"}}}}}
```

We'll order the users in the database so that we can get cursors, but we'll treat the comments query as if it was written without the order by.

I'd also like to skip the ordering when you don't provide a `first`, `last`, `before`, or `after` param, but we already return cursors in that case and it might be considered a breaking change to stop.

### DIFF
--- a/server/src/instant/db/instaql.clj
+++ b/server/src/instant/db/instaql.clj
@@ -24,7 +24,7 @@
    [instant.util.tracer :as tracer]
    [instant.util.uuid :as uuid-util]
    [instant.system-catalog :as system-catalog]
-   [medley.core :refer [update-existing-in dissoc-in]]
+   [medley.core :refer [update-existing-in]]
    [instant.storage.s3 :as instant-s3]
    [instant.comment :as c])
   (:import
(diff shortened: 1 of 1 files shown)


server/src/instant/db/instaql.clj
```diff
@@ -863,11 +863,8 @@
 (defn page-info-of-form [{:keys [state] :as ctx}
                          {:keys [etype level option-map] :as _form}]
   (let [{:keys [order limit first last offset before after]} option-map]
-    ;; We don't need to do extra work to order the results if we're returning
```
dwwoelfel, line 866: I removed this comment, because we do check for the `order` field now.

nezaj (approved):
Nice!

PR #1433, opened 2025-07-26
Add better error for invalid id

nezaj:
Sometimes LLMs don't use uuids in transacts. Before we gave a generic spec failure. This PR adds a more helpful error for users to copy and paste to get unstuck.

**Before**

CleanShot 2025-07-25 at 17 57 17@2x

**After**

CleanShot 2025-07-25 at 17 56 51@2x


dwwoelfel (approved):
LGTM!

PR #1434, opened 2025-07-28
Detect Bun in cli for newer versions of Bun.

drew-harris:
Bun recently switched to a non-binary lockfile format for new projects. If that is in use, this will detect it.


stopachka (approved):
Niice!

dwwoelfel (approved):
Cool!

PR #1436, opened 2025-07-28
[docs] Fix reference to step 3.5 in Sign in with Apple

stopachka:
A user pointed out that in one of our steps for sign in with apple, we were talking about a "step 3.5", which was non-existent for their flow. 

Ref: https://x.com/maxrusakovic/status/1944073418204119395

This is because our conditional param was not working, when we included `<!-- prettier-ignore -->` beside it. 

I went ahead and removed the prettier-ignore, and the conditional now works. 


nezaj (approved):
Nice!

PR #1437, opened 2025-07-28
Use the same cache key for docs and dash selected app

stopachka:
In one of our user sessions, a user mentioned how he was surprised that the selected app across dash and docs weren't in sync.

I went ahead and made it so both the docs and the dash use the same local storage key to store the app id. 

Hence, they now stay in sync!

**Note: you may be wondering, why keep using getLocal and setLocal, instead of useLocalStorage?** 

If we switch to `useLocalStorage`, we can't quite keep the useEffect logic as it currently is, since setting our cached item will trigger a re-render. I wanted to keep this change incremental. 


nezaj (approved):
Nice!

PR #1438, opened 2025-07-28
validate queries and transactions

drew-harris:
This pr adds runtime validation for queries and transactions. When queries or transactions are run, if they are invalid, a helpful error will be thrown synchronously.

### DIFF
--- a/client/packages/version/src/version.ts
+++ b/client/packages/version/src/version.ts
@@ -1,6 +1,6 @@
 // This is the shared version for all of the js packages
 // Update the version here and merge your code to main to
 // publish a new version of all of the packages to npm.
-const version = 'v0.20.18';
+const version = 'v0.20.19';
 
 export { version };


client/packages/core/__tests__/src/queryValidation.test.ts
```diff
@@ -0,0 +1,792 @@
...
+  q: unknown,
+  schema: InstantSchemaDef<any, any, any> | null = testSchema,
+) => {
+  expect(() => validateQuery(q, schema ?? undefined)).toThrow();
+};
+
+test('validates top level types', () => {
+  beValid({});
+  beWrong('Testing');
+  beWrong(8392);
```
stopachka, line 102: Maybe we also test `beWrong([])` ?

client/packages/core/__tests__/src/queryValidation.test.ts
```diff
@@ -0,0 +1,792 @@
...
+      doesNotExist: {},
+    },
+  });
+
+  beWrong({
+    posts: {
+      unlinkedWithAnything: {},
+    },
+  });
+
```
stopachka, line 149: Another fun won for beWrong: 

```
beWrong({ posts: [] })
```

client/packages/core/__tests__/src/queryValidation.test.ts
```diff
@@ -0,0 +1,792 @@
...
+    posts: {
+      comments: {},
+    },
+  });
+});
+
+test('dollar sign object', () => {
+  beWrong({
+    posts: {
+      $where: {
```
stopachka, line 160: Nice!

client/packages/core/__tests__/src/queryValidation.test.ts
```diff
@@ -0,0 +1,792 @@
...
+    posts: {
+      $: {
+        where: {
+          title: { $not: 'Draft' },
+        },
+      },
+    },
+  });
+
+  // Valid $gt, $lt, $gte, $lte operators (would need indexed fields in real usage)
```
stopachka: Should we throw if we know the field isn't indexed?

client/packages/core/__tests__/src/queryValidation.test.ts
```diff
@@ -0,0 +1,792 @@
...
+      $: {
+        where: {
+          name: { $in: ['John', 123] },
+        },
+      },
+    },
+  });
+
+  // Invalid 'in' operator - wrong type in array
+  beWrong({
```
stopachka, line 363: Would add a test that this passes with no schema

client/packages/core/__tests__/src/transactionValidation.test.ts
```diff
@@ -0,0 +1,369 @@
...
+  // Valid update with multiple fields
+  beValid(tx.users[userId].update({ name: 'Jane', bio: 'Updated bio' }));
+
+  // Invalid update - wrong type
+  // @ts-expect-error
+  beWrong(tx.users[userId].update({ name: 123 }));
+
+  // Invalid update - unknown attribute
+  // @ts-expect-error
+  beWrong(tx.users[userId].update({ unknownField: 'value' }));
```
stopachka: A part of me hesitates in erroring out when we receive extra fields. It feels unrichhickey-like -- but thinking it through I do think it makes sense. Wanted to flag in case others disagree with me. @nezaj @tonsky @dwwoelfel

client/packages/core/__tests__/src/transactionValidation.test.ts
```diff
@@ -0,0 +1,369 @@
...
+  // Valid link
+  beValid(tx.users[userId].link({ posts: postId }));
+
+  // Valid link with array
+  beValid(tx.users[userId].link({ posts: [postId, id()] }));
+
+  // Invalid link - unknown link
+  // @ts-expect-error
+  beWrong(tx.users[userId].link({ unknownLink: postId }));
+
```
stopachka, line 202: Maybe we also fail if link isn't a uuid?

client/packages/core/__tests__/src/transactionValidation.test.ts
```diff
@@ -0,0 +1,369 @@
...
+  beWrong(tx.users[userId].unlink({ unknownLink: postId }));
+});
+
+test('validates entity existence', () => {
+  const unknownId = id();
+
+  // Invalid entity
+  beWrong({
+    __ops: [['create', 'unknownEntity', unknownId, { field: 'value' }]],
+    __etype: 'unknownEntity',
```
stopachka: I would write this as `unknownNamespace`

client/packages/core/__tests__/src/transactionValidation.test.ts
```diff
@@ -0,0 +1,369 @@
...
+    tx.users[userId].create({
+      name: 'John',
+      email: 'john@example.com',
+      junk: { anything: 'goes' },
+    }),
+  );
+
+  // Invalid create - wrong type
+  // @ts-expect-error
+  beWrong(tx.users[userId].create({ name: 123, email: 'john@example.com' }));
```
stopachka, line 135: Maybe we also validate that the input id is a uuid?

client/packages/core/src/queryValidation.ts
```diff
@@ -0,0 +1,380 @@
...
+  'fields',
+  'aggregate',
+];
+
+const getAttrType = (attrDef: DataAttrDef<any, any, any>): string => {
+  if (attrDef.valueType === 'string') return 'string';
+  if (attrDef.valueType === 'number') return 'number';
+  if (attrDef.valueType === 'boolean') return 'boolean';
+  if (attrDef.valueType === 'date') return 'date';
+  if (attrDef.valueType === 'json') return 'object';
```
stopachka: Thought: 

Perhaps we should just return attrDef.valueType || 'uknown' 

Main reason: it would be nice if we could avoid the slight indirection from json to object. 

We could also return an enum type here rather than string.

client/packages/core/src/queryValidation.ts
```diff
@@ -0,0 +1,380 @@
...
+  return 'unknown';
+};
+
+const isValidValueForType = (
+  value: unknown,
+  expectedType: string,
```
stopachka: We could then use that enum here

client/packages/core/src/queryValidation.ts
```diff
@@ -0,0 +1,380 @@
...
+const validateWhereClauseValue = (
+  value: unknown,
+  attrName: string,
+  attrDef: DataAttrDef<any, any, any>,
+  entityName: string,
+): void => {
+  const expectedType = getAttrType(attrDef);
+  const isAnyType = attrDef.valueType === 'json';
+
+  if (typeof value === 'object' && value !== null && !Array.isArray(value)) {
```
stopachka: maybe we can make this a variable `isComplexObject` to simplify the reading

client/packages/core/src/queryValidation.ts
```diff
@@ -0,0 +1,380 @@
...
+    // For any type, allow complex objects without treating them as operators
+    if (isAnyType) {
+      return; // Any type accepts any value, including complex objects
+    }
+
+    const operators = value as Record<string, unknown>;
+
+    for (const [op, opValue] of Object.entries(operators)) {
+      switch (op) {
```
stopachka: It may make sense to extract this bit into a helper function, to make reading this part of the function a bit easier

client/packages/core/src/queryValidation.ts
```diff
@@ -0,0 +1,380 @@
...
+        case 'in':
+        case '$in':
+          if (!Array.isArray(opValue)) {
+            throw new QueryValidationError(
+              `Operator '${op}' for attribute '${attrName}' in entity '${entityName}' must be an array, but received: ${typeof opValue}`,
```
stopachka: One thought I had: 

It may be nice to keep track of a `path` variable, so as we validate things, we can then tell the user 'where' the error happened. 

We do something like this on the backend. 

https://github.com/instantdb/instant/blob/main/server/src/instant/db/instaql.clj#L223

client/packages/core/src/queryValidation.ts
```diff
@@ -0,0 +1,380 @@
...
+              );
+            }
+          }
+          break;
+        case '$not':
+        case '$gt':
+        case '$lt':
+        case '$gte':
+        case '$lte':
+          if (!isValidValueForType(opValue, expectedType, isAnyType)) {
```
stopachka: We may want to wrap this into a helper function like `assertValidValueForType`, so we don't have to repeat it

client/packages/core/src/queryValidation.ts
```diff
@@ -0,0 +1,380 @@
...
+      const availableLinks = Object.keys(currentEntity.links);
+      throw new QueryValidationError(
+        `Link '${linkName}' does not exist on entity '${currentEntityName}' in dot notation path '${dotPath}'. Available links: ${availableLinks.length > 0 ? availableLinks.join(', ') : 'none'}`,
+      );
+    }
+
+    currentEntityName = link.entityName;
+  }
+
+  // Validate the final attribute
```
stopachka, line 216: Nit: You don't need to final attribute necessarily. 

For example, you could write: 

```
{ comments: { $: { where: { post: post.id } } } }
```

Both backend and frontend I _think_ will still support this

stopachka (approved):
Great work! Some nits, but overall LGTM

client/packages/core/__tests__/src/transactionValidation.test.ts
```diff
@@ -367,3 +367,18 @@ test('validates without schema', () => {
   beValid(originalTx.randomEntity[userId].update({ anyField: 123 }), null);
   beValid(originalTx.randomEntity[userId].link({ anyLink: id() }), null);
 });
+
+test('validates UUID format for entity IDs', () => {
+  // Valid UUID should pass
+  const validUuid = id();
```
stopachka, line 373: One thing I would double check: 

Lookups should still be valid.

client/packages/core/__tests__/src/transactionValidation.test.ts
```diff
@@ -367,3 +367,18 @@ test('validates without schema', () => {
   beValid(originalTx.randomEntity[userId].update({ anyField: 123 }), null);
   beValid(originalTx.randomEntity[userId].link({ anyLink: id() }), null);
 });
+
+test('validates UUID format for entity IDs', () => {
```
stopachka, line 371: We may also want this test on `link`

client/packages/core/src/queryValidation.ts
```diff
@@ -46,93 +50,126 @@ const isValidValueForType = (
...
-            throw new QueryValidationError(
-              `Cannot use '$isNull: true' on required attribute '${attrName}' in entity '${entityName}'`,
-            );
-          }
-          break;
-        default:
-          throw new QueryValidationError(
-            `Unknown operator '${op}' for attribute '${attrName}' in entity '${entityName}'`,
-          );
+    const assertValidValue = (
```
stopachka: Nit: Perhaps we should move this function out into the top-level? 

Writing function inline are nice when you want to use enclose some surrounding variables. But in this case I think we don't do that. Future readers may wonder: "which variable is being enclosed", and it's generally hard to discover.

client/packages/core/__tests__/src/queryValidation.test.ts
```diff
@@ -0,0 +1,800 @@
...
+  });
+});
+
+test('where clause dot notation validation', () => {
+  // Valid dot notation - users.posts.title
+  beValid({
+    users: {
+      $: {
+        where: {
+          'posts.title': 'Some Title',
```
stopachka, line 628: I would add a test here for 

```
    users: {
      $: {
        where: {
           posts: id()
```


```
    users: {
      $: {
        where: {
          'posts.comments': id()
```

client/packages/core/__tests__/src/queryValidation.test.ts
```diff
@@ -0,0 +1,800 @@
...
+        where: {
+          'author.name': { $isNull: true },
+        },
+      },
+    },
+  });
+
+  // Don't need final attributes
+  beValid({
+    comments: { $: { where: { post: 'hi' } } },
```
stopachka: In this case I _think_ it would have to be a uuid

stopachka (approved):
Aweesome!!

client/packages/core/src/transactionValidation.ts
```diff
@@ -122,6 +123,16 @@ const validateOp = (
 
   const [action, entityName, _id, args] = op;
 
+  // _id should be a uuid
+  if (!Array.isArray(_id)) {
+    const isUuid = validateUuid(_id);
```
stopachka: nit: would call this `validateUUID`

stopachka:
Alright good to land @drew-harris !

PR #1441, opened 2025-07-28
Fix formatting for apple docs conditionals

dwwoelfel:
Two changes:

1. Adds an else branch to the nav `{% conditional %}` tags
2. Uses the else branch in the apple doc. We copy the shared elements from the list into the else branch so that we don't have to put a conditional in between list items. The conditional inside of the items gives us weird gaps.

Before:

image

After:

image


stopachka (approved):
Ah, thank you @dwwoelfel !

PR #1443, opened 2025-07-29
Use `= any(values)` instead of `in values`

dwwoelfel:
Updates our datalog queries to use `= ANY` instead of `IN` or multiple `OR`s, e.g. `value = ANY '{"a","b"}' instead of `value in ("a", "b")` or `value = "a" OR value = "b"`. 

We did this in a previous PR for matching the attr ids: https://github.com/instantdb/instant/pull/973

The benefits are:
1. Fewer parameters
2. Less data over the wire
3. Better query plans for `$in` queries

If you do a query with `$in`, we'll create a big list of ORs and postgres will do a big bitmap-or scan on all of the individual values. 

For this query:

```js
{$files: {$: {where: {paths: {$in: list-of-paths}}}
```
The query plan before looks like:

```
    ->  Bitmap Heap Scan on triples t0  (cost=36.06..37.22 rows=1 width=88) (actual time=1.736..2.089 rows=20 loops=1)
          Recheck Cond: (((app_id = :app-id) AND (attr_id = :attr-id) AND (value = :v1) AND ave) OR ((app_id = :app-id) AND (attr_id = :attr-id) AND (value = :v2) AND ave) ...
          Heap Blocks: exact=20
          ->  BitmapOr  (cost=36.06..36.06 rows=1 width=0) (actual time=1.712..1.717 rows=0 loops=1)
                ->  Bitmap Index Scan on ave_index  (cost=0.00..1.80 rows=1 width=0) (actual time=0.328..0.328 rows=1 loops=1)
                      Index Cond: ((app_id = :app-id) AND (attr_id = :attr-id) AND (value = :v1))
                ->  Bitmap Index Scan on ave_index  (cost=0.00..1.80 rows=1 width=0) (actual time=0.031..0.031 rows=1 loops=1)
                      Index Cond: ((app_id = :app-id) AND (attr_id = :attr-id) AND (value = :v2))
                ...
```                

The query plan using `= any` can just use a single index condition

```
    ->  Index Scan using ave_index on triples t0  (cost=0.69..37.16 rows=1 width=88) (actual time=0.095..0.496 rows=20 loops=1)
          Index Cond: ((app_id = :app-id) AND (attr_id = :attr-id) AND (value = ANY ('{"v1, "v1",...}')))
```

### DIFF
--- a/server/src/instant/jdbc/sql.clj
+++ b/server/src/instant/jdbc/sql.clj
@@ -21,7 +21,6 @@
    (java.time Instant LocalDate LocalDateTime)
    (java.util UUID)
    (javax.sql DataSource)
-   (org.postgresql.jdbc PgArray)
    (org.postgresql.util PGobject PSQLException)))
 
 (set! *warn-on-reflection* true)
@@ -60,7 +59,7 @@
         "bit" (Long/parseLong value 2)
         value))))
 
-(defn <-pgarray [^PgArray a]
+(defn <-array [^Array a]
   (let [type (.getBaseTypeName a)
         vs (.getArray a)]
     (case type
@@ -87,9 +86,9 @@
 
 
 (extend-protocol rs/ReadableColumn
-  PgArray
-  (read-column-by-label [^PgArray v _] (<-pgarray v))
-  (read-column-by-index [^PgArray v _2 _3] (<-pgarray v))
+  Array
+  (read-column-by-label [^Array v _] (<-array v))
+  (read-column-by-index [^Array v _2 _3] (<-array v))
 
   PGobject
   (read-column-by-label [^PGobject v _] (<-pgobject v))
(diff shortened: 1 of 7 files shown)


stopachka (approved):
yasss 🔥

server/src/instant/jdbc/sql.clj
```diff
@@ -73,10 +35,36 @@
...
 
+(defn <-array [^Array a]
+  (let [type (.getBaseTypeName a)
+        vs (.getArray a)]
+    (case type
+      ("json" "jsonb") (mapv <-json vs)
+      (vec vs))))
+
+(defn- create-pg-array [^PreparedStatement s pgtype clazz vs]
+  (.createArrayOf (.getConnection s) pgtype (into-array clazz vs)))
```
dwwoelfel, line 46: Instead of writing our own stringify-er for arrays, this uses's the pg driver's default stringify-er.

PR #1449, opened 2025-07-30
[Docs] Clarify index vs unique

nezaj:
We got some feedback that the docs where confusing about `unique` vs `indexed` -- hoping to clarify with this PR!


dwwoelfel (approved):
Nice!

PR #1453, opened 2025-07-31
soft-delete attrs

stopachka:
This makes `delete-attr` soft delete, rather than hard-delete. This is a win for two reasons: 

1. delete-attr can now be very fast. This means we won't have trouble if we make it a first class citizen in cli commands. 
2. And it's reversible. Which means if an LLM makes this change accidentally, we can always revive the triples. 

To do it: 
1. I added `deletion_marked_at` column on attrs
2. When run delete-attr, we 
  a. Mark deletion_marked_at
  b. Remove `is_indexed` `is_required` `is_unique` markings (so there's no inadvertent errors about them)
  c. Rename etype+label (so we don't get uniqueness violations if the user creates the same attr again) 
3. When we run restore-attr, we 
  a. Remove deletion_marked_at, and put the names of our etype+labels back
4. To run sweeps, we've updated our app_deletion_sweeper to also delete attrs

Some notes:
1. I made restore-attr a part of transact, as we want invalidator to trigger. Invalidator only triggers when we have a `transaction` table entry in our wal record. I thought it would be best to keep the creation of that record to one place. Plus it means we can more easily allow restores in a UI down the road.
2. There's one gotcha in invalidator. When we restore, we need to write `[:ea _ _ _]`, to bust object queries


server/src/instant/db/model/attr.clj
```diff
@@ -577,7 +578,151 @@
...
+  (with-cache-invalidation app-id
+    (sql/do-execute!
+     ::restore-multi!
+     conn
+     (sql/format
+      "with restored_attrs as (
+        update attrs
+        set 
+          deletion_marked_at = null,
+          is_unique = false,
```
dwwoelfel: It's ok to keep the unique flag, because all of the data in the triples table will still satisfy the unique constraint.

server/src/instant/db/model/attr.clj
```diff
@@ -577,7 +578,151 @@
...
+  (with-cache-invalidation app-id
+    (sql/do-execute!
+     ::soft-delete-multi!
+     conn
+     (sql/format
+      "with soft_deleted_attrs as (
+        update attrs
+        set 
+          deletion_marked_at = now(),
+          etype = id::text || '_deleted$' || etype,
```
dwwoelfel, line 672: Maybe we could add another jsonb column to attr to store whether it used to be `unique` and `required`? We could use that information later to offer a better restore experience.

stopachka: I like the idea! 

WDYT think about a general `metadata` jsonb?

server/resources/migrations/75_attr_soft_deletes.up.sql
```diff
@@ -0,0 +1,7 @@
+alter table
+  attrs
+add
+  column deletion_marked_at timestamp with time zone;
+
+create index if not exists idx_attrs_deletion_marked_at on attrs (deletion_marked_at);
```
dwwoelfel: It might save us some space to add `where deletetion_marked_at is not null`

server/src/instant/db/model/attr.clj
```diff
@@ -577,7 +578,151 @@
...
+            else null 
+          end,
+          reverse_label = case 
+            when reverse_label is not null 
+            then substring(reverse_label from position('$' in etype) + 1)
+            else null 
+          end
+        where 
+          app_id = ?app-id 
+          and id in (
```
dwwoelfel: You should be able to provide the array as a param directly. You can do `id = any(?attr-ids)` here and then do `(with-meta ids {:pgtype "uuid[]"})` when you provide the param.

server/src/instant/db/permissioned_transaction.clj
```diff
@@ -548,17 +553,18 @@
 
                 check-commands
                 (io/warn-io :check-commands
-                  (concat
-                   (attr-checks ctx attr-changes)
+                            (concat
```
dwwoelfel: Do you know if there is some way to make conjure indent `io/warn-io` the same way it indents a `defn`?

server/src/instant/flags.clj
```diff
@@ -296,10 +296,10 @@
                                      [:e2e-logging :invalidator-every-n])
                              10000)))))
 
-(defn app-deletion-sweeper-disabled? []
+(defn hard-deletion-sweeper-disabled? []
   (-> (query-result)
```
dwwoelfel: Probably safer to move this down and use `toggled?` directly, e.g.

```clj
(defn hard-deletion-sweeper-disabled? []
  (toggled? :hard-deletion-sweeper-disabled?))
```

server/src/instant/db/permissioned_transaction.clj
stopachka: I think we can! Will update our kondo with this 🫡

server/src/instant/flags.clj
stopachka: Oo nice! Missed this

server/.lsp/config.edn
```diff
@@ -0,0 +1 @@
+{:cljfmt {:indents {#re "io/warn-io" [[:block 1]]}}}
```
stopachka, line 1: cc @dwwoelfel looks like clojure-lsp uses clj-fmt to format things. Adding this rule fixed the issue in permissioned-transaction.clj!

server/src/instant/util/exception.clj
```diff
@@ -480,9 +480,18 @@
 
 (defn translate-and-throw-psql-exception!
   [^PSQLException e]
-  (let [{:keys [server-message condition table] :as data} (pgerrors/extract-data e)
-        hint (select-keys data [:table :condition :constraint])]
+  (let [{:keys [server-message condition table column] :as data} (pgerrors/extract-data e)
+        hint (select-keys data [:table :condition :constraint :column])]
     (case condition
+      :not-null-violation
```
stopachka: (update: actually I will do something different here. I will throw a custom exception!)

Issue #790, opened 2025-02-01
Explorer namespaces list fails to stay up-to-date


Mertzenich:
# Description

I noticed a bug with the namespace list while I was experimenting in the sandbox. In the "Explorer" tab of the dashboard, namespaces that were added while another tab was open will not appear until a refresh or if further changes are made while the explorer is still open. Note that if you have both the explorer and the sandbox open at the same time the namespace list will update as it should, the problem occurs when the explorer is not open while a namespace is created.

# Steps (Text)

1. Navigate to the Dashboard
2. Navigate to the Sandbox tab
3. Check the "Write to DB" box and execute a transaction which creates a new namespace (i.e. `db.transact(tx.goals[id()].update({ title: 'eat' }));` in a blank app)
4. Navigate to the Explorer tab
5. Notice that the new "goals" namespace does not appear in the list
6. Refresh the page
7. Notice that the new namespace now appears

# Steps (Images)

![Image](https://github.com/user-attachments/assets/d06db7a4-d9e0-4753-b50f-8f26dfa16f7a)

![Image](https://github.com/user-attachments/assets/fd0aff7f-c028-4646-abbb-71691868f153)

![Image](https://github.com/user-attachments/assets/2af8848a-cf2c-466e-968b-2ffc098bdc51)

![Image](https://github.com/user-attachments/assets/3c954a6a-32f4-4955-88f5-ad089c5b729f)

![Image](https://github.com/user-attachments/assets/d378c59b-ec99-4414-8ea2-13e77023d492)

nezaj:
@Mertzenich thank you for this detailed issue and repro! Put up a fix for this here https://github.com/instantdb/instant/pull/1454

nezaj:
This is now shipped!

kiluazen:
The stale namespace bug you wrote up here is one of the reasons I started scoping live writes.

[Lens](https://lens.kushalsm.com/) is a schema-aware client for InstantDB. Subscriptions instead of polling, so namespaces and rows update without a refresh. Links also render as labels, not "1 link" placeholders.

Not shipping yet but adding user-one slots. The page has a way to ping me if your sandbox app could use it.

best,
[kushal](https://lens.kushalsm.com/)

PR #1454, opened 2025-07-31
[Dash] Fix namespace updates no propagating

nezaj:
We got a repro in https://github.com/instantdb/instant/issues/790 which shows that shows when users created a new namespace in the Sandbox tab, it wouldn't appear in the Explorer tab until they refreshed the page.

### Solution

Moved the schema subscription to the dashboard level to capture all namespace changes, eliminating the issue where newly mounted tabs wouldn't receive the current namespace state

### Changes:

1. Created `DashboardContent` component in `/www/pages/dash/index.tsx` that:

   - Calls `useSchemaQuery` at the dashboard level
   - Passes namespace data as props to child components
   - Ensures namespace updates are always being listened to

2. Created `DevtoolContent` component in `/www/pages/_devtool/index.tsx` with the same pattern

Thought it was also nicer to encapsulate these into their own components -- `Dashboard` and `DevtoolWithData` were kinda big

### Testing

- Create a namespace in Sandbox tab → immediately visible in Explorer tab
- Do the same and see the devtool also sees the update
- Schema, Perms, and Query Inspector tabs all receive live updates

**Before**

https://github.com/user-attachments/assets/c5bb62e0-a7a5-48a2-9fd3-d5017d4b3093

**After**

https://github.com/user-attachments/assets/62f21064-553a-42ec-bbeb-b330a684efa7

### DIFF
--- a/client/www/components/dash/Perms.tsx
+++ b/client/www/components/dash/Perms.tsx
@@ -10,16 +10,17 @@ import { InstantApp, DashResponse, SchemaNamespace } from '@/lib/types';
 import { Button, Content, JSONEditor, SectionHeading } from '@/components/ui';
 import { HomeButton } from '@/pages/dash';
 import { InstantReactWebDatabase } from '@instantdb/react';
-import { useSchemaQuery } from '@/lib/hooks/explorer';
 
 export function Perms({
   app,
   db,
   dashResponse,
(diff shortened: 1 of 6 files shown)


dwwoelfel (approved):
LGTM!

PR #1456, opened 2025-07-31
Tweak header and footer navs

nezaj:
I realize the only links we have to the tutorial at the moment are the button in the hero unit and in the home tab of the dashboard. Thought it would be nice to also add it into the header nav and be more explicit in the CTA.

Also removed some links in the footer which I thought was a little crowded


stopachka (approved):
SGTM!

PR #1457, opened 2025-07-31
fix random non-error being thrown

drew-harris:


client/packages/core/src/Reactor.js
```diff
@@ -310,7 +310,12 @@ export default class Reactor {
     if (ok) {
       dfd.resolve({ status, eventId });
     } else {
-      dfd.reject({ status, eventId, ...errDetails });
+      dfd.reject(
+        new Error(
+          'Transaction failed on the server: ' +
```
stopachka: One problem here, is that this will cause a really big text string as the error. 

You can look at InstantAPIError as one example where we throw ~something like what is in `errDetails`. 

I _think_ errDetails will always have `message`, which we can use as the error message.

--- 

Once we convert this into an error, one thing we may want to think about. Errors seem to hide the other details, like hints, etc. Right now it's nice that this shows up in the console. Maybe we can do some research about that.

@dwwoelfel if you have thoughts

drew-harris:
New Formatting
image

client/packages/core/src/Reactor.js
dwwoelfel: I think this should throw an `InstantAPIError` unless there's a reason not to.

client/packages/core/src/Reactor.js
```diff
@@ -292,25 +294,37 @@ export default class Reactor {
 
   /**
    * @param {'enqueued' | 'pending' | 'synced' | 'timeout' |  'error' } status
-   * @param string eventId
-   * @param {{message?: string, hint?: string, error?: Error}} [errDetails]
+   * @param {string} eventId
+   * @param {{message?: string, type?: string, status?: number, hint?: unknown}} [errorMsg]
```
stopachka, line 300: Aside, do we actually know that all callsites to this truly send message,type,status,hint? May be good to double check

client/packages/core/src/Reactor.js
```diff
@@ -292,25 +294,37 @@ export default class Reactor {
...
+      console.error('Mutation failed', { status, eventId, ...errorMsg });
     }
     if (!dfd) {
       return;
     }
     if (ok) {
       dfd.resolve({ status, eventId });
     } else {
-      dfd.reject({ status, eventId, ...errDetails });
+      // Check if error comes from server or client
```
stopachka, line 317: My thinking is, we should return the same `error` in both cases. 

In both cases imo we should throw dfd.reject(new InstantError(...)) 

Separately, we can make InstantAPIError extend InstantError. 

cc @dwwoelfel @nezaj if you have other thoughts

stopachka (approved):
One comment to discuss but otherwise LGTM!

client/packages/core/src/Reactor.js
drew-harris: Just double checked, all good 👍

client/packages/core/src/Reactor.js
drew-harris: I think it's a useful distinction to tell whether the error is from the server or the client and since InstantAPIError extends InstantError, checking if its InstantError will work 100% of the time

stopachka: My one worry here is that we may end up having client errors in the future that have a `type`. 

This feels bike-sheddy though, so will defer to you! Let's ship as is, we can change moving forward

PR #1458, opened 2025-08-01
[wip] count-min-sketch for attrs

dwwoelfel:
Adds count-min sketch for attrs to keep track of attr counts.

Still TODO
- [ ] Lots of small todos in the code
- [ ] Define the migrations for populating the initial sketches
- [ ] Set up the process for claiming the aggregation slot and monitoring it to claim once it is dropped by another machine

Stretch goal
- [ ] Keep track of data usage at the same time (tricky because we don't know how postgres defines the size)


dwwoelfel:
Replaced by https://github.com/instantdb/instant/pull/1485

PR #1459, opened 2025-08-01
Add a test that shows the current transaction behavior

dwwoelfel:
Adds a test that shows what happens when deletes are mixed with updates.

I added a test for the undesirable behavior we currently have, where delete followed by update followed by delete will not apply the last delete. That happens because the actions are grouped by their action, so all the deletes happen first.


tonsky:
Incorporated into #1386

PR #1463, opened 2025-08-01
useUser hook, <SignedIn> <SignedOut>, top level hooks

drew-harris:


client/packages/react/src/InstantReactAbstractDatabase.tsx
```diff
@@ -261,6 +261,14 @@ export default abstract class InstantReactAbstractDatabase<
     return state;
   };
 
+  useUser = () => {
```
stopachka: nit: I would add a docblock on the function with an example. This way when a user hovers over it in intellisense, they can see how to use it right away

client/packages/react/src/InstantReactAbstractDatabase.tsx
```diff
@@ -345,4 +353,22 @@ export default abstract class InstantReactAbstractDatabase<
   }> => {
     return this._core.queryOnce(query, opts);
   };
+
+  SignedIn = ({ children }: { children: ReactNode }) => {
```
stopachka: Three ideas here: 

1. nit: I would extract these out as standalone functions, and pass in `db` as the argument. 

Reasoning: since this is a React component, it can be surprising to use the `this` variable. 

2. Let's add extra props with defaults, to let users handle isLoading and error? 

Maybe we should add some extra props, like an option to render something when it's in an `isLoading` or `error` state. We could have defaults for these too. I _think_ for isLoading we'll want a null state. 

But what should we do for an error? We could just show some basic component. I'd be curious what others think. 

cc @dwwoelfel @nezaj @tonsky 

3. I would add a docblock with an example, so users can see how to use it in their editor

client/packages/react/src/InstantReactWebDatabase.ts
```diff
@@ -1,5 +1,5 @@
 import type { InstantConfig, InstantSchemaDef } from '@instantdb/core';
-import InstantReactAbstractDatabase from './InstantReactAbstractDatabase.ts';
+import InstantReactAbstractDatabase from './InstantReactAbstractDatabase.tsx';
 
 export default class InstantReactWebDatabase<
```
stopachka, line 4: nit: I _think_ we need to delete this file?

client/packages/react/src/index.ts
```diff
@@ -135,3 +149,7 @@ export {
   type UploadFileResponse,
   type DeleteFileResponse,
 };
+
+declare global {
```
stopachka: cc @dwwoelfel @nezaj @tonsky @drew-harris  for discussion 

I do worry about the top-level interface. 

The default experience will now be to write something like: 

```js
// db.ts 

import { schema } from '@instantdb/schema'; 

const db = init({ 
  appId: process.env.INSTANT_APP_ID,
  schema 
});

declare global {
  interface InstantRegister {
    schema: typeof schema;
  }
}
```

This works, but declare globals are used so infrequently, that most users don't know how to write the syntax. 

Once we do, it may get a bit more difficult to write more dynamic db's too. For example, to overwrite the type for a useQuery, you would have to write

```js
import { useQuery } from '@instantdb/react'; 

useQuery<InstaQLParams<typeof schema>({ ... })
```

--- 

One alternative approach: 

We keep everything but the global interface. One downside for now will be that users will have to write 

```
<db.SignedIn>
</db.SignedIn>
<db.SignedOut>
</db.SignedOut>
```

Which _does_ feel un-reacty too.

What do you guys think?

client/sandbox/react-native-expo/app/play/auth-and-context.tsx
```diff
@@ -0,0 +1,169 @@
+import {
```
stopachka: Niice! @drew-harris for the example you have for web, could we add it as a page in the nextjs sandbox too? Will make it easier to dogfood this PR

client/packages/react/src/index.ts
drew-harris: I think a benefit of the global declaration is that we could put it in the generated schema file so users won't have to think about it, but i agree on it being a weird pattern.
Are there more cases than I thought of users with 'dynamic databases'? Not quite sure what that means other than the code that *we* are writing. 

I'm fine with splitting up this pr into an auth gate pr and later on a top level interface pr once we've ironed it that idea more.

client/packages/react/src/InstantReactAbstractDatabase.tsx
```diff
@@ -345,4 +369,64 @@ export default abstract class InstantReactAbstractDatabase<
   }> => {
     return this._core.queryOnce(query, opts);
   };
+
+  /**
+   * Only render children if the user is signed in.
+   * Optional `loading` prop will render if the user is loading.
+   *
```
stopachka, line 380: Nice! Do we want to link this to the auth docs?

client/packages/react/src/InstantReactAbstractDatabase.tsx
```diff
@@ -345,4 +369,64 @@ export default abstract class InstantReactAbstractDatabase<
...
+    loading?: ReactNode;
+    error?: ReactNode;
+  }> = ({ children, loading, error }) => {
+    const auth = this.useAuth();
+    if (auth.user) {
+      return <>{children}</>;
+    } else if (loading && auth.isLoading) {
+      return <>{loading}</>;
+    } else if (error && auth.error) {
+      return <>{error}</>;
```
stopachka: Should we have a default `error` component?

client/packages/react/src/InstantReactAbstractDatabase.tsx
```diff
@@ -345,4 +369,64 @@ export default abstract class InstantReactAbstractDatabase<
...
+   *    <MyComponent />
+   *  </db.SignedIn>
+   */
+  SignedIn: React.FC<{
+    children: ReactNode;
+    loading?: ReactNode;
+    error?: ReactNode;
+  }> = ({ children, loading, error }) => {
+    const auth = this.useAuth();
+    if (auth.user) {
```
stopachka: Nit: I would re-write it like this: 

```
`if (isLoading) ...` 
`if (error) ...` 
`if (!user) return null` 
return <>{children}</> 
```

The idea being that all the previous states are a kind of `guard` for this main state.

client/packages/react/src/InstantReactAbstractDatabase.tsx
```diff
@@ -345,4 +369,64 @@ export default abstract class InstantReactAbstractDatabase<
...
+   *    <MyComponent />
+   *  </db.SignedOut>
+   */
+  SignedOut: React.FC<{
+    children: ReactNode;
+    loading?: ReactNode;
+    error?: ReactNode;
+  }> = ({ children, loading, error }) => {
+    const auth = this.useAuth();
+    if (!auth.isLoading && !auth.user && !auth.error) {
```
stopachka: Would re-structure guard style.

client/packages/react/src/InstantReactAbstractDatabase.tsx
```diff
@@ -345,4 +369,64 @@ export default abstract class InstantReactAbstractDatabase<
...
+      return <>{children}</>;
+    } else if (loading && auth.isLoading) {
```
stopachka: We may not need to check for `loading` too -- i.m.o null is renderable no?

stopachka (commented):
Niice! i.m.o just about ready to go. Let's include the docs updates?

client/packages/react/src/InstantReactAbstractDatabase.tsx
```diff
@@ -261,6 +259,32 @@ export default abstract class InstantReactAbstractDatabase<
...
+   *    const user = db.useUser()
+   *    return <div>Logged in as: {user.email}</div>
+   *  }
+   *
+   *  <db.SignedIn>
+   *    <UserDisplay />
+   *  </db.SignedIn>
+   *
+   */
+  useUser = () => {
```
stopachka: one nit: let's type this output with `User`? 

Reasoning: it can be dangerous to let typescript infer output types. For example, if we didn't explicitly type the output, and the internals somehow changed to not return `User`, we wouldn't get an error, but it would be a breaking change

client/packages/react/src/InstantReactAbstractDatabase.tsx
```diff
@@ -345,4 +369,74 @@ export default abstract class InstantReactAbstractDatabase<
...
+      return <>{loading}</>;
+    }
+    if (auth.error) {
+      if (error) {
+        return <>{error}</>;
+      }
+      return <>{auth.error.message}</>;
+    }
+    return null;
+  };
```
stopachka: This is more what I meant: 

```suggestion
  SignedIn: React.FC<{
    children: ReactNode;
    loading?: ReactNode;
    error?: ReactNode;
  }> = ({ children, loading, error }) => {
    const auth = this.useAuth();
    if (auth.isLoading) { 
      return loading || null;
    }
    if (auth.error) { 
      return error || <>{auth.error.message}</>;
    }
    if (!auth.user) return null; 

    return <>{children}</>
  };
```

client/packages/react/src/InstantReactAbstractDatabase.tsx
```diff
@@ -345,4 +369,74 @@ export default abstract class InstantReactAbstractDatabase<
...
+   *  <db.SignedOut loading={<div>Loading...</div>}>
+   *    <MyComponent />
+   *  </db.SignedOut>
+   */
+  SignedOut: React.FC<{
+    children: ReactNode;
+    loading?: ReactNode;
+    error?: ReactNode;
+  }> = ({ children, loading, error }) => {
+    const auth = this.useAuth();
```
stopachka, line 425: (Would do same style as above.)

client/packages/react/src/InstantReactAbstractDatabase.tsx
```diff
@@ -261,6 +259,32 @@ export default abstract class InstantReactAbstractDatabase<
...
+   *
+   *  <db.SignedIn>
+   *    <UserDisplay />
+   *  </db.SignedIn>
+   *
+   */
+  useUser = () => {
+    const { user } = this.useAuth();
+    if (!user) {
+      throw new Error('useUser must be used within an auth-protected route');
```
stopachka: Should we throw an `InstantError`?

client/www/pages/docs/auth/frontend.md
```diff
@@ -0,0 +1,87 @@
+---
+title: Frontend
+description: Accessing information about auth on the client.
```
stopachka: i.m.o we don't really need this page. Most users can figure this out by choosing one of the auth methods. 

i.m.o in each auth method we already have a section that shows useAuth, SignedIn, etc etc. Let's update those sections to use `useUser`, db.SignedIn, etc?

client/www/pages/docs/auth/magic-codes.md
```diff
@@ -177,27 +177,6 @@ on the backend during permission checks.
 
 On the client, `useAuth` will set `isLoading` to `false` and populate `user` -- huzzah!
 
-## useAuth
```
stopachka, line 180: i.m.o we can include what we can replace this with something like `SignedIn and SignedOut components`

client/www/pages/docs/auth/magic-codes.md
```diff
@@ -219,23 +198,3 @@ db.auth.signInWithMagicCode({ email: sentEmail, code }).catch((err) => {
 ```
 
 You can then use `auth.signInWithMagicCode` to authenticate the user with the magic code they provided.
-
-## Sign out
```
stopachka, line 223: I would keep these here.

client/www/pages/docs/auth/magic-codes.md
```diff
@@ -219,23 +198,3 @@ db.auth.signInWithMagicCode({ email: sentEmail, code }).catch((err) => {
...
-
-```javascript
-db.auth.signOut();
-```
-
-Use `auth.signOut` from the client to invalidate the user's refresh token and
-sign them out.You can also use the admin SDK to sign out the user [from the
-server](/docs/backend#sign-out).
-
-## Get auth
```
stopachka, line 233: Maybe `useUser` can have a section here

client/www/pages/docs/auth/frontend.md
```diff
@@ -0,0 +1,87 @@
+---
+title: Frontend
```
nezaj: Instead of a separate page for this what if we put this as part of the content for the main Auth page?

financialvice:
wtf this man is on fire

financialvice:
loving all of this

client/www/markdoc/tags.js
```diff
@@ -69,6 +69,9 @@ const tags = {
   },
   'nav-group': {
     render: NavGroup,
+    attributes: {
+      margin: { type: Boolean, default: true },
```
stopachka: nit: let's remove margin

client/www/pages/docs/auth.md
```diff
@@ -28,3 +68,36 @@ Instant comes with support for auth. We currently offer [magic codes](/docs/auth
             /%}
 
 {% /nav-group %}
+
+## Low Level API
+
```
stopachka, line 75: I would add something like: 

"Sometimes you need finer control over auth states. In those cases you can use some of the lower-level API."

client/www/pages/docs/auth/frontend.md
```diff
@@ -0,0 +1,89 @@
+---
```
stopachka: imo we can delete this?

client/www/pages/docs/auth/magic-codes.md
```diff
@@ -177,27 +177,6 @@ on the backend during permission checks.
 
 On the client, `useAuth` will set `isLoading` to `false` and populate `user` -- huzzah!
```
stopachka, line 173: We should update the cases in the docs where we use `db.useAuth`, in exchange for `db.useUser` and db.SignedIn et al?

client/www/pages/docs/auth/apple.md
```diff
@@ -265,35 +265,37 @@ const APP_ID = '__APP_ID__';
...
-        />
-      </View>
-    );
-  }
-  return <Login />;
+  return (
+    <>
+      <db.SignedIn
+        loading={
+          <View style={styles.container}>
```
stopachka: nit: I would remove loading.

PR #1464, opened 2025-08-01
[Sandbox] Type rn sandbox apps

nezaj:
Also noticed our react native sandbox apps are in JS. Figured it would be nicer to have these as typescript


dwwoelfel (approved):
LGTM!

PR #1466, opened 2025-08-02
Update using-llms.md: fix the not-found link

martin0258:
### Current behavior
After clicking the `whirlwind tutorial` link, it redirected to the home page `instantdb.com` as the link is not found.
image
image

### Expected behavior and the fix
Use the right link to the tutorial and it should navigate to it.


stopachka (approved):
Thank you Martin!

PR #1467, opened 2025-08-03
Wait for a merge to main to publish packages

dwwoelfel:
This fixes the problem of needing to make a new commit when we publish the package versions.

Instead of storing the version in the `version.md` file and updating the version.js files in the deploy script, we have a new `@instantdb/version` package that stores the version. 

I tried to come up with other ways to propagate the version, but everything I could find would force us to modify how we build.

Now when you want to publish packages, you just need to update `client/packages/version/src/version.ts` with a new version. 

I would have liked to have put the version in a json file, but I couldn't find a way to make it work with both commonjs and esm.

When you merge to main, we'll check for changes to the `version.ts` publish the package.

The only change we'll see is that the version that we see from the dashboard will be the current version (`v0.20.16`) instead of e.g. `v0.20.16-dev`.


client/version.md
```diff
@@ -1 +1,2 @@
-v0.20.16
+Don't edit the version here any more.
```
stopachka, line 1: QQ: is the idea that we'll delete this later?

stopachka (approved):
So cool!!

client/version.md
dwwoelfel: Yep!

PR #1469, opened 2025-08-04
[www] Sandbox output virtualization

ayoung19:
Stacked on #1468.


ayoung19:
Actually this improves performance but not enough because for large results the real performance hog is the rendering of `Json`. Closing for now.

PR #1472, opened 2025-08-04
Fix mcp deploy

dwwoelfel:
The mcp deploy was failing because the docker image was missing the new version package.


nezaj (approved):
Nice!

PR #1473, opened 2025-08-04
Adds a count-min-sketch implementation

dwwoelfel:
I'm pulling apart this PR into smaller pieces to make it easier to review: https://github.com/instantdb/instant/pull/1458

This PR adds a [count-min-sketch](https://en.wikipedia.org/wiki/Count%E2%80%93min_sketch) implementation with some tests.

It uses [xxhash](https://xxhash.com/) for hashing the inputs, which seems to be a popular choice because it is fast and it's easy to create a bunch of pairwise-independent hashes by incrementing the seed.

When we put items in the sketch, we use the `checked-data-type` to let us normalize dates. If we didn't do that then we would return the wrong result if a user put 0 in the db, but then queried for "1970-01-01T00:00:00Z". When you add a checked-data-type to an existing column, we'll see that change in the wal listener. We'll remove the old value from the sketch and add the new value with the new type.

### DIFF
--- a/server/src/instant/db/attr_sketch.clj
+++ b/server/src/instant/db/attr_sketch.clj
@@ -1,0 +1,141 @@
+(ns instant.db.attr-sketch
+  (:require
+   [clojure.pprint]
+   [instant.db.model.triple :as triple])
+  (:import
+   (java.nio ByteBuffer)
+   (net.openhft.hashing LongHashFunction)))
+
+(set! *warn-on-reflection* true)
+
+(defrecord Sketch [width
+                   depth
+                   bins
+                   total
+                   total-not-binned])
+
+;; Update printer so that it doesn't print a ton numbers for the bins
+(defmethod clojure.pprint/simple-dispatch Sketch [x]
+  (let [bin-display (symbol (format "<bins %d>" (count (:bins x))))]
+    (#'clojure.pprint/pprint-map (into {} (assoc x :bins bin-display)))))
+
+(def ln2 (Math/log 2))
+
+(defn make-sketch
+  ([] (make-sketch {}))
+  ([{:keys [confidence error-rate]
+     :or {confidence 0.998
+          error-rate 0.001}}]
+   (let [width (int (Math/ceil (/ 2 error-rate)))
+         depth (int (Math/ceil (/ (* -1 (Math/log (- 1 confidence)))
+                                  ln2)))]
+     (map->Sketch {:width width
+                   :depth depth
+                   :bins (vec (repeat (* width depth) 0))
+                   :total 0
+                   :total-not-binned 0}))))
+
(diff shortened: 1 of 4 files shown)


server/src/instant/db/attr_sketch.clj
```diff
@@ -0,0 +1,143 @@
...
+(set! *warn-on-reflection* true)
+
+(defrecord Sketch [width
+                   depth
+                   bins
+                   total
+                   total-not-binned])
+
+;; Update printer so that it doesn't print a ton numbers for the bins
+(defmethod clojure.pprint/simple-dispatch Sketch [x]
```
stopachka, line 17: Oo heck yeah!

stopachka (approved):
🤩🤩🤩 biyutiful

server/test/instant/db/attr_sketch_test.clj
```diff
@@ -0,0 +1,86 @@
...
+    (is (every? zero? (:bins sketch)))
+
+    (is (= 1 (:total (cms/add sketch nil "val"))))))
+
+(deftest ignores-values-we-cant-match-on
+  (let [sketch (-> (cms/make-sketch)
+                   (cms/add nil 1)
+                   (cms/add nil {:json :data}))]
+    (is (= 2 (:total sketch)))
+    (is (thrown-with-msg? AssertionError
```
stopachka: curious q: do we want to throw AssertionError? 

(I remember that AssertionError doesn't get caught by the standard `catch Exception e`.)

server/src/instant/db/attr_sketch.clj
```diff
@@ -0,0 +1,143 @@
...
+      ;; the sketch. We could use bytes, but this will still work if the
+      ;; uuid gets passed to us as a string somehow
+      java.util.UUID [:string (str x)]
+      nil)))
+
+(defn hash-val [^Long seed ^Long hash-idx data-type val]
+  (let [xx (LongHashFunction/xx3 (+ seed hash-idx))]
+    (case data-type
+      :long (.hashLong xx val)
+      :double (.hashBytes xx (.. (ByteBuffer/allocate 8)
```
nezaj: Nit. Claude suggests doing `:double (.hashLong xx (Double/doubleToRawLongBits val))`

The idea is that `ByteBuffer.allocate(8)` takes about ~20-30ns (heap allocation) whereas `Double.doubleToRawLongBits()` takes 1-2ns and no allocations

dwwoelfel: Cool!

server/test/instant/db/attr_sketch_test.clj
dwwoelfel: Updated to a regular exception

PR #1476, opened 2025-08-05
[Explorer] Support mulit-line string inputs

nezaj:
Saw this requested a few times and thought it would be straightforward to add!

**Before**

https://github.com/user-attachments/assets/ec66ae4a-b1ab-43e7-9af7-ccf51346d35d

**After**

https://github.com/user-attachments/assets/dbc5c52e-e698-490a-9be1-d9dcea289010

### DIFF
--- a/client/www/components/dash/explorer/EditRowDialog.tsx
+++ b/client/www/components/dash/explorer/EditRowDialog.tsx
@@ -60,8 +60,18 @@ function ResizingTextArea({
   const textareaRef = useRef<HTMLTextAreaElement>(null);
 
   const autoResize = (element: HTMLTextAreaElement) => {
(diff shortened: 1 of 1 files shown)


client/www/components/dash/explorer/EditRowDialog.tsx
```diff
@@ -788,6 +788,15 @@ export function EditRowDialog({
     });
   };
 
+  // Auto-resize textareas when dialog opens
+  useLayoutEffect(() => {
```
stopachka: Maybe we extract this code out into a `ResizingTextArea`? or something like that? 

(For this kind of extending text area usually the way to go use to use contenteditable inputs, etc. But heck if this works well let's just use it!)

PR #1477, opened 2025-08-05
Adds support for postgres copy out

dwwoelfel:
This adds support for postgres copy. We'll use this to populate the attr sketches when we first create the new logical replication slot for the aggregator.

It also upgrades cheshire and jackson to fix a bug with parsing bytes. I looked through the change history and didn't see any breaking changes in the documentation that would affect us.

When we create the replication slot, postgres will give us a snapshot identifier. We can use that snapshot identifier to call `set transaction snapshot :snapshot-id` and get a view into the database as of the time that the snapshot was created. Then we can call `copy` on that connection to fetch all of the triples and build up the attr sketches. Once the sketches are saved, we can start listening to changes on the new slot. This will give us a consistent view of the database without any missing data in the sketches.

In the future, we may also be able to use copy for things like exports, but I've found it to be a bit flaky. You can't used it with the conn-pool, because if it's not closed properly (if there is an error, there doesn't seem to be a way to close it properly without closing the underlying connection), then it will cause subsequent queries to fail.

### DIFF
--- a/server/src/instant/jdbc/copy.clj
+++ b/server/src/instant/jdbc/copy.clj
@@ -1,0 +1,181 @@
+(ns instant.jdbc.copy
+  (:require
+   [instant.util.json :as json])
+  (:import
+   (java.nio ByteBuffer)
+   (java.time Duration Instant LocalDate ZoneOffset)
+   (java.util UUID)
+   (org.postgresql.jdbc PgConnection)))
+
+;; ------------------
+;; Decoders for types
+
+;; binary decoders that consume the input
+
+(defn- advance-buf [^ByteBuffer bb ^Integer len]
+  (.position bb (int (+ (.position bb) len))))
+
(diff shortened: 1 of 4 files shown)


server/src/instant/jdbc/copy.clj
```diff
@@ -0,0 +1,187 @@
...
+
+(defn- bin-decode-jsonb [^ByteBuffer bb ^Integer byte-len]
+  ;; jsonb starts with a version byte, then the data comes after
+  (let [version (.get bb)]
+    (when (not= 1 version)
+      (throw (ex-info "Invalid version byte for jsonb" {:version version}))))
+  (bin-decode-json bb (dec byte-len)))
+
+(def pg-diff (Duration/between Instant/EPOCH
+                               (.atStartOfDay (LocalDate/of 2000 1 1)
```
stopachka, line 41: huh! Interesting

PR #1478, opened 2025-08-06
add @types/react to react-native-playground

drew-harris:
Fixes typescript errors when trying to use react in the react-native playground package.


client/packages/core/src/package.json
```diff
@@ -0,0 +1,3 @@
+{
+  "type": "module"
```
stopachka: nit: I _think_ we need to delete this

PR #1479, opened 2025-08-06
Validate pagination params are only used at top level

nezaj:
What it says on the tin!

For the query

```typescript
const todosQuery = {
  todos: {
    owner: { $: { limit: 1 } },
  },
} satisfies InstaQLParams<AppSchema>;
```

CleanShot 2025-08-06 at 14 13 23@2x

### DIFF
--- a/client/packages/core/src/queryValidation.ts
+++ b/client/packages/core/src/queryValidation.ts
@@ -391,7 +391,7 @@ const validateDollarObject = (
   for (const param of paginationParams) {
     if (dollarObj[param] !== undefined && !isTopLevel) {
       throw new QueryValidationError(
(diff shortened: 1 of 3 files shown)


client/packages/core/src/queryValidation.ts
```diff
@@ -399,6 +418,7 @@ const validateEntityInQuery = (
   entityName: string,
   schema: IContainEntitiesAndLinks<any, any>,
   path: string,
+  isTopLevel: boolean = false,
```
stopachka: It might be nicer for this to be a `depth` field, since for the same cost we'll get more info

nezaj: Updated!

drew-harris (approved):
LGTM

PR #1481, opened 2025-08-07
[client] validate appId

stopachka:
**Problem**

If you gave an invalid appId, we would pass this all the way along to the server. The error takes a bit to propagate, so it's harder for humans and LLMs to see it. 

**Solution** 

We now immediately throw. 

**Note** 

I moved this lower down in Reactor, under the `isClient` check. 

This way, if the user is running a build or a test for example, and they don't intend to actually run db, they can still pass in an empty appId. Our validations will never get called. 


nezaj (approved):
LGTM!

PR #1485, opened 2025-08-07
Aggregator for attr sketches

dwwoelfel:
Adds a new wal slot and processor for populating the attr sketches.

The aggregator is a lot like our invalidator, except that there is a single replication slot that never gets deleted. Instead of each machine reading from their own replication slot, the machines race to acquire the slot and only the winner receives changes from the slot.

When one of the instances acquires the slot, it will listen for changes to triples and periodically flush those changes to the attr sketches in the database. When it updates the sketches, it also updates a `wal_aggregator_status` record with the latest lsn. This ensures that we never miss an update and we never write the same data twice.

The first machine to start the aggregator will create the replication slot. When you create the slot, you can get a snapshot that will give you a view into the database at the time the slot was created. We use that snapshot to fetch all triples from the database (using the new [copy support](https://github.com/instantdb/instant/pull/1477)) and initialize the attr sketches. Then we start processing the replication stream, catching up on any data we missed since the snapshot was created.

The sketch bins are compressed with zstd before we store them in the database. I tried both zstd and lz4. In my testing, zstd was faster and compressed better. The compressed size for bins of an empty sketch of the default size is only 70 bytes.

Minor changes:
1. Adds `-Dclojure.core.async.go-checking=true` flag in dev/test to help us catch improper use of blocking operations in a go block
2. Adds support for storing `nil` in the sketches (we store a hash of the null byte. Postgres won't store in a jsonb column, so there's no change of a collision)
3. Tells wal2json which tables we care about. It will filter out updates from other tables and save us some i/o.
4. Adds a new `chunked-chan` that will batch items and deliver them every N ms or M items.

In the future, we may want to replace one invalidator slot on each machine with a single aggregator slot on one machine. The aggregator can distribute the changes to the rest of the instances through hazelcast, sqs/sns, nats, or something else.

We can also use the aggregator to collect database size usage efficiently.

## Deploy plan
We need to do this in 2 steps because our old slot-cleaner would delete the aggregator slot.

1. [ ] Deploy these changes and wait for them to fully deploy
2. [ ] Push a new PR that enables the aggregator

Next steps (future work):
1. Replace the count queries for pg_hint_plan with lookups in the attr sketches and release pg_hint_plan
2. Build an attr sketch that grows as the data grows
3. Use the aggregator to collect db_size for each attr.

### DIFF
--- a/server/src/instant/db/attr_sketch.clj
+++ b/server/src/instant/db/attr_sketch.clj
@@ -86,7 +86,8 @@
         clojure.lang.LazySeq nil ;; top-level arrays
         (tracer/with-span! {:name "attr-sketch/unknown-type"
                             :attributes {:warning true
-                                         :value x}}
+                                         :value x
+                                         :type (type x)}}
           nil)))))
 
 (defn hash-val [^Long seed ^Long hash-idx data-type val]


server/Makefile
```diff
@@ -58,11 +58,6 @@ restart:
 	@echo "Restarting beanstalk..."
 	aws elasticbeanstalk restart-app-server --environment-name Instant-prod-env
 
-dev-with-prod-db:
```
dwwoelfel, line 61: Removing this because you could mess up the sketches in prod if you run with the prod db locally.

server/resources/migrations/77_add_sketch.up.sql
```diff
@@ -0,0 +1,67 @@
...
+  created_at timestamp with time zone not null default now(),
+  updated_at timestamp with time zone not null default now()
+);
+
+create trigger update_updated_at_trigger
+before update on wal_aggregator_status
+for each row
+execute function update_updated_at_column();
+
+alter table triples replica identity full;
```
dwwoelfel, line 45: Gives us the full row in the `identity` column so that we can see the old value in the wal listener.

server/resources/migrations/77_add_sketch.up.sql
```diff
@@ -0,0 +1,45 @@
...
+create index on attr_sketches (attr_id);
+
+create trigger update_updated_at_trigger
+before update on attr_sketches
+for each row
+execute function update_updated_at_column();
+
+-- This acts as a singleton table where we keep track of which lsn
+-- we've completed processing. The updates to the attr_sketches will
+-- happen in the same transaction where we update the wal_aggregator
```
stopachka, line 27: Nice idea!

server/resources/migrations/77_add_sketch.up.sql
stopachka: Wow! Very useful. I think this will open up much nicer topics down the line

server/src/instant/db/attr_sketch.clj
```diff
@@ -143,3 +167,282 @@
...
+
+(defn compress-bins
+  "Returns the bins as bytes compressed with zstd."
+  ^bytes [sketch]
+  (let [bb (ByteBuffer/allocate (* (:width sketch)
+                                   (:depth sketch)
+                                   Long/BYTES))]
+    (doseq [l (:bins sketch)]
+      (.putLong bb l))
+    (Zstd/compress (.array bb))))
```
stopachka, line 232: 🔥

server/src/instant/db/attr_sketch.clj
```diff
@@ -143,3 +167,282 @@
...
+  (let [dst (byte-array (* width depth Long/BYTES))]
+    (Zstd/decompress dst compressed-bin-bytes)
+    (bytes->longs dst)))
+
+;;; Queries
+
+(defn qualify-col [ns col]
+  (keyword (format "%s.%s" (name ns) (name col))))
+
+(defn qualify-cols [ns cols]
```
stopachka, line 251: Cool idea!

server/src/instant/db/attr_sketch.clj
```diff
@@ -143,3 +167,282 @@
...
+                    :set {:lsn :?lsn
+                          :process-id :?process-id}
+                    :where [:and
+                            [:= :lsn :?previous-lsn]
+                            [:= :slot-name :?slot-name]]
+                    :returning :*}]
+                  [:check-empty
+                   {:select [[[:case [[:exists {:select :1 :from :update-wal-aggregator-status}]]
+                               true
+                               :else [:raise_exception_message [:inline "lsn is not what we expected, another machine may have stolen the replication slot"]]]
```
stopachka, line 424: nice!

server/src/instant/db/attr_sketch.clj
```diff
@@ -143,3 +167,282 @@
...
+
+(defn insert-initial-sketches!
+  "Only should be used during bootstrapping when we initially
+   add attr sketches to the database."
+  [conn {:keys [sketches
+                lsn]}]
+  (let [params (reduce (fn [acc {:keys [app-id attr-id sketch]}]
+                         (let [{:keys [width depth total total-not-binned]} sketch]
+                           (-> acc
+                               (update :app-id conj app-id)
```
stopachka, line 446: This parallel array took me for a spin. Interesting idea!

server/src/instant/reactive/aggregator.clj
```diff
@@ -0,0 +1,462 @@
...
+        collect-changes (fn [batch]
+                          (reduce (fn [acc row]
+                                    (let [key (select-keys row [:app-id :attr-id])
+                                          record (select-keys row [:value :checked-data-type])]
+                                      (update-in acc [key record] (fnil inc 0))))
+                                  {}
+                                  batch))
+        {sketches-by-key :sketches}
+        (transduce
+          ;; Take 1M triples at a time
```
stopachka: 🦁

server/src/instant/util/async.clj
```diff
@@ -263,3 +262,65 @@
...
+   put when the channel is closed.
+
+   This only works if you remember to close the close-signal channel
+   when you close the channel itself."
+  [close-signal-ch ch val]
+  `(a/alt! [[~ch ~val]] ([res#] res#)
+           ~close-signal-ch false))
+
+(defn chunked-chan
+  "Takes a `flush-ms` and a `max-size`. Returns an `in` and `out` chan.
```
stopachka, line 278: Biytuiful!

server/test/instant/reactive/aggregator_test.clj
```diff
@@ -0,0 +1,236 @@
+(ns instant.reactive.aggregator-test
```
stopachka, line 1: So cool that we have tests for this!

server/src/instant/jdbc/wal.clj
```diff
@@ -195,14 +254,19 @@
...
+                 :aggregator [:triples])
+        add-tables (clojure.string/join "," (map #(str "public." (name %)) tables))
+        builder (-> replication-conn
                     (.getReplicationAPI)
                     (.replicationStream)
                     (.logical)
                     (.withSlotOption "include-lsn" true)
                     (.withSlotOption "format-version" version)
-                    (.withStartPosition (LogSequenceNumber/valueOf start-lsn))
+                    (.withSlotOption "add-tables" add-tables)
```
stopachka, line 271: Niice!

stopachka (approved):
Learned a lot reading this PR. Awesome work @dwwoelfel !

PR #1487, opened 2025-08-08
[sm] a few improvements to db.SignedIn

stopachka:
- We realized that loading and error states didn't make sense in these components
- Updated docstrings
@drew-harris @nezaj @dwwoelfel @tonsky


nezaj (approved):
LGTM!

PR #1490, opened 2025-08-08
Tweak llm rules

nezaj:
Did a pass on the live tutorial and noticed a few fixes

* The frontmatter was off for cursor and windsurf
* Updated instructions for cursor to specify `.mdc` format
* Used `<div></div>` instead of `<></>` in App component.
* Fixed SSR issue for `intern/llm-example`


dwwoelfel (approved):
LGTM!

PR #1492, opened 2025-08-08
[RULES] Tweak seed rules

nezaj:
Made a few more tweaks. To address these issues I've been seeing

* Sometimes `useProfile` is used outside of `db.SignedIn`
* Seeding data on the client can lead to infinite loops due to permission failures.
* Standardize on using `CRITICAL` instead of `IMPORTANT` for hints

### DIFF
--- a/client/www/lib/intern/llm-example/llm-rules-template.md
+++ b/client/www/lib/intern/llm-example/llm-rules-template.md
@@ -76,10 +76,16 @@ and react native. Instant also offers a javascript admin SDK that can be used on
 the backend.
 
 If you want to use Instant with react you should only use `@instantdb/react`. For react-native you should
-only use `@instantdb/react-native`. For the admin SDK you should only use
+only use `@instantdb/react-native`. For scripts or server environments you should only use the admin SDK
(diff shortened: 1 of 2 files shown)


client/www/lib/intern/llm-example/llm-rules-template.md
```diff
@@ -76,10 +76,16 @@ and react native. Instant also offers a javascript admin SDK that can be used on
 the backend.
 
 If you want to use Instant with react you should only use `@instantdb/react`. For react-native you should
-only use `@instantdb/react-native`. For the admin SDK you should only use
+only use `@instantdb/react-native`. For scripts or server environments you should only use the admin SDK
 `@instantdb/admin`. For other client-side frameworks or vanilla js you should only use `@instantdb/core`
 
-You cannot use Instant on the backend outside of the admin SDK at the moment.
+CRITICAL: To use the admin SDK you MUST get an admin token for the app. You can
+get the admin token with the MCP tool via `create-app` or `get-app`. The admin
```
stopachka: Aside, I _think_ get-app may not return the admin-token?

nezaj: Good catch!

PR #1498, opened 2025-08-10
fix cursor link

stopachka:
looks like cursor added a `/en` to their install-mcp link 


nezaj (approved):
Ship!

PR #1502, opened 2025-08-11
fix lookup strings being tested as uuids

drew-harris:
This PR fixes the issue of lookup attributes being tested as UUIDs and throwing errors for valid transactions

### DIFF
--- a/client/packages/core/__tests__/src/transactionValidation.test.ts
+++ b/client/packages/core/__tests__/src/transactionValidation.test.ts
@@ -1,6 +1,6 @@
 import { i } from '../../src/schema';
 import { validateTransactions } from '../../src/transactionValidation.ts';
-import { tx as originalTx, TxChunk } from '../../src/instatx.ts';
+import { lookup, tx as originalTx, TxChunk } from '../../src/instatx.ts';
 import id from '../../src/utils/uuid.ts';
 import { expect, test } from 'vitest';
 import { InstantSchemaDef } from '../../src';
@@ -385,4 +385,11 @@ test('validates UUID format for entity IDs', () => {
   // Test for links
   beValid(tx.users[validUuid].link({ posts: id() }));
   beWrong(tx.users[validUuid].link({ posts: 'not-a-uuid' }));
+});
+
+test('allows lookup values in square bracket', () => {
+  beValid(
+    tx.users[lookup('email', 'john@example.net')].update({ name: 'John' }),
+  );
+  beValid(tx.users[lookup('email', 'john@example.net')].link({ posts: id() }));
 });
--- a/client/packages/core/__tests__/src/transactionValidation.test.ts
+++ b/client/packages/core/__tests__/src/transactionValidation.test.ts
@@ -393,3 +393,8 @@ test('allows lookup values in square bracket', () => {
   );
   beValid(tx.users[lookup('email', 'john@example.net')].link({ posts: id() }));
 });
+
+test('allows lookup values in link', () => {
+  beValid(tx.users[id()].link({ posts: lookup('title', 'Hello') }));
+  beWrong(tx.users[id()].link({ posts: 'non lookup or uuid' }));
+});
--- a/client/packages/core/src/queryValidation.ts
+++ b/client/packages/core/src/queryValidation.ts
@@ -3,7 +3,7 @@ import {
   DataAttrDef,
   ValueTypes,
 } from './schemaTypes.ts';
-import { validateUUIDOrLink } from './transactionValidation.ts';
+import { isValidEntityId } from './transactionValidation.ts';
 
 export class QueryValidationError extends Error {
   constructor(message: string, path?: string) {
@@ -225,7 +225,7 @@ const validateDotNotationAttribute = (
 
   // Handle 'id' field specially - every entity has an id field
   if (finalAttrName === 'id') {
-    if (!validateUUIDOrLink(value)) {
+    if (!isValidEntityId(value)) {
       throw new QueryValidationError(
         `Invalid value for id field in entity '${currentEntityName}'. Expected a UUID, but received: ${value}`,
         path,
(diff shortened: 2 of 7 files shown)


client/packages/core/src/transactionValidation.ts
```diff
@@ -1,7 +1,17 @@
-import { TransactionChunk, Op } from './instatx.ts';
+import { TransactionChunk, Op, isLookup } from './instatx.ts';
 import { IContainEntitiesAndLinks, DataAttrDef } from './schemaTypes.ts';
 import { validate as validateUUID } from 'uuid';
 
+export const validateUUIDOrLink = (value: unknown): boolean => {
```
nezaj: Curious, why call `OrLink`?

Maybe we can just call this isValidEntityId?

nezaj (approved):
LGTM! Just a couple asks and a question

Could we add a transaction validation test for lookups in core and also to `sandbox/admin-sdk-express/src/index.ts`

client/packages/core/src/transactionValidation.ts
```diff
@@ -141,7 +151,7 @@ const validateOp = (
 
   // _id should be a uuid
   if (!Array.isArray(_id)) {
-    const isUuid = validateUUID(_id);
+    const isUuid = isValidEntityId(_id);
```
nezaj: I think for this one we should _only_ check that this is a uuid. We probably don't want to check it's a lookup here?

client/packages/core/src/transactionValidation.ts
```diff
@@ -104,15 +114,15 @@ const validateLinkOperation = (
...
+          if (!isValidEntityId(uuid)) {
             throw new TransactionValidationError(
               `Invalid UUID in link '${linkName}' for entity '${entityName}'. Expected a UUID, but received: ${uuid}`,
             );
           }
         }
       } else {
         // Handle single UUID
-        if (!validateUUID(linkValue)) {
+        if (typeof linkValue === 'string' && !isValidEntityId(linkValue)) {
```
nezaj: Nit. Do we need the typeof check here?

client/sandbox/admin-sdk-express/src/index.ts
```diff
@@ -74,6 +74,27 @@ async function testTransact() {
   console.log(JSON.stringify(res, null, 2));
 }
 
+async function testTransactWithLookup() {
```
nezaj: Curious if you run this does it actually pass? I'm getting the following error

```
InstantAPIError: Validation failed for lookup: title is not a unique attribute on todos
 ```
 
We don't have these tests automatically running so I had to add `testTransactWithLookup()` and then run `pnpm run dev`

Comment (drew-harris) [reply client/sandbox/admin-sdk-express/src/index.ts]: Didn't notice that we had a schema defined in the other file, moved this function into that file and made sure it works with the schema, heres a script i ran to seed the data
https://gist.github.com/drew-harris/15d359fd0abd9ad0d856a38081f7f324

File: client/packages/core/src/transactionValidation.ts [hunk 5]
```diff
@@ -104,15 +114,15 @@ const validateLinkOperation = (
       if (Array.isArray(linkValue)) {
         // Handle array of UUIDs
         for (const uuid of linkValue) {
-          if (!validateUUID(uuid)) {
+          if (!isValidEntityId(uuid)) {
             throw new TransactionValidationError(
               `Invalid UUID in link '${linkName}' for entity '${entityName}'. Expected a UUID, but received: ${uuid}`,
```
stopachka: Should this say 'Invalid entity id...expected a UUID or a lookup?'

client/packages/core/__tests__/src/transactionValidation.test.ts
```diff
@@ -386,3 +386,10 @@ test('validates UUID format for entity IDs', () => {
   beValid(tx.users[validUuid].link({ posts: id() }));
   beWrong(tx.users[validUuid].link({ posts: 'not-a-uuid' }));
 });
+
+test('allows lookup values in square bracket', () => {
```
stopachka, line 390: Let's add a test for lookups in link too?

client/sandbox/admin-sdk-express/src/with-schema.ts
```diff
@@ -64,6 +64,8 @@ process.on('unhandledRejection', (reason, promise) => {
   console.error('Unhandled Rejection at:', promise, 'reason:', reason);
 });
 
+testTransactWithLookup();
```
nezaj: The convention in this file has been to actually leave the tests commented out. If you look below you'll see a block for tests (e.g. `testTransact`)

client/sandbox/admin-sdk-express/src/with-schema.ts
```diff
@@ -95,6 +97,26 @@ async function testTransact() {
   console.log(JSON.stringify(res, null, 2));
 }
 
+async function testTransactWithLookup() {
```
nezaj, line 98: I'm curious does this test actually succeed? I don't think it will.

client/sandbox/admin-sdk-express/src/with-schema.ts
```diff
@@ -95,6 +97,26 @@ async function testTransact() {
...
+  const todoAId = id();
```
nezaj: Do we need `todoAId` and `todoBId` if we don't use them in this test?

client/sandbox/admin-sdk-express/src/with-schema.ts
drew-harris: It succeeds for me, given that the preconditions are there, should I add the "seed" code into the test?

drew-harris:
force pushed to fix the test being uncommented out

client/packages/core/src/queryValidation.ts
```diff
@@ -3,8 +3,7 @@ import {
   DataAttrDef,
   ValueTypes,
 } from './schemaTypes.ts';
-
-import { validate as validateUUID } from 'uuid';
+import { isValidEntityId } from './transactionValidation.ts';
```
nezaj: Ah queryValidation shouldn't use `isValidEntityId`, we should just use `validateUUID` for this file

PR #1503, opened 2025-08-11
Better date parsing

dwwoelfel:
This catches more of the edge cases in our date parsing to make it match more closely with postgres.

Handles every edge case that's currently in the database on the server-side and returns the same value as postgres. The client is a little harder to test, but it passes all of the cases that we've identified in the test cases.

Adds support for "special" keywords on the server, like "today" and "tomorrow". I haven't implemented those on the client because we don't have a good way to agree on what `tomorrow` is between the server and client.

Updates our config in dev to use `UTC` as the default timezone with `-Duser.timezone=UTC` . That will keep us more in line with production. JDBC automatically sets the timezone on the postgres connection using that flag.


client/packages/core/src/utils/dates.ts
```diff
@@ -8,9 +10,22 @@ function localDateTimeStrToInstant(s) {
   return new Date(s + 'Z');
 }
 
+const localDateStrRe = /^(\d+)[\./-](\d+)[\./-](\d+)$/;
+
 function localDateStrToInstant(s) {
-  // Parse date and set to start of day in UTC
-  return new Date(s + 'T00:00:00Z');
+  const [match, part1, part2, part3] = s.match(localDateStrRe);
```
nezaj: I think if `s.match(...)` returns null then this destructuring will break

client/packages/core/src/utils/dates.ts
```diff
@@ -101,17 +116,49 @@ function usDateTimeStrToInstant(s) {
...
   zenecaDateStrToInstant,
   dowMonDayYearStrToInstant,
   usDateTimeStrToInstant,
   rfc1123ToInstant,
-  localDateStrToInstant,
   localDateTimeStrToInstant,
-  zonedDateTimeStrToInstant,
   iso8601IncompleteOffsetToInstant,
   offioDateStrToInstant,
+  zonedDateTimeStrToInstant,
```
nezaj, line 161: Claude suggested we put these on top since they're more likely to validate first

```
zonedDateTimeStrToInstant,
localDateTimeStrToInstant,
localDateStrToInstant,
```

client/packages/core/src/utils/dates.ts
dwwoelfel: It's kind of ok, because these are designed to throw, but I'll handle the case.

nezaj (approved):
LGTM w/ two minor comments!

client/packages/core/src/utils/dates.ts
nezaj: Ah yea makes sense!

client/packages/core/src/utils/dates.ts
dwwoelfel: We want those to come last because they'll parse things like `04/14/2024` in the current timezone, and we want to force them to parse in UTC to match what the backend will do.

PR #1504, opened 2025-08-11
[Sandbox] Infinite transact bug

nezaj:
In this sandbox example we try to create todos when none exist (e.g seeding data on the client) but the permissions prevent us from creating them, which causes an infinite loop of trying to create todos.

I've seen LLMs make this bug before. Wanted to put up a repro for this so we can think on how to address it.

As an aside -- I thought it would be nice to start separating some of the sandboxes into a "bug" section. This way we can easily see what repros we have for known bugs. Now you'll be able to see these at `localhost:4000/bugs`

https://github.com/user-attachments/assets/b9fefb2e-5052-49d4-b89c-8ec84aff779d


dwwoelfel (approved):
LGTM!

PR #1506, opened 2025-08-12
[Rules] Remove permissions

nezaj:
Right now it's very easy for LLMs to write permissions that will cause errors. I think until we can reliably improve the permission experience we should remove them from the tutorial. This will create a significantly better getting started experience.

I used this example prompt from getmocha.com

> Create a Twitter-like social feed with mock tweets, profile pictures, and engagement metrics. Include features like likes, retweets, and comments. Focus on the main timeline view. It should be interactive as if the user is logged in and they can submit a tweet. Use mock data so we can get a feeling for the design with a non-empty feed.

Here's the video after one-shotting with permission vs without. The one with permissions effectively no-ops. At least the one without permissions lets you take actions and create new data

**With permissions**

https://github.com/user-attachments/assets/9b47c1ba-a161-413f-a634-fe90f5b91b74

**Without permissions**

https://github.com/user-attachments/assets/4672e516-6b49-46d1-8d7f-84b84aa260a1


stopachka (approved):
Maybe we can add a specific note to the LLM: once you are done, you can prompt the user about perms, and you can then read this file. 

But that can be for later. 

LGTM!

PR #1508, opened 2025-08-12
strict query validation - new types

drew-harris:
This PR creates a new ValidQuery type to validate query objects. All of the possible querying methods across the packages have been changed to use this new type.

### DIFF
--- a/client/packages/version/src/version.ts
+++ b/client/packages/version/src/version.ts
@@ -1,6 +1,6 @@
 // This is the shared version for all of the js packages
 // Update the version here and merge your code to main to
 // publish a new version of all of the packages to npm.
-const version = 'v0.20.28';
+const version = 'v0.21.0';
 
 export { version };


client/packages/core/__tests__/src/schema.test.ts
```diff
@@ -118,11 +119,11 @@ test('runs without exception', () => {
   queryResult?.users[0].friends[0]._friends[0].bio;
   queryResult?.users[0].posts[0].author?.junk;
 
-  // NOTE: @ts-expect-error has been replaced with #ts-expect-error
+  // NOTE: @ts-expect-error has been replaced with @ts-expect-error
```
stopachka: You can probably delete this?

client/packages/core/src/queryTypes.ts
```diff
@@ -438,6 +438,213 @@ type InstaQLOptions = {
...
+        : never
+      : Path extends keyof Schema['entities'][EntityName]['links']
+        ? DataAttrDef<string, false, false>
+        : never;
+
+type ValidQuery<
+  Q extends object,
+  S extends IContainEntitiesAndLinks<any, any>,
+> = S extends InstantUnknownSchemaDef
+  ? InstaQLParams<S>
```
stopachka, line 646: We can ship this version as is, but we may want to do some research about unifying things. It would suck to have to keep all the previous code to handle just this case.

client/packages/core/src/queryTypes.ts
```diff
@@ -438,6 +438,213 @@ type InstaQLOptions = {
...
+ * Extracts the attribute definition from a valid dot path.
+ * Given a dot path like "user.posts.title", this will resolve to the type of the "title" attribute.
+ * If the path is just a link name (e.g., "posts"), it defaults to the id field.
+ * Returns DataAttrDef or never
+ */
+type ExtractAttrFromDotPath<
+  Path extends string | number | symbol,
+  Schema extends IContainEntitiesAndLinks<any, any>,
+  EntityName extends keyof Schema['entities'],
+> = Path extends keyof Schema['entities'][EntityName]['attrs']
```
stopachka, line 626: I def worry about the kind of errors typescript seems to be returning when things go wrong. Let's see how the user feedback goes. We may want to look into improving error messages.

client/packages/core/__tests__/src/schema.test.ts
drew-harris: theres still a few cases that technically have invalid queries that pass but i will fix the comment to make it more clear

client/packages/core/src/queryTypes.ts
drew-harris: very much agreed

client/packages/core/__tests__/src/schema.test.ts
stopachka: (In this case I would keep the # on L122)

client/packages/core/src/queryTypes.ts
drew-harris: Agreed, especially some of the <Exactly> type of warnings say "not assignable to undefined".

client/packages/core/__tests__/src/schema.test.ts
drew-harris: was a search-and-replace mistake haha

drew-harris:
Thought about trying to create a generic type for people who want to wrap the query function but there are too many different circumstances to do it cleanly. 
Think we should deploy and assist people who are trying to do this on how to write the validation generic and take what we learned and put it in the docs.

PR #1511, opened 2025-08-13
Add tracer/with-new-trace-root

dwwoelfel:
Adds a nicer way to do `(binding [tracer/*span* nil]` to make it more obvious what we're doing.

Instead of:

```clojure
(binding [tracer/*span* nil]
  (do-thing))
```

You do:

```clojure
(tracer/with-new-span-root
  (do-thing))
```


stopachka (approved):
Good idea!

PR #1512, opened 2025-08-13
Add more debugging data

dwwoelfel:
Various debugging improvements:

1. Adds a tag to sql query in indexing-jobs
2. Moves the tracer around the try-catch in indexing jobs so that we can see what caused the errors in honeycomb
3. Adds `x-amzn-trace-id` and `x-amzn-cf-id` to `handle-receive`'s tracer attrs to make it easier to trace the connection back to the load balancer and cdn
4. Doesn't skip the websocket request in the tracer. The idea was that we shouldn't wrap a span around the websocket handler because it's long-running and all of its children would be attached to the trace. I was expecting to have to use [`tracer/with-new-trace-root`](https://github.com/instantdb/instant/pull/1511) on it, but it turns out not to be necessary. The handler completes immediately and just gets handed off to an internal undertow handler.


stopachka (approved):
Nice!

PR #1514, opened 2025-08-13
Link unlink permission checks

tonsky:
Design doc: https://www.dropbox.com/scl/fi/4sishcjbh9b219nvag0c8/Linkunlink-permissions.paper?rlkey=6qbov9fg0v0mzwai29voh5ch8&dl=0

Current API:

```
{:posts
 {:allow
  {:link   {"author" "..."}
   :unlink {"author" "..."}}}
 :users
 {:allow
  {:link   {"posts" "..."}
   :unlink {"posts" "..."}}}}
```

Notes:

- link/unlink permissions are specified _per attribute_. So post → users can have different rule than post → comment
  - attribute name is used instead of etype because two etypes can be connected by multiple different attributes
- link/unlink permissions are 2-sided. In the example above, rules are specified for both posts → users and for users → posts.
  - If both are specified, & (AND) is used
  - If one or both directions are missed, fallback is used, which is `update` check for forward + `view` check for reverse
    - So if we only specify `link` forward, we will check `link` forward and `view` reverse. If we only specify reverse, we will check `update` forward + `link` reverse
- The way transaction is written (`tx.posts[].link({users: ...})` or `tx.users[].link({posts: ...})` does not matter, they will raise the same 2-sided perm checks in both cases
- in addition to `data` and `newData`, we’ll introduce `linkedData` that will represent other side of the relation. In the example above, e.g. posts → author will have `linkedData` defined as if it was `newData.ref(author)`
- `delete-entity` produces unlink check for the other side (the side that is not deleted). On the deleted side, only `delete` check is run


stopachka:
API sounds good to me!

client/sandbox/react-nextjs/pages/play/linkunlink.tsx
```diff
@@ -0,0 +1,201 @@
...
+            </div>
+          </div>
+        );
+      })}
+    </div>
+  );
+}
+
+export default function Page() {
+  return <EphemeralAppPage schema={schema} perms={perms} Component={Example} />;
```
stopachka: cc @tonsky 

Tried playing with link/unlink, by creating this play page. I did the following steps: 

1. Clicked 'Seed', to add users and groups
2. Picked the 'current user' as an `Owner`. 
3. Tried to click 'add' on someone. 

I got a perms-pass failure, with a hint on `["groups" :object]` 

I am not sure if my example is wrong through. Do you mind checking?

tonsky: @stopachka updated your example. Main problem was that you were using auth.id but that doesn’t work without real authentication. You were setting currentUser to local state but it didn’t affect anything

stopachka: Ah! Nice catch, thanks @tonsky !

client/sandbox/react-nextjs/pages/play/linkunlink_ruleparams.tsx
```diff
@@ -0,0 +1,203 @@
+import { useState } from 'react';
+import { init, tx, id, i } from '@instantdb/react';
+import config from '../../config';
+import EphemeralAppPage from '../../components/EphemeralAppPage';
+
```
stopachka, line 5: cc @tonsky made another example with ruleparams -- thought this would be nice, since we can more quickly test "acting" as different users. If you want to keep the other one feel free 🫡

client/sandbox/react-nextjs/pages/play/linkunlink_ruleparams.tsx
```diff
@@ -0,0 +1,203 @@
...
+    },
+  },
+});
+
+const perms = {
+  groups: {
+    allow: {
+      view: 'true',
+      create: 'true',
+      update: 'true',
```
stopachka, line 37: cc @tonsky

QQ: If I set `update` to false, linking and unlinking will fail. 

Is the rule something like: 
For a `link` or `unlink`, the `update` check has to pass, as well as the `link` check?

tonsky: It fall backs to previous behavior: update for forward direction, view for reverse direction. So in the end, there are always two checks:

- for forward direction, if `link` is defined, then `link`, otherwise `update`
- for forward direction, if `link` is defined, then `link`, otherwise `view`

client/sandbox/react-nextjs/pages/play/linkunlink_ruleparams.tsx
tonsky: Shouldn’t you change the rules to make use of ruleParams?

client/sandbox/react-nextjs/pages/play/linkunlink_ruleparams.tsx
tonsky: Tried to explain it here https://github.com/instantdb/instant/pull/1514/files#diff-85bc61244af9920daebe28894bbc3f5a33e8df53f127fc5b7e243471784ea027

client/sandbox/react-nextjs/pages/play/linkunlink_ruleparams.tsx
stopachka: Oops -- was doing some git foo and lost my change there! Updated

client/sandbox/react-nextjs/pages/play/linkunlink_ruleparams.tsx
stopachka: > for forward direction, if link is defined, then link, otherwise update

In this case `groups` has the link perm defined

Should this then ignore the `update` check for `groups`?

tonsky: If you set update on groups to false, it should not matter, yes

stopachka: Okay, tested once more...and you are totally right. Maybe my git foo had me in a bad state before. Thanks!

stopachka (approved):
Great work @tonsky !

server/src/instant/util/test.clj
```diff
@@ -236,3 +266,29 @@
...
+                                          (->> right
+                                             (partition 2)
+                                             (map first)
+                                             (mapcat #(coll/collect symbol? %)))
+                                          [])
+                         (symbol? left) [left]
+                         :else          (coll/collect symbol? left))]
+      sym))))
+
+(defmacro test-matrix
```
stopachka, line 285: 🔥

PR #1517, opened 2025-08-14
Add ne op

nezaj:
Noticed LLMs often hallucinate a `$ne` op for `$not` -- discussed with team and figured we could add it in!

This includes support on both the client and the backend so that `$ne` works with the admin SDK too.

Verified this works as expected in sandbox and also adding a smoke test in the admin sdk.

Will put up docs in a follow-up PR after deploying

### DIFF
--- a/client/packages/core/__tests__/src/instaql.test.js
+++ b/client/packages/core/__tests__/src/instaql.test.js
@@ -1144,8 +1144,9 @@ test('$isNull with reverse relations', () => {
   ).toEqual(['Lonely shelf']);
 });
 
-test('$not', () => {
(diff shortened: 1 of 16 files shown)


client/packages/core/PR.md
```diff
@@ -0,0 +1,48 @@
+# Add $ne operator as an alias for $not
```
stopachka: RM?

stopachka (approved):
SGTM!

dwwoelfel (commented):
Could we replace $ne -> $not in the instaql layer when we parse the query instead so that we don't have to handle `$ne` in so many places?

PR #1519, opened 2025-08-15
[Docs] Update $not -> $ne

nezaj:
Updates docs to reference `$ne` instead of `$not`

Also adds a deprecation note to `$not`, hence the version bump


stopachka (approved):
LGTM!

PR #1521, opened 2025-08-15
[LLM] Add option map and remove required

nezaj:
What it says on the tin! This should help reduce hallucinations and not hint the llm to use required


stopachka (approved):
LGTM!

PR #1523, opened 2025-08-15
Fix for converting like pattern to regex

dwwoelfel:
Before we were just replacing `%` with `.*` and `_` with `.` and then creating a pattern out of it. If you had some regex special characters in the pattern, then we would include those as part of the pattern.

Now we walk the pattern and quote any strings in the pattern, so that it matches the string exactly.

Also added some additional tests.


stopachka (approved):
Heeck yeah!

PR #1525, opened 2025-08-15
Add batch-lookup for cache

dwwoelfel:
Adds a way to lookup items from the cache in a batch. This will allow us to put the attr sketches in a ttl cache and fetch all of the ones we don't have in the cache in a single sql request.

Instead of something like:

```clojure
(for [attr-id attr-ids]
  ;; One fetch per missing attr
  (cache/lookup-or-miss cache attr-id (fn [a] (fetch-sketch a))))
```

We can do:

```clojure
;; One fetch for all missing attrs
(cache/lookup-or-miss-batch attr-ids (fn [aids] (fetch-sketches aids)))
```

Also upgrades core.cache to be resilient to "stampedes", where a bunch of simultaneous requests cause us to do extra fetching. The new update stores a delay in the cache, so you only fetch once. However, this has a big problem. If the value-function throws an error when it is fetching the item to put in the cache, then every time you lookup the item in the cache, you'll get the same error. If your cache doesn't have a lot of churn, then you might throw every time you access the item until you restart the server. One bad sql request for an app's attrs could take down all requests for that app.

```clojure
;; Put an error in the cache with the default lookup-or-miss
(is (thrown? Exception
             (cache/lookup-or-miss cache-atom :wrapped (fn [_]
                                                         (throw (Exception. "oops"))))))

;; Now we'll never get that error out of the cache
(is (thrown? Exception
             (cache/lookup-or-miss cache-atom :wrapped (fn [_] :ok)))))

```

I fixed it by writing our own `lookup-or-miss` that catches errors before returning and evicts the error from the cache before continuing.

### DIFF
--- a/server/src/instant/util/cache.clj
+++ b/server/src/instant/util/cache.clj
@@ -183,7 +183,8 @@
                                                 :a a}
                                                (set es))
 
-                                       new-values (delay (batch-value-fn missing))]
+                                       new-values (when (seq missing)
+                                                    (delay (batch-value-fn missing)))]
                                    (reduce (fn [a e]
                                              (let [v (delay (get @new-values e))]
                                                (vswap! our-results conj v)


nezaj (approved):
Really nice!

PR #1529, opened 2025-08-18
fix query validation for complex links

drew-harris:
Allows for complex queries on links.


client/packages/core/__tests__/src/queryValidation.test.ts
```diff
@@ -994,3 +994,26 @@ test('pagination parameters can only be used at top-level namespaces', () => {
...
   });
 });
+
+test('relations with complex objects', () => {
+  beValid({
+    users: {
+      $: {
+        where: {
+          posts: {
+            $isNull: true,
```
stopachka, line 960: would add the test for $not

PR #1531, opened 2025-08-18
Add delay for autoresize in explorer

nezaj:
What it says on the tin! This should fix the multi-line not resizing in prod


client/www/components/dash/explorer/EditRowDialog.tsx
```diff
@@ -77,6 +77,12 @@ function ResizingTextArea({
   useLayoutEffect(() => {
     if (textareaRef.current) {
       autoResize(textareaRef.current);
+      // Small delay to ensure autoResize runs after the component has rendered
+      setTimeout(() => {
```
stopachka: If we want to run this _after_ a component renders, why do we use `useLayoutEffect`? If we used `useEffect` would it become unnecessary?

PR #1535, opened 2025-08-18
use query always returns a result

drew-harris:
Updates the `useQuery` hook to get its initial data from `getPreviousResultOrEmpty`, a new reactor function that will send back the data from an empty store with pending mutations applied. Also updates the types for the return type of useQuery to ensure that data is always defined.

### DIFF
--- a/client/packages/core/src/Reactor.js
+++ b/client/packages/core/src/Reactor.js
@@ -732,6 +732,48 @@ export default class Reactor {
   // ---------------------------
   // Queries
 
+  // Only called by the react package
+  getPreviousResultOrEmpty = (q) => {
+    if (!q) {
+      return { data: null };
+    }
+
+    if (typeof window === 'undefined') {
+      return {
+        data: Object.keys(q).reduce((acc, key) => {
+          acc[key] = [];
+          return acc;
+        }, {}),
+      };
+    }
+
+    const hash = weakHash(q);
+    const data = this.dataForQuery(hash);
+    if (data) {
+      return data;
+    }
+    // create store and apply pending mutations
+    const store = s.createStore(
+      {},
+      [],
+      this.config.useDateObjects,
+      this._linkIndex,
+      this.config.useDateObjects,
+    );
+
+    const pendingMutations = this.pendingMutations.currentValue;
+    const mutations = this._rewriteMutationsSorted(
+      store.attrs,
+      pendingMutations,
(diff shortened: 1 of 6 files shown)


client/packages/core/src/Reactor.js
```diff
@@ -733,6 +733,51 @@ export default class Reactor {
   // ---------------------------
   // Queries
 
+  // Only called by the react package
+  getPreviousResultOrEmpty = (q) => {
+    if (!q) {
+      return null;
+    }
+    if (typeof window !== undefined) {
```
stopachka: Here do you mean === undefined?

client/packages/core/src/Reactor.js
```diff
@@ -733,6 +733,51 @@ export default class Reactor {
...
+        data: Object.keys(q).reduce((acc, key) => {
+          acc[key] = [];
+          return acc;
+        }, {}),
+      };
+    }
+
+    const hash = weakHash(q);
+    const data = this.dataForQuery(hash);
+    if (!data) {
```
stopachka: would invert this: if (data) return data 

then the rest can be unnested

client/packages/react/src/InstantReactAbstractDatabase.tsx
```diff
@@ -197,10 +204,8 @@ export default abstract class InstantReactAbstractDatabase<
   useQuery = <Q extends ValidQuery<Q, Schema>>(
     query: null | Q,
     opts?: InstaQLOptions,
-  ): InstaQLLifecycleState<
-    Schema,
-    Q,
-    NonNullable<Config['useDateObjects']>
+  ): DataAlwaysDefined<
```
stopachka: Should we update InstaQLLifecycleState itself? (I am not sure why DataAlwaysDefined is necessary)

drew-harris: we'd have to duplicate the whole type because the non-react packages also use the type

client/packages/core/src/Reactor.js
drew-harris: oof good catch

client/packages/react/src/package.json
```diff
@@ -0,0 +1,3 @@
+{
+  "type": "module"
```
stopachka: nit: rm

PR #1541, opened 2025-08-19
Log info from hazelcast's jmx metrics to gauges

dwwoelfel:
Gets properties that hazelcast publishes to jmx and sends them to honeycomb with our gauges.

Adds a dependency on https://github.com/clojure/java.jmx, which provides a nicer interface to jmx info.


stopachka (approved):
Heck yeah!

PR #1542, opened 2025-08-19
Fix delete cascade that goes into system namespaces

tonsky:
Report: https://discord.com/channels/1031957483243188235/1407291069015785522

Issue is that we validate reserve names after cascade delete expansion (which is more correct, as we actually check what we are deleting). In the old version we were only checking it before expansion (so technically you can delete from any namespace that is cascade linked to e.g. users)

### DIFF
--- a/server/src/instant/db/permissioned_transaction_new.clj
+++ b/server/src/instant/db/permissioned_transaction_new.clj
@@ -33,7 +33,12 @@
           :when (#{:add-triple :deep-merge-triple :retract-triple :delete-entity} op)
           :when (and etype
                      (string/starts-with? etype "$")
-                     (not (and admin? (= etype "$users")))
+                     (not (and admin?
+                               (#{"$users"
+                                  "$userRefreshTokens"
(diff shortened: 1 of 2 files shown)


server/src/instant/db/permissioned_transaction_new.clj
```diff
@@ -33,7 +33,12 @@
           :when (#{:add-triple :deep-merge-triple :retract-triple :delete-entity} op)
           :when (and etype
                      (string/starts-with? etype "$")
-                     (not (and admin? (= etype "$users")))
+                     (not (and admin?
```
nezaj: Per discord discussion no need to check the tables and just simplify this to `admin?`

The check below still handle `$files` separately

nezaj (approved):
LGTM w/ one note!

server/src/instant/db/permissioned_transaction_new.clj
tonsky: Yeah makes sense

PR #1544, opened 2025-08-19
Set hazelcast timeouts to expire before we give up on startup

dwwoelfel:
This should make us more resilient to issues where we can't connect to Hazelcast on startup.

Hazelcast was set to wait 5 minutes before giving up on joining the cluster, but we give up on startup after 1 minute. If we had an issue joining the cluster, then we would quit before hazelcast timed out, leaving the instance in a limbo state until eb decided to kill it.

Now we'll wait 2 minutes before giving up and we set all of the timeouts to lower values.

If the join timeout expires, then the instance will start its own cluster, which we can resolve by cycling the instances. While we're doing that, clients won't have a consistent view of presence.


nezaj (approved):
LGTM!

PR #1545, opened 2025-08-19
Enable pg hint plan by default (gated behind a feature flag)

dwwoelfel:
Adds a new instant-config toggle to enable pg_hint_plan by default, with an instant-config flag to disable hints for individual queries by their hashes.

The testing is looking good and I've already enabled it for the 20 best queries.

Changes in this PR:
1. Enables hints for the lookup subqueries in datalog queries.

When you do a lookup, we do something like:

```sql
where entity_id = (select entity_id from triples where attr_id = :lookup-attr-id and value = :lookup-value)
```

Lookups are unique, so we know that should be an `av_index` lookup. It complicated the `where-clause` function a bit because now it needs to pass pg-hints back up. There are other subqueries that we'll want to add hints to also.

2. Uses hints during tests and adds some helpers to get the sketches during tests, since they update incrementally in the background.

Easier to review with whitespace turned off: https://github.com/instantdb/instant/pull/1545/files?w=1

### DIFF
--- a/server/test/instant/db/instaql_test.clj
+++ b/server/test/instant/db/instaql_test.clj
@@ -2,7 +2,6 @@
   (:require
    [clojure.string :as string]
    [clojure.test :as test :refer [are deftest is testing]]
-   [instant.config :as config]
    [instant.data.bootstrap :as bootstrap]
    [instant.data.constants :as constants]
    [instant.data.resolvers :as resolvers]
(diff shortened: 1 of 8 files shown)


server/test/instant/db/instaql_test.clj
```diff
@@ -2768,132 +2767,132 @@
...
-            (is (= #{true} (run-query :boolean {:etype {:$ {:where {:boolean {:$gte true}}}}})))
-            (is (= #{} (run-query :boolean {:etype {:$ {:where {:boolean {:$lt false}}}}})))
-            (is (= #{false} (run-query :boolean {:etype {:$ {:where {:boolean {:$lt true}}}}})))
-            (is (= #{false true} (run-query :boolean {:etype {:$ {:where {:boolean {:$lte true}}}}})))
-            (is (= #{true} (run-query :boolean {:etype {:$ {:where {:boolean true}}}})))
-            (is (= #{false} (run-query :boolean {:etype {:$ {:where {:boolean {:$not true}}}}})))
-
-            (testing "uses index"
-              (is (= "triples_boolean_type_idx" (run-explain :boolean true))))))))))
+        (with-sketches app
```
stopachka, line 2774: cool!

stopachka (approved):
Ohmagad!

PR #1548, opened 2025-08-20
[Docs] Tweak storage docs

nezaj:
I noticed LLMs would keep using `upload` instead of `uploadFile` for the storage admin sdk even though `upload` is the deprecated method.

Realized we probably needed to update that section in the docs. Lo and behold!

### DIFF
--- a/client/www/pages/docs/storage.md
+++ b/client/www/pages/docs/storage.md
@@ -550,7 +550,7 @@ options otherwise the default content-type will be `application/octet-stream`.
 import fs from 'fs';
 
 // Read a file from disk and upload to storage, return the uploaded file id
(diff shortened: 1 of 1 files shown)


client/www/pages/docs/storage.md
```diff
@@ -552,11 +549,13 @@ options otherwise the default content-type will be `application/octet-stream`.
 ```tsx
 import fs from 'fs';
 
+// Read a file from disk and upload to storage, return the uploaded file id
 async function upload(filepath: string) {
```
stopachka: Nit: should we call this function uploadFile too?

nezaj: Sure!

PR #1550, opened 2025-08-20
Exit on failure during startup

dwwoelfel:
This PR will kill the jvm if we encounter an exception during startup. This should help us be more resilient to failures that temporarily take out the JVM.

If you need to do some debugging on the failure, you can set `EXIT_ON_ERROR=false` to prevent the jvm from exiting. It also won't exit in dev mode.

### DIFF
--- a/server/src/instant/core.clj
+++ b/server/src/instant/core.clj
@@ -287,8 +287,6 @@
                            flags/queries
                            flags/query-results)))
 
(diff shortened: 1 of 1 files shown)


tonsky:
LGTM
Surprised JVM doesn’t exit with non-zero code by default on exception

PR #1551, opened 2025-08-20
Add support for uploading streams to admin sdk

nezaj:
As part of stress testing uploads, noticed we don't support uploading streams from the admin sdk but we do support it via API. With this PR the admin SDK can now also upload streams

```typescript
async function testUploadFileStream(
  src: string,
  dest: string,
  contentType?: string,
) {
  const fp = path.join(__dirname, src);
  const stream = fs.createReadStream(fp);
  const fileSize = fs.statSync(fp).size;
  const data = await db.storage.uploadFile(dest, stream, {
    contentType: contentType,
    fileSize,
  });
  console.log('Uploaded with stream:', data);
}
```

Separately, I noticed after we migrated to `tshy` we changed our package.json for tests to be `module`. This mean using `__dirname` was broken. Fixed that here!

If this looks good will add docs in a follow-on!

### DIFF
--- a/client/packages/admin/src/index.ts
+++ b/client/packages/admin/src/index.ts
@@ -1,3 +1,5 @@
+import { Readable } from 'stream';
+
 import {
   tx,
   lookup,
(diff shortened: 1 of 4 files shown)


dwwoelfel (approved):
LGTM!

PR #1554, opened 2025-08-20
Collapse ors with  $isNull: true for indexed attrs

dwwoelfel:
This will help queries that do something like `{:or [{:status "pending"} {:status {:$isNull true}}]}`.

If `status` is indexed, we'll convert it to `{:status {:$in [nil "pending"]}}`, saving a CTE and the or gather CTE.

For the query I'm testing with, it drops the time from 140 ms to 8ms.


nezaj (approved):
Neat!

PR #1558, opened 2025-08-21
[Rules] Reference instant-rules for Claude only

nezaj:
Noticed we referenced `instant-rules` for the other editors/tools but this should be specific to Claude only


stopachka (approved):
Nice!

PR #1559, opened 2025-08-21
subscribeQuery in the admin SDK

dwwoelfel:
Adds `subscribeQuery` over SSE for the admin SDK.

Only allows a single query per connection. Unlike websockets, the SSE connection is send-only. The client can't send new `add-query` commands through the SSE connection. If we want to multiplex, we will need to have some way to send an `add-query` POST request and have it route to the machine that holds their session (possibly through hazelcast?).

It uses our regular session middleware with a couple of small differences: 
1.`init` and `add-query` are triggered on the backend in `on-open` instead of having the user send `init` and `add-query`.
2. The session has `:sse-conn` instead of `:ws-conn`, then store/send-event dispatches based on which is in the session

### Async iterator

```ts
const sub = db.subscribeQuery({goals: {}});

for await (const payload of sub) {
  if (payload.type === 'error') {
    console.log('error', error);
    sub.close();
  } else {
    console.log('data', payload.data)
  }
}

// To end the subscription:
sub.close();
```

### Callback

```ts
const sub = db.subscribeQuery({goals: {}}, (payload) => {
  if (payload.type === 'error') {
    console.log('error', error);
    sub.close();
  } else {
    console.log('data', payload.data)
  }
});

// To end the subscription:
sub.close();
```

Also updates undertow from 2.2.20.Final to 2.3.18.Final. I didn't see any changes in the changelog that would affect us, but I'll keep an eye on things after deploying.

### DIFF
--- a/client/packages/admin/src/index.ts
+++ b/client/packages/admin/src/index.ts
@@ -843,7 +843,7 @@ class InstantAdminDatabase<
   /**
    * Use this to to get a live view of your data!
    *
-   * @see https://instantdb.com/docs/instaql
+   * @see https://www.instantdb.com/docs/backend
    *
    * @example
    *  // create a subscription to a query
--- a/client/packages/admin/src/index.ts
+++ b/client/packages/admin/src/index.ts
@@ -77,6 +77,8 @@ import {
   subscribe,
   SubscribeQueryCallback,
   SubscribeQueryResponse,
+  SubscribeQueryPayload,
+  SubscriptionReadyState,
 } from './subscribe.ts';
 
 type DebugCheckResult = {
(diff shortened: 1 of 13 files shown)


client/packages/admin/src/index.ts
```diff
@@ -829,6 +840,71 @@ class InstantAdminDatabase<
     });
   };
 
+  /**
+   * Use this to to get a live view of your data!
+   *
+   * @see https://instantdb.com/docs/instaql
```
stopachka: Nit: should we link instant-on-the-backend here?

client/packages/admin/src/index.ts
```diff
@@ -829,6 +840,71 @@ class InstantAdminDatabase<
...
+   *      sub.close();
+   *    } else {
+   *      console.log(payload.data);
+   *    }
+   *  }
+   *
+   *  // Stop the subscription
+   *  sub.close();
+   *
+   *  // Createa a subscription with a callback
```
stopachka:
```suggestion
   *  // Create a subscription with a callback
```

client/packages/admin/src/index.ts
```diff
@@ -829,6 +840,71 @@ class InstantAdminDatabase<
...
+   *
+   * @see https://instantdb.com/docs/instaql
+   *
+   * @example
+   *  // create a subscription to a query
+   *  const query = { goals: { $: { where: { title: "Get Fit" } } } }
+   *  const sub = db.subscribeQuery(query);
+   *
+   *  // iterate through the results with an async iterator
+   *  for await (const payload of sub) {
```
stopachka, line 854: Wow! For later: I realize we could this in core too.

server/src/instant/reactive/session.clj
```diff
@@ -514,17 +541,20 @@
...
-                         :app-id app-id
-                         :client-event-id client-event-id
-                         :status 500
-                         :original-event (merge original-event
-                                                debug-info)
-                         :message (str "Yikes, something broke on our end! Sorry about that."
-                                       " Please ping us (Joe and Stopa) on Discord and let us know!")
-                         :session-id sess-id})))
+    (if (= :error (:op original-event))
+      ;; Don't send an error if we failed to send the error or we'll get into an
```
stopachka, line 545: nice!

stopachka (approved):
Biyutiful!!

client/packages/admin/src/index.ts
dwwoelfel: Good idea! Done!

PR #1560, opened 2025-08-22
Enable optimistic create for admin SDK

nezaj:
What it says on the tin! Adds a test for this too.

Now doing something like this should succeed!

```ts
db.transact(db.tx.test[id()].create({a: 1}))
```


stopachka (approved):
Niice!

PR #1562, opened 2025-08-22
[server] order triple inserts

stopachka:
We had a bunch deadlock-detected errors when an instant startup did a big push of transactions. Instant is being used as a queue, which creates a lot of contention: 

https://ui.honeycomb.io/instantdb/environments/prod/datasets/instant-server/result/BxYcNG8M9Z1?vs=hideCompare

**Here's one intuition as to why:**

Imagine two transactions that work on `entity-a` and `entity-b`

```
tx1: take lock on entity-a 
tx2: take lock on entity-b
tx1: wait for lock on entity-b
tx2: wait for lock on entity-a
```

Postgres takes a lock when we do an insert on a row. If we don't order correctly, we could cause a deadlock. 

**Note:** I was not able to repro this locally. 

But, we are seeing those errors. Let's see if shipping this fixes it. 

### DIFF
--- a/server/src/instant/db/model/triple.clj
+++ b/server/src/instant/db/model/triple.clj
@@ -449,7 +449,8 @@
         lookup-ref-inserts
         {:insert-into [[:triples triple-cols]
                        {:select triple-cols
-                        :from :enhanced-lookup-refs}]
+                        :from :enhanced-lookup-refs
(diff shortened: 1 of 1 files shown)


nezaj (approved):
🙏🙏

PR #1569, opened 2025-08-25
Increase span attr limit

dwwoelfel:
Increases the span attribute limit from 128 to 1900. Honeycomb's API limit is 2000, so this leaves a little headroom for refinery to add attrs. 

This will fix a problem where we're missing some of the attributes for the gauges.


stopachka (approved):
SGTM!

PR #1571, opened 2025-08-26
[order] one more repro about flickering orders

stopachka:
I noticed a similar issue that @nezaj described in our ordering colors bug, when building ChatGPTBlox. 

Added one more repro here. This one gives a bit more 'color' (pun intended):

CleanShot 2025-08-26 at 12 21 12@2x

What this shows is happening: 

When we order by a field that we changed, the item that was changed seems to 'disappear' from the result temporarily.

The order history shows how the list renders. I sent a message, and saw (curiously enough) a _different_ chat disappear from the list: 

image


stopachka:
I am going to do a much more comprehensive play page, with docs on what is and isn't happening. Will drop this for now

PR #1575, opened 2025-08-27
Try ReentrantLock in reactive store

tonsky:
We seem to have long pauses (10+ seconds) in reactive store

image

These seem to be related to high congestion on a lock (see lock_time_ms above)

One thing we can try is ReentrantLock.

On its own, it could be faster. That’s what we will be testing by default.

Another option is to try making it fair. Javadoc (highlight mine):

> When set true, under contention, locks favor granting access to the longest-waiting thread. Otherwise this lock does not guarantee any particular access order. Programs using fair locks accessed by many threads may display lower overall throughput (i.e., are slower; often much slower) than those using the default setting, but have __smaller variances in times to obtain locks__ and __guarantee lack of starvation__.

I made the fairness enabled through a flag, in case we see a catastrophic uptick in latency.

Deployment plan:

- push flags config cd server/flags-config && pnpm i && pnpm run push

### DIFF
--- a/server/src/instant/flags.clj
+++ b/server/src/instant/flags.clj
@@ -20,6 +20,7 @@
             :use-patch-presence {}
             :refresh-skip-attrs {}
             :new-permissioned-transact {}
+            :store-fair-lock {}
             :drop-refresh-spam {}
             :promo-emails {}
             :rate-limited-apps {}
@@ -121,6 +122,24 @@
              :default-value default-value
              :disabled? disabled?}))
 
+        store-fair-lock
+        (when-let [hz-flag (-> (get result "store-fair-lock")
+                               first)]
+          (let [disabled-apps (-> hz-flag
+                                  (get "disabled-apps")
+                                  (#(map parse-uuid %))
+                                  set)
+                enabled-apps (-> hz-flag
+                                 (get "enabled-apps")
+                                 (#(map parse-uuid %))
+                                 set)
+                default-value (get hz-flag "default-value" false)
+                disabled? (get hz-flag "disabled" false)]
+            {:disabled-apps disabled-apps
+             :enabled-apps enabled-apps
+             :default-value default-value
+             :disabled? disabled?}))
+
         promo-code-emails (set (keep (fn [o]
                                        (get o "email"))
                                      (get result "promo-emails")))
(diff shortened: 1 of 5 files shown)


server/src/instant/flags.clj
```diff
@@ -121,6 +122,24 @@
              :default-value default-value
              :disabled? disabled?}))
 
+        store-fair-lock
```
stopachka, line 125: Thought: would the nifty `flags` namespace that @dwwoelfel popped up recently be useful for this? We could keep all of this in a json field there, so we don't have to create a new namespace

server/src/instant/reactive/store.clj
```diff
@@ -90,11 +93,25 @@
    :datalog-query/delayed-call {} ;; delay with datalog result (from query.clj)
    :datalog-query/topics {:db/type :db.type/list-of-topics}})
 
-(defn make-app-conn [_app-id]
-  (d/create-conn schema))
+(defn create-conn [schema app-id]
+  (doto (d/create-conn schema)
+    (alter-meta! assoc
+                 :lock   (ReentrantLock. (flags/store-fair-lock? app-id))
```
stopachka, line 99: nit: whitespace

```suggestion
                 :lock (ReentrantLock. (flags/store-fair-lock? app-id))
```

server/src/instant/util/lang.clj
```diff
@@ -20,3 +20,11 @@
       (when (and val# (not (instance? Var$Unbound val#)))
         (~method val#)
         (Var/.doReset var# nil)))))
+
+(defmacro with-reentrant-lock [lock-expr & body]
```
stopachka, line 24: cool!

stopachka (approved):
LGTM!

server/src/instant/flags.clj
tonsky: I like the flexibility of disabled/disabled-apps/enabled-apps/enabled. Maybe we can create a universal namespace for this. I’ll do it next time

stopachka: I _think_ you can implement this with the current `flags` namespace. i.e, store a json blob in the `value` as 

```js
{"disabled": ..., "disable-apps": ... }
```

PR #1576, opened 2025-08-27
Make aggregator resilient to huge transactions

dwwoelfel:
When we get a giant transaction from a COPY or an app delete, we collect all of the changes for the transaction in memory and then apply them in the invalidator.

This can cause the jvm to run out of memory and crash. With the invalidator, we'll create a new slot the next time we start up that skips over the giant transaction, so it's self-healing in a way.

The aggregator will keep trying and failing to process the giant transaction.

This PR fixes the problem for the aggregator by ignoring transaction boundaries and just putting the `update`, `delete`, and `insert` records directly into the process channel. It doesn't wait for all of the changes for the transaction and just handles them as they come. This is fine for the aggregator--it doesn't need to care about transactions. We can't do the same for the invalidator because we need the `tx-id` from the `transactions` table.

### DIFF
--- a/server/src/instant/jdbc/wal.clj
+++ b/server/src/instant/jdbc/wal.clj
@@ -395,7 +395,10 @@
             (recur (.read stream)))
 
           (:insert :update :delete)
-          (let [put-result (a/alt!! [[to record]] :put)]
+          (let [put-result (a/alt!! [[to record]] :put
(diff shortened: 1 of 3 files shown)


stopachka (approved):
Heeck yeahh!!

PR #1577, opened 2025-08-27
[soft-deletes] api

stopachka:
Added: 
1. An API to get soft-deleted attrs

Given the already existing `restore-attr` in transact, we now have all the machinery needed to add soft-deletes to the UI. 

I made a play page to showcase it: 

CleanShot 2025-08-27 at 14 12 34@2x


server/src/instant/admin/routes.clj
```diff
@@ -593,6 +593,29 @@
...
+;; Soft deletions 
+
+(defn soft-deleted-attrs-get [req]
+  (let [{:keys [app-id]} (req->app-id-authed! req :data/read)
+        soft-deleted-attrs (attr-model/get-soft-deleted-by-app-id
+                            (aurora/conn-pool :read)
+                            app-id)]
+    (response/ok {:soft-deleted-attrs soft-deleted-attrs})))
+
+(defn hard-delete-attrs-post [req]
```
dwwoelfel: Why do we need to provide a hard delete to the user?

If we handle deletes in the background ourselves, then we can manage transaction sizes and pause deletes when the system is overloaded.

stopachka: I thought in the admin API it would be good to expose this to devs (Some like the control even if it doesn't really matter for them in this case). I do see your point though -- will delete! If folks complain we can figure it out

stopachka: Removed @dwwoelfel !

server/src/instant/admin/routes.clj
```diff
@@ -593,6 +593,16 @@
...
 
+;; --------------- 
+;; Soft deletions 
+
+(defn soft-deleted-attrs-get [req]
+  (let [{:keys [app-id]} (req->app-id-authed! req :data/read)
+        soft-deleted-attrs (attr-model/get-soft-deleted-by-app-id
+                            (aurora/conn-pool :read)
+                            app-id)]
+    (response/ok {:soft-deleted-attrs soft-deleted-attrs})))
```
```suggestion
    (response/ok {:attrs soft-deleted-attrs})))
```

This will be easier for other clients since it has no dashes

client/sandbox/react-nextjs/pages/play/soft-delete-test.tsx
```diff
@@ -0,0 +1,278 @@
...
+          {
+            method: 'GET',
+            headers: {
+              'app-id': appId,
+              authorization: `Bearer ${adminToken}`,
+            },
+          },
+        );
+        const data = await response.json();
+        setSoftDeletedAttrs(data['soft-deleted-attrs'] || []);
```
nezaj: Question, it looks the API returns `{:attrs soft-deleted-attrs}` -- how does this work?

stopachka: https://github.com/instantdb/instant/pull/1577/files/aebd807e89f7a387151934d34bf81c1ae4f56ead#diff-76efc8493523f11efd3fb55c34d4e06dc52982b80e60ebda7ea3448cbe642f8b

I accepted this commit, and had a todo to update here, but haven't gotten to this PR yet. Will address shortly 🤜🤛

server/test/instant/admin/routes_test.clj
```diff
@@ -1195,5 +1198,48 @@
...
+               {:body {:steps [[:delete-attr aid-2]]}
+                :headers {"app-id" (str app-id)
+                          "authorization" (str "Bearer " admin-token)}})
+
+            ;; Check that attr appears 
+            after-delete (soft-deleted-attrs-get
+                          {:headers {"app-id" (str app-id)
+                                     "authorization" (str "Bearer " admin-token)}})
+            _ (is (= 200 (:status after-delete)))
+            soft-deleted-attrs (-> after-delete :body :soft-deleted-attrs)
```
nezaj: I think this failure is related to my question above

PR #1578, opened 2025-08-27
Split overview stats into two sections

dwwoelfel:
Splits the overview components so that they can load separately. Coded by claude: https://claude.ai/share/b2202725-9b6a-4c3b-a8d6-fb0c694e735d

To view, login at https://instant-www-js-separate-loading-jsv.vercel.app, then view at https://instant-www-js-separate-loading-jsv.vercel.app/intern/overview


stopachka (approved):
Heeck yeah!

PR #1579, opened 2025-08-27
Prevent logging parameters in auto explain

dwwoelfel:
This applies a connection-level setting to prevent auto_explain from logging parameters. It is applied whenever hikari creates a new connection for its pool.

We would set it in the db config, but aurora doesn't give us a way to modify that setting.


stopachka (approved):
LGTM!

dwwoelfel:
Closing because aurora won't let you set that param. New approach at https://github.com/instantdb/instant/pull/1580

PR #1580, opened 2025-08-27
Don't log bins in auto-explain

dwwoelfel:
Alternative to https://github.com/instantdb/instant/pull/1579 because aurora won't let us set `auto_explain.log_parameter_max_length`

This will help reduce the amount of postgres logs we produce.


stopachka (approved):
Nice!

PR #1582, opened 2025-08-28
Don't log bins in auto-explain (take 2)

dwwoelfel:
Follow up to https://github.com/instantdb/instant/pull/1580. That approach doesn't work because you can't use set-setting in a write connection unless auto-commit is off. This PR puts the set_config directly in the CTE (which is a lot easier than making the connection not use auto-commit).

It also disables handling empty updates in the aggregator in dev. When the aggregator processes an update, it writes to the database, which will trigger the aggregator on the next cycle. So in dev the aggregator will run in a loop every 10 seconds. It was enabled in the first place in https://github.com/instantdb/instant/pull/1576 so that the tests would know when all of their writes had been acknowledged. Now we just ignore those empty updates in dev.


stopachka (approved):
Great idea!

PR #1583, opened 2025-08-28
Link unlink permission checks, take 2

tonsky:
Changed the logic so that `delete` checks `link`/`unlink` only if they are explicitly defined (no fallback to `update`/`view` rules in that case)

For create/update, other side is still checked for view, but this is current behavior, so it should not break anything

#1514 #1581


stopachka (approved):
Let's do it!

PR #1584, opened 2025-08-28
add claude code message

drew-harris:
It feels awkward pausing before claude code outputs any messages, this adds an extra log to let the user know claude is starting. 

image


stopachka (approved):
🚢

PR #1586, opened 2025-08-28
Tweak deadlock-prevention ordering

dwwoelfel:
Updates the ordering on inserts to match the primary key. The ordering to avoid deadlocks was introduced in https://github.com/instantdb/instant/pull/1562. Making the ordering match the primary key seems to work better. 

I also managed to create a test case that demonstrates the deadlocks. You can see it fail before the ordering change here: https://github.com/instantdb/instant/actions/runs/17309198926/job/49139603660#step:10:30360

Other ideas in case we're still seeing deadlocks after this change:

1. Split out `av` inserts into their own CTEs and order the inserts to match the `av_index` column order (similar for `eav_uuid_index`). We may just be screwed here if we have to lock for the primary key and av index separately and those lock in different orders--not 100% sure how it works.
2. Do a `select * from triples for update` on the enhanced triples. This also worked if I ordered by the primary key.

### DIFF
--- a/server/src/instant/db/model/triple.clj
+++ b/server/src/instant/db/model/triple.clj
@@ -304,7 +304,8 @@
         lookup-ref-inserts
         {:insert-into [[:triples triple-cols]
                        {:select triple-cols
-                        :from :enhanced-lookup-refs}]
+                        :from :enhanced-lookup-refs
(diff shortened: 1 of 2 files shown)


stopachka (approved):
biyutiful!

PR #1589, opened 2025-08-29
Fix unlinking on delete for system namespaces

tonsky:
See https://discord.com/channels/1031957483243188235/1196584090552512592/1410865450975498320

I removed reverse unlink check during delete for namespaces that don’t explicitly define unlink in #1583. What I missed was that we have “default rule” for system namespace and that was confusing my “is unlink defined” logic


stopachka (approved):
SGTM!

PR #1590, opened 2025-08-29
Skip non existing entities checks

tonsky:
Report: https://discord.com/channels/1031957483243188235/1410891418276335628

There was an issue when a user did `db.tx.ns[<existing-id>].link({attr: <non-existing-id>})`, we were trying to run a link/unlink/view permission check on non-existing-id.

Normally that should’ve been an exception, but for compatibility reasons we didn’t throw in previous versions so we shouldn’t start now


stopachka (approved):
Nice!

PR #1591, opened 2025-08-29
Improve reconnect handling for admin subscribe

dwwoelfel:
Two changes here:
1. If admin-init succeeds, set the retry interval on the connection to 500ms (configurable with an instant-config flag). The default is 3 seconds, which is a bit too long. 
2. Only deliver an error on disconnect if we don't reconnect after 5 seconds.


stopachka (approved):
Super cool!

PR #1598, opened 2025-09-02
Update create doc-string

nezaj:
A user reported the current doc-string is incorrect -- addressing it here!


stopachka (commented):
Thank you!

PR #1599, opened 2025-09-02
typedoc api reference documentation

drew-harris:


drew-harris:
Closing for now because I don't think it can be very useful without a lot of work, keeping the branch though

PR #1601, opened 2025-09-02
Restore Attrs UI

drew-harris:
image
image

### DIFF
--- a/client/www/components/dash/explorer/EditNamespaceDialog.tsx
+++ b/client/www/components/dash/explorer/EditNamespaceDialog.tsx
@@ -139,9 +139,11 @@ export function EditNamespaceDialog({
                 <div className="flex gap-3 items-center">
                   <span className="py-0.5 font-bold">{attr.name}</span>
                   {notes.notes[attr.id]?.message && (
-                    <div className="text-xs px-2 text-gray-500">
-                      {notes.notes[attr.id].message}
-                    </div>
+                    <InfoTip>
+                      <div className="text-xs px-2 text-gray-500">
+                        {notes.notes[attr.id].message}
+                      </div>
+                    </InfoTip>
                   )}
                 </div>
                 {attr.name !== 'id' ? (
--- a/client/www/components/dash/explorer/RecentlyDeletedAttrs.tsx
+++ b/client/www/components/dash/explorer/RecentlyDeletedAttrs.tsx
@@ -163,14 +163,14 @@ const getConstraintMessage = (attr: SoftDeletedAttr): string | null => {
       attr.metadata.soft_delete_snapshot.is_indexed &&
       attr.metadata.soft_delete_snapshot.is_required
     ) {
-      return 'note: index and required constraints were dropped after restoring';
+      return 'Index and required constraints were dropped after restoring';
     }
 
     if (attr.metadata.soft_delete_snapshot.is_indexed) {
-      return 'note: indexed constraint was dropped after restoring';
+      return 'Indexed constraint was dropped after restoring';
     }
     if (attr.metadata.soft_delete_snapshot.is_required) {
-      return 'note: required constraint was dropped after restoring';
+      return 'Required constraint was dropped after restoring';
     }
     return null;
   }


drew-harris:
Note: also adds a min-width to the namespaces sidebar in the explorer, felt very cramped before

Before:
image
After 
image

client/www/components/dash/explorer/EditNamespaceDialog.tsx
```diff
@@ -1563,6 +1553,10 @@ function EditAttrForm({
   }
 
   async function deleteAttr() {
+    // update the recently deleted attr cache
+    setTimeout(() => {
```
stopachka, line 1577: Nit: would move this below the `await` on pushOps, so you are guaranteed that this will fetch once reactor has pushed

client/www/components/dash/explorer/RecentlyDeletedAttrs.tsx
```diff
@@ -0,0 +1,146 @@
...
+import { useDashFetch } from '@/lib/hooks/useDashFetch';
+import useSWR from 'swr';
+import { errorToast } from '@/lib/toast';
+import { InstantReactWebDatabase } from '@instantdb/react';
+import { ClockIcon } from '@heroicons/react/24/outline';
+import { ArrowUturnLeftIcon } from '@heroicons/react/24/solid';
+import { add, formatDistanceToNow } from 'date-fns';
+import { useEffect } from 'react';
+
+const getNamesByNamespace = (attr: DBAttr): Record<string, string> => {
```
stopachka: nit: maybe let's call this `softDeletedAttr`, to make it clear this is really only for soft-deleted attrs

client/www/components/dash/explorer/RecentlyDeletedAttrs.tsx
```diff
@@ -0,0 +1,146 @@
...
+  const dialog = useDialog();
+
+  const restoreAttr = async (attrId: string) => {
+    if (!db) return;
+    if (!data) return;
+    try {
+      await db._core._reactor.pushOps([['restore-attr', attrId]]);
+      mutate({
+        attrs:
+          data.attrs.filter((attr) => (attr.id === attrId ? false : true)) ??
```
stopachka: nit: I think you can just write here `attr.id !== attrId`

stopachka (approved):
LGTM! 

Maybe as a next step: 

Can we open a dropbox paper for to get the team in on a quick dogfooding session?

(We could use the vercel preview url here to test it out, though we may need to hardcode the grace-period-days temporarily)

client/www/components/dash/explorer/EditNamespaceDialog.tsx
drew-harris: its not guaranteed since pushOps doesn't wait for a response confirming its been deleted, which is why i added the setTimeout but i moved it to be after to at least read better.

client/www/components/dash/explorer/RecentlyDeletedAttrs.tsx
drew-harris: theres nothing about this function that is specific to soft deleted attrs, but ive changed the name and the type to make it clear this is the only situation its used currently.

client/www/components/dash/explorer/RecentlyDeletedAttrs.tsx
drew-harris: 🤦‍♂️

client/www/components/dash/explorer/EditNamespaceDialog.tsx
stopachka: I _think_ if the user is online `pushOps` will await until we get a `transact-ok`, which effectively means the db has confirmed the delete 

https://github.com/instantdb/instant/blob/main/client/packages/core/src/Reactor.js#L582

drew-harris:
image

drew-harris:
image

client/www/components/dash/explorer/RecentlyDeletedAttrs.tsx
```diff
@@ -32,13 +33,16 @@ export const RecentlyDeletedAttrs: React.FC<{
...
-}> = ({ namespace, appId, db }) => {
+  notes: ReturnType<typeof useAttrNotes>;
+}> = ({ namespace, appId, db, notes }) => {
   const { data, mutate, error } = useRecentlyDeletedAttrs(appId);
 
   const [expandedAttr, setExpandedAttr] = useState<string | null>(null);
 
   const dialog = useDialog();
 
+  // TODO: if restoring attr with previous contraints
```
nezaj, line 44: I don't think we need this anymore?

drew-harris:
image

PR #1602, opened 2025-09-02
Add link/unlink types to the dashboard editor

tonsky:
Report: https://discord.com/channels/1031957483243188235/1258878823529844766/1412446308492644494

### DIFF
--- a/client/packages/core/src/rulesTypes.ts
+++ b/client/packages/core/src/rulesTypes.ts
@@ -8,29 +8,19 @@ type InstantRulesAttrsAllowBlock = {
   delete?: string | null | undefined;
 };
 
-export type InstantRulesAllowBlock<
-  Schema extends InstantSchemaDef<any, any, any> = InstantUnknownSchema,
-> = InstantRulesAttrsAllowBlock & {
-  link?:
-    | { [EntityName in keyof Schema['entities']]?: string }
-    | null
-    | undefined;
-  unlink?:
-    | { [EntityName in keyof Schema['entities']]?: string }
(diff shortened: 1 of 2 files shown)


client/www/components/dash/Perms.tsx
```diff
@@ -168,6 +168,14 @@ const rulesSchema = (namespaces: SchemaNamespace[] | null) => {
           update: { type: 'string' },
           delete: { type: 'string' },
           view: { type: 'string' },
+          link: {
```
stopachka, line 171: I _think_ you'll need to update core/ruleTypes as well

tonsky:
@stopachka added

client/packages/core/src/rulesTypes.ts
```diff
@@ -1,21 +1,36 @@
...
+};
+
+export type InstantRulesAllowBlockWithoutLinks = InstantRulesAttrsAllowBlock;
+
 export type InstantRules<
   Schema extends InstantSchemaDef<any, any, any> = InstantUnknownSchema,
 > = {
-  $default?: { bind?: string[]; allow: InstantRulesAllowBlock };
-  attrs?: { bind?: string[]; allow: InstantRulesAllowBlock };
+  $default?: { bind?: string[]; allow: InstantRulesAllowBlock<Schema> };
```
stopachka: I would remove inferring from schema for now, as it could be backwards incompatible. If we add it way may want to do some more testing

tonsky: Done

PR #1603, opened 2025-09-02
Initial support for orgs

dwwoelfel:
Creates tables to hold orgs and adds a hidden page on the dashboard where we can start testing it.

Changes in this PR:
1. Creates the orgs table and the org members table and updates subscriptions to allow them to attach to orgs.
2. Creates dash routes for fetching, creating, and deleting orgs

The org members has a similar structure to the app members with the same set of roles. The main difference is that the org has no creator_id--instead we create an org_member with role "owner". That way you can have multiple "creators" and your org isn't tied to a single person.

Still a lot to do for full support for orgs:
1. Add/remove members
2. Add app to an org
3. Transfer apps between orgs
4. Add subscriptions to an org
5. Update permissions everywhere so that org access propagates to app access
6. UI

### DIFF
--- a/client/www/components/dash/Orgs.tsx
+++ b/client/www/components/dash/Orgs.tsx
@@ -1,0 +1,89 @@
+import config from '@/lib/config';
+import { TokenContext } from '@/lib/contexts';
+import { jsonFetch } from '@/lib/fetch';
+import { useContext } from 'react';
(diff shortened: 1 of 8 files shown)


stopachka (approved):
Niice!!

PR #1606, opened 2025-09-03
custom claude code path + show cursor after ctrl c

drew-harris:
Small pr to allow people to set CLAUDE_CODE_PATH environment variable
I tried using the `which` command to find it but that doesn't play nice with claude code sdk. 

Also fixed an issue where ctrl-c kept hiding the cursor.


nezaj (approved):
Nice!

PR #1607, opened 2025-09-03
[Docs] Update getting started

nezaj:
We got some feedback that our updated getting started docs weren't as helpful. This PR reverts the getting started docs back to the way they were.

Separately this adds a new doc page specifically for create-instant-app.

I also added a new markdown tag for tabbed commands. For now this is only useful for single-line commands, but it would be nice to have another component for mulit-line

CleanShot 2025-09-03 at 12 06 33@2x

### DIFF
--- a/client/www/components/docs/TabbedSingle.jsx
+++ b/client/www/components/docs/TabbedSingle.jsx
@@ -1,0 +1,83 @@
+import { useState, useContext } from 'react';
+import { ClipboardDocumentIcon } from '@heroicons/react/24/outline';
+import { SelectedAppContext } from '@/lib/SelectedAppContext';
+
+/**
+ * TabbedSingle - A component for displaying single-line content with tab selection.
+ *
+ * Useful for things like showiing one-liners for different platforms /
+ * framworks / etc. (e.g. Next.js, Expo, Vanilla TS, etc.)
+ *
+ * @param {Object} tabs - Object with tab keys containing { label, content }
+ * @param {string} defaultTab - Optional default selected tab key
+ */
+export function TabbedSingle({ tabs, defaultTab }) {
+  // Defensive check for tabs
+  if (!tabs || Object.keys(tabs).length === 0) {
+    return null;
+  }
+
+  const [selectedTab, setSelectedTab] = useState(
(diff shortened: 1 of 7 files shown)


client/www/components/docs/TabbedSingle.jsx
```diff
@@ -0,0 +1,89 @@
...
+  // Defensive check for tabs
+  if (!tabs || Object.keys(tabs).length === 0) {
+    return null;
+  }
+
+  const defaultValue = defaultTab || Object.keys(tabs)[0];
+
+  // Use localStorage if storageKey is provided, otherwise use regular state
+  const [selectedTab, setSelectedTab] = storageKey
+    ? useLocalStorage(`tabbed-single-${storageKey}`, defaultValue)
```
stopachka: This goes against the rules of hooks -- you can't conditionally render hooks. If `storageKey` ever changed this would throw an error. I would just settle on localStorage

nezaj: Updated!

client/www/components/docs/TabbedSingle.jsx
```diff
@@ -0,0 +1,87 @@
...
+ * @param {Object} tabs - Object with tab keys containing { label, content }
+ * @param {string} defaultTab - Optional default selected tab key
+ * @param {string} storageKey - Required unique key for localStorage persistence
+ */
+export function TabbedSingle({ tabs, defaultTab, storageKey }) {
+  const defaultValue = defaultTab || (tabs && Object.keys(tabs)[0]) || '';
+
+  // Always call hooks before any conditional returns
+  const [selectedTab, setSelectedTab] = useLocalStorage(
+    `tabbed-single-${storageKey}`,
```
dwwoelfel, line 21: I think you could drop the storage key. If someone selects a package manager, they probably want to use that same package manager everywhere.

nezaj: I use `TabbedSingle` also for choosing frameworks (Next, Expo Go, Vanilla TS) which is why I wanted a separate key. Great point though, renamed the storage keys to `app-framework` and `pkg-manager` to be more generic!

PR #1608, opened 2025-09-03
Change link unlink fallback

tonsky:
Per @nezaj request, link/unlink now work like this:

- if `link` is not defined for either direction, `update`/`view` fallback is used
- if `link` is defined for one direction, only that `link` permission is checked. The other side is not using `update`/`view` fallback (this is the part that changed)
- if `link` is defined for both directions, both are checked and results are ANDed

Same logic for `unlink`

Second change relates to `delete`. Before, when we were deleting entity and that entity had links, we were checking both `delete` for entity and `unlink` for the other side of the relation. After this PR, we only check `delete`.

Third change is introduction of `$default` for links. If you don’t want to list all attribute names manually, you can use `$default` as attribute name & override selected attributes:

```json
"todos": {
  "allow": {
    "update": "...",
    "link": {
      "author": "...",
      "$default": "..."
    }
  }
}
```

This PR also updates documentation for link/unlink on our website.

### DIFF
--- a/client/www/pages/docs/permissions.md
+++ b/client/www/pages/docs/permissions.md
@@ -224,8 +224,7 @@ You can use `$default` as the namespace:
 
 You can specify default link/unlink permission (these will apply to all attributes) and override it for specific attributes:
 
-```
-json
+```json
 {
   "users": {
     "allow": {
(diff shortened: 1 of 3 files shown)


client/www/pages/docs/permissions.md
```diff
@@ -206,6 +222,22 @@ You can use `$default` as the namespace:
 }
 ```
 
+You can specify default link/unlink permission (these will apply to all attributes) and override it for specific attributes:
+
+```
+json
```
nezaj:
```suggestion
```json
```

nezaj (approved):
LGTM w/ one nit!

PR #1610, opened 2025-09-03
Orgs Part 2

dwwoelfel:
This adds invites to orgs that match closely with the app invites. It reuses all of the route handlers and replaces the `app-member-invites` namespace with a `member-invites` namespace that works for both apps and orgs.

It still uses two separate tables. I considered putting `app_member_invites` and `org_member_invites` into a shared `member_invites` table, but having two foreign keys, one of which would be null on each row, seemed like a hassle and prone to error. It could probably be done with the right constraints, though.

Adds a route for getting all of the apps, invites, and members for the org, similar to `GET /dash`.

Adds the ability to create an app on an org.


client/www/components/dash/Orgs.tsx
```diff
@@ -45,22 +93,43 @@ export default function Orgs() {
 
   const orgs = dashResponse.data.orgs;
 
+  console.log(expandedOrgs);
```
nezaj, line 96: Nit. Guessing we don't want this?
```suggestion
```

dwwoelfel: This is just a debugging page, so I left it here to make it easier to debug.

nezaj (approved):
Awesome work! Love the tests!

PR #1613, opened 2025-09-03
Denormalize subscription_id on app

dwwoelfel:
This PR makes it easier to determine if an app is on a paid plan.

Instead of querying for the newest row in the instant_subscriptions table, we update the app with the subscription_id when we create it.

Includes a migration to update all of the existing apps. If someone subscribes between the time I run the migration and the time that I deploy the code, I'll run the migration again to update their row.


nezaj (approved):
SGTM!

PR #1614, opened 2025-09-03
Fix subscription error

dwwoelfel:
We were getting an error processing subscriptions because the user_id was a string instead of a uuid. 

I made it a uuid and also cast it to a uuid in the query for good measure.


nezaj (approved):

stopachka (approved):
Nice!

PR #1619, opened 2025-09-04
Subscriptions for orgs

dwwoelfel:
Adds subscriptions for orgs that closely match the structure of subscriptions for apps.

We reuse the same `instant_subscriptions` and `instant_stripe_customers` tables. Each table has a new `org_id` column and a constraint that ensures they have exactly one of `app_id` or `org_id` (sub `user_id` for `app_id` in the stripe_customers table).

We tie the stripe customer to an org instead of a user so that if a user leaves the org, we don't lose billing for the org. I added a billing_email field to the orgs table so that users could set a generic email (e.g. billing@) (no route for updating that yet).


stopachka (approved):
Heeck yeah!

PR #1622, opened 2025-09-04
[conc] set limit to 20

stopachka:
Our tasks were often getting stuck in dev. Adding --concurrency=20 _seems_ to fix it.

See: https://github.com/vercel/turborepo/issues/7924#issuecomment-2129262139

By default turbo has a concurrency limit of 10. When the `parallel` flag is turned on, theoretically the `concurrency` option is ignored. But as this user has reported, that is not the case. 

Aside: 
We may want to remove some of our dev tasks -- i.e the nuxt template, et al.

@drew-harris @dwwoelfel @nezaj @tonsky

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -13,7 +13,7 @@
   },
   "scripts": {
     "build": "turbo run build --cache-dir=.turbo --no-update-notifier",
-    "dev": "turbo run dev --filter=!@instantdb/mcp --parallel --no-update-notifier",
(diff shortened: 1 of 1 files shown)


drew-harris:
agreed on removing some dev tasks

maybe a dev:all script if thats needed otherwise keep it minimal for the 95% of things we are working on

nezaj (approved):
🚢

PR #1624, opened 2025-09-05
Clean up old flags

tonsky:
Namespaces to delete after publish in `instant-config`:

- use-patch-presence
- refresh-skip-attrs
- store-fair-lock
- threading
- view-checks
- hazelcast
- drop-refresh-spam
- app-users-to-triples-migration


tonsky:
Removed all flags from prod

PR #1626, opened 2025-09-05
[order] support optimistic updates for `order: desc`

stopachka:
## Context

Consider a table `docs` with `title` and `updatedAt`

```js
{ docs: { $: { order: { updatedAt: 'desc' } } } }
```

Say this returns the following docs:

```js
{ id: doc2id, title: Doc 2, updatedAt: 2 }
{ id: doc1id, title: Doc 1, updatedAt: 1 }
```

The `pageInfo` in this case would be: 

```js
{ 'start-cursor': [doc2id, updatedAtAttrId, 2, ts] }
```

## Problems

We have two problems here: 

### 1) Failed optimistic update for doc3

Imagine if the user adds `Doc 3, updatedAt: 3`

The user would expect Doc 3 to show up right away. **However, right now Doc3 would get filtered out.** 

Why?

Because `pageInfo` still says `[doc2id, updatedAt, 2]`

And in this case updatedAt: 3 would fall out of bounds for this pageInfo. Look at [isBefore](https://github.com/instantdb/instant/blob/main/client/packages/core/src/instaql.js#L533):

```js
updatedAt 3 > updatedAt 2 // therefore it is 'before' this start cursor
```

### 2) Flickering UI for an update to doc2

Now consider what happens if you change the `updatedAt` for doc2

The user would expect Doc2 to have a new updatedAt. 

However, right now Doc2 would get filtered out

Why?

This is for the same reason as our first problem. Doc2 would get an optimistic update, and “fall out” of the cursor.

## Root Cause

The problem is that though our result set changes with optimistic updates, our cursors are static -- we wait for the server to tell us about the cursors.

## Solution

We can use the fact that this is a 'leading' query to our advantage. The query is "at the beginning" of it's search. So if we get any value that 'fits' in the beginning, then we can safely apply the optimistic update

## Future work

Imagine two simultaneous queries: `Query 1: offset: 0, limit: 10` `Query 2: offset: 10, limit: 10`

We can get an optimistic update on Query 1, but we can't for Query 2 (it is not leading). 

However, since we _do_ have the dataset needed on the client, technically we _could_ make optimistic updates work. We would need to be much smarter about tracking than we are right now though. Likely it would require something more similar to Single Store.

--- 

### DIFF
--- a/client/packages/core/__tests__/src/instaql.test.js
+++ b/client/packages/core/__tests__/src/instaql.test.js
@@ -887,6 +887,94 @@ test('pagination first', () => {
   expect(books.length).toEqual(10);
 });
 
+test('Pagination order:desc without offsets work with optimistic updates', () => {
+  function storeWithUpdatedNicole() {
+    const chunk = tx.users[lookup('handle', 'nicolegf')].update({
+      createdAt: '2025-09-05 18:53:07.993689',
+    });
+
+    const txSteps = instaml.transform({ attrs: store.attrs }, chunk);
+    return transact(store, txSteps);
+  }
+
+  function storeWithBob() {
+    const chunk = tx.users[randomUUID()].update({
+      fullName: 'bob',
+      email: 'bob@instantdb.com',
+      handle: 'bob',
+      createdAt: '2025-09-05 18:53:07.993689',
+    });
+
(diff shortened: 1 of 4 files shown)


dwwoelfel (approved):
Nice!

PR #1628, opened 2025-09-05
Clean separation of personal apps and org apps

dwwoelfel:
This PR creates a cleaner separation between apps in orgs and personal apps.

Requests to `/dash`:
  - `apps` will only return the apps that are not in an org
  - `orgs` will return all paid orgs you are a member of + any orgs where you are a member of a paid app. If the app you're a member of is paid or if the org is paid, then you'll be able to see the base org details.

Requests to `/dash/orgs/:org-id`
  - If you are a member of the org, then you can see everything (org, apps, members, and invites)
  - If you are a member of any app in the org (paid app or paid org), then you can see the org and the apps you're a member of. `members` and `invites` will be empty

cc @drew-harris

### DIFF
--- a/server/src/instant/model/org.clj
+++ b/server/src/instant/model/org.clj
@@ -49,8 +49,11 @@
                                                                 :o.created-at
                                                                 :o.updated-at
                                                                 [[:inline "app-member" :role]]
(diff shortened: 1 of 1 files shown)


nezaj (approved):
Nice!

PR #1629, opened 2025-09-06
Add ability skip permissions checks for cascade delete

nezaj:
Adds a `skipCascadePermissionCheck` configuration option to InstantDB that allows cascading deletes to bypass permission checks on related entities while still enforcing permissions on the primary entity being deleted.

### Problem

When using cascade deletes with restrictive permissions on related entities, legitimate delete operations would fail. For example, if comments have `delete: false` permissions but are set to cascade-delete when their parent post is deleted, deleting the post would fail even though the user has permission to delete the post itself.

### Solution

- Added `skipCascadePermissionCheck` boolean option to the init configuration
- Modified server-side transaction processing to track original vs cascaded deletes
- When enabled, cascaded deletes skip permission checks while original deletes are still validated
- Preserves security by only bypassing checks for system-initiated cascade operations, not user-initiated deletes

You can enable just by doing `init({ ..., skipCascadePermissionCheck: true})`

Created side-by-side ephemeral app demo showing the difference in behavior with and without the flag!

https://github.com/user-attachments/assets/315bf0b7-eed5-4bf4-81e2-8ae66f9ae43b

### DIFF
--- a/client/sandbox/react-nextjs/pages/play/cascade-permission-demo.tsx
+++ b/client/sandbox/react-nextjs/pages/play/cascade-permission-demo.tsx
@@ -207,7 +207,9 @@ function AppInstance({
 
       setSuccess('Created post with 2 comments');
     } catch (err) {
-      setError(`Failed to create: ${err instanceof Error ? err.message : String(err)}`);
+      setError(
+        `Failed to create: ${err instanceof Error ? err.message : String(err)}`,
+      );
     } finally {
       setLoading(false);
     }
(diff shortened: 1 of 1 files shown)


tonsky:
I have another idea: what if we add new “cascadeDelete” permission? Which defaults to normal “delete” but you can also set it to “true” and get the behavior that you described here?

nezaj:
Closing this and continuing on in https://github.com/instantdb/instant/pull/1639

PR #1632, opened 2025-09-08
Replace core.cache with Caffeine

tonsky:
core.cache is based on persistent data structure, which have subpar performance and create a lot of garbage, thus producing unnecessary pressure on GC

Caffeine is one of the most popular caches in Java world. Expect better performance, less garbage, better eviction algorithm, and nicer APIs (e.g. I was able to replace both MultiEvictLRU, lookup-or-miss, LRU + TTL, and lookup-or-miss-batch with very clean and straightforward implementations)

I also bumped attr-cache and app-cache to 512 instead of 256, feels like it’s time (monthly active apps sit around 512, so it’s safer + I doubt these things take too much memory anyways)

### DIFF
--- a/server/test/instant/util/cache_test.clj
+++ b/server/test/instant/util/cache_test.clj
@@ -63,7 +63,7 @@
 (deftest lookup-or-miss-batch-rejects-errors
   (let [cache (cache/make {:max-size 2})]
     (is (thrown? Exception
-                 (cache/get-all cache [:instant] (fn [_]
+                 (cache/get-all-sync cache [:instant] (fn [_]
                                                    (throw (Exception. "oops"))))))
 
     (is (= {:instant :instant} (cache/get-all cache [:instant] (fn [x] (zipmap x x)))))))
--- a/server/src/instant/util/cache.clj
+++ b/server/src/instant/util/cache.clj
@@ -21,6 +21,23 @@
     value-fn       (.build value-fn)
     (not value-fn) (Caffeine/.build)))
 
+(defn invalidate
+  "Discards any cached value for the key. The behavior of this operation is
+   undefined for an entry that is being loaded (or reloaded) and is otherwise
+   not present"
+  [^Cache cache key]
+  (when (some? key)
+    (.invalidate cache key)))
+
+(defn invalidate-all
+  "invalidate-all/1 invalidates all keys.
+   invalidate-all/2 invalidates only passed keys."
+  ([^Cache cache]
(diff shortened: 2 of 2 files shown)


server/test/instant/util/cache_test.clj
```diff
@@ -1,161 +0,0 @@
-(ns instant.util.cache-test
```
dwwoelfel, line 1: Can we add these tests back for the new cache?

The stampede tests, the tests that we reject errors from the cache, and the test that batched-get returns things not in the cache are all important. I think the only one that's not interesting is the `wrapped-ttl-test` test.

tonsky: These would be testing caffeine itself. Tests are usually needed for stuff that’s expected to change. We are not expected to modify caffeine, and I don’t think we should test it

tonsky: I mean, I did check that errors don’t persist, but I don’t think we need a test that checks that this fact doesn’t change over time (why would it?)

> batched-get returns things not in the cache

This is literally part of the javadoc. Again, why would we want to check external library with each test run? I can understand end-to-end tests, but unit?

tonsky: @dwwoelfel I restored the file but some tests work differently and I’m not sure what were you trying to test. Can you take a look please and fix/keep only parts that you think are relevant?

server/test/instant/util/cache_test.clj
```diff
@@ -1,129 +1,115 @@
...
-      (is (= :instant (ucache/lookup-or-miss cache-atom :instant (fn [x] x)))))))
+    (is (= :instant (cache/get cache :instant (fn [x] x))))))
 
 (deftest lookup-or-miss-batch-rejects-errors
-  (let [cache-atom (cache/lru-cache-factory {} :threshold 2)]
+  (let [cache (cache/make {:max-size 2})]
     (is (thrown? Exception
-                 (ucache/lookup-or-miss-batch cache-atom [:instant] (fn [_]
-                                                                      (throw (Exception. "oops"))))))
+                 (cache/get-all cache [:instant] (fn [_]
```
dwwoelfel: I think this test will fail if `get-all-sync` is called. There's nothing to evict the delay with the error from the cache.

tonsky: You are absolutely right! Here’s the revised version that evicts errors from the cache.

tonsky:
Merged as part of #1644

PR #1634, opened 2025-09-08
Keep track of triples size in attr_sketches

dwwoelfel:
This PR will allow us to replace our expensive query to determine an app's database size with a much more efficient query.  O(triples_count) -> O(attrs_count).

Query before:

```sql
select sum(pg_column_size(triples)) from triples where app_id = :app_id
```

Query after: 
```sql
select sum(pg_triples_size) from attr_sketches where app_id = :app_id
```

We add two new columns: 1. pg_size to triples and 2. triples_pg_size to the attr sketches.

There is a new statement-level trigger that will populate the `pg_size` after insert/update on each triple. Then the aggregator process will pick that up and add/subtract it from the `triples_pg_size` on the attr sketch for the triple's attr_id.

It adds a little extra work on each transaction that affects triples, but I saved a little bit of work by converting the files sweeper from a row-level trigger to a statement-level trigger. We handle all of the rows for a single transaction all at once instead of running the trigger for each row. In my testing, it costs about 3ms/1000 triple rows.

Another way to solve the problem would be to skip the trigger and just calculate the size of the triple row in the aggregator. I couldn't find a way to reliably match postgres' pg_column_size. The value gets compressed when it gets toasted and I wasn't able to replicate that.

The deploy plan is a little complicated:

- [ ] `make prod-up-one` 
  - Runs migration 83 to add the new columns
- [ ] Deploy the backend and wait for all old machines to die
- [ ] `create index concurrently if not exists triples_pg_size_idx on triples (pg_size) where pg_size is null;`
  - Create the index for the migration concurrently
- [ ] `make prod-up`
  - Runs migration 84, which will noop since we already add the index, and migration 85 which will add the triggers
- [ ] Update the config toggle `run-pg-size-migration` to true and monitor to make sure it's not too much load on the db
  - If it is, we have two config flags to modify: 
    - `migrate-pg-size-delay-ms`: adds a pause between batches
    - `migrate-pg-size-batch-size`: defaults to 1000, we can raise or lower
- [ ] Later, after the migration finishes, drop the index on pg_size and remove the migration code

### DIFF
--- a/server/src/instant/db/indexing_jobs.clj
+++ b/server/src/instant/db/indexing_jobs.clj
@@ -50,7 +50,8 @@
                          :ms-since-creation (ms-since (:created_at job) now)
                          :ms-since-update (ms-since (:updated_at job) now)}
                         (when (and (= "completed" (:job_status job))
-                                   (:done_at job))
+                                   (:done_at job)
+                                   (:created_at job))
                           {:run-time-ms (ms-since (:created_at job)
                                                   (.toInstant ^Timestamp (:done_at job)))}))}))
 


server/resources/migrations/85_triples_size_trigger.up.sql
```diff
@@ -0,0 +1,118 @@
...
+
+create or replace function triples_update_batch_trigger()
+returns trigger as $$
+begin
+  -- Don't let this trigger cause itself to fire. We let it fire
+  -- twice because postgres 16 has some bug (fixed in 17) where
+  -- pg_column_size on insert is different than pg_column_size on
+  -- update, possibly due to how it handles nulls? If that happens,
+  -- then the second time it fires it will get the right value.
+  if pg_trigger_depth() > 2 then
```
dwwoelfel, line 59: We can make this a little faster in postgres 17 because we won't need to allow a second run though the trigger.

stopachka: Nice!

server/src/instant/util/hsql.clj
```diff
@@ -1,6 +1,60 @@
 (ns instant.util.hsql
   (:require [honey.sql :as hsql]))
 
+;; ----------------------
```
dwwoelfel, line 4: I moved registering the pg-hints clause from the `pg-hints` ns to the hsql util ns so that we would be sure to have it defined for preformatted queries.

PR #1636, opened 2025-09-09
Propogate rule params on cascade delete

nezaj:
Fixes cascade deletes to properly inherit `ruleParams` from their parent entities, enabling dynamic permission contexts to work correctly with cascade operations.

### Problem

When deleting entities with cascade relationships, the `ruleParams` passed for the parent deletion weren't propagated to cascaded children. This caused permission checks to fail for legitimate cascade operations. For example, if both posts and comments require `data.localId == ruleParams.localId` for deletion, deleting a post with `ruleParams: { localId: "user123" }` would fail because the cascaded comment deletes didn't receive those parameters.

### Solution

- Modified the cascade delete SQL query to track parent-child relationships through a `parent_id` column in the recursive CTE
- Each cascaded entity now inherits the `ruleParams` from its specific parent that triggered the cascade
- Supports multiple parent deletions with different `ruleParams` in a single transaction - each cascade chain maintains its parent's permission context independently

This enables patterns like:
```javascript
// Each post's comments inherit their parent's ruleParams
db.transact([
  db.tx.posts[postA].ruleParams({ localId: "alice" }).delete(),
  db.tx.posts[postB].ruleParams({ localId: "bob" }).delete()
])
```


stopachka (approved):
SGTM!

PR #1638, opened 2025-09-10
Return org owner as creator for org apps

dwwoelfel:
This PR does two things:

1. When we fetch the app's creator on an org app, we return the org's creator.
2. We prevent an org from removing its last owner.

We accomplish (2) with a trigger on org_members. It prevents deleting or updating the role on a member if the change would remove the last owner (unless the transaction also deletes the org).


stopachka (approved):
Heeck yeah!

PR #1639, opened 2025-09-10
Add `skipCascadePermissionCheck` rule for cascade delete operations

nezaj:
Based on our discussion, adds a new permission rule `skipCascadePermissionCheck` that allows parent entities to bypass permission checks on their cascaded deletions

### Problem

When using cascade deletes with restrictive permissions on related entities, legitimate delete operations would fail. For example, if comments have `delete: "false"` permissions but are set to cascade-delete when their parent post is deleted, deleting the post would fail even though the user has permission to delete the post itself.

### Solution

- Added `skipCascadePermissionCheck` as a permission rule that can be evaluated like any other CEL expression
- Modified `expand-delete-entity-cascade` to mark cascaded deletes
- Evaluates `skipCascadePermissionCheck` for parent entities before running permission checks
- Cascaded deletes skip permission checks only if their parent's `skipCascadePermissionCheck` rule passes
- Supports both entity-specific and `$default` rules with proper fallback behavior
- Adds type for our dashboard editor and in `ruleTypes` for our packages

### Usage

```javascript
// Simple boolean flag
const perms = {
  posts: {
    allow: {
      delete: "auth.id == data.creatorId",
      skipCascadePermissionCheck: "true"
    }
  },
  comments: {
    allow: {
      delete: "false"  // Comments can't be deleted directly, but cascade with posts
    }
  }
};

// With permission expression
const perms = {
  posts: {
    allow: {
      delete: "auth.id == data.creatorId",
      skipCascadePermissionCheck: "auth.id == data.creatorId"  // Only skip for post owner
    }
  }
};

// Using $default
const perms = {
  $default: {
    allow: {
      skipCascadePermissionCheck: "auth.ref('$users.role.type') == 'admin'"  // Admins skip cascade checks
    }
  }
};
```

Also verified this works with the play example (included in PR)!

If this looks good will bump up the version and update docs!

### DIFF
--- a/server/src/instant/db/permissioned_transaction.clj
+++ b/server/src/instant/db/permissioned_transaction.clj
@@ -193,190 +193,221 @@
     (transient entities-map)
     tx-steps)))
 
+(defn evaluate-skip-cascade-checks
+  "Evaluates skipCascadePermissionCheck rules for parent deletes and returns 
+   a set of ids who's children don't need to be checked"
+  [{:keys [rules] :as ctx} entities-map rule-params-map tx-steps]
+  (let [parent-delete-checks (for [{:keys [op eid etype cascade-delete?]} tx-steps
+                                   :when (and (= :delete-entity op)
+                                              (not cascade-delete?))
+                                   :let [skip-expr (get-in (:code rules)
+                                                           [etype "allow" "skipCascadePermissionCheck"])]
+                                   :when skip-expr]
+                               {:etype    etype
+                                :eid      eid
+                                :program  (rule-model/get-program! rules
+                                                                   [[etype "allow" "skipCascadePermissionCheck"]])
+                                :bindings {:data (get entities-map {:eid eid :etype etype})
+                                           :rule-params (get rule-params-map {:eid eid :etype etype})}})
+        ;; Evaluate the skip rules (but don't throw on failure)
+        results (when (seq parent-delete-checks)
+                  (cel/eval-programs! (assoc ctx :preloaded-refs (cel/create-preloaded-refs-cache))
+                                      parent-delete-checks))]
+    (set (for [{:keys [eid result]} results
+               :when result]
+           eid))))
+
 (defn pre-checks
   "Checks that run before tx: update, delete for attrs & objects"
-  [{:keys [attrs admin? rules]}
+  [{:keys [attrs admin? rules] :as ctx}
    entities-map
    updated-entities-map
    rule-params-map
    tx-steps]
-  (for [tx-step tx-steps
-        :let [{:keys [op eid aid etype value rev-etype]} tx-step
-              key                {:eid eid :etype etype}
-              entity             (get entities-map key)
-              ref?               (some? rev-etype)
-              [_ _ fwd-label]    (:forward-identity (attr-model/seek-by-id aid attrs))
-              [_ _ rev-label]    (:reverse-identity (attr-model/seek-by-id aid attrs))
-              rev-key            {:eid value :etype rev-etype}
-              rev-entity         (when ref?
-                                   (get entities-map rev-key))
-              rule-params        (get rule-params-map key)
-              link-program       (when ref?
-                                   (rule-model/get-program! rules [[etype "allow" "link" fwd-label]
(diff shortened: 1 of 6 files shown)


server/src/instant/db/permissioned_transaction.clj
```diff
@@ -193,190 +193,222 @@
...
-                [{:scope    :attr
-                  :action   :restore
-                  :etype    "attrs"
-                  :program  {:result admin?}}]
+  (let [;; IDs of parent entities whose children should skip permission checks
+        skip-cascade-parents (evaluate-skip-cascade-checks ctx entities-map rule-params-map tx-steps)
+        ;; Filter our cascade deletes that don't need checks
+        should-skip? (fn [tx-step]
+                       (and (:cascade-delete? tx-step)
+                            (contains? skip-cascade-parents (:cascade-parent-id tx-step))))]
```
tonsky, line 233: You should look up by parent-id + parent-etype, just id is not enough

server/src/instant/db/permissioned_transaction.clj
```diff
@@ -193,190 +193,222 @@
...
+                                                        ["$default" "allow" "skipCascadePermissionCheck"]])]
+                                   :when skip-program]
+                               {:etype    etype
+                                :eid      eid
+                                :program  skip-program
+                                :bindings {:data (get entities-map {:eid eid :etype etype})
+                                           :rule-params (get rule-params-map {:eid eid :etype etype})}})
+        ;; Evaluate the skip rules (but don't throw on failure)
+        results (when (seq parent-delete-checks)
+                  (cel/eval-programs! (assoc ctx :preloaded-refs (cel/create-preloaded-refs-cache))
```
tonsky, line 215: Should preload cache be shared with pre-checks? Technically they both look at the same version of database, so no reason to fetch stuff twice

server/src/instant/db/permissioned_transaction.clj
```diff
@@ -193,190 +193,222 @@
     (transient entities-map)
     tx-steps)))
 
+(defn evaluate-skip-cascade-checks
+  "Evaluates skipCascadePermissionCheck rules for parent deletes and returns 
+   a set of ids who's children don't need to be checked"
```
dwwoelfel, line 198:
```suggestion
   a set of ids whose children don't need to be checked"
```

server/src/instant/db/permissioned_transaction.clj
```diff
@@ -193,190 +193,222 @@
...
+  [{:keys [rules] :as ctx} entities-map rule-params-map tx-steps]
+  (let [parent-delete-checks (for [{:keys [op eid etype cascade-delete?]} tx-steps
+                                   :when (and (= :delete-entity op)
+                                              (not cascade-delete?))
+                                   :let [skip-program (rule-model/get-program!
```
dwwoelfel, line 203: Could we put the logic for determining the skip cascade check into rule model? Something like:

```clj
// rule-model.clj
(defn get-skip-cascade-permission-check [rules etype]
  (get-program! rules ...))
```

Figuring out where the rule is defined seems like an implementation detail that the rule-model namespace should be aware of and not the transact namespace.

server/src/instant/db/permissioned_transaction.clj
```diff
@@ -193,190 +193,222 @@
     (transient entities-map)
     tx-steps)))
 
+(defn evaluate-skip-cascade-checks
```
dwwoelfel, line 196: I think it might be more efficient overall to only do one round through `cel/eval-programs`, even if it means doing extra permission checks for the skipped entities that we throw away later.

Right now the process is (please correct me if I'm reading this wrong):

1. Run the `skipCascadePermissionCheck` permission check
2. Filter out the entities that don't need to be checked
3. Run the remaining permission checks

With the following process, we can avoid the extra round-trip to the database
1. Run the permission checks + the skipCascadePermissionCheck
2. Replace the result for the entities that don't need to be checked with "true"

It also might be nice for observability later to be able to tell a user "your delete would have failed the permission check, but it was bypassed due to the `skip` check". We could also return that information in `transact-perms-check`

nezaj:
Will close this for now to pick up later

PR #1641, opened 2025-09-10
link/unlink refinement take 3

tonsky:
- Removed newData.ref (it wasn’t really working because it feels like it should run on the new version of the DB but in fact it runs on the old one)


tonsky:
Also added:

- Allow specifying ruleParams on either side of the link (from @stopachka friction log)

PR #1642, opened 2025-09-10
Add json encoders for cel types

dwwoelfel:
Adds JSON encoders for cel types so that you can debug cel rules in the sandbox.

Example of it in action: 

The rule is set to `data` and you can see the data in the output of the rule on the right.

image


stopachka (commented):
Biyutiful!

PR #1643, opened 2025-09-11
Trace in-flight-keys in attr-sketch

tonsky:
#1632 started a discussion whether we should make our cache synchronous or not (see https://discord.com/channels/1031957483243188235/1414690924864344134)

This PR adds a metric to see how many keys are actually requested while they are being queried from Postgres


tonsky:
Added as a part of #1644

PR #1644, opened 2025-09-11
Caffeine async

dwwoelfel:
Adds `make-async` that uses an async cache from caffeine and writes -async methods for all of the cache operations.

`AsyncCache` doesn't inherit from `Cache`, so I wrote a whole new set of functions for it. It might be nicer to write our own wrapper for a cache so that we don't need two functions for everything, but I didn't want to get bogged down.

### DIFF
--- a/server/src/instant/db/attr_sketch.clj
+++ b/server/src/instant/db/attr_sketch.clj
@@ -338,7 +338,10 @@
    {:max-size 4096
     :ttl      (* 1000 60 5)}))
 
-(defn lookup* [conn keys]
+(def in-flight-keys
+  (atom #{}))
+
+(defn lookup*-impl [conn keys]
   (if-not (seq keys)
     {}
     (let [params {:params {:app-ids (with-meta (mapv :app-id keys) {:pgtype "uuid[]"})
@@ -357,6 +360,12 @@
        {}
        rows))))
 
+(defn lookup* [conn keys]
+  (swap! in-flight-keys into keys)
+  (let [res (lookup*-impl conn keys)]
+    (swap! in-flight-keys #(reduce disj % keys))
+    res))
+
 (defn lookup
   "Takes a set of {:app-id attr-id} maps and fetches sketches, if they exist.
    Returns a map with key {:app-id :attr-id} and value :sketch-record"
@@ -364,7 +373,13 @@
    (lookup (aurora/conn-pool :read) keys))
   ([conn keys]
    (if (= conn (aurora/conn-pool :read))
-     @(cache/get-all-async lookup-cache keys #(lookup* conn %))
+     (do
+       (tracer/with-new-trace-root
+         (tracer/with-span! {:name "attr-sketch/in-flight-keys"
+                             :attributes {:count     (count keys)
+                                          :in-flight (count (filter @in-flight-keys keys))}}
+           nil))
+       @(cache/get-all-async lookup-cache keys #(lookup* conn %)))
      (lookup* conn keys))))
 
 (defn- create-empty-sketch-rows!
--- a/server/src/instant/reactive/aggregator.clj
+++ b/server/src/instant/reactive/aggregator.clj
@@ -119,7 +119,7 @@
                                        ref-k (assoc! ref-k 1))
                                      (cms/add-batch sketch (persistent! triples))
                                      (cms/add-batch reverse-sketch (persistent! reverse-triples))
-                                     (+ triples-pg-size (:pg-size triple)))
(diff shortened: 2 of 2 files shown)


server/src/instant/util/cache.clj
```diff
@@ -21,6 +24,21 @@
...
+   :ttl        <number>          eviction time, ms
+   :on-remove  <fn [k v cause]>  removal listener"
+  ^AsyncLoadingCache [{:keys [max-size max-weight weigher ttl on-remove]}]
+  (cond-> (Caffeine/newBuilder)
+    max-size   (.maximumSize max-size)
+    max-weight (.maximumWeight max-weight)
+    weigher    (.weigher weigher)
+    ttl        (.expireAfterWrite (Duration/ofMillis ttl))
+    on-remove  (.removalListener on-remove)
+    true       (Caffeine/.buildAsync)))
```
tonsky, line 40: This is missing value-fn, I think it should be here for feature parity

server/src/instant/util/cache.clj
```diff
@@ -21,6 +24,21 @@
     value-fn       (.build value-fn)
     (not value-fn) (Caffeine/.build)))
 
+(defn make-async
```
tonsky, line 27: Why do we need separate function instead of just a key in options map?

server/src/instant/util/cache.clj
```diff
@@ -29,6 +47,14 @@
   (when (some? key)
     (.invalidate cache key)))
 
+(defn invalidate-async
+  "Discards any cached value for the key. The behavior of this operation is
+   undefined for an entry that is being loaded (or reloaded) and is otherwise
+   not present"
+  [^AsyncCache cache key]
+  (when (some? key)
+    (.invalidate (.synchronous cache) key)))
```
tonsky, line 56: Why duplicate every method if we can just have one `as-synchronous` that converts from async to sync and then all the normal methods work and we don’t need to duplicate everything?

server/src/instant/util/cache.clj
```diff
@@ -51,12 +86,30 @@
...
 
+(defn get-async
+  "Returns a completeable future with the value associated with the key
+   in this async cache. This method provides a simple substitute for the
+   conventional “if cached, return; otherwise create, cache and return”
+   pattern."
+  [^AsyncLoadingCache cache key ^Function value-fn]
+  (if (some? key)
+    (.get cache key value-fn)
+    (CompletableFuture/completedFuture nil)))
```
tonsky, line 97: What’s the use-case for getting completable future from the cache instead of actual value?

server/src/instant/util/cache.clj
dwwoelfel: The async builder doesn't support value-fn as far as I can tell.

server/src/instant/util/cache.clj
dwwoelfel: It has a different return type (AsyncCache instead of a Cache)

server/src/instant/util/cache.clj
dwwoelfel: I don't know how to make the type hints work for that, since `AsyncCache` doesn't inherit from `Cache`.

server/src/instant/util/cache.clj
dwwoelfel: Good point, might as well just deref here instead of having the caller do that. Will update.

server/src/instant/util/cache.clj
tonsky: https://javadoc.io/static/com.github.ben-manes.caffeine/caffeine/3.2.2/com.github.benmanes.caffeine/com/github/benmanes/caffeine/cache/Caffeine.html#buildAsync(com.github.benmanes.caffeine.cache.CacheLoader) ?

server/src/instant/util/cache.clj
tonsky:
```
(defn as-synchronous ^Cache [^AsyncCache cache]
  (.synchronous cache))
```

server/src/instant/util/cache.clj
tonsky: Why is that a problem?

tonsky: Actually I don’t want to split hairs, if you think it should be two methods let it be two methods

server/src/instant/util/cache.clj
tonsky: The idea is clients would do

```
(cache/invalidate (cache/as-sync async-cache))
```

instead of

```
(cache/invalidate-async async-cache)
```

Up to you

PR #1645, opened 2025-09-11
[docs] update permissions

stopachka:
- Removed notes on link/unlink until we figure it out
- Updated `creatorId` to `creator`. (I found myself preferring creator when building manyminiapps)

@dwwoelfel @nezaj @tonsky @nezaj


stopachka:
Undid the creatorId change -- we can discuss and solve that separately!

PR #1659, opened 2025-09-12
Transfer app to org

dwwoelfel:
Adds the ability to transfer an app from a personal account to an org or from an org to another org.

If the app is paid and it's transferred to a paid org, then we'll cancel the plan for the app and credit the org with a credit for the remaining time on the app's subscription.

The transfer route is at `/dash/apps/:app_id/transfer_to_org/:org_id` and it returns a `credit` field. That field will have a (possibly null) negative number with the credit if there was a credit applied to the org.

The webhooks are aware of the difference between a user canceling a subscription and us canceling a subscription when the app is transferred. We'll see a message in discord that looks like:

```
Paid app `test team app` transferred to unpaid app in paid org `empty-org`
```

We'll also get notified in discord when we give a customer a credit.

Incidental changes:
1. I split the stripe namespace into `instant.stripe` and `instant.stripe-webhook` so that I could use the stripe helpers in the model code
2. I added a some test helpers to allow us to create subscriptions in stripe. Now `with-pro-app` and friends can add payment methods to customers and create subscriptions in stripe's test env.
3. Create a single reusable stripe client. I think this improves the latency for stripe api calls, maybe it creates an http connection pool or something.

Still to do in followup PRs:
1. Transfer members. This PR was big enough that I think it would be better to split it up.
2. Credits for paid apps when you add a new subscription to an org.


.github/workflows/clojure.yml
```diff
@@ -163,7 +164,7 @@ jobs:
         cli: 1.11.3.1463
 
     - name: Cache Clojure Dependencies
-      uses: actions/cache@v3
+      uses: actions/cache/restore@v4
```
dwwoelfel, line 167: With v4 we can do restore-only in the lint and build-check jobs. They don't download as many deps, so our cache was missing a lot of stuff that the tests needed.

server/deps.edn
```diff
@@ -78,7 +78,7 @@
 
         com.auth0/java-jwt {:mvn/version "4.4.0"}
         com.auth0/jwks-rsa {:mvn/version "0.22.1"}
-        com.stripe/stripe-java {:mvn/version "25.1.0"}
+        com.stripe/stripe-java {:mvn/version "29.5.0"}
```
dwwoelfel, line 79: New version includes the preview invoice API, which is how we determine how much of a credit to give the org.

server/src/instant/stripe.clj
```diff
@@ -191,23 +160,82 @@
...
+(defn subscription ^Subscription [subscription-id]
+  (-> (stripe-client)
+      (.subscriptions)
+      (.retrieve subscription-id)))
+
 (defn subscriptions []
   (let [params (-> (SubscriptionListParams/builder)
                    (.addExpand "data.discounts")
+                   (.setStatus SubscriptionListParams$Status/ACTIVE)
+                   (.setLimit 100)
```
dwwoelfel, line 172: Should help us fetch the subscriptions faster, previously it was fetching even canceled subscriptions, and only fetched 10 at a time.

server/src/instant/fixtures.clj
```diff
@@ -212,108 +213,118 @@
...
-              :creator-id (:id owner)
-              :id app-id
-              :admin-token (random-uuid)})
-        stripe-customer (instant-stripe-customer-model/get-or-create-for-user! {:user owner})
-        owner-req (mock-app-req app owner)
-        _ (instant-subscription-model/create!
-           {:user-id (:id owner)
-            :app-id app-id
-            :subscription-type-id 2 ;; Pro
+(defn with-pro-app [create-fake-objects? owner f]
```
stopachka, line 216: awesome!

PR #1663, opened 2025-09-13
Update create instant app title

nezaj:
I noticed when running create instant app the logo is slightly off. Regardless of whether we are in light or dark mode, we want to render the same title.

**Before**
CleanShot 2025-09-12 at 18 02 12@2x

**After**
CleanShot 2025-09-12 at 18 02 30@2x

This also works as expected for light mode

CleanShot 2025-09-12 at 18 08 09@2x


drew-harris:
Now that they are both the same, we should move down the terminal theme sequence query to be lower so it doesn't block the logo appearing -> faster experience

PR #1664, opened 2025-09-15
Test optimizations

dwwoelfel:
Adds a few optimizations for the tests:

1. Updates the order we check the rows in the triples_batch_update postgres function. We were doing `join_row.col = triples.col`, and postgres was using the triples_created_at index. Switching the order (`triples.col = join_row.col`) seems to make postgres use the primary key. Only seems to be an issue during the tests.
2. Reduce the batch size for some indexing job tests so that we don't have to update as many triples to test everything.
3. Fixes an error calculating the job span attrs when we don't get back a job.

Much easier to review with whitespace turned off.


tonsky:
Looks good!

PR #1667, opened 2025-09-15
Fewer js runs

dwwoelfel:
Only run the js jobs if something changed in client (or the js.yml workflow file).

Also removes a flaky cache test and makes the sweeper test less resource-intensive.


stopachka (approved):
Niice!

PR #1668, opened 2025-09-15
Only run clojure tests on push

dwwoelfel:
Skips the PR runs for clojure tests. 

I'm not sure they're all that useful because we already have the tests run on every push.


stopachka (approved):
🚊nice!

PR #1670, opened 2025-09-16
Cancel and credit pro app subscriptions when org upgrades

dwwoelfel:
Cancels all of an org's pro app subscriptions and credits the org when the org adds a startup plan.


stopachka (approved):
Niice!

PR #1671, opened 2025-09-16
Update order rules

nezaj:
Got reports of `order` errors with claude and though it would be helpful to add explicit rule about how to use order (before we just mentioned you should only use it on indexed fields)


dwwoelfel (approved):
LGTM!

PR #1672, opened 2025-09-16
Restrict updating and removing members

dwwoelfel:
Restrict updating and removing the org members to members that are at your level or below. That means that if you're an owner, you can remove anybody and update anyone's role, but if you're an admin you can't modify owners.


nezaj (approved):
Nice!

PR #1673, opened 2025-09-16
org ui!

drew-harris:
Refactors the dashboard to support user organizations. Most of the app controls are left untouched, created new pages for user settings and organization settings and the ability to switch organizations.


client/www/components/dash/org-management/OrgBilling.tsx
```diff
@@ -0,0 +1,147 @@
...
+          )}
+        </div>
+      </div>
+      <SectionHeading className="pt-8">Billing</SectionHeading>
+      {isFreeTier ? (
+        <div className="flex flex-col gap-2">
+          <Button variant="primary" onClick={onUpgrade}>
+            Upgrade to Pro
+          </Button>
+          <div className="italic text-sm w-full bg-purple-100 text-purple-800 rounded border border-purple-400 px-2 py-1">
```
drew-harris, line 145: This text needs to be different

client/www/components/dash/org-management/RenameOrg.tsx
```diff
@@ -0,0 +1,84 @@
...
+        if (prev.type === 'personal') {
+          return;
+        }
+        prev.org.title = value;
+        return prev;
+      },
+    );
+  };
+
+  if (dash.data.workspace.type === 'personal') {
```
drew-harris, line 47: replace with redirect

client/www/components/dash/CLILoginDialog.tsx
drew-harris, line 1: this is probably completely broken

drew-harris: its not!

client/www/components/dash/Billing.tsx
```diff
@@ -120,6 +124,11 @@ export default function Billing({ appId }: { appId: string }) {
     return <Loading />;
   }
 
+  if (orgIsPaid) {
```
dwwoelfel, line 127: Instead of redirecting, maybe we should show the billing page and just have a link to the org billing page?

That way it's easy to find the billing for an app and you're not confused why the billing page went away.

Would also be good to have some explanation somewhere on the billing page for org apps that they can either add a plan to each individual app or they can add a plan to the org for all apps. And explain that all members of an org have access to all of the apps in a paid org.

client/www/components/dash/MainDashLayout.tsx
```diff
@@ -0,0 +1,165 @@
...
+import Head from 'next/head';
+import { ReactNode, useEffect, useState } from 'react';
+import { FullscreenLoading } from '../ui';
+import { FullscreenErrorMessage } from '@/pages/dash';
+import { useAuthToken } from '@/lib/auth';
+import Auth from './Auth';
+import { TokenContext } from '@/lib/contexts';
+import { CLILoginDialog } from './CLILoginDialog';
+import { useTicketSystem } from '@/lib/hooks/useTicketSystem';
+import { createInitializedContext } from '@/lib/createInitalizedContext';
```
dwwoelfel: Missing an `i` in the file name: `createInitalizedContext` -> `createInitializedContext`

client/www/components/dash/org-management/OrgBilling.tsx
```diff
@@ -0,0 +1,147 @@
...
+            Pro offers 10GB of storage, backups, multiple team members for apps,
+            and priority support.
+          </div>
+        </div>
+      ) : (
+        <Button variant="primary" onClick={onManageBilling}>
+          Manage Pro subscription
```
dwwoelfel: For the plan on an org, it's called a "Startup" subscription instead of a "Pro" subscription.

client/www/pages/dash/index.tsx
```diff
@@ -417,123 +327,59 @@ function Dashboard() {
...
+          }
+        }),
+      );
+    } else {
+      await dashResponse.refetch();
+    }
     const _appId = _apps[0]?.id;
     nav({ s: 'main', app: _appId, t: 'hello' });
   }
-  if (screen === 'personal-access-tokens') {
```
dwwoelfel, line 466: We should have some backwards compatibility so that links to `/dash?s=personal-access-tokens` and other still go to the correct page.

client/www/pages/dash/index.tsx
```diff
@@ -219,6 +221,22 @@ function Dashboard() {
...
+  useEffect(() => {
+    if (screen === 'new') {
+      router.replace('/dash/new');
+    }
+    if (screen === 'invites') {
+      router.replace('/dash/user-settings?tab=invites');
+    }
+    if (screen === 'user-settings') {
+      if (tab === 'oauth-apps') {
+        router.replace('/dash/user-settings?tab=oauth');
```
dwwoelfel, line 236: Should this get a return afterwards?

PR #1675, opened 2025-09-16
Update stripe urls for org billing

dwwoelfel:
Updates the url from `/dash?t=billing&org=:org-id` to `/dash/org?tab=billing&org=:org-id`.

Also combines success and cancel url into a single return-url since we use the same url for both.


stopachka (approved):
Niice!

PR #1678, opened 2025-09-17
Allow .link calls without .update({})

tonsky:
During my experiments with task-tracker I found that we can’t currently write

```js
db.tx.namespace[id].link({ ... })
```

At the very least you have to write

```js
db.tx.namespace[id].udpate({}).link({ ... })`
```

This seems like an oversight

### DIFF
--- a/server/src/instant/admin/model.clj
+++ b/server/src/instant/admin/model.clj
@@ -93,14 +93,14 @@
     eid))
 
 (defn- with-id-attr [attrs etype eid steps]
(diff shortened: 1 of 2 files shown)


tonsky:
Damn! Will check. I always forget we have this code in multiple places. Would be great to unify it

tonsky:
@stopachka you were right, ported the changes there. Review?

stopachka (approved):
Lesgoo!

PR #1681, opened 2025-09-18
Fix optimistic update for entity that exists but triple with attr does not

tonsky:
Reported by Kosmik https://discord.com/channels/1031957483243188235/1417133699094548601

Core of the issue is strict update mode on the client was filtering out triples if entity existed but triple with previous value didn't. Observed result was that in case like this

```
db.transact(
  db.tx.stickers[id].create({ v: 0 });
);

db.transact(
  db.tx.stickers[id].update({ v: 1, resizable: true });
);
```

after second transact you were getting query results with `{ v: 1 }` but without `resizable`

### DIFF
--- a/client/packages/core/src/store.js
+++ b/client/packages/core/src/store.js
@@ -674,17 +674,24 @@ function findTriple(store, rawTriple) {
 }
 
 export function transact(store, txSteps) {
-  const txStepsFiltered = txSteps.filter(([action, ...rawTriple]) => {
+  const txStepsFiltered = txSteps.filter(([action, eid, attrId, value, opts]) => {
     if (action !== 'add-triple' && action !== 'deep-merge-triple') {
       return true;
     }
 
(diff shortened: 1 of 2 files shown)


client/packages/core/src/store.js
```diff
@@ -674,17 +674,24 @@ function findTriple(store, rawTriple) {
...
       return true;
     }
 
-    const exists = findTriple(store, rawTriple);
+    let exists = false;
+
+    const attr = getAttr(store.attrs, attrId);
+    if (attr) {
+      const idAttr = getPrimaryKeyAttr(store, attr['forward-identity'][1]);
+      exists = undefined !== findTriple(store, [eid, idAttr.id, eid]);
```
stopachka: Here could we write `!!findTriple(...` ?

stopachka (approved):
Nice find!

PR #1682, opened 2025-09-18
Disable stripe tests if stripe-secret is not set

tonsky:
According to @dwwoelfel, Stripe API is slow. It’s currently so slow that it adds extra minute to the test runtime, effectively doubling it (52 -> 107 sec)

It also seems unnecessary to test external API on every run. Hopefully Stripe is not changing it in backwards-incompatible way

So this change does two things:

- If `:stripe-secret` is not set (it was not set for me, possibly for others too), tests that rely on it are skipped, making overall test suite pass again
- Cuts local test time in half, which seems important for local dev (1 minute less wait __per test run__!)

### DIFF
--- a/server/test/instant/dash/routes_test.clj
+++ b/server/test/instant/dash/routes_test.clj
@@ -23,129 +23,128 @@
 (use-fixtures :each silence-routes-exceptions)
 
 (deftest app-invites-work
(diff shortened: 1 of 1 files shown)


dwwoelfel (approved):
LGTM!

PR #1684, opened 2025-09-19
Run clojure tests in parallel

dwwoelfel:
Splits our tests across 5 machines and runs them in parallel to speed up the tests.

The tests are sorted by how long they took on the last run on main and then evenly distributed among the test nodes. Each node runs 1/5 of the tests. The logic to choose the tests happens on each node, so it's important that they all share the same view of the test timings. We achieve that by fetching the timings in an earlier job and storing the timings in the workflow artifacts. Then each machine will pull the same timings data from the artifacts.

On a successful test run on `main`, we publish the test timings to an Instant app, overwriting the previous timings. It would be nice to have some kind of rolling average, but I didn't see an easy way to implement that.

Moves the test config from `dev-resources/circleci_test/config.clj` into `test_core.clj` so that we could specify the test vars to run.

I initially started to remove circleci.test, but it became a bigger project than I wanted for this PR, so it stays for now.

The [last run on this branch took 2:49](https://github.com/instantdb/instant/actions/runs/17849417070) vs [7:05 on the last run before this](https://github.com/instantdb/instant/actions/runs/17837673054).

I also fixed up annotations (we were including `test` in the file path twice before)

image

It even adds it to unchanged files

image


tonsky:
Very clever! Use instant to store test results of instant is also brilliant

PR #1686, opened 2025-09-19
Rollback links on new entities, keep links on existing entities without update

tonsky:
Stuff around implicit id creation turned out to be too hard for me, and I don’t want to spend too much time there. This rolls back this change:

```
db.tx.ns[id()].link({ ... })
```

You can still do this instead:

```
db.tx.ns[id()].create({}).link({ ... })
```

↑ this feels more natural anyway

or this (this wasn't possible before, was throwing)

```
db.tx.ns[existing-id].link({ ... })
```

All this as part of an effort to fix https://discord.com/channels/1031957483243188235/1417133699094548601
Rolls back parts of #1678, for context: #1681

### DIFF
--- a/client/sandbox/react-nextjs/pages/play/query-updates.tsx
+++ b/client/sandbox/react-nextjs/pages/play/query-updates.tsx
@@ -1,3 +1,8 @@
+/**
+ * See that updates to existing triple and non-existing triple
+ * when transacted together are applied together #1681 #1686
+ */
+
 import { i, init, tx } from '@instantdb/react';
 import { useEffect } from 'react';
 import config from '../../config';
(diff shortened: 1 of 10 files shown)


client/sandbox/react-nextjs/pages/play/query-updates.tsx
```diff
@@ -0,0 +1,244 @@
+import { i, init, tx } from '@instantdb/react';
```
stopachka, line 6: nit: would rename this file to something like `stickers`, to keep the users ~semi-private

client/sandbox/task-tracker/next-env.d.ts
```diff
@@ -0,0 +1,5 @@
+/// <reference types="next" />
+/// <reference types="next/image-types/global" />
```
stopachka: Was this intended?

PR #1699, opened 2025-09-22
Add data['$action] and linkedData['$action] to link permissions, add task-tracker to sandbox

tonsky:
As an attempt to make link/unlink permission checks more useful, we need to distinguish when objects are created vs when they already exist during permission check.

For example:

- When creating a project, I can set myself as admin. But only during create

During link, there are 4 possibilities. For example, project → admin

- project is new, user is new
- project is new, user exists
- project exists, user is new
- project exists, user exists

Potentially our clients might want to write permissions based on arbitrary combination of these properties. So it seemed that the most versatile solution is to expose them on `data` and `linkedData`. This way, if you don’t need them, your `link` permissions look simple. If you do need them, you can make your permissions as complicated as you need. Potentially, you can have different permissions for all 4 of the scenarios above.

Solution:

- During `link` check, you get access to `data['$action']` and `linkedData['$action']`
- Both of them can be `'create'` or `'update'` independently

This PR does not add any documentation, with the idea that we might want to test/play with it first.

This PR also add task-tracker project to the sandbox. This might be used to further test permissions. E.g., for setting admin, it has this check:

```js
projects: {
  bind: [
    'isAdmin',       'auth.id in data.ref("admins.id")',
    'isNewProject',  'data["$action"] == "create"',
    'linkingMyself', 'linkedData.id == auth.id',
  ],
  allow: {
    link: {
      admins: 'isNewProject ? linkingMyself : isAdmin',
    }
  }
}
```


client/sandbox/task-tracker/package.json
```diff
@@ -0,0 +1,29 @@
...
+  "devDependencies": {
+    "instant-cli": "workspace:*",
+    "@tailwindcss/postcss": "^4",
+    "@types/node": "^20",
+    "@types/react": "^19",
+    "@types/react-dom": "^19",
+    "tailwindcss": "^4",
+    "typescript": "^5"
+  },
+  "packageManager": "pnpm@10.15.1"
```
stopachka: Nit: I would remove this, just in case it causes issues. Would also set the same versions for next react et al as in `react-nextjs`.  

Main reason: sometimes turbo/pnpm screws up when it has multiple versions of the same library.

PR #1700, opened 2025-09-22
Make platform sdk aware of orgs

dwwoelfel:
Adds two new api functions to the platform sdk:

1. `api.getOrgs()`
  a. Returns all of the orgs that the authed user is a member of
2. `api.getAppsForOrg(orgId)`
  a. Returns a list of apps that the user has access to through the org
  
Also adds support for creating an app inside of an org. Just pass in a valid `orgId` to `api.createApp`.

Updates the permissions for delete app so that admins can delete apps in orgs.


stopachka (approved):
Nice!

PR #1702, opened 2025-09-22
Redirect to org if the app is in another org

dwwoelfel:
If you go directly to an org's `app` and it's not in the apps from the dash response, we'll find the org that the app belongs to and redirect to the app with the org.

Test plan:

1. Go to the url for an app in an org, copy the url and make sure to remove any `org` param.
2. Go to your personal account and click on an app. 
3. Paste in the url you copied and make sure that it landed on the same app from step 1

cc @drew-harris


drew-harris (approved):
didn't test for the exact scenario this fixes, but no issues on the orgs i have setup locally

PR #1705, opened 2025-09-23
Dash: Sort apps, clean up app styles

tonsky:
Report: https://discord.com/channels/1031957483243188235/1258878823529844766/1419961650487754773

Before:

Screenshot 2025-09-23 at 18 04 45

After:

Screenshot 2025-09-23 at 18 04 38

(also cleaned up styles a bit)

### DIFF
--- a/client/www/pages/dash/index.tsx
+++ b/client/www/pages/dash/index.tsx
@@ -731,6 +731,13 @@ function ExplorerTab({
   );
 }
 
+function compareAppTitles(a: InstantApp, b: InstantApp): number {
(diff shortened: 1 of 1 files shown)


client/www/pages/dash/index.tsx
```diff
@@ -751,6 +751,13 @@ function AppCombobox({
     ? apps.filter((a) => a.title.toLowerCase().includes(appQuery))
     : apps;
 
+  const sortedApps = filteredApps.toSorted((a, b) =>
```
dwwoelfel: A `useMemo` might be good here if we're re-rendering this a lot

PR #1706, opened 2025-09-23
Make org banner match app banner

dwwoelfel:
image


drew-harris:
very smart!!!! tested locally

PR #1711, opened 2025-09-23
better debugTransact function return type

drew-harris:
adds a return type to debugTransact in the admin sdk


client/packages/admin/src/index.ts
```diff
@@ -124,6 +124,22 @@ type ImpersonationOpts =
   | { token: AuthToken }
   | { guest: boolean };
 
+type DebugTransactResult = {
```
nezaj, line 127: Perhaps we should only gauranteee `tx-id` and `all-checks-ok?` so it's easier to be backwards compatible if we make changes in the future?

PR #1712, opened 2025-09-23
[client] reduce the number of packages we run by default

stopachka:
Our dev script used to run just about _every_ package in dev. This would cause slower builds. We only really need to work on a few core packages during development. So: 
- I made it so the `dev` script 
- only runs the core packages: `@instantdb/core`, `@instantdb/react`, `@instantdb/react-native`, `@instantdb/admin`
- and the two next sandboxes we use all the time: `sb-react-nextjs`, `www` 

If we want to add more packages, we can either include them in a filter: 

```bash
make dev --filter create-instant-app
```

Or cd into the project and run it directly

```bash
cd create-instant-app && npm run dev
```

Extra notes:
- I renamed the package names in sandbox, as lint was telling me you couldn't have `/` in a package name. 

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -13,7 +13,7 @@
   },
   "scripts": {
     "build": "turbo run build --cache-dir=.turbo --no-update-notifier",
-    "dev": "turbo run dev --filter=@instantdb/core --filter=@instantdb/react --filter=@instantdb/react-native --filter=@instantdb/admin --filter=www --filter=sb-react-nextjs --filter=sb-admin-sdk-express --parallel --concurrency=20 --no-update-notifier",
+    "dev": "turbo run dev --filter=@instantdb/core --filter=@instantdb/react --filter=@instantdb/react-native --filter=@instantdb/admin --filter=instant-www --filter=sb-react-nextjs --filter=sb-admin-sdk-express --parallel --concurrency=20 --no-update-notifier",
     "test": "turbo run test:ci --no-update-notifier",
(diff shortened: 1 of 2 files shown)


drew-harris:
big fan

client/Makefile
```diff
@@ -3,7 +3,7 @@ MAKE = make $(MAKEFLAGS)
 
 dev:
 	@echo "Booting up frontend..."
-	pnpm run dev --ui tui
+	pnpm run dev --ui tui $(filter --filter%,$(MAKECMDGOALS))
```
nezaj, line 6: TIL!

stopachka: I def had Claude help me on that. I asked him and he said: 


> ⏺ The makefile change I proposed uses $(filter --filter%,$(MAKECMDGOALS)) which:
>   1. $(MAKECMDGOALS) - Contains all the targets/arguments passed to make
>  2. filter --filter% - Filters to only keep arguments that start with "--filter"
>  3. This allows you to run: make dev --filter @instant/core --filter @instant/react
>  4. Those --filter arguments get passed through to: pnpm run dev --ui tui --filter @instant/core
>  --filter @instant/react
>
> This enables you to use pnpm's workspace filtering to only run dev on specific packages.

tonsky:
Does filter work for you? I’m getting

```
[~/ws/instant/client] make dev --filter sb-task-tracker
/Applications/Xcode.app/Contents/Developer/usr/bin/make: unrecognized option `--filter'
```

PR #1715, opened 2025-09-23
tailwind prettier class sorting

drew-harris:
automatically sort classnames with prettier based on predefined rules set by tailwind

note: requires `pnpm i` due to package updates


client/www/package.json
```diff
@@ -90,11 +92,20 @@
...
     "postcss": "^8.4.5",
-    "prettier-plugin-tailwindcss": "^0.1.1",
+    "prettier": "3.6.2",
+    "prettier-plugin-tailwindcss": "^0.6.14",
     "tailwindcss": "^3.0.7",
     "typescript": "5.5.4"
   },
   "engines": {
     "node": "22.x"
+  },
```
dwwoelfel, line 102: @drew-harris why do we need the extra prettier config here and in `client/www/.prettierrc`. Shouldn't it pick up the config from `client/.prettierrc`?

drew-harris: just making sure things aren't finicky, like if someone uses a different editor and only opens up the www folder or something

PR #1720, opened 2025-09-23
Fix devtools for apps in orgs

dwwoelfel:
If an app is in an org, the devtools won't be able to find it.

This does an extra fetch when we load the devtools to determine which org the app is in.

Easier to review with whitespace mode turned off https://github.com/instantdb/instant/pull/1720/files?w=1

### DIFF
--- a/client/www/pages/_devtool/index.tsx
+++ b/client/www/pages/_devtool/index.tsx
@@ -26,7 +26,6 @@ import {
 import Auth from '@/components/dash/Auth';
 import { FullscreenErrorMessage, getRole, isMinRole } from '@/pages/dash/index';
 import { TrashIcon, XMarkIcon } from '@heroicons/react/24/solid';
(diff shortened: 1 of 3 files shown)


client/www/pages/_devtool/index.tsx
```diff
@@ -137,14 +182,18 @@ function DevtoolAuthorized({ appId }: { appId: string }) {
     );
   }
 
+  if (!dashResponse.ready) {
```
drew-harris: this is unnecessary because createInitializedContext will hide children

drew-harris: same thing with the error state

drew-harris:
Works locally, my only feedback is the unreachable loading and error ui

client/www/pages/_devtool/index.tsx
dwwoelfel: Thanks! Fixed!

PR #1723, opened 2025-09-24
Guest auth

tonsky:
Doc https://www.dropbox.com/scl/fi/dhqmqlg69welgjv8e8zmx/Anon-Perms.paper

Current plan:

- [x] Add `db.auth.signInAnonymously`
- [ ] Add `$users.isGuest` attribute
- [ ] Add `auth.isGuest` to permissions
- [x] Support upgrade from magic link flow
- [ ] Magic link: move `$user` creation from `sendMagicCode` to `signInWithMagicCode`
- [ ] Support upgrade from oauth flows


server/src/instant/runtime/magic_code_auth.clj
```diff
@@ -58,24 +58,31 @@
...
-(defn create! [{:keys [app-id email]}]
-  (let [existing-u (app-user-model/get-by-email {:app-id app-id :email email})
-        {user-id :id :as u} (or existing-u
-                                (next-jdbc/with-transaction [conn (aurora/conn-pool :write)]
-                                  (let [app-user (app-user-model/create! conn {:id (random-uuid)
-                                                                               :app-id app-id
-                                                                               :email email})]
-                                    (transaction-model/create! conn {:app-id app-id})
-                                    app-user)))
+(defn create! [{:keys [app-id email user-id]}]
```
dwwoelfel: We could get into trouble if we let users claim an anonymous account with just the user id. Devs often create a public `profile` namespace that shares the id with the user id. An attacker could list profiles and try to claim any new ids they find.

One idea would be to store a random secret locally and store a hash of the secret on the user (or on a new linked `$anonUserProof` system ns). Then when you claim the user, you send the secret and we check that the hash matches.

tonsky: Addressed: we now send refreshToken instead of userId to verifyMagicCode

client/packages/core/src/authAPI.ts
```diff
@@ -58,6 +68,20 @@ export async function verifyRefreshToken({
...
+  appId,
+}: SharedInput): Promise<VerifyResponse> {
+  const res = await jsonFetch(`${apiURI}/runtime/auth/sign_in_guest`, {
+    method: 'POST',
+    headers: { 'content-type': 'application/json' },
+    body: JSON.stringify({
+      'app-id': appId,
+    }),
+  });
+  return res;
```
dwwoelfel, line 82: The `VerifyResponse` type contains a `user` and the`user` has a non-null `email` field, which will be undefined here.

dwwoelfel (commented):
I don't see anything wrong with the backend code or overall logic.

What is the plan for the types on the client, where `user` may not have an `email` field?

client/packages/core/src/authAPI.ts
tonsky: Right! Updated the type

tonsky:
Technically email is already not a required attribute. Updated TS types to reflect that

PR #1725, opened 2025-09-24
[experiment] update mux

stopachka:
We sometimes noticed a `TypeError: j.default.getLevel is not a function` in prod. 

I _think_ it's related to mux. I wasn't able to figure out how to repro this. But I thought, let's go ahead and update Mux -- maybe they fixed the bug, we are 1 major version behind. 

I also updated our `moduleResolution` to `bundler`, so typescript 5.0 can look up the types that come packages with the newer mux library. 

@dwwoelfel @nezaj @drew-harris @tonsky

### DIFF
--- a/client/www/tsconfig.json
+++ b/client/www/tsconfig.json
@@ -9,7 +9,7 @@
     "noEmit": true,
     "esModuleInterop": true,
     "module": "esnext",
-    "moduleResolution": "node",
+    "moduleResolution": "bundler",
     "resolveJsonModule": true,
     "isolatedModules": true,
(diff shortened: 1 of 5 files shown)


client/www/tsconfig.json
```diff
@@ -9,7 +9,7 @@
     "noEmit": true,
     "esModuleInterop": true,
     "module": "esnext",
-    "moduleResolution": "node",
+    "moduleResolution": "bundler",
```
stopachka, line 12: I checked the latest tsconfig from create-next-app, and they also set to bundler

PR #1726, opened 2025-09-25
Add support for login with LinkedIn

dwwoelfel:
Adds login with Linkedin.

UI is mostly codex. Linkedin doesn't give id tokens, so it required a small backend change to fetch the email from the tokeninfo endpoint.


stopachka (approved):
Whaat amazing!

PR #1728, opened 2025-09-25
Saved Sandboxes

drew-harris:
Uses local storage to save named presets for the sandbox, and also keeps the current state of the sandbox in base64 url query param so its easily sharable. 
Also replaces the select component with one from shadcn so we can show an empty state.


drew-harris:
image

client/www/components/dash/Sandbox.tsx
```diff
@@ -1,31 +1,71 @@
...
+const base64Parser = createParser({
+  parse(value) {
+    try {
+      return JSON.parse(atob(value));
+    } catch {
+      return '';
+    }
+  },
+  serialize(value) {
+    return btoa(JSON.stringify(value));
```
dwwoelfel, line 64: You might want to make this url-safe, `btoa(JSON.stringify(value)).replace(/+/g, '-').replace(/\//g, '_').replace(/=/g, '');` here and `JSON.parse(atob(value).replace(/-/g, '+').replace(/_/g, '/'));` above.

client/www/components/dash/Sandbox.tsx
```diff
@@ -73,6 +235,8 @@ export function Sandbox({
       apiURI: config.apiURI,
     });
 
+    console.log('CORE DB INITIALIZED', coreDb);
```
dwwoelfel: debug

client/www/components/dash/Sandbox.tsx
```diff
@@ -35,30 +75,152 @@ export function Sandbox({
...
+
+    if (matchingSandbox) {
+      setSelectedSandbox(matchingSandbox.name);
+      setHasUnsavedWork(false);
+    } else {
+      setSelectedSandbox(null);
+    }
+  }, []);
+
+  useEffect(() => {
```
dwwoelfel: Could we debounce this? It seems like a lot of work to do on every key stroke.

client/www/components/dash/Sandbox.tsx
```diff
@@ -35,30 +75,152 @@ export function Sandbox({
...
+  const [sandboxCodeValue, setSandboxValue] = useQueryState(`code`, {
+    ...base64Parser,
+  });
+
+  const [useAppPerms, setUseAppPerms] = useQueryState(
+    'useAppPerms',
+    parseAsBoolean.withDefault(true),
   );
+
+  const [permsValue, setPermsValue] = useQueryState('perms', {
```
dwwoelfel: It would be nice if we could remove `perms` from the URL if the user unchecked "use saved app rules", but I don't see an easy way to do that with `useQueryState`.

client/www/components/dash/Sandbox.tsx
```diff
@@ -81,10 +245,18 @@ export function Sandbox({
...
       }, 10);
       if (sandboxCodeValue) return;
+      console.log('USER EFFECT', JSON.stringify(sandboxCodeValue));
       const schema = dbAttrsToExplorerSchema(_oAttrs);
       const ns = schema.at(0);
+      console.log('hihi', ns);
       setSandboxValue(initialSandboxValue(ns?.name || 'example'));
     });
+
+    return () => {
```
dwwoelfel: I think you can just do `return unsubAttrs` here.

client/www/components/dash/Sandbox.tsx
drew-harris: The underlying library for useQueryState handles this for us

client/www/components/dash/Sandbox.tsx
drew-harris: I half agree, but also think its still valuable to send both over by default so anyone testing can toggle on and off variations of their perms to test things.

drew-harris:
~~TODO: FIX BOOLEAN NEW ROW BUTTON IN EXPLORER~~

client/www/components/dash/Sandbox.tsx
dwwoelfel: I still think it would be a good idea because then you never have to worry about whether the query param gets double encoded.

client/www/components/dash/Sandbox.tsx
```diff
@@ -81,10 +261,14 @@ export function Sandbox({
         }
       }, 10);
       if (sandboxCodeValue) return;
+      console.log('USER EFFECT', JSON.stringify(sandboxCodeValue));
```
dwwoelfel: debug code

client/www/components/dash/Sandbox.tsx
```diff
@@ -81,10 +261,14 @@ export function Sandbox({
...
       const schema = dbAttrsToExplorerSchema(_oAttrs);
       const ns = schema.at(0);
+      console.log('hihi', ns);
```
dwwoelfel: debug code

client/www/components/dash/Sandbox.tsx
```diff
@@ -35,30 +78,167 @@ export function Sandbox({
...
+          runAsUser: runAsUserEmail,
+          useAppPerms,
+        },
+      ]);
+    }
+    setHasUnsavedWork(false);
+  };
+
+  const loadPreset = (name: string) => {
+    console.log('loading preset', name);
```
dwwoelfel: Debug code

client/www/components/dash/Sandbox.tsx
```diff
@@ -198,307 +380,508 @@ export function Sandbox({
...
+        </form>
+      </Dialog>
+      <div className="flex w-full justify-between border-b bg-white p-2 px-3 dark:border-b-neutral-600 dark:bg-neutral-800">
+        <PresetManager />
+        <div className="flex items-center gap-8">
+          <IconButton
+            variant="subtle"
+            label="Copy link to sandbox"
+            onClick={() => {
+              navigator.clipboard.writeText(window.location.toString());
```
dwwoelfel: It would be safer to construct this url ourselves, since there could be some delay between your change to the sandbox and the URL updating.

client/www/lib/hooks/useSavedQueryState.tsx
```diff
@@ -0,0 +1,42 @@
+import { debounce } from 'lodash';
+import { useQueryState, UseQueryStateOptions, UseQueryStateReturn } from 'nuqs';
+import { useCallback, useEffect } from 'react';
+
+export const useSavedQueryState = <
```
dwwoelfel, line 5: This isn't working for me in the vercel preview. Steps to replicate: 

1. Go to sandbox, enter in "hello"
2. Click on the Explorer tab
3. Click on the sandbox tab, code gets replaced by default value.

I think useQueryState may do something weird if you call setValue before it has had time to initialize? If you get the item from localStorage and set the defaultValue in `useQueryState` as `localStorageValue || defaultValue`, then it seems to work.

drew-harris: interesting, i didn't set the defaultvalue there because if the current value matches the default then it doesn't add a param at all, but i realized i can turn that off so ill just do that

dwwoelfel (approved):
LGTM!

PR #1729, opened 2025-09-25
Make cli aware of orgs

dwwoelfel:
When you do `create-app`, it will ask you which org to pull from (if you have orgs). 

image

For pull, it will ask you which org to check if you want to import an app.

image

Demo: https://asciinema.org/a/q0p1jbTA4s8CEzOYpZu420FB6

Also fixes a bug in the dashboard where it can't find an app if you go to the direct url after being in an org.


stopachka (approved):
SGTM!

PR #1730, opened 2025-09-25
Postpone user creation during magic code auth

tonsky:
Right now, we create new $user when someone requests magic code. This seems too early (what if they fail to validate it?). This PR moves $user creation to when user comes back to verify code they got in email

This is required for guest auth promotion but I figured it’s easier to digest in a separate PR


server/src/instant/db/model/attr.clj
```diff
@@ -980,13 +980,14 @@
...
   (let [wrapped-attrs (if (instance? Attrs attrs)
                         attrs
                         (wrap-attrs attrs))
-        n [(name etype) (name label)]]
-    (:id (or (seek-by-fwd-ident-name n wrapped-attrs)
-             (seek-by-rev-ident-name n wrapped-attrs)))))
+        n             [(name etype) (name label)]
+        attr          (or (seek-by-fwd-ident-name n wrapped-attrs)
+                          (seek-by-rev-ident-name n wrapped-attrs))]
+    (assert (uuid? (:id attr)) (str "Expceted uuid, got: " (pr-str (:id attr))))
```
dwwoelfel:
```suggestion
    (assert (uuid? (:id attr)) (str "Expected uuid, got: " (pr-str (:id attr))))
```

server/src/instant/model/app_user_magic_code.clj
```diff
@@ -9,38 +9,44 @@
...
-                  [:add-triple id (resolve-id :codeHash) (-> code
-                                                             crypt-util/str->sha256
-                                                             crypt-util/bytes->hex-string)]
-                  [:add-triple id (resolve-id :$user) user-id]])
-      (assoc (get-entity id)
-             :code code)))))
+(def ttl-ms
+  (* 24 60 60 1000))
+
+(defn expired? [magic-code]
```
dwwoelfel, line 22: Why remove the ability to pass in the date?

server/src/instant/model/app_user_magic_code.clj
```diff
@@ -9,38 +9,44 @@
...
+   (create! (aurora/conn-pool :write) params))
+  ([conn {:keys [app-id email id code]}]
+   (let [id   (or id (random-uuid))
+         code (or code (rand-code))]
+     (update-op
+      conn
+      {:app-id app-id
+       :etype etype}
+      (fn [{:keys [transact! get-entity]}]
+        (transact! [{:id       id
```
dwwoelfel, line 38: You'll still need to set the user-id while this deploys. Otherwise anyone who goes through the login process during the deploy might get an error.

server/src/instant/model/app_user_magic_code.clj
tonsky: It wasn’t used

server/src/instant/model/app_user_magic_code.clj
tonsky: Split this PR into two (#1733). Now should be fully backwards-compatible as long as all email-less magicCodes expire

server/src/instant/db/model/attr.clj
dwwoelfel:
```suggestion
    (assert (uuid? (:id attr)) (str "Expected uuid, got: " (pr-str (:id attr))))
```

server/src/instant/system_catalog.clj
```diff
@@ -164,13 +164,7 @@
    (make-attr "$magicCodes" "email"
               :unique? false
               :index? true
-              :checked-data-type :string)
-   ;; TODO remove after migrating to $magicCodes.email
-   (make-attr "$magicCodes" "$user"
```
dwwoelfel, line 169: You'll have to manually remove this from the system attrs.

tonsky: Or we can leave it

PR #1731, opened 2025-09-26
Show active connection count on dashboard

tonsky:
Somebody [mentioned it on Discord](https://discord.com/channels/1031957483243188235/1421062620810711050) and I thought it should be easy to implement and a fun idea. It was

Screenshot 2025-09-26 at 16 38 24

### DIFF
--- a/server/src/instant/machine_summaries.clj
+++ b/server/src/instant/machine_summaries.clj
@@ -2,10 +2,13 @@
   (:require
    [instant.util.hazelcast :as hz]
    [instant.reactive.ephemeral :as eph]
-   [instant.reactive.store :as rs])
+   [instant.reactive.store :as rs]
+   [instant.util.cache :as cache])
   (:import
    (com.hazelcast.cluster Member)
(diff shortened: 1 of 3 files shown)


server/src/instant/machine_summaries.clj
```diff
@@ -38,17 +41,39 @@
                 @fut]))))
 
 (comment
-  (get-all-session-reports (eph/get-hz)))
+  (get-session-reports (eph/get-hz)))
+
+(def session-reports-cache
+  (cache/make
+   {:ttl      30000
```
dwwoelfel: We could probably lower this to 5 seconds here and in the UI to make it feel more snappy.

tonsky: This is almost 100k of data transferred over entire cluster to each machine, you still think 5 sec is okay?

dwwoelfel: Let's put it in a `flags/flag` and then we can tweak it later.

PR #1738, opened 2025-09-26
Support orgs in create-instant-app

nezaj:
Following up from https://github.com/instantdb/instant/pull/1732, this adds support for creating/importing apps with orgs.

If you don't have orgs this extra prompt will be skipped.

### When creating apps
- If you have orgs, prompts: "Would you like to create the app in an organization?"
- Shows list of orgs (plus "No organization" option)
- Creates app in selected org

### When importing apps
- If you have orgs, prompts: "Would you like to import an app from an organization?"
- Shows list of orgs (plus "No organization" option)
- Fetches and displays apps from selected org
- Imports the selected app

### Implementation
- Added `fetchDashboard` to get both apps and orgs in one call
- Added `fetchOrganizationApps` to get apps from a specific org
- Filters out orgs where user only has `app-member` role (can't create apps there)
- Updated `createApp` to accept optional `orgId` parameter

https://github.com/user-attachments/assets/afa4d1f1-7082-4931-ad45-b5dc35b1536f

### DIFF
--- a/client/packages/create-instant-app/src/login.ts
+++ b/client/packages/create-instant-app/src/login.ts
@@ -31,10 +31,14 @@ export const promptForAppName = async (program: Project) => {
   return title.trim();
 };
 
-export const createApp = async (title: string, authToken: string) => {
+export const createApp = async (
+  title: string,
+  authToken: string,
+  orgId?: string | null,
+) => {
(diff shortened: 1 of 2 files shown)


stopachka (approved):
SGTM!

PR #1739, opened 2025-09-27
[expo] upgrade to SDK 54

stopachka:


stopachka:
Will get to this later

PR #1742, opened 2025-09-29
Handle new date format

dwwoelfel:
Handles dates in format m/d/yyyy where year is a large value, like 3/12/4444


tonsky:
somebody really uses years > 2100?

PR #1743, opened 2025-09-29
Add Founding Firebase

stopachka:
- Added the transcript for Founding Firebase
- Added a `::transcript` modifier, so all timestamps link to youtube


nezaj (approved):
Let's run it!

PR #1744, opened 2025-09-29
Cache app selector for docs

nezaj:
One of our users reported that clicking the "Docs" button from the dashboard doesn't actually open a new tab. I thought this would be a simple fix but noticed that we had two other problems.

1) We would get react hydration issues when doing a refresh on /docs
2) We no longer cache the selected app correctly -- if you look at prod right now and select an app in the dashboard and then go to the docs page the same app won't be selected.

This was because we weren't accounting for org id in the doc app selector.

This PR fixes all three issues (docs pop out, no more hydration issue, and the same app is selected across dashboard/docs)

Update: Added logic to let you select apps from your orgs as well!

CleanShot 2025-09-29 at 17 26 20@2x


stopachka (approved):
SGTM!

PR #1745, opened 2025-09-29
Handle `$isNull` sketch counts for unindexed fields

dwwoelfel:
When we do a count for `$isNull`, we do `(check sketch nil nil)`. That gets the count of triples that have `null` for their value. This works great for indexed blob fields because we always put a `nil` into the triples table for every indexed field. 

It doesn't work so well for unindexed fields and refs. If you create an entity and leave off the field, then there will be no triple for that field and we won't see it in the sketch.

This PR solves that problem by subtracting the count of the `id` triples from the count of the field's triples, to get the number of triples that are "undefined". We add the number of nulls and number of undefined's to get the total number of `$isNull`.

### DIFF
--- a/server/test/instant/db/datalog_test.clj
+++ b/server/test/instant/db/datalog_test.clj
@@ -9,7 +9,7 @@
             [instant.data.resolvers :as resolvers]
             [instant.jdbc.sql :as sql]
             [instant.fixtures :refer [with-movies-app with-zeneca-app with-zeneca-checked-data-app]]
-            [instant.util.test :refer [in-memory-sketches-for-app make-attrs]]))
+            [instant.util.test :refer [in-memory-sketches-for-app]]))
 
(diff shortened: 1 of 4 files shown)


server/test/instant/db/datalog_test.clj
```diff
@@ -439,5 +441,142 @@
...
+          ;; comparators just take 1/2 since we don't have a way to do real comparisons
+          (is (= 2 (estimate-rows [:av '?e (resolvers/->uuid r :users/handle) {:$comparator
+                                                                               {:op :$gt
+                                                                                :value "alex"
+                                                                                :data-type :string}}]))))
+
+        (testing "$isNull"
+          (doseq [{:keys [settings props]} isnull-cases]
+            (testing settings
+              (tool/def-locals)
```
stopachka: nit: def-locals

PR #1747, opened 2025-09-30
[admin] allow omitting admin token

stopachka:
**Problem**

Right now, you need to provide an `adminToken` to use `@instantdb/admin`

**But, there are cases where you don’t really need it:** 

1. If you provide a user token with asUser({ token: '...' });
2. If you say you want to run a query as a guest asUser( { guest: true }); 

**Solution**

1. We now make `adminToken` optional in `init`
2. If you explicitly use `asUser({ token })` or `asUser({ guest: true })`, queries and transacts will work. Otherwise, they will fail. 

Got claude to help write tests in admin, and made a little sandbox app too: 

image

### DIFF
--- a/client/packages/admin/src/index.ts
+++ b/client/packages/admin/src/index.ts
@@ -94,7 +94,7 @@ type DebugCheckResult = {
 
 type Config = {
   appId: string;
-  adminToken: string;
+  adminToken?: string;
   apiURI?: string;
   useDateObjects?: boolean;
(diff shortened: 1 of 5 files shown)


dwwoelfel (approved):
LGTM, but it would be nice to have error handling in the library so that we could give a more specific error message when you make a query without the token.

stopachka:
@dwwoelfel feedback

PR #1748, opened 2025-09-30
[easy] _core -> core

stopachka:
Context: 

Sometimes react users want to do things like `subscribeQuery`. Right now we expose this through the `db._core` property.

But really `core` is a public interface. 

1. Deprecated `_core `
2. Added a new `core`


dwwoelfel (approved):
Nice!

PR #1753, opened 2025-09-30
Better cost function for choosing datalog index

dwwoelfel:
Adds a new cost function for determining the best index. It's behind a feature flag so that we can quickly roll back if it causes issues.

The old cost function treated `[e v]` as equivalent to `[v e]`, but if you have fewer `e`, then it's much better to start with the `e` first because the `e` is unique. You know that there only be at most one v per e. The `v` might point to many different `e`s. 

Still a long way to go to create a better model of the work the database will actually do, but this brings us a little closer.

### DIFF
--- a/.github/workflows/clojure.yml
+++ b/.github/workflows/clojure.yml
@@ -110,8 +110,10 @@ jobs:
           POSTGRES_PASSWORD: test_password
           POSTGRES_DB: test_db
         options: >-
-          --no-healthcheck
-
+          --health-cmd "exit 0"
+          --health-interval 1s
+          --health-timeout 1s
+          --health-retries 1
 
     steps:
     - name: Checkout
--- a/.github/workflows/clojure.yml
+++ b/.github/workflows/clojure.yml
@@ -109,6 +109,9 @@ jobs:
           POSTGRES_USER: test_user
           POSTGRES_PASSWORD: test_password
           POSTGRES_DB: test_db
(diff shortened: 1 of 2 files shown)


.github/workflows/clojure.yml
```diff
@@ -109,6 +109,11 @@ jobs:
           POSTGRES_USER: test_user
           POSTGRES_PASSWORD: test_password
           POSTGRES_DB: test_db
+        options: >-
```
dwwoelfel, line 112: Hoping this will make the step start faster. We don't care about its health here, we'll wait for it to be heathy  in the "Prepare database" step. Instead of waiting, it can be loading in the background while we checkout and prepare java.

server/src/instant/db/datalog.clj
```diff
@@ -1212,6 +1221,33 @@
...
+(defn path-cost-with-joins-new
+  "Tries to estimate the work we'll be doing for an individual index,
+  taking into account joins. It should correlate with the cost of a
+  nested loop."
+  [index]
+  (let [costs (:index-costs index)
+        index-lookup-cost (reduce (fn [acc {:keys [cost col]}]
+                                    (let [next-cost (* acc cost)]
+                                      (if (contains? (:unique-cols index) col)
+                                        (reduced next-cost)
```
stopachka, line 1233: 🔥

PR #1754, opened 2025-09-30
INSTANT_ADMIN_TOKEN -> INSTANT_APP_ADMIN_TOKEN

stopachka:
I was e2e testing the new admin changes, and realized the .env var that gets created is called `INSTANT_APP_ADMIN_TOKEN`. 

I went ahead and updated the chalk text, as well as a few docblocks where we use process.env.  


dwwoelfel (approved):
LGTM!

PR #1755, opened 2025-09-30
Publish our postgres image to github's container registry

dwwoelfel:
We get occasional rate-limit errors when we pull the image from amazon's container registry. It doesn't fail the build, because we retry and start again, but it slows it down.

With this change, we'll push the image to the github container registry, which seems to be a bit faster.

It's also nice that we can have multiple package names--with the amazon registry, everything went into the `instantdb` package, and we just had to give it different tags.

### DIFF
--- a/.github/workflows/clojure.yml
+++ b/.github/workflows/clojure.yml
@@ -102,7 +102,7 @@ jobs:
 
     services:
       postgres:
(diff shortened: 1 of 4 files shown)


stopachka (approved):
Very cool!

PR #1758, opened 2025-09-30
Allow single-digit month in triple dates

dwwoelfel:
Updates the date-time parsers to allow single digit month values, e.g. `2025-9-29T23:59:59.999Z`


nezaj (approved):
LGTM!

PR #1750, opened 2025-09-30
Guest auth part 2

tonsky:
Before: #1723

Other types of sign in should support promoting from guest user

- [x] Google Popup
- [x] Google Redirect
- [x] LinkedIn Redirect
- [x] Apple Popup
- [x] Apple Redirect
- [x] Clerk

This PR also adds `createAuthorizationURLAsync` because guest promotion needs refreshToken and that lives in `getCurrentUser` which is async :( If anyone have better ideas how to work around that I’m all ears

We also changed scope to GenericOAuthClient from `email` to `email openid` which fixes LinkedIn and seem to not affect other providers

Finally, an uber Sign In page that collects everything:

Screenshot 2025-09-30 at 21 21 58

I’ll look into native sign in tomorrow, chances are they also need massaging, perhaps in next PR (+ docs)

PR #1759, opened 2025-10-01
Add client-id and auth-code to $oauthCodes

tonsky:
Pre-requisite for #1750

We want to postpone user creation from `/oauth/callback` to `/oauth/token`. To do that, we need to store `$oauthClient` and `$authCode` in `$oauthCodes` and wait for all in-flight requests to expire (10-60 minutes)

Needs to be deployed before #1750

### DIFF
--- a/server/src/instant/runtime/routes.clj
+++ b/server/src/instant/runtime/routes.clj
@@ -350,78 +350,90 @@
 
 (defn oauth-callback [{:keys [params] :as req}]
   (try
-    (let [return-error (fn return-error [msg & params]
-                         (throw (ex-info msg (merge {:type :oauth-error :message msg}
-                                                    (apply hash-map params)))))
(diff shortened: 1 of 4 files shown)


server/src/instant/runtime/routes.clj
```diff
@@ -350,78 +350,90 @@
...
+                           (throw (ex-info msg (merge {:type :oauth-error :message msg}
+                                                      (apply hash-map params)))))
+
+          _              (when (:error params)
+                           (return-error (:error params)))
+
+          state-param    (or (:state params)
+                             (return-error "Missing state param in OAuth redirect."))
+
+          ;; _app-id unused for now, but will be used when we have app_oauth_redirects in triples
```
dwwoelfel, line 363: We could remove this comment now

PR #1760, opened 2025-10-01
Postpone user creation to oauth-token-callback

tonsky:
Follow-up to #1759
Pre-requisite to #1750

Should be deployed separately at least 60 minutes after #1759


dwwoelfel:
Moving to https://github.com/instantdb/instant/pull/1763

PR #1763, opened 2025-10-01
Postpone user creation to oauth-token-callback

dwwoelfel:
(creating a new PR for this so that I don't annoy anybody with notifications)

Easier to review with whitespace turned off https://github.com/instantdb/instant/pull/1763/files?w=1

Picks up from https://github.com/instantdb/instant/pull/1760, but it gets the user-info in the initial callback instead of waiting for the token exchange. This does two things: 1. we keep the error handling for invalid tokens in the same place, and 2. we have access to the email address, so that we can show the user who they are going to log in as.

The background is that we want to delay user creation until we get to the token exchange. Anonymous users can pass us their instant token during that step and we can convert them to real users. We could convert them in the oauth callback, but then we'd have to pass the instant tokens in the URL and we don't want to do that.

This PR adds the user-info into the `$oauthCode` table in the oauth callback. When the user exchanges the code for a token, we use that data to upsert the user.

### Deploy plan

1. Deploy this PR
2. Wait an hour for all of the oauth codes without user-info to expire
3. Flip the `delay-user-creation-to-token-exchange` flag toggle
4. Push up a new PR that removes the old code (I have todos scattered in the code so that we don't forget anything)

### DIFF
--- a/server/src/instant/model/app_oauth_code.clj
+++ b/server/src/instant/model/app_oauth_code.clj
@@ -18,7 +18,8 @@
                  code-challenge-method
                  code-challenge
                  client-id
-                 auth-code]}]
+                 user-id
+                 user-info]}]
    (update-op
     conn
     {:app-id app-id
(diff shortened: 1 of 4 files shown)


stopachka (approved):
Awesome!

PR #1766, opened 2025-10-02
Cleanup after postpone user creation in oauth flow

dwwoelfel:
We've flipped all of the flags and we're now delaying user creation to the token exchange step in the oauth flow.

This PR removes all of the temporary code we had to support the transition.

It also includes logs at startup for system attrs that don't match our config and for system attrs that we no longer want.


stopachka (approved):
Biyutiful!

PR #1768, opened 2025-10-02
Guest auth part 2

dwwoelfel:
(creating a new PR for this so that I don't annoy anybody with notifications)

Picks up from https://github.com/instantdb/instant/pull/1750

Adds support for promoting a guest user from the OAuth redirect flows and the sign in token flows (used by clerk, google button, and apple popup).

Also updates scope for GenericOAuthClient from `email` to `email openid`, which fixes LinkedIn and seems to not affect other providers.

### Linking orphaned guest $users

When you promote a guest account, but an account already exists, we'll add a new link from the guest $user to the real $user through a new `linkedPrimaryUser` field, whose reverse link is called `linkedGuestUsers`. We don't allow users to edit the system tables, so you have to be an admin to set that field.

I modified the default view rule for $users so that they can view the guest accounts that are linked to them.

This gives developers a way to transfer data to the new user. For example, say you have todos with a creator field that links to $users and your view rule is `data.creator == auth.id`. If you want the user to be able to transfer the data they created as a guest, then you just need to modify the rule to `data.creator == auth.id || data.creator in auth.ref('$user.linkedGuestUsers.id')`. 

### Playground

The playground is at `/play/guest`:

https://github.com/user-attachments/assets/ab1dfbf7-4ede-4cf7-9844-afbed2efc0ea


server/src/instant/db/model/attr.clj
```diff
@@ -164,7 +164,9 @@
   (cache/make {:max-size 512}))
 
 (defn evict-app-id-from-cache [app-id]
-  (cache/invalidate attr-cache app-id))
+  (if (= app-id system-catalog-app-id)
+    (cache/reset attr-cache)
```
dwwoelfel, line 168: This fixes a bug where the new system attrs would never show up. Now we swap out everything when the system attrs change.

server/src/instant/model/rule.clj
```diff
@@ -147,7 +147,7 @@
     (let [compiler (cel/action->compiler action)]
       (if (and (= "$users" etype)
                (#{"view" "link" "unlink"} action))
-        (let [code "auth.id == data.id"
+        (let [code "auth.id == data.id || (data.linkedPrimaryUser != null && auth.id == data.linkedPrimaryUser)"
```
dwwoelfel, line 150: This is the rule change that allows users to see the list of guest users that are linked to them.

Users can't modify the data in the system tables (except for $files.path), so we don't have to worry about them linking to users they shouldn't. When we open up that ability, we should just prevent users from editing values for any of the system attributes (excepting $files.path).

server/src/instant/runtime/routes.clj
```diff
@@ -509,13 +520,18 @@
...
-                                                                         :id current-refresh-token-id}))
+                                (app-user-refresh-token-model/get-by-id
+                                 {:app-id app-id
+                                  :id current-refresh-token-id}))
+
+        social-login (upsert-oauth-link! {:email       email
+                                          :sub         sub
+                                          :app-id      (:app_id client)
+                                          :provider-id (:provider_id client)
+                                          :guest-user-id (:user_id current-refresh-token)})
```
dwwoelfel: I think this is wrong, it might be an actual user sometimes. Will fix.

dwwoelfel: Fixed via https://github.com/instantdb/instant/pull/1768/commits/ff93e140bc24eab537350361865f1b39846d6428

server/src/instant/db/model/attr.clj
stopachka: Oh cool!

server/src/instant/system_catalog.clj
```diff
@@ -76,12 +77,14 @@
    "content-disposition" "cdisp"
    "location-id" "lid"
    "key-version" "kv"
-   "userInfo" "userInfo"})
+   "userInfo" "userInfo"
+   "linkedGuestUsers" "lgu"
```
stopachka, line 81: one alternative name suggestion: 

users has many `guestAccounts` 
users has one `primaryAccount` 

WDYT @dwwoelfel ?

stopachka (approved):
Beautiful!

server/src/instant/system_catalog.clj
dwwoelfel: Discussed on Discord, decided to leave it as-is.

PR #1770, opened 2025-10-03
[CLI] Add logout command

nezaj:
What it says on the tin! Useful for testing different states of `create-instant-app`


dwwoelfel (approved):
Nice!

PR #1771, opened 2025-10-03
Add an `isGuest` field to user

dwwoelfel:
Adds an `isGuest` field to the user so that you don't have to check `user.type === 'guest'`. 

I decided to do it on the backend instead of the client because the user shows up in so many different places on the client that it would be very easy to miss one.


nezaj (approved):
LGTM!

PR #1775, opened 2025-10-03
Docs for anonymous auth

dwwoelfel:
Includes a sandbox to sanity check that transferring data from the guest user to the primary user works. I couldn't use the regular `guest` sandbox because that one relies on having a stable sandbox app that you can tie all of the oauth clients to. I needed an ephemeral app that you can define permissions for.

For the sandbox, I needed to include `$users` in my ephemeral app's schema. I made a small change to how we plan schema so that the system catalog links are removed. I also moved the function that filters out the system changes `plan!` instead of making every caller do the filtering themselves.


client/www/pages/docs/auth/magic-codes.md
```diff
@@ -16,14 +16,14 @@ for an example with vanilla JS, check out this [sandbox](https://github.com/inst
 
 Open up your `app/page.tsx` file, and replace the entirety of it with the following code:
 
-```javascript {% showCopy=true %}
-"use client";
+```jsx {% showCopy=true %}
```
dwwoelfel, line 19: I changed this from `javascript` to `jsx` and now prettier can format it.

server/src/instant/dash/routes.clj
```diff
@@ -1272,19 +1270,13 @@
...
-   entities))
-
 (defn schema-push-plan-post [req]
   (let [{{app-id :id} :app} (req->app-and-user-accepting-platform-tokens! :collaborator
                                                                           :apps/read
                                                                           req)
         client-defs         (-> req
                                 :body
-                                :schema
-                                (update :entities remove-system-namespaces))
```
dwwoelfel, line 1287: `schema-model/plan!` handles removing the system namespaces now

nezaj (approved):
Great docs!

client/www/pages/docs/auth/guest-auth.md
```diff
@@ -0,0 +1,349 @@
...
+    },
+  },
+};
+```
+
+You can then query for the data owned by those guest users and transfer it to the primary account. The specific implementation will depend on your application's data model.
+
+To enable your user to access the data stored by the guest users, you can update your rules to grant access:
+
+```diff
```
nezaj, line 283: TIL !

PR #1776, opened 2025-10-03
Fix on delete direction for linked guest users

dwwoelfel:
Had the on-delete direction reversed. It should delete the guest users when you delete the primary user, but it shouldn't delete the primary user if you delete the guest user.

Already updated the attr in production. Includes a test.


nezaj (approved):
Nice!

PR #1778, opened 2025-10-03
Show the guest fields on $users in the explorer

dwwoelfel:
Now that we have docs for guest users, I think it's safe to show them in the explorer.

image

Also updates our server flags:
1. Updates to the latest versions
2. pulled the schema


stopachka (approved):
niice!

PR #1779, opened 2025-10-03
Update system catalog to match db

dwwoelfel:
Somehow we had size as required in the db, but not in the catalog. This updates it to match prod.


nezaj (approved):
SGTM!

PR #1781, opened 2025-10-06
Fix the cost of joins

dwwoelfel:
We were adding the joins instead of multiplying them, which was causing us to choose bad indexes.

Put it behind a feature flag.

No change across my set of test queries, but a huge improvement in the query that uri reported https://discord.com/channels/1031957483243188235/1148284450992574535/1424333410255507556

### DIFF
--- a/server/src/instant/db/datalog.clj
+++ b/server/src/instant/db/datalog.clj
@@ -790,14 +790,10 @@
           :cols [:a :v]
           :idx-key :ave}
 
-         (when (and (flags/toggled? :ave-with-e-index)
-                    ;; Can't use this with the old cost
-                    ;; function because it generates bad joins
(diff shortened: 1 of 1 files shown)


server/src/instant/db/datalog.clj
```diff
@@ -1261,7 +1262,7 @@
                                         next-cost)))
                                   1
                                   (:path costs))
-        join-cost (reduce + 0 (vals (:join-remaining costs)))
+        join-cost (reduce * 1 (vals (:join-remaining costs)))
```
dwwoelfel, line 1265: This is the actual change. 

The change to the previous function is just us changing `path-cost-with-joins-old` to be the previous `path-cost-with-joins-new`

PR #1782, opened 2025-10-06
Remove ave_index (part 1 of 2)

dwwoelfel:
Now that the ave_with_e_index has been in production for a while, we can start the process to remove the `ave_index`. 

It will take two steps.

1. Stop using the index. That's this PR. It adds a flag to stop using the ave_index.
2. Run a migration to drop the ave_index from the db. That will be in a new PR if everything goes fine with this one.


stopachka (approved):
Woohoo!

PR #1784, opened 2025-10-06
schema migrations

drew-harris:


client/packages/cli/src/index.js
```diff
@@ -353,7 +359,7 @@ program
     console.log(`\n✅ Created app "${chalk.green(appTitle)}"\n`);
     console.log(`App ID: ${chalk.cyan(appId)}`);
     console.log(
-      `App Admin Token (double click first 4 digits to select): ${chalk.dim(appToken.substring(0, 4))}${chalk.hidden(appToken.substring(4))}`,
```
stopachka, line 356: TIL on this chalk.hidden

client/packages/cli/src/renderSchemaPlan.ts
```diff
@@ -0,0 +1,260 @@
+import {
+  convertTxSteps,
+  MigrationTx,
+  MigrationTxTypes,
+} from '@instantdb/platform';
+import chalk from 'chalk';
+import { promptOk } from './index.js';
+
+// Hack to prevent using @instantdb/core as a dependency for cli
```
stopachka, line 11: Another kind of hack, but maybe platform should re-export this?

client/packages/cli/tsconfig.json
```diff
@@ -3,7 +3,9 @@
   "compilerOptions": {
     "outDir": "dist",
     "rootDir": "src",
-    "skipLibCheck": true // `unconfig` currently imports a broken type from `@antfu/utils`
+    "sourceMap": true,
```
drew-harris, line 6: Should I remove sourcemaps?, i only turned them on for the debugger

client/packages/cli/src/renderSchemaPlan.ts
```diff
@@ -0,0 +1,260 @@
...
+  while (i < planSteps.length) {
+    const step = planSteps[i];
+
+    if (step.type === 'add-attr') {
+      const namespace = step.identifier.namespace;
+      const namespaceAttrs: string[] = [];
+      let hasIdAttr = false;
+      let j = i;
+
+      while (j < planSteps.length) {
```
stopachka: The lookahead here is a bit hard to think through. I wonder if we can re-structure the code here. 

Some ideas: 

**Intermediate step**
First, we group by (namespace, add-attr | delete-attr) 
Then we walk through the groups, and flatMap them into groupedSteps. 

**Reduce**
An alternative could be to use a reduce, which keeps `groupedSteps` and a `processed` set. 

WDYT @drew-harris

client/packages/cli/src/renderSchemaPlan.ts
```diff
@@ -0,0 +1,260 @@
...
+            `Deletion of namespace ${step.namespace}`,
+          );
+        }
+
+        console.log(
+          `${chalk.bgRed(' - DELETE NAMESPACE ')} ${step.namespace}${step.attrs.length > 0 ? ` (${step.attrs.join(', ')})` : ''}`,
+        );
+        break;
+      case 'add-attr':
+        if (step['value-type'] === 'ref' && step['reverse-identity']) {
```
stopachka, line 217: nit: I would move this into a variable like `isLink`, to make it simpler.

For the slightly more complicated cases, it may be nicer to move them into a separate function. Sometimes moving into separate functions can feel performative, but in case statements, they can be helpful, as inside the body you don't have to worry about `break`

client/packages/platform/__tests__/src/migrations.test.ts
```diff
@@ -0,0 +1,398 @@
...
+  entities: {
+    songs: i.entity({
+      title: i.string(),
+      artist: i.string(),
+    }),
+  },
+});
+
+const createChooser = (
+  pickThese: (RenamePromptItem<string> | string)[],
```
stopachka, line 33: cool!

client/packages/platform/__tests__/src/migrations.test.ts
```diff
@@ -0,0 +1,398 @@
...
+  MigrationTx,
+  MigrationTxTypes,
+  RenamePromptItem,
+  RenameResolveFn,
+} from '../../src/migrations';
+
+const simpleSchemaBefore = i.schema({
+  entities: {
+    songs: i.entity({
+      deleteme: i.string(),
```
stopachka: nit: I would call this some realistic name, as I notice in the code we sometimes rename this.

client/packages/platform/__tests__/src/migrations.test.ts
```diff
@@ -0,0 +1,398 @@
...
+        }),
+      },
+    }),
+    i.schema({
+      entities: {},
+    }),
+    createChooser(['artist']),
+  );
+  console.log(result);
+  expectTxType(result, 'delete-attr', 2);
```
stopachka, line 79: side note: I wonder if we should do snapshot tests for these. Perhaps it may be good to know the "full structure"?

client/packages/platform/__tests__/src/migrations.test.ts
```diff
@@ -0,0 +1,398 @@
...
+        }),
+        songs: i.entity({
+          name: i.string(),
+        }),
+      },
+      links: {},
+    }),
+    createChooser([]),
+  );
+  console.log(result);
```
stopachka, line 368: nit: are the console.logs expected?

client/packages/platform/src/migrations.ts
```diff
@@ -0,0 +1,744 @@
...
+  'remove-index': convertSimpleConstraintUpdate,
+  'remove-unique': convertSimpleConstraintUpdate,
+  'remove-required': convertSimpleConstraintUpdate,
+
+  'delete-attr': (from, existing) => {
+    const found = findAttrFromExisting(from, existing);
+    if (found) {
+      return found.id;
+    }
+    throw new Error(`Attribute ${from.namespace}.${from.attrName} not found`);
```
stopachka: nit: would reverse the order, so errors work like 'guards'

client/packages/platform/__tests__/src/migrations.test.ts
drew-harris: i just left them in to be helpful for debugging the tests, they should only show when you run a single test at a time (so won't clog up ci logs)

client/packages/cli/src/util/loadConfig.ts
```diff
@@ -10,7 +10,7 @@ export async function loadConfig<T>(
   const res = await _loadConfig(opts);
   // Unconfig seems to add an __esModule property to the config object
   // Removing it.
-  if (typeof res.config === 'object' && '__esModule' in res.config) {
+  if (typeof res.config === 'object' && '__esModule' in res.config!) {
```
drew-harris, line 13: note: added this because i enabled `strictNullChecks` in tsconfig

client/packages/cli/src/renderSchemaPlan.ts
drew-harris: I like the group + flatmap idea more, ill do that

client/packages/cli/src/renderSchemaPlan.ts
drew-harris: All of the places where I check if its a link are based on different object shapes so I can't cleanly check them all with one function without basically making a new case statement, and I wouldn't get the type narrowing. I see your point about the break statement but this feels like the best way.

client/packages/cli/src/index.js
```diff
@@ -1286,146 +1292,102 @@ function linkOptsPretty(attr) {
...
-      }
+  let showRawSteps = false;
+  try {
+    await renderSchemaPlan(diffResult, currentAttrs);
+  } catch (error) {
+    if (error instanceof CancelSchemaError) {
+      console.info('Schema migration cancelled!');
+      return { ok: true };
     }
+    console.error('Error displaying schema plan', error);
```
stopachka: Perhaps here, we should actually re-throw the error. i.m.o we don't really have confidence to continue after this stage. 

In this way, we can simplify the code that follows, with a guarantee that nothing has gone wrong. 

For showRawSteps, perhaps we can skip that alltogether, and rely on --verbose to see the steps.

client/packages/platform/src/migrations.ts
```diff
@@ -0,0 +1,760 @@
...
+  'reverse-identity'?: Identifier | null;
+  'on-delete'?: InstantDBAttrOnDelete | null | undefined;
+  'on-delete-reverse'?: InstantDBAttrOnDelete | null | undefined;
+};
+
+export type MigrationTxTypes = {
+  'delete-attr': { identifier: Identifier };
+  'update-attr': { identifier: Identifier; partialAttr: AttrWithIdentifier };
+  'add-attr': {
+    identifier: Identifier;
```
stopachka, line 32: nit: I wonder if there's some way to clean up how the attr typedefs. 

It looks like this, plus, AttrWithIdentifier all work over the underlying 'Attr' structure.

I wonder if we should have a single definition, and then use pick or something. 

What do you think?

client/packages/platform/src/migrations.ts
```diff
@@ -0,0 +1,760 @@
...
+  required: { identifier: Identifier };
+  'remove-required': { identifier: Identifier };
+  'check-data-type': {
+    identifier: Identifier;
+    'checked-data-type': InstantDBCheckedDataType;
+  };
+  'remove-data-type': { identifier: Identifier };
+};
+
+type EasyMigrationTypes =
```
stopachka: Nit: I would call this something that aligns with what's going on in the backend. Maybe `JobMigrationType` ?

client/packages/platform/src/migrations.ts
```diff
@@ -0,0 +1,760 @@
...
+      return (
+        attr['forward-identity'][1] === ident.namespace &&
+        attr['forward-identity'][2] === ident.attrName
+      );
+    }) || null
+  );
+};
+
+const convertSimpleConstraintUpdate: ConvertPlanStepFn<any> = (
+  from,
```
stopachka: nit: perhaps instead of `from` we should call this `migTx` or something?

Reason: `from` and `existing` feel similar in meaning

client/packages/platform/src/migrations.ts
```diff
@@ -0,0 +1,760 @@
...
+const CONVERTERS: AllConvertPlanStepFns = {
+  index: convertSimpleConstraintUpdate,
+  unique: convertSimpleConstraintUpdate,
+  required: convertSimpleConstraintUpdate,
+  'remove-index': convertSimpleConstraintUpdate,
+  'remove-unique': convertSimpleConstraintUpdate,
+  'remove-required': convertSimpleConstraintUpdate,
+
+  'delete-attr': (from, existing) => {
+    const found = findAttrFromExisting(from.identifier, existing);
```
stopachka: nit, you can probably extract the `found + assert` as `getExisitngAttrThrowing` (name can be whatever you like). It's repeated quite a bit.

client/packages/platform/src/migrations.ts
```diff
@@ -0,0 +1,760 @@
...
+  }
+  return {
+    'attr-id': found.id,
+    'forward-identity': found['forward-identity'],
+  };
+};
+
+// converts migration operations from Identifier (namespace/name) based
+// into transaction steps that use database ids
+const CONVERTERS: AllConvertPlanStepFns = {
```
stopachka, line 100: (Note: we shouldn't implement it this way, but thought i'd share the idea for background.) 

Another way of implementing this function, could be to do a `recursive walk`. Effectively, you put 'identifiers' wherever you need them, then you have a 'mapper' that detects identifiers and replaces them. 

This approach simplifies what follows, but then we'd end up with problems in places like `add-attr`, where we want to separate out the steps. (If of our migration system produced the steps as separate, then this would be easier to implement). 

This is fine and simple to read through though, so no need to change.

client/packages/platform/__tests__/src/migrations.test.ts
```diff
@@ -0,0 +1,398 @@
...
+});
+
+test('delete and add - intent', async () => {
+  const result = await diffSchemas(
+    simpleSchemaBefore,
+    simpleSchemaAfter,
+    createChooser([]),
+  );
+  console.log(result);
+  expectTxType(result, 'delete-attr', 3);
```
stopachka, line 89: (discussed, but would be a bit more explicit about the checks, i.e what attrs are deleted, how the structure, looks, etc)

client/packages/cli/tsconfig.json
stopachka: Hmm good q! I guess it doesn't hurt to keep em?

server/src/instant/dash/routes.clj
```diff
@@ -1284,6 +1285,19 @@
...
+                                                                          :apps/write
+                                                                          req)
+
+        input-steps (ex/get-param! req [:body :steps] #(when (coll? % ) %))
+        coerced (tx/coerce! input-steps)
+        plan-result (schema-model/apply-plan-with-deletes! app-id coerced)
+       ]
+    (response/ok plan-result))
+  )
+
```
```suggestion
        plan-result (schema-model/apply-plan-with-deletes! app-id coerced)]
    (response/ok plan-result)))

```

PR #1788, opened 2025-10-06
fix the sidebar again

drew-harris:
Not sure why it keeps disappearing but might be something to do with the use of opacity? Setting it to a solid color this time

BEFORE:
image

AFTER:
image


stopachka (approved):
SGTM!

PR #1789, opened 2025-10-06
Update type for admin sdk's `getUser` to omit the refresh token

dwwoelfel:
We don't expose the auth token when we fetch the user in the admin sdk. This updates the type to match what we actually do.

An alternative would be to generate a new refresh_token each time you call `getUser`, but that seems like a bad idea.


stopachka (approved):
SGTM!

PR #1796, opened 2025-10-07
Add error handling to the persister

dwwoelfel:
Adds a try/catch when we fetch data from the store. This will prevent an error in the store from completely blocking an app from working.

Prerequisite for https://github.com/instantdb/instant/pull/1793, so that users can roll back to this version if there is something wrong with that version.


stopachka (approved):
Nice!

PR #1799, opened 2025-10-07
Guard against the case where etype is missing

dwwoelfel:
Not sure how this happens and there's not enough info in the logs to reproduce. Hopefully this will get us further along to a better error message (right now it just throws `Cannot invoke "Object.toString()" because "s" is null`.


nezaj (approved):
SGTM!

PR #1800, opened 2025-10-07
add temporary tshy package.jsons to .gitignore

drew-harris:


.gitignore
```diff
@@ -218,6 +218,13 @@ server/resources/metrics/*-usage.png
 # emails
 server/resources/emails/*
 
-# algolia 
+# algolia
 client/www/scripts/algolia-objects.json
 client/packages/mcp/.env
+
```
stopachka, line 224: let's add a note like 'tshy'?

drew-harris: smart

PR #1802, opened 2025-10-07
Pass through inference? for admin subscribe query

dwwoelfel:
This fixes hasOne links in admin subscribe query. We weren't propagating the `inference?` field, so we were always returning an array.


stopachka (approved):
Biyutiful!

PR #1803, opened 2025-10-07
Add copy as markdown button for docs

nezaj:
Adds a copy as markdown button for our docs! Stopa mentioned he wanted this and I thought it would be useful too!

CleanShot 2025-10-07 at 16 40 57@2x


stopachka (approved):
SGTM!

PR #1804, opened 2025-10-07
Sandbox and Query Inspector improvements

dwwoelfel:
### Query Inspector

Passes a schema (generated from the attrs) to the query inspector so that users won't get arrays for `hasOne` links.

### Sandbox

1. Adds a generated `instant.schema.ts` to the monaco editor to give us better types for `db.query` and `db.transact`.

2. Runs prettier when you do `Opt+Shift+F`. It's even aware of how wide the pane is, so it will shrink or expand depending on the width. We delay loading the prettier library until you actually use the shortcut.

image
image
image

### DIFF
--- a/client/www/components/dash/Sandbox.tsx
+++ b/client/www/components/dash/Sandbox.tsx
@@ -596,6 +596,7 @@ export function Sandbox({
             >
               <div className="flex-1">
                 <Editor
+                  key={app.id}
                   theme={darkMode ? 'vs-dark' : 'light'}
                   height={'100%'}
                   path="sandbox.ts"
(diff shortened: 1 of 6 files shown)


client/www/components/dash/Sandbox.tsx
```diff
@@ -311,6 +342,34 @@ export function Sandbox({
     consoleRef.current?.scrollTo(0, consoleRef.current.scrollHeight);
   }, [output]);
 
+  const prettify = async (editor: Parameters<OnMount>[0]) => {
+    const code = editor.getValue();
+    const [prettier, tsPlugin, estreePlugin] = await Promise.all([
+      import('prettier/standalone'),
```
stopachka, line 347: cool!

stopachka (approved):
Beautiful!

PR #1812, opened 2025-10-09
Add Github OAuth

nezaj:
Saw another user request for this and was inspired by dww's earlier use of codex for LinkedIn. Gave it a shot to do a similar thing for Github (did this with Claude).

Github uses OAuth2 vs OpenID so needed to handle it differently. Github also has a notion of "private" emails so needed to special handle that.

To help with abstracting the differences between OAuth/OID clients this adds an `oauth_providers.clj` where we config settings for Github/Google/Apple/LinkedIn/etc -- I think this should make it easier to add other oauth2 clients as well.

This PR also includes the following:

* Github Clients section in the dashboard
* Docs explaining how to add it yourself

Tested this works in both web and expo. Also verified Google and LinkedIn still work as expected in web/expo

* [Gist example for web](https://gist.github.com/nezaj/b7037ce828989289b9e8d1db9e2e4ea5)
* [Gist example for expo](https://gist.github.com/nezaj/9ddb42fcdeaa53de04ccd100037f7d27)

If you want to give it a spin, you should be able to [follow the docs](https://instant-www-js-gh-auth-jsv.vercel.app/docs/auth/github-oauth) and get it working! Only difference is to set the authorization callback in your github app settings to your local machine

CleanShot 2025-10-08 at 17 11 28@2x

**web**

https://github.com/user-attachments/assets/4da4da11-c4d7-4124-b48c-c64b5c6a3ff4

**expo**

https://github.com/user-attachments/assets/04422e68-7201-441f-a104-bb04b0d5dba8

### DIFF
--- a/server/src/instant/model/app_oauth_client.clj
+++ b/server/src/instant/model/app_oauth_client.clj
@@ -103,8 +103,8 @@
       (Secret.)))
 
 (defn ->OAuthClient [oauth-client]
-  (if (:discovery_endpoint oauth-client)
-    ;; OIDC provider with discovery endpoint
+  (cond
+    (:discovery_endpoint oauth-client)
     (oauth/generic-oauth-client-from-discovery-url
      {:app-id (:app_id oauth-client)
       :provider-id (:provider_id oauth-client)
@@ -114,12 +114,14 @@
       :discovery-endpoint (:discovery_endpoint oauth-client)
       :meta (:meta oauth-client)})
 
-    ;; Otherwise, assume GitHub, our only non-OIDC provider which
-    ;; doesn't use discovery endpoints.
+    (= "github" (get (:meta oauth-client) "providerName"))
     (oauth/map->GitHubOAuthClient
      {:app-id (:app_id oauth-client)
       :provider-id (:provider_id oauth-client)
       :client-id (:client_id oauth-client)
       :client-secret (when (:client_secret oauth-client)
                        (decrypted-client-secret oauth-client))
-      :meta (:meta oauth-client)})))
+      :meta (:meta oauth-client)})
+
+    :else
+    (throw (ex-info "Unsupported OAuth client" {:oauth-client oauth-client}))))
--- a/client/www/components/dash/auth/GitHub.tsx
+++ b/client/www/components/dash/auth/GitHub.tsx
@@ -144,6 +144,7 @@ export function AddClientForm({
         clientName,
         clientId,
         clientSecret,
+        meta: { providerName: 'github' },
       });
       onAddClient(resp.client);
     } catch (e) {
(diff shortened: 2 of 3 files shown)


server/src/instant/auth/oauth.clj
```diff
@@ -25,6 +25,78 @@
...
+                              meta]
+  OAuthClient
+  (create-authorization-url [_ state redirect-url extra-params]
+    (let [base-params {:scope (:default-scope github-config)
+                       :response_type "code"
+                       :state state
+                       :redirect_uri redirect-url
+                       :client_id client-id}
+          params (merge base-params
+                        (or (select-keys extra-params allowed-extra-params)
```
dwwoelfel: Do the allowed-extra-params work with `GitHub`? If not, I'd just remove that part.

server/src/instant/auth/oauth.clj
```diff
@@ -25,6 +25,78 @@
...
+                                          {:throw-exceptions false
+                                           :as :json
+                                           :coerce :always
+                                           :headers (merge (:headers github-config)
+                                                           {"Authorization" (str "Bearer " access-token)})})]
+              (if-not (clj-http/success? user-resp)
+                {:type :error :message (get-in user-resp [:body :message] "Failed to fetch user info from GitHub.")}
+                (let [user-data (:body user-resp)
+                      user-id (:id user-data)
+                      email (:email user-data)
```
dwwoelfel: We might want to hit [`/user/emails`](https://docs.github.com/en/rest/users/emails?apiVersion=2022-11-28#list-email-addresses-for-the-authenticated-user) to get the primary verified email. The docs from the `user` endpoint that I found don't mention whether the user.email field is verified.

server/src/instant/auth/oauth.clj
nezaj: Updated!

server/src/instant/model/app_oauth_client.clj
```diff
@@ -102,21 +103,23 @@
...
       :client-secret (when (:client_secret oauth-client)
                        (decrypted-client-secret oauth-client))
-      :discovery-endpoint discovery-endpoint
+      :discovery-endpoint (:discovery_endpoint oauth-client)
       :meta (:meta oauth-client)})
-    (oauth/map->GenericOAuthClient
+
+    ;; Otherwise, assume GitHub, our only non-OIDC provider which
+    ;; doesn't use discovery endpoints.
+    (oauth/map->GitHubOAuthClient
```
dwwoelfel, line 118: We should be able to check by the provider name whether it is GitHub. But that might be a little annoying to fetch, so you could also just put that it's `github` in the `meta` field on the oauth client.

server/src/instant/auth/oauth.clj
nezaj: Updated!

server/src/instant/model/app_oauth_client.clj
nezaj: Updated!

PR #1816, opened 2025-10-09
Add instructions for codex in the docs

nezaj:
This does the following:

* Adds instructions for how to use the instant mcp with codex
* Re-orders sections so it's Claude, Codex, Cursor, Windsurf, Zed, Others
* Removes local MCP instructions -- feels like most major clients support remote MCP now

### DIFF
--- a/client/www/pages/docs/using-llms.md
+++ b/client/www/pages/docs/using-llms.md
@@ -20,17 +20,6 @@ You can verify you set up the rules correctly by asking your LLM "How do you
 make queries and transactions in InstantDB?" If everything is set up correctly,
 you should see a response with information about `db.useQuery` and `db.transact`
 
-### Cursor
-
-Save these rules for your framework at the root of your project in `.cursor/rules/instant.mdc`:
-
-- [Next.js rules](/llm-rules/next/cursor-rules.md)
-- [Expo rules](/llm-rules/expo/cursor-rules.md)
-
-You may need to restart Cursor for them to take effect.
-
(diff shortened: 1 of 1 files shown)


client/www/pages/docs/using-llms.md
```diff
@@ -20,21 +20,26 @@ You can verify you set up the rules correctly by asking your LLM "How do you
...
-### Claude Code
+If you already had claude running, restart it for the rules to take effect.
+
+### Codex
 
 Save these instructions for your framework at the root of your project as
-`CLAUDE.md`:
+`AGENTS.md`:
 
 - [Next.js instructions](/llm-rules/next/claude.md)
```
nezaj, line 44: The format for codex is the same as claude, only difference is codex calls their file `AGENTS.md` instead of `CLAUDE.md`

client/www/pages/docs/using-llms.md
```diff
@@ -209,111 +237,6 @@ Or our SSE endpoint
 https://mcp.instantdb.com/sse
 ```
 
-## Local MCP server
```
dwwoelfel, line 212: You might want to keep the local instructions. The github copilot agent wants a local mcp server, for example: https://docs.github.com/en/copilot/how-tos/use-copilot-agents/coding-agent/extend-coding-agent-with-mcp#example-sentry

nezaj: Gotcha! Removed this commit!

PR #1819, opened 2025-10-09
feat: Add profile image support to OAuth flow

ayxliu19:
Add profile image (picture) field throughout the OAuth authentication pipeline, allowing applications to display user profile pictures from OAuth providers like Google.

Changes:
- Extract 'picture' claim from OAuth ID tokens in both Google-specific and generic OAuth flows
- Store profile image URL in $users system catalog as 'image' attribute
- Add image field to User type in client TypeScript definitions
- Update app_user model to support creating and updating image field
- Pass picture data through OAuth redirect and token exchange flows
- Display user profile image in sandbox demo UI

Screenshot 2025-10-09 at 3 46 39 PM

Test Google OAuth Login

  1. Navigate to the sandbox demo at http://localhost:4000/play/googleloginbutton
  2. Click the Google sign-in button
  3. Complete the Google OAuth flow
  4. After successful authentication, you should see:
    - Your profile picture displayed as a rounded avatar (16x16 size)
    - Your email address next to it
    - Your user ID below

### DIFF
--- a/client/packages/version/src/version.ts
+++ b/client/packages/version/src/version.ts
@@ -1,6 +1,6 @@
 // This is the shared version for all of the js packages
 // Update the version here and merge your code to main to
 // publish a new version of all of the packages to npm.
-const version = 'v0.22.13';
+const version = 'v0.22.14';
 
 export { version };


server/src/instant/system_catalog.clj
```diff
@@ -52,6 +52,7 @@
   {"id" "id"
    "email" "email"
    "type" "type"
+   "image" "image"
```
stopachka: Could we call this `imageURL`? 

Main reasoning: 
- It matches a bit of Clerk's API, so users (and LLMs) can intuit the right name: 
  - https://clerk.com/docs/guides/development/image-optimization
- People may think an `image` is a an object, where you can define width, height etc. (This is done in graphql schemas)

ayxliu19: Makes sense!

server/src/instant/model/app_user.clj
```diff
@@ -86,6 +88,17 @@
       (transact! [[:add-triple id (resolve-id :email) email]])
       (get-entity id)))))
 
+(defn update-image!
```
stopachka: Small nit: would remove this, since it's unused.

PR #1820, opened 2025-10-10
SSE fallback when websockets are missing

dwwoelfel:
Adds fallback to SSE when the WebSocket fails.

## How it works

The client will try to connect over the websocket. If the websocket connection fails before we receive any messages, we'll try to connect over SSE. Each time we reconnect, we'll try to connect over the websocket again. This way if there is some temporary network glitch, the user won't be stuck on the slower SSE connection. 

There are very few changes to reactor. We created a new connection interface that implements a common set of functions for both `WebSocket` and `EventSource`.

When the user first connects via SSE, we will send a message down the pipe with a `machine-id,` `session-id`, and `sse-token`. On the client, we'll wait until we get that initial message to mark our connection as open because we need the params before we can send messages.

When we send a message with the SSE interface, we'll make a POST request with the `machine-id`, `session-id`, and `sse-token` params. The server that receives the request will use the `machine-id` to identify which instance to direct the message to. That will be the machine that is holding on to the session. It will execute a task on that instance with Hazelcast. 

The hazelcast task will check that the `sse-token` is the correct token for the session-id. If not, it will reject the request. The token prevents other users from hijacking the SSE session. If the token is correct, then we'll put the message into the receive queue (just as if it had come through the `on-message` handler for a websocket).

### React Native

React Native doesn't like the `eventsource` package, so I embedded the EventSource implementation from https://github.com/binaryminds/react-native-sse/blob/master/src/EventSource.js (with some heavy modification to match our needs). 

It also rearranging the imports to `react-native` from `@instantdb/react`. The `react` package needs to include the `eventsource` package, but react native will complain if it's in the `react` `index.js`. We only need to import `ReactAbstractDatabase` for react native, so I created a new `index.native.js` that react-native can import from.

## Incidental changes:

1. Updates typescript to 5.9.3 to fix an compile error I was getting.
2. Includes a caddyfile.dev and instructions for running multiple instances + ssl with caddy. Caddy makes it easier to test SSE. It uses http/2/3, so you don't hit connection limits, and it can route to multiple instances, so you can test with multiple jvms.

### DIFF
--- a/client/packages/version/src/version.ts
+++ b/client/packages/version/src/version.ts
@@ -1,6 +1,6 @@
 // This is the shared version for all of the js packages
 // Update the version here and merge your code to main to
 // publish a new version of all of the packages to npm.
-const version = 'v0.22.14';
+const version = 'v0.22.15';
 
 export { version };


client/packages/core/src/Connection.ts
```diff
@@ -0,0 +1,199 @@
...
+  }
+}
+
+type SSEInitParams = {
+  machineId: string;
+  sessionId: string;
+  sseToken: string;
+};
+
+export class SSEConnection implements Connection<EventSource> {
```
stopachka: Beautiful!

client/packages/admin/src/index.ts
```diff
@@ -733,7 +733,7 @@ class Storage {
     );
     const { ok } = await fetch(presignedUrl, {
       method: 'PUT',
-      body: file,
+      body: file as unknown as BodyInit,
```
dwwoelfel, line 736: This was a type error with the upgrade to typescript 5.9.3

server/README.md
```diff
@@ -81,7 +82,37 @@ make compile-java
...
-# Setting up local https
+# Setting up local https and SSE
+
+There are two options for local https.
+
+Use undertow if you don't want any other dependencies and you're not testing SSE or multiple hazelcast instances.
+
+Use caddy if you want to test SSE locally (without hitting connection limits) or want to test with multiple local instant server instances.
+
+## caddy
```
stopachka, line 93: cool!

PR #1827, opened 2025-10-13
Add dark mode for explorer links

nezaj:
Adds dark themes to the link box and it's tool tips

**Before**
CleanShot 2025-10-13 at 10 08 20@2x
CleanShot 2025-10-13 at 10 08 32@2x
CleanShot 2025-10-13 at 10 09 06@2x

**After**
CleanShot 2025-10-13 at 10 10 35@2x
CleanShot 2025-10-13 at 10 35 39@2x
CleanShot 2025-10-13 at 10 10 08@2x

### DIFF
--- a/client/www/components/dash/explorer/EditRowDialog.tsx
+++ b/client/www/components/dash/explorer/EditRowDialog.tsx
@@ -451,7 +451,7 @@ function LinkCombobox({
           portal={false}
           unmount={false}
           static={true}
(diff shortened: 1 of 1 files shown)


client/www/components/dash/explorer/EditRowDialog.tsx
drew-harris: this should have a `dark:border-neutral-700` to match our other dropdowns too

PR #1833, opened 2025-10-15
Batching transact! in the datascript store

dwwoelfel:
Adds batching to `transact!` in `instant.reactive.store`.

We're spending a lot of time waiting for the lock to release. This PR reduces the amount of work we do in `transact!` by batching the transactions.

When we call `transact!`, we'll put the tx-data on a queue and then dispatch a job on a thread (one virtual thread per datascript db) that will perform up to 50 transacts at a time. Instead of doing `d/transact!` for each one, we'll do `d/with`. That way we only have to do one `reset!` instead of 50 `swap!`s. 

It's behind a flag for now. The default is to use the old transact!. We can turn it on for individual apps or turn it on globally.

One thing I'm worried about is that we use a virtual thread per datascript db to execute the transacts. If it causes issues, we could switch to a regular thread, but we'd have to come up with some way to share threads between multiple datascript dbs.

Also makes `remove-session!` more efficient. We lock around less code and we have a faster check for whether this is the last session for the app.

### DIFF
--- a/server/src/instant/reactive/store.clj
+++ b/server/src/instant/reactive/store.clj
@@ -96,35 +96,10 @@
    :datalog-query/delayed-call {} ;; delay with datalog result (from query.clj)
    :datalog-query/topics {:db/type :db.type/list-of-topics}})
 
-(defn create-conn [schema app-id]
-  (doto (d/create-conn schema)
-    (alter-meta! assoc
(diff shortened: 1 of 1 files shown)


stopachka (approved):
Biyutiful!

PR #1835, opened 2025-10-15
instantify cli

drew-harris:
image
image
image
image

image

image

UI refresh for all instant cli apps (cia and instant-cli)
New widget to connect apps (create ephemeral, create app, select app, change org)
Fixes a typo in the resolve renames ui (and redoes it in general)

Left to do: 
- [x] make sure -y works for cli
- [x] create an org scoped app in instant cli (works but bad ui)
- [x] Some tweaks on individual promptOks

### DIFF
--- a/client/packages/version/src/version.ts
+++ b/client/packages/version/src/version.ts
@@ -1,6 +1,6 @@
 // This is the shared version for all of the js packages
 // Update the version here and merge your code to main to
 // publish a new version of all of the packages to npm.
-const version = 'v0.22.20';
+const version = 'v0.22.21';
 
 export { version };


stopachka:
Exciting!!

client/packages/create-instant-app/src/scaffold.ts
```diff
@@ -1,52 +1,51 @@
...
+          ],
+          defaultValue: 'abort',
+          modifyOutput: UI.ciaModifier(),
+        }),
+      );
 
-      if (p.isCancel(overwriteDir) || overwriteDir === 'abort') {
-        spinner.stop('Aborting installation...');
+      if (overwriteDir === 'abort') {
+        // TODO: REPLACE
```
stopachka: flagging in case todo needs to be removed

drew-harris: removed the comment

client/packages/cli/src/ui/index.ts
```diff
@@ -0,0 +1,1127 @@
...
+  type SelectProps<T> = {
+    options: {
+      value: T;
+      label: string;
+    }[];
+    promptText: string;
+    modifyOutput?: ModifyOutputFn;
+    defaultValue?: T;
+  };
+  export class Select<T> extends Prompt<T> {
```
stopachka, line 98: small nit: I would consider moving these out of the namespace definition, writing them first, then inlining the names. 

Here's my reasoning: 
- This adds a 2 sp padding on all the code
- From a first glance, it can feel a bit scary, because it makes one think: is there some kind of state sharing going on, and hence why it's indented under something? 
- It will also make it easier to read what all the exports of the namespace are. 

WDYT?

client/packages/cli/src/renderSchemaPlan.ts
```diff
@@ -268,9 +268,19 @@ export const renderSchemaPlan = (
         if (step.partialAttr['value-type'] === 'ref') {
           addLine(renderLinkUpdate(step, prevAttrs));
         } else {
-          addLine(
-            `${chalk.bgYellow.black(' * UPDATE ATTR ')} ${createDotName(step)}`,
-          );
+          if (
+            step.partialAttr['forward-identity']?.attrName &&
```
stopachka: nit: would consider moving this into a variable name, like `isRename`, or even making that a little helper function

client/packages/cli/src/toggle.ts
```diff
@@ -1,76 +0,0 @@
-// From:
-// https://github.com/skarahoda/inquirer-toggle
-import type { Theme } from '@inquirer/core';
```
stopachka, line 3: 🔥🫡

client/packages/cli/src/ui/lib.ts
```diff
@@ -0,0 +1,345 @@
...
+}
+
+export class CancelledPromptError extends Error {
+  constructor(message?: string) {
+    super(message);
+    this.name = 'CancelledPromptError';
+  }
+}
+
+export async function renderUnwrap<T>(view: Prompt<T>): Promise<T> {
```
stopachka, line 285: nit: would consider adding a docstring to this. It requires some context to figure out what this is meant to do.

stopachka (approved):
Awesome work @drew-harris !

client/packages/cli/src/ui/index.ts
drew-harris: It was definitely more of a stylistic choice i made to make it easier to import the components in the future since most of this stuff won't need to be edited much in the future baring bugfixes. Nearly everything is exported from the module and if people want the full list then they can just search "export" in their editor. 
As for the state sharing thats possible in any module at the top level too but i see how the indentation makes it feel "unsafer" 

I think its worth keeping these sacrifices though for how easy it is to import UI.Select and still get the tree shaking.

stopachka: SGTM!

PR #1836, opened 2025-10-15
Include session info in subscribe query payloads

dwwoelfel:
Updates the admin sse connection to send down the `sse-init` message.

This allows us to expose the session_id when we deliver payloads, which can be helpful for debugging.

Payloads look like this:
image

### DIFF
--- a/client/packages/version/src/version.ts
+++ b/client/packages/version/src/version.ts
@@ -1,6 +1,6 @@
 // This is the shared version for all of the js packages
 // Update the version here and merge your code to main to
 // publish a new version of all of the packages to npm.
-const version = 'v0.22.18';
+const version = 'v0.22.19';
 
 export { version };


server/src/instant/util/tracer.clj
```diff
@@ -103,6 +103,7 @@
         {:keys [code-ns code-line code-file]} source
         default-attributes (cond-> {"host.name" @config/hostname
                                     "process-id" @config/process-id
+                                    "machine-id" config/machine-id
```
nezaj, line 106: Should this be `@config/machine-id` ?

nezaj (approved):
LGTM w/ one question!

server/src/instant/util/tracer.clj
nezaj: Ah looked into config.clj and see it should not!

dwwoelfel: No, we don't create it in a delay because it's just a uuid. The other ones require some work to generate.

PR #1837, opened 2025-10-16
Replace YourKit with Java Mission Control

dwwoelfel:
Our free YourKit license expired, so this PR replaces it with java mission control.

We start recording when the jvm starts and keep up to 1gb of recording data, which will allow us to look back in time if we notice an issue on the server (if we're quick enough).

Tested in staging.


nezaj (approved):
🚢!

PR #1840, opened 2025-10-17
disable scrolling on number input

drew-harris:
disables scrolling to change the number on a number input. 
tested in safari and chrome


stopachka (approved):
Nit: maybe we move this out of `patchFirefoxClicks`, and something like `patchWheelNavigation`, with a note as to why? 

Reasoning: we may want to delete patchFirefoxClicks later, if firefox ever fixes this

PR #1841, opened 2025-10-17
Fix bug validating required fields with shared entity ids

dwwoelfel:
Fixes a bug in validate-required! when you modify an entity with the same id as the entity you're deleting. We weren't properly checking the etype on the modified entity.


stopachka (approved):
Biyutiful!

PR #1842, opened 2025-10-17
allow login in create-instant-app

drew-harris:
Adds a new option to login during the scaffold so an existing user can connect an app / make a permanent one

### DIFF
--- a/client/packages/version/src/version.ts
+++ b/client/packages/version/src/version.ts
@@ -1,6 +1,6 @@
 // This is the shared version for all of the js packages
 // Update the version here and merge your code to main to
 // publish a new version of all of the packages to npm.
-const version = 'v0.22.23';
+const version = 'v0.22.24';
 
 export { version };


client/packages/create-instant-app/src/login.ts
```diff
@@ -225,3 +258,53 @@ export const tryConnectApp = async (
 
   return selectedApp;
 };
+
+function sleep(ms: number) {
+  return new Promise((resolve) => setTimeout(resolve, ms));
+}
+async function waitForAuthToken({
```
stopachka, line 265: Now that we re-use UI across cli and cia, maybe we should start exposing these kind of functions too? 

imo it would be nice if we could re-use this logic

client/packages/create-instant-app/src/login.ts
```diff
@@ -168,6 +175,32 @@ export const tryConnectApp = async (
       }),
     );
 
+    if (choice === 'login') {
+      const registerRes = await fetchJson<any>({
```
stopachka, line 179: Could we extract this part into a function used across cia and cli, or does cli have more specific stuff in it? 

If possible, would re-use, but if a bit of a burden, would skip

client/packages/create-instant-app/src/login.ts
drew-harris: I'm all for sharing functionality especially now since we are importing things from the cli but that should be done in a separate pr since it'd involve a major refactor since everything has console.logs and error handling that doesn't match how cia does it.

client/packages/create-instant-app/src/login.ts
drew-harris: Just like previous response, the cli version has console.log messages everywhere that doesn't match the format for cia and a different style of error handling.

client/packages/create-instant-app/src/login.ts
stopachka: SGTM!

PR #1844, opened 2025-10-17
renames via flags

drew-harris:
Ability to specify renames when pushing the schema
Example: `instant-cli push --rename posts.author:posts.creator`

### DIFF
--- a/client/packages/version/src/version.ts
+++ b/client/packages/version/src/version.ts
@@ -1,6 +1,6 @@
 // This is the shared version for all of the js packages
 // Update the version here and merge your code to main to
 // publish a new version of all of the packages to npm.
-const version = 'v0.22.22';
+const version = 'v0.22.23';
 
 export { version };


client/packages/cli/src/index.js
```diff
@@ -365,6 +369,10 @@ program
   .command('push-perms', { hidden: true })
   .argument('[app-id]')
   .description('Push perms to production.')
+  .option(
```
nezaj: Does push-perms needs rename?

client/packages/cli/src/index.js
```diff
@@ -386,6 +394,10 @@ program
     "Don't check types on the server when pushing schema",
   )
   .option('-t --title', 'Title for the created app')
+  .option(
+    '--rename [renames...]',
+    'List of full attribute names separated by a ":"\n Example:`push --rename posts.author:posts.creator`',
```
nezaj: Nit. Might be nice to give an example of renaming two things so someone knows how to do mulitple renames

client/packages/cli/src/index.js
```diff
@@ -353,6 +353,10 @@ program
     '--skip-check-types',
     "Don't check types on the server when pushing schema",
   )
+  .option(
+    '--rename [renames...]',
```
nezaj: imo we don't need to support `--rename`` for `push-schema` since it's legacy for ~1yr now and we want people on the new `push` command

client/packages/cli/src/index.js
```diff
@@ -1444,6 +1455,10 @@ async function pushSchema(appId, _opts) {
   try {
     const groupedSteps = groupSteps(diffResult);
     const lines = renderSchemaPlan(groupedSteps, currentAttrs);
+    if (program.optsWithGlobals().yes) {
```
nezaj, line 1478: Curious, why this check / console? If we include `yes` don't we want to avoid prompts?

client/packages/cli/src/index.js
```diff
@@ -2040,3 +2055,77 @@ async function validateAppLinked() {
     process.exit(1);
   }
 }
+
+function buildAutoRenameSelector(opts) {
```
nezaj: Might be nice to summarize this behavior in a doc-string. Also could be nice to have some simple tests to illustrate how this would work

client/packages/cli/src/index.js
drew-harris: that flag shouldn't be available for this command, removing it

client/packages/cli/src/index.js
drew-harris: 👍

client/packages/cli/src/index.js
drew-harris: the promptOk wont show the fancy migration output, so if I don't print it out it will look like nothing happened

client/packages/cli/src/index.js
drew-harris: tests added

PR #1845, opened 2025-10-17
Import all exact counts

dwwoelfel:
Makes allExactCounts an import instead of storing it in the public directory.

Benefits to this approach:
1. Type safety
2. It's part of the js bundle, so if you ship a change to the format of allExactCounts, people who are roaming the site won't be affected
3. Less code
4. The browser might preload it?


stopachka (approved):
I like it!

PR #1847, opened 2025-10-20
[Dash] Hide app id in tooltip

nezaj:
Got a report about this from a user. We'll now show the redacted value in the tooltip when in a hidden state

**Before**

CleanShot 2025-10-20 at 09 32 12@2x

**After**

CleanShot 2025-10-20 at 09 37 54@2x

### DIFF
--- a/client/www/components/ui.tsx
+++ b/client/www/components/ui.tsx
@@ -909,7 +909,7 @@ export function SmallCopyable({
               truncate: !multiline,
               'whitespace-pre-wrap break-all': multiline,
             })}
-            title={hideValue || hidden ? 'Copy App ID to Clipboard' : value}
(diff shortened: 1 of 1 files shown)


client/www/components/ui.tsx
```diff
@@ -995,7 +995,7 @@ export function Copyable({
           truncate: !multiline,
           'whitespace-pre-wrap break-all': multiline,
         })}
-        title={value}
+        title={hideValue || hidden ? redactedValue(value) : value}
```
dwwoelfel:
```suggestion
        title={hideValue || hidden ? 'Copy App ID to Clipboard' : value}
```

This is what we do for the other app id input https://github.com/instantdb/instant/pull/1735/files

nezaj: Updated to `Copy to clipboard` !

PR #1848, opened 2025-10-20
Track bytes sent/recieved for a sql query

dwwoelfel:
Adds tracking for sql queries, counting how much data we send and receive through the underlying socket.

We'll use this for datalog query caching. We want to limit the size of the data in our cache, but we don't know how much memory the result of `d/query` uses. We could walk the datastructure to count the size of all of the nodes, but that would slow down our queries. Instead, we can get a pretty good approximation by counting the amount of data that postgres used to transfer the query result.

### How it works

The postgres driver lets you define a `SocketFactory` that it will use to generate the socket for its data stream. We create a new java class called `instant.WrappedSocket` that creates "CountingSocket"s. They're just extensions of the default `Socket` class, but they count the number of bytes that flow through them.

We associate the socket to the connection through a couple of weak maps.

Then when we run a query we get the count of bytes by subtracting the socket's byte count before and after we run the query.

There could be some miscounting if we start handling postgres notifications through the connections in our connections pool, but I don't think we'll do that.

There's a nice new `(tool/recompile-java)` function in dev that makes it a little easier to iterate on java classes. You still end up restarting the jvm quite a bit, but you don't have to do it as often.

### DIFF
--- a/server/src/instant/jdbc/socket_track.clj
+++ b/server/src/instant/jdbc/socket_track.clj
@@ -14,15 +14,22 @@
 
 (def ^:dynamic *connection-id* nil)
 
+;; WeakMap to store connection-id -> Socket
+;; Autoevicts when the socket is GC'd
 (defonce socket-map (-> (MapMaker.)
                         (.weakValues)
                         (.makeMap)))
 
+;; WeakMap to store PgConnection -> connection-id
+;; Autoevicts when the connection is GC'd
 (defonce connection-map (-> (MapMaker.)
                             (.weakKeys)
                             (.makeMap)))
 
-(defn socket-for-connection ^CountingSocket [conn]
+(defn socket-for-connection
+  "Gets the socket for the connection, unwrapping the connection
+   to get the underlying PgConnection if necessary."
+  ^CountingSocket [conn]
   (when-let [conn (cond (instance? HikariProxyConnection conn)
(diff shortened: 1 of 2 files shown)


server/build.clj
```diff
@@ -8,6 +8,17 @@
 (def basis (delay (b/create-basis {:project "deps.edn"})))
 
 (defn compile-java [_]
+  ;; First build the socketutil directory because instant.jdbc.socket-track
+  ;; depends on it
+  (b/javac {:src-dirs ["src/java/instant/socketutil"]
```
stopachka, line 13: Nice that we have this kind of control!

server/src/instant/jdbc/aurora.clj
```diff
@@ -191,10 +193,13 @@
   "Creates a new connection for the connection pool and sets the idle_in_transaction_session_timeout.
    Defaults to one minute, but can be modified with a flag in an emergency."
   [config]
-  (let [conn (next-jdbc/get-connection config)]
-    (next-jdbc/execute! conn ["select set_config('idle_in_transaction_session_timeout', ?::text, false)"
-                              (flags/flag :idle-in-transaction-session-timeout (* 1000 60))])
-    conn))
+  (binding [socket-track/*connection-id* (random-uuid)]
```
stopachka, line 196: man are bindings powerful!

PR #1849, opened 2025-10-20
Quieter exceptions when session is missing

dwwoelfel:
If we get an sse message for a session that is already closed, then we'll make a bunch of noise in the errors channel.

There's no need for us to be noisy about it--the client will reconnect if it gets the error (it's more likely that they navigated away from the page before the message finished sending).

We will still add info about the instant error we encounted, but those will just be regular attributes on the span instead of a full error event with stack trace.


nezaj (approved):
LGTM!

PR #1850, opened 2025-10-20
instant-cli claim command for ephemeral apps

drew-harris:
run `instant-cli` claim to transfer an ephemeral app into your account

### DIFF
--- a/client/packages/version/src/version.ts
+++ b/client/packages/version/src/version.ts
@@ -1,6 +1,6 @@
 // This is the shared version for all of the js packages
 // Update the version here and merge your code to main to
 // publish a new version of all of the packages to npm.
-const version = 'v0.22.21';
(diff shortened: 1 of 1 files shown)


client/packages/cli/src/index.js
```diff
@@ -449,6 +454,31 @@ program
...
+  if (!envResult.adminToken) {
+    error('No admin token found in environment variables.');
+    return process.exit(1);
+  }
+
+  const appId = envResult.appId.value;
+  const adminToken = envResult.adminToken.value;
+
+  console.log(`Found ${chalk.green(envResult.appId.envName)}: ${appId}`);
+  console.log(
```
dwwoelfel: Maybe we shouldn't log the admin token in case someone is doing this on a stream or something.

PR #1852, opened 2025-10-20
custom names for expo scaffold

drew-harris:
Uses the users custom project name in the app.json and the _layout.ts file


stopachka (commented):
Yess

PR #1855, opened 2025-10-20
Add title input to create-app option

Elijer:
I intentionally returned to a02c847a (0.22.20) because it seems like `create-app` was removed since then, and I don't know what's up with that.

But this small change would allow a title to be passed in.


nezaj:
This is now supported via `init-without-files` https://github.com/instantdb/instant/pull/1949

Thanks for flagging this!

PR #1859, opened 2025-10-21
Capped datalog query cache

dwwoelfel:
This replaces our datalog query cache in the datascript store with a Caffeine `AsyncLoaderCache`.

Unlike our current cache, items will get evicted if they exceed a maximum "weight". We determine the weight of each item by tracking how many bytes the underlying sql query generated. It's not exactly the same as the memory that the actual result uses, but it gets us in the ballpark. We multiply by a constant (configurable with a flag) to get a number that's closer to the actual number.

Our old cache had a nice feature where it would cancel pending queries if all of the subscribers left. It took a lot of effort, but I managed to preserve that behavior. When the datalog-query is evicted from the datascript store or when the last listener is interrupted, we cancel the query unless it has any threads waiting on the result. Canceling was a little tricky, but I managed to make it work with a custom `ExecutorService` that uses our tracked futures.

We attach lots of stats for the cache to the `transact` span when we evict anything. That will help us keep track of how the cache is doing. We can configure the maximum weight, weight multiple, and ttl through flags if we need to tweak anything.

### DIFF
--- a/server/src/instant/reactive/store.clj
+++ b/server/src/instant/reactive/store.clj
@@ -232,7 +232,7 @@
    queries that nobody is waiting on a result for.
 
    Should run after `transact!`"
-  [conn {:keys [db-before tx-data]}]
+  [conn {:keys [tx-data]}]
   (let [deleted-datalog-query-ids (keep (fn [datom]
                                           (when (and (= (:a datom)
                                                         :datalog-query/app-id+query)


server/src/tool.clj
```diff
@@ -354,7 +354,7 @@
   `(do
      (let [name# (quote ~value)
            result# ~value]
-       (println (pr-str name#) "is" (pr-str result#))
+       (println "INSPECT" (pr-str name#) "is" (pr-str result#))
```
dwwoelfel, line 357: This makes it easier to find the inspect in the logs

PR #1860, opened 2025-10-21
[cli] Remove checkVersion

stopachka:
- Removed unused function `linkOptsPretty` 
- Removed `checkVersion`

In the future, we can have http endpoints refuse if we need some version requirement. No need for the client too. 

### DIFF
--- a/client/packages/create-instant-app/package.json
+++ b/client/packages/create-instant-app/package.json
@@ -32,14 +32,16 @@
     "@clack/prompts": "^0.11.0",
     "@ianvs/prettier-plugin-sort-imports": "^4.2.1",
     "@instantdb/version": "workspace:*",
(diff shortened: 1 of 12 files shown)


client/packages/cli/src/index.js
```diff
@@ -514,7 +515,7 @@ async function checkVersion() {
 }
 
 async function handlePush(bag, opts) {
-  await checkVersion();
+  checkVersion();
```
stopachka: What do you guys think? cc @dwwoelfel @nezaj @drew-harris

drew-harris:
maybe checkVersion should block if the `--yes` flag is passed, since that seems the most likely scenario for a race

stopachka:
Good idea @drew-harris !

PR #1865, opened 2025-10-21
[cli] do not try to install or prompt in push and pull

stopachka:
When you run `init`, we don't _require_ that you log in. We detect if you aren't logged in, and log you in. Similarily, we don't require an app id. If you haven't provided one, we let you create or pick one. 

Previously, we used the same logic across `push` and `pull`. This doesn't quite make sense in those cases, because when you are running `push` or `pull`, you have an explicit intention about some particular app. It would be better if we threw here, since it would tell the user when they are making a mistake early. It would also make non-interactive CLI a bit easier to build. 

I went ahead and updated things, so only `init` prompts to install packages, log in, etc al. The rest will throw an error. 

### DIFF
--- a/client/packages/version/src/version.ts
+++ b/client/packages/version/src/version.ts
@@ -2,6 +2,6 @@
 // Update the version here and merge your code to main to
 // publish a new version of all of the packages to npm.
 
-const version = 'v0.22.28';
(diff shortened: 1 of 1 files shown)


stopachka:
good idea!

PR #1866, opened 2025-10-21
new app selector ui

drew-harris:
https://github.com/user-attachments/assets/47894f7a-e5f0-4342-b0d6-6c3e8bba20d3

https://github.com/user-attachments/assets/ca89c81a-7253-490d-b4a2-bf56e395c4fe

### DIFF
--- a/client/packages/version/src/version.ts
+++ b/client/packages/version/src/version.ts
@@ -2,6 +2,6 @@
 // Update the version here and merge your code to main to
 // publish a new version of all of the packages to npm.
 
-const version = 'v0.22.27';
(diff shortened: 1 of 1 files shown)


client/packages/cli/src/ui/index.ts
```diff
@@ -262,6 +263,9 @@ ${inputDisplay}`;
             });
           }
         }
+        if (keyInfo.name === 'tab') {
+          return;
+        }
```
drew-harris, line 268: fixes a bug where pressing tab in a text input messes up the terminal output

PR #1867, opened 2025-10-21
[Bug] Add cascade delete repro

nezaj:
What it says on the tin!

https://github.com/user-attachments/assets/4fd0e291-3f22-457b-946d-b37768fa5167


dwwoelfel (approved):
LGTM!

PR #1870, opened 2025-10-21
[cli] allow auth with admin tokens

stopachka:
Right now the only way to auth with CLI is to run `login`, or provide a personal access token. 

But sometimes, users already have an admin token provided. In that case, we can _just_ use the admin token to auth. 

This is helpful in two scenarios:

1. Now you can push and pull schema on ephemeral apps
2. For folks who are building app builders, they can use schema push and pull more easily. 

### DIFF
--- a/client/packages/version/src/version.ts
+++ b/client/packages/version/src/version.ts
@@ -2,6 +2,6 @@
 // Update the version here and merge your code to main to
 // publish a new version of all of the packages to npm.
 
-const version = 'v0.22.30';
+const version = 'v0.22.31';
 
 export { version };


client/packages/cli/src/index.js
```diff
@@ -1905,7 +1905,7 @@ async function readLocalSchemaFileWithErrorLogging() {
 
 async function readConfigAuthToken() {
   const options = program.opts();
-  if (options.token) {
+  if (typeof options.token === 'string') {
```
stopachka, line 1909: I did this as a little trick, to have typescript correctly infer the output of this function as `string | undefined`

server/src/instant/dash/routes.clj
```diff
@@ -95,11 +95,24 @@
...
                           :role least-privilege}))))
 
-(defn req->app-and-user-accepting-platform-tokens! [least-privilege scope req]
+(defn req->app-and-user-accepting-admin-or-platform! [least-privilege scope req]
   (let [token (http-util/req->bearer-token! req)]
     (if (token-util/is-platform-token? token)
       (req->superadmin-user-and-app! scope least-privilege req)
-      (req->app-and-user! least-privilege req))))
+      (let [app-id (ex/get-param! req [:params :app_id] uuid-util/coerce)]
+        (or (when-let [user (instant-user-model/get-by-refresh-token {:refresh-token token
```
stopachka: A part of me wished we could "know upfront" for admin tokens. 

I considered having the client hint it, but went against it for the following reasons: 
- The user _could_ provide a token with the `--token` flag. At that point, we would get an ambiguous input, and would have to do this anyways. 

Maybe in the future it would be really cool if we could have admin tokens as adm_xxx

server/src/instant/dash/routes.clj
```diff
@@ -95,11 +95,19 @@
                           :app-id app-id
                           :role least-privilege}))))
 
-(defn req->app-and-user-accepting-platform-tokens! [least-privilege scope req]
-  (let [token (http-util/req->bearer-token! req)]
-    (if (token-util/is-platform-token? token)
-      (req->superadmin-user-and-app! scope least-privilege req)
-      (req->app-and-user! least-privilege req))))
+(defn req->app-accepting-admin-or-platform! [least-privilege scope req]
+  (let [token (http-util/req->bearer-token! req)
```
dwwoelfel: One small suggestion:

If you use `req->superadmin-app!` from superadmin.routes:

```clojure
(try (req->superadmin-app! scope least-privilege req)
     (catch Exception _
         (when-let [user ...]
             ...)))
```

Then you'll also get free support for personal access tokens.

dwwoelfel (approved):
One suggestions, but LGTM!

server/src/instant/dash/routes.clj
stopachka: Great idea @dwwoelfel !

PR #1878, opened 2025-10-22
editable system namespaces

stopachka:
### [Try it out](https://www.dropbox.com/scl/fi/u1mogvo6l072fbnmcet0c/system-namespaces-feedback.paper?rlkey=wzixc82k2q72twy4025ua7ygz&dl=0)

This PR lets users create extend system namespaces. 

**What you can do**

1. Create custom column attributes: like `$users.fullName` or `$files.isFavorite` 
2. Create links where the forward identity is on the system namespace

This unblocks a bunch of cases where users would have had to create a separate namespace before. 

**General limits**
1. You can't mark a new system attribute as `isRequired`. This would make it impossible for the backend to create $users and $files. For now, we don't allow this.

**Update behavior** 
This opens up new behavior for `db.transact` on $users and $files objects. 
1. For both `$users` and `$files`, you can transact and update custom columns. 
2. For `$users`, admins can change reserved columns like `email`. _But_, normal users can not. 
3. For `$files`, _no one_ but the instant backend can change reserved columns, like `size` et al.

**Adding new system namespaces** 
You may wonder, how will we add new system namespaces moving forward? 
I propose a two step process. 
1. First, reserve it's idents, and deploy that. This will make it so users can't create that ident
2. Then, push an update actually creating it. We have a check that prevents creating system idents that aren't unique.

**Future considerations** 
1. We need to make it possible users to add initial data during creation. If we can do that, we can make it possible to create required columns
4. We need a way to hide specific columns with permissions. This will make it _much_ easier for users to share the `users` table.
5. We may want to "push" down the special handling we do for email and file updates, into permissions. 

**Related fixes**
1. I added the error UI for Indexing Jobs. If we get an `unexpected error`, we will render something: 
image
2. I refactored how we threw validation errors. We repeated the hint in both the `message` and the `hint.message` fields. This caused the UI to sometimes show duplicated text.

**Todos** 
- [ ] Update Server README to include instructions about adding new system catalog attributes.

CleanShot 2025-10-27 at 11 36 40@2x

### DIFF
--- a/client/www/components/dash/explorer/RecentlyDeletedAttrs.tsx
+++ b/client/www/components/dash/explorer/RecentlyDeletedAttrs.tsx
@@ -15,16 +15,23 @@ type SoftDeletedAttr = DBAttr & {
   'deletion-marked-at': string;
 };
 
+const deletedMarker = '_deleted$';
+const removeDeletedMarker = (s: string): string => {
+  const idx = s.indexOf(deletedMarker);
+  if (idx === -1) return s;
+  return s.slice(idx + deletedMarker.length);
+};
+
 const getNamesByNamespace = (
   softDeletedAttr: SoftDeletedAttr,
 ): Record<string, string> => {
   const result: Record<string, string> = {};
-  result[softDeletedAttr['forward-identity'][1].split('$')[1]] =
-    softDeletedAttr['forward-identity'][2].split('$')[1];
+  const [_, fwdEtype, fwdLabel] = softDeletedAttr['forward-identity'];
+  result[removeDeletedMarker(fwdEtype)] = removeDeletedMarker(fwdLabel);
 
   if (softDeletedAttr['reverse-identity']) {
-    result[softDeletedAttr['reverse-identity'][1].split('$')[1]] =
-      softDeletedAttr['reverse-identity'][2].split('$')[1];
+    const [_, revEtype, revLabel] = softDeletedAttr['reverse-identity'];
+    result[removeDeletedMarker(revEtype)] = removeDeletedMarker(revLabel);
   }
   return result;
 };


client/www/components/dash/explorer/EditNamespaceDialog.tsx
```diff
@@ -1717,9 +1710,9 @@ function EditAttrForm({
               the new name.
             </Content>
             <TextInput
-              disabled={isSystemCatalogNs}
+              disabled={isSystemCatalogAttr}
               title={
-                isSystemCatalogNs
+                isSystemCatalogAttr
                   ? `Attributes in the ${attr.namespace} namespace can't be edited.`
```
stopachka: TODO: make sure to change the error message

server/src/instant/db/transaction.clj
```diff
@@ -365,21 +356,30 @@
              missing
              [{:message (str "Updating entities that don't exist: " (string/join ", " (map :eid missing)))}])))))))
 
-(defn prevent-$files-updates
+(def editable-system-ident-names #{(list "$users" "id")
+                                   (list "$files" "id")
+                                   (list "$files" "path")})
+
+(defn prevent-system-attr-updates
```
stopachka: TODO I may have to move "move" up $users, and "move down" $files, similar to how it was.

client/sandbox/cli-nodejs/src/instant.perms.ts
```diff
@@ -0,0 +1,24 @@
+// Docs: https://www.instantdb.com/docs/permissions
+
+import type { InstantRules } from '@instantdb/core';
+
+const rules = {
```
stopachka: remove

client/sandbox/cli-nodejs/src/instant.schema.ts
```diff
@@ -0,0 +1,42 @@
+// Docs: https://www.instantdb.com/docs/modeling-data
+
+import { i } from '@instantdb/core';
```
stopachka: remove

server/src/instant/db/model/attr.clj
```diff
@@ -259,22 +261,46 @@
...
-   reserved namespaces.
-   We need this so that users don't clash with special namespaces, like the
-   $users table and $files rules."
+(defn validate-system-catalog-inserts!
+  "When we run a system catalog migration, we are inserting attrs that affect all 
+  existing apps. 
+  
+  What if a new insert has an ident that conflicts with some existing app's attr? 
+  
+  This prevents that from happening."
```
dwwoelfel: You might want to add a note here that it will kill the machine when it tries to deploy if there is a conflict.

server/src/instant/db/model/attr.clj
```diff
@@ -259,22 +261,46 @@
...
+  [conn attrs]
+  (let [idents  (mapcat ident-names attrs)
+        conflicts (sql/execute!
+                   conn
+                   (hsql/format
+                    {:with [[[:to-insert {:columns [:etype :label]}]
+                             {:values idents}]]
+                     :select :*
+                     :from [[:idents :i]]
```
dwwoelfel: I'd just look at attrs here instead of idents. Now that we're storing `etype` and `label` on the attr, I don't think there's any need for the idents. If we can remove all references in the code, maybe we can get rid of them.

server/src/instant/db/model/attr.clj
```diff
@@ -259,22 +261,46 @@
...
+                     :where [:in
+                             [:composite :i.etype :i.label]
+                             {:select [:etype :label]
+                              :from :to-insert}]}))]
+
+    (when-let [{:keys [etype label attr_id]} (first conflicts)]
```
dwwoelfel, line 297: Can we also dump a message into discord? I think we'll eventually end up with a failing deploy when someone tries to add a new system attr and we might not notice these errors.

server/src/instant/db/model/attr.clj
```diff
@@ -259,22 +261,46 @@
...
+
+(defn validate-system-ident-names!
+  "Prevents users from creating attrs that use catalog idents"
   [attrs]
   (doseq [attr attrs]
-    (when-let [fwd-etype (-> attr :forward-identity second)]
-      (when (string/starts-with? fwd-etype "$")
+    (let [fwd (vec (fwd-ident-name attr))
+          rev (vec (rev-ident-name attr))
+          found (some system-catalog/all-ident-names [fwd rev])]
```
dwwoelfel: If you used `contains?` instead of `some` here, then the reader wouldn't have to check that `all-ident-names` was a set to know if this is efficient.

server/src/instant/db/permissioned_transaction.clj
```diff
@@ -31,22 +31,78 @@
   (-> check
       (update :program :display-code)))
 
-(defn validate-reserved-names!
-  "Throws a validation error if the users tries to add triples to the $users table"
-  [{:keys [admin?]} tx-step-maps]
+(def editable-attrs #{["$users" "id"]
```
dwwoelfel: A docstring would be handy here.

server/src/instant/db/permissioned_transaction.clj
```diff
@@ -31,22 +31,78 @@
...
+
+   Here's what we do: 
+
+   1. There are _some_ system columns that we let users edit. 
+      
+      1. The $files 'path' attribute can be changed. 
+
+      2. The `id` attribute can flow through. 
+         
+         This is because the user may have their own columns defined, and
```
dwwoelfel, line 86: Are there existing users editing the ids? I'm a little confused about the use-case here.

server/src/instant/db/permissioned_transaction.clj
```diff
@@ -230,158 +286,158 @@
...
-                  :action   :delete
-                  :etype    "attrs"
-                  :program  {:result admin?}}]
-
-                (= :restore-attr op)
-                [{:scope    :attr
-                  :action   :restore
-                  :etype    "attrs"
-                  :program  {:result admin?}}]
+               (= :update-attr op)
```
dwwoelfel, line 287: Could you undo all of these whitespace changes?

server/src/instant/system_catalog.clj
```diff
@@ -331,6 +331,10 @@
               :unique? false
               :index? false
               :checked-data-type :string)])
+(make-attr "$users" "content-type"
```
dwwoelfel: What's this?

server/src/instant/system_catalog.clj
```diff
@@ -341,3 +345,18 @@
                        $oauth-code-attrs
                        $oauth-redirect-attrs
                        $files-attrs))
+
+(def reserved-ident-names
+  "Add ident names you'd like to add here. Once it's added, users can't create 
+   attrs with this idents. 
+   
+   After this is in production, you can then actually create the attr in question."
+  [])
```
dwwoelfel: An example would be handy here, so that we don't have to look up the structure of an `ident-name`.

server/src/instant/util/exception.clj
```diff
@@ -625,6 +625,10 @@
         (throw+ {::type ::validation-failed
                  ::message "There must be at least one member of the org that is an owner."})
 
+        "required_set_on_reserved_attrs"
+        (throw+ {::type ::validation-failed
+                 ::message "You can't create a required system attribute. Make sure it's optional."})
```
dwwoelfel: Maybe we could link to the docs? I'll bet some users don't know the syntax to make attrs optional.

server/src/instant/system_catalog.clj
dwwoelfel: It might be a good idea to make this pull from a feature flag instead. That way you wouldn't have to do multiple deploys to create a new system namespace.

server/src/instant/db/model/attr.clj
stopachka: Good point!

server/src/instant/db/model/attr.clj
stopachka: Good point! 

I went with: 

```
(sql/execute!
                   conn
                   (hsql/format
                    {:with [[[:to-insert {:columns [:etype :label]}]
                             {:values idents}]]
                     :select [[:i.etype :etype]
                              [:i.label :label]
                              [:a.id :attr-id]]
                     :from [[:to-insert :i]]
                     :join [[:attrs :a] [:or
                                         [:= [:composite :i.etype :i.label]
                                          [:composite :a.etype :a.label]]
                                         [:= [:composite :i.etype :i.label]
                                          [:composite :a.reverse-etype :a.reverse-label]]]]}))
```

This checks only on `attrs`. I had to make the where clause a bit more involved, to check for the reverse etype + label as well.

server/src/instant/db/model/attr.clj
stopachka: Great idea! 🫡

server/src/instant/db/model/attr.clj
stopachka: Fair! 

Changed made: 

```
         found (or (when (system-catalog/reserved-ident-name? fwd)
                            fwd)
                          (when (and rev (system-catalog/reserved-ident-name? rev))
                            rev))]

```

I unraveled the `some` here into an `or` check. This way it's clear that the list there isn't expected to be large. 

I made a special `system-catalog/reserved-ident-name?` function. This has two wins: 
1. It will let us use feature flags for reserved names (coming soon!)
2. Makes it irrelevant whether you pass in a vec or a seq. We'll do a lookup in system-catalog's inner function. This way callers won't have to worry about the efficiency.

server/src/instant/db/permissioned_transaction.clj
stopachka: Updated!

server/src/instant/db/permissioned_transaction.clj
stopachka: Updated the readme on this! 

The reason we allow the `id` triple, is because from the client we always send an 

`[:add-triple id :$users/id id]` when making updates 

Hence I let it pass through. Let me know if you think otherwise @dwwoelfel 🫡

server/src/instant/system_catalog.clj
stopachka: Oops! Was doing some repl testing! Will remove

server/src/instant/db/permissioned_transaction.clj
stopachka: Done!

server/src/instant/util/exception.clj
stopachka: Nice idea! Done

server/src/instant/system_catalog.clj
stopachka: Great idea! This way users can't guess what the name will be either > : )

stopachka: Done!

server/src/instant/db/transaction.clj
```diff
@@ -217,15 +217,6 @@
 (comment
   (gen/generate (s/gen ::tx-steps)))
 
-(defn prevent-system-catalog-attrs-updates! [op tx-steps]
```
dwwoelfel, line 220: We still need some form of this, or else users will be able to modify properties on the existing system attrs (e.g. changing the type).

stopachka: Hmm good point. I had the check `validate-system-ident-names!` in attr.clj. 

Two ideas: 

We could either: 

(a) run the `validate-system-ident-names` in `update-multi!` as well. 

(b) Move `validate-system-ident-names` up into a check here, so it never even gets to attr.clj 

I am leaning towards (a), so the check is as close to the underlying db as possible.

dwwoelfel: One alternate idea is to move the check to the database, similar to how we prevent deleting system attrs: https://github.com/instantdb/instant/blob/cfaa8a41a54cece3da48077ede176a784454313c/server/resources/migrations/34_hardcoded_objects.up.sql#L49

Then you never have to worry about it getting removed in a refactor.

server/resources/migrations/92_editable_attrs.up.sql
```diff
@@ -0,0 +1,78 @@
...
+  if disable_trigger then
+    return new;
+  end if;
+
+  -- Prevent updates for system catalog app
+  if old.app_id = 'a1111111-1111-1111-1111-111111111ca7' then
+    raise exception 'Updating attrs on the system catalog app is not allowed. Set the `instant.allow_system_catalog_app_attr_update` setting to true to override.';
+  end if;
+
+  return old;
```
dwwoelfel:
```suggestion
  return new;
```

server/resources/migrations/92_editable_attrs.up.sql
```diff
@@ -0,0 +1,78 @@
...
+  if disable_trigger then
+    return new;
+  end if;
+
+  -- Prevent updates for system catalog app
+  if old.app_id = 'a1111111-1111-1111-1111-111111111ca7' then
+    raise exception 'Updating idents on the system catalog app is not allowed. Set the `instant.allow_system_catalog_app_ident_update` setting to true to override.';
+  end if;
+
+  return old;
```
dwwoelfel:
```suggestion
  return new;
```

client/packages/platform/src/schema.ts
```diff
@@ -434,9 +434,41 @@ export class SchemaValidationError extends Error {
...
+  }
+
   for (const link of Object.values(schema.links)) {
+    if (
+      systemCatalogIdentNames[link.forward.on] &&
+      !systemCatalogIdentNames[link.forward.on].has(link.forward.label) &&
+      link.forward.required
+    ) {
+      throw new SchemaValidationError(
+        `The ${link.forward.on} namespace is managed by the system and can't require new attributes yet.` +
```
dwwoelfel:
```suggestion
        `The ${link.forward.on} namespace is managed by the system and can't modify required constraints yet.` +
```

It kind of seems like the error is saying that you can't add new attributes.

server/resources/migrations/92_editable_attrs.up.sql
```diff
@@ -0,0 +1,78 @@
...
+returns trigger as $$
+begin
+  if new.is_required
+    and starts_with(new.etype, '$')
+    and new.app_id <> 'a1111111-1111-1111-1111-111111111ca7' then
+        raise exception 'required_set_on_reserved_attrs';
+    end if;
+    return new;
+end;
+$$ language plpgsql;
```
dwwoelfel, line 13:
```suggestion
create or replace function prevent_required_on_reserved_attrs()
returns trigger as $$
begin
  if new.is_required
     and starts_with(new.etype, '$')
     and new.app_id <> 'a1111111-1111-1111-1111-111111111ca7'
  then
    raise exception 'required_set_on_reserved_attrs';
  end if;
  
  return new;
end;
$$ language plpgsql;
```

server/resources/migrations/92_editable_attrs.up.sql
```diff
@@ -0,0 +1,78 @@
...
+  exception when undefined_object then
+    disable_trigger := false;
+  end;
+
+  if disable_trigger then
+    return new;
+  end if;
+
+  -- Prevent updates for system catalog app
+  if old.app_id = 'a1111111-1111-1111-1111-111111111ca7' or new.app_id = 'a1111111-1111-1111-1111-111111111ca7' then
```
dwwoelfel, line 37: minor, but we could probably save some cycles if we moved the check for `disable_trigger` into this branch of the `if`.

stopachka: Good idea!

drew-harris:
Seems like we don't have soft-deletes on removing custom system namespaces, just wanna flag that if that's supposed to be implemented

stopachka:
Nice catch @drew-harris ! Fixed

PR #1881, opened 2025-10-23
Add security headers to api.instantdb.com

dwwoelfel:
This will get us an A on https://securityheaders.com/

The only place where this should affect anything are the OAuth landing pages and I tested that they all work normally. But just in case, I added a flag to disable them if they cause some unexpected problem.


nezaj (approved):
Nice!

PR #1884, opened 2025-10-24
Don't delete pending mutations that are in progress

dwwoelfel:
The PR that fixed flickers https://github.com/instantdb/instant/pull/1137 has some code that deletes pending mutations once they've been confirmed.

When it cleans up the mutations, it also runs this code:

```ts
      // backwards compat for mutations with no `confirmed`
      if (deleted && timeless) {
        for (const [eventId, mut] of Array.from(prev.entries())) {
          if (!mut.confirmed) {
            prev.delete(eventId);
          }
        }
      }
```

It's labeled as "backwards compat" and it has the unfortunate side effect of deleting pending mutations that the user is still waiting on. If you delete one mutation that we've acknowledged and has expired and you have another query that is still pending, then it will delete the pending mutation. I added a test to demonstrate.

Since it's backwards compat code from a PR from 6 months ago, I think we can just remove it.


stopachka (approved):
Woohoo!

PR #1889, opened 2025-10-28
[home] make it so you can't copy the $ from the npx create-instant-app command

stopachka:
A user gave me some feedback that if you copy the create-instant-app command manually, the $ will come with it. 

Made it so the $ is select-none. 

@nezaj @dwwoelfel @drew-harris


dwwoelfel (approved):
Nice!

PR #1890, opened 2025-10-28
[presence-fix][1] do not overwrite with initialData, if we _already_ have a result

stopachka:
Consider: 
1. joinRoom: initialPresence: "foo" 
2. subscribePresence: initialData: "bar" 

Previously, subscribePresence would 'reset' the initial presence in this room. 

**Solution**

I made it so when we consider setting initial data, we _only_ do it if we don't have data for this room. 

**Note**

I also added a little sandbox I am using to play with more changes to presence.


dwwoelfel (approved):
LGTM!

PR #1892, opened 2025-10-28
server routes for ssr

drew-harris:
adds a new route under /runtime/triples that allows getting triples and attrs for a query. Should have same permission level as add-query over the websocket

### DIFF
--- a/server/src/instant/runtime/routes.clj
+++ b/server/src/instant/runtime/routes.clj
@@ -606,16 +606,18 @@
 (defn query-triples [req]
   (let [query (ex/get-param! req [:body :query] #(when (map? %) %))
         auth-token (http-util/req->bearer-token req)
-        app-id  (req->app-id-untrusted! req)
-        user (when auth-token (app-user-model/get-by-refresh-token! {:refresh-token auth-token :app-id app-id}))
+        app-id (req->app-id-untrusted! req)
+        user (when auth-token
+               (app-user-model/get-by-refresh-token! {:refresh-token auth-token
+                                                      :app-id app-id}))
         attrs (attr-model/get-by-app-id app-id)
         ctx {:db {:conn-pool (aurora/conn-pool :read)}
-                    :app-id app-id
-                    :attrs attrs
-                    :datalog-query-fn d/query
-                    :datalog-loader (d/make-loader)
-                    :current-user user
-                    :versions (-> req :body :versions)}
+                  :app-id app-id
+                  :attrs attrs
+                  :datalog-query-fn d/query
+                  :datalog-loader (d/make-loader)
+                  :current-user user
+                  :versions (-> req :body :versions)}
         nodes (iq/permissioned-query ctx query)
         result (collect-instaql-results-for-client nodes)]
     (response/ok {:result result :attrs attrs})))
@@ -633,7 +635,7 @@
                                                   {:decoder parse-cookie}))
   (POST "/runtime/oauth/callback" [] (wrap-cookies oauth-callback
                                                    {:decoder parse-cookie}))
-  (POST "/runtime/triples" [] query-triples)
+  (POST "/runtime/query" [] query-triples)
 
   (POST "/runtime/oauth/token" [] oauth-token-callback)
   (POST "/runtime/:app_id/oauth/token" [] oauth-token-callback)


server/src/instant/runtime/routes.clj
```diff
@@ -594,6 +603,23 @@
...
                   (str config/server-origin "/runtime/" app-id "/oauth/token")})))
 
+(defn query-triples [req]
+  (let [query (ex/get-param! req [:body :query] #(when (map? %) %))
+        auth-token (http-util/req->bearer-token req)
+        app-id  (req->app-id-untrusted! req)
+        user (when auth-token (app-user-model/get-by-refresh-token! {:refresh-token auth-token :app-id app-id}))
+        attrs (attr-model/get-by-app-id app-id)
+        ctx {:db {:conn-pool (aurora/conn-pool :read)}
+                    :app-id app-id
```
dwwoelfel: Formatting is weird here. `:app-id` has an extra two spaces

server/src/instant/runtime/routes.clj
```diff
@@ -594,6 +603,23 @@
...
+                    :attrs attrs
+                    :datalog-query-fn d/query
+                    :datalog-loader (d/make-loader)
+                    :current-user user
+                    :versions (-> req :body :versions)}
+        nodes (iq/permissioned-query ctx query)
+        result (collect-instaql-results-for-client nodes)]
+    (response/ok {:result result :attrs attrs})))
```
dwwoelfel, line 623: Kind of a shame to send the full attrs each time. Not suggesting we do this now, but would it be possible to filter to just the attrs used in the triples, or do the attrs get put into the store?

server/src/instant/runtime/routes.clj
```diff
@@ -607,6 +633,7 @@
                                                   {:decoder parse-cookie}))
   (POST "/runtime/oauth/callback" [] (wrap-cookies oauth-callback
                                                    {:decoder parse-cookie}))
+  (POST "/runtime/triples" [] query-triples)
```
dwwoelfel: I would call this `/runtime/query` because that describes what it is doing, while `triples` describes what it returns.

In other places we allow you to return the result as a tree instead of triples, and we might want that here in the future.

server/src/instant/runtime/routes.clj
```diff
@@ -594,6 +603,23 @@
                   :token_endpoint
                   (str config/server-origin "/runtime/" app-id "/oauth/token")})))
 
+(defn query-triples [req]
+  (let [query (ex/get-param! req [:body :query] #(when (map? %) %))
+        auth-token (http-util/req->bearer-token req)
+        app-id  (req->app-id-untrusted! req)
```
dwwoelfel:
```suggestion
        app-id (req->app-id-untrusted! req)
```

server/src/instant/runtime/routes.clj
```diff
@@ -594,6 +603,23 @@
...
+        user (when auth-token (app-user-model/get-by-refresh-token! {:refresh-token auth-token :app-id app-id}))
```
dwwoelfel:
```suggestion
        user (when auth-token
               (app-user-model/get-by-refresh-token! {:app-id app-id
                                                      :refresh-token auth-token}))
```

Keeps line from getting too long.

server/src/instant/runtime/routes.clj
```diff
@@ -594,6 +603,23 @@
...
+        attrs (attr-model/get-by-app-id app-id)
+        ctx {:db {:conn-pool (aurora/conn-pool :read)}
+                    :app-id app-id
+                    :attrs attrs
+                    :datalog-query-fn d/query
+                    :datalog-loader (d/make-loader)
+                    :current-user user
+                    :versions (-> req :body :versions)}
```
dwwoelfel:
```suggestion
                  :app-id app-id
                  :attrs attrs
                  :datalog-query-fn d/query
                  :datalog-loader (d/make-loader)
                  :current-user user
                  :versions (-> req :body :versions)}
```

dwwoelfel (commented):
Code LGTM, left some formatting and naming comments.

server/src/instant/runtime/routes.clj
drew-harris: agreed that probably not best to send the full thing every time
the attrs get put into the store

PR #1895, opened 2025-10-29
[presence][4] do not send a notif in join-room-ok

stopachka:
Previously, when we got a `join-room-ok`, we would: 
1. Set room.isConnected to true 
2. Send an update

**But consider the following case**

- Room: me: { user1 } , others: { }

**User 2 joins the room**

User 2 will see: 

1. isLoading: true, me: { user1 } , others: { }
2. isLoading: false, me: { user1 } , others: { }
3. isLoading: false, me: { user1 } , others: { user2 }

**Here, step 2. is not fully correct**

Even though we got `join-room-ok`, we did not actually get a presence update yet. We shouldn't be triggering an update in this case. 

**The fix**
I no longer trigger presence updates on join-room-ok . Instead, we can rely on patch-presence and friends

**Note** 

There can be a race though, were we get a `patch-presence` _before_ we get a `join-room-ok`. To get around this, I make sure we always say we are connected, when we receive a `patch-presence`. 

**Future** 

Maybe extracting out the room logic into a `room actor` of sorts would help linearize the steps. 

### DIFF
--- a/client/packages/version/src/version.ts
+++ b/client/packages/version/src/version.ts
@@ -1,6 +1,6 @@
 // This is the shared version for all of the js packages
 // Update the version here and merge your code to main to
 // publish a new version of all of the packages to npm.
-const version = 'v0.22.38';
+const version = 'v0.22.39';
 
 export { version };


client/packages/core/src/Reactor.js
```diff
@@ -671,10 +675,16 @@ export default class Reactor {
...
 
-        joinedRoom.isConnected = true;
-        this._notifyPresenceSubs(loadingRoomId);
+        this._trySetRoomConnected(loadingRoomId, true);
         this._flushEnqueuedRoomData(loadingRoomId);
         break;
+      }
+      case 'leave-room-ok': {
+        const roomId = msg['room-id'];
+        debugger;
```
dwwoelfel: debugger

PR #1899, opened 2025-10-30
Add app examples page

nezaj:
Alrighty this is now ready for review!

[Preview](https://instant-www-js-showcase-jsv.vercel.app/examples)

This PR does the following:

* Adds a new page with examples for a todo, microblog, and chat app
* Each app has it's own detail page
* Content for an app is written in markdown
* Code referenced for each app is in a separate repo at https://github.com/instantdb/instant-examples

The content for each app varies

* The todo app has a video, getting started instructions, and a code walkthrough
* The microblog has getting started instructions and a code walkthrough
* The chat app only has getting started instructions

I recorded a video walking through the microblog code but I didn't think it was great. I also recorded about ~40 minutes making a chat app from scratch, but wasn't too happy with it either.

I included a little feedback box at the end of each guide, perhaps if folks want more videos and guides they can let us know and we can invest more!

One thing that would be nice is adding a "See it live" button. I wasn't sure though what would be an easy way to 1) make sure apps were not cross-contaminated (to avoid users potentially seeing profanity) 2) not need to maintain two code bases for each app (e.g. if we did something like ephemeral apps for the live version, it would be annoying to need to update both that and the example code we give users for getting started)

### DIFF
--- a/client/www/_examples/chat.md
+++ b/client/www/_examples/chat.md
@@ -21,17 +21,16 @@ If you haven't already, be sure to log into the Instant CLI
 pnpx instant-cli login
 ```
 
-Create a new app in the [Instant Dashboard](https://www.instantdb.com/dash)
-and copy the `INSTANT_APP_ID` and `INSTANT_APP_ADMIN_TOKEN` into your `.env` file.
+Now let's initialize a new app with the Instant CLI.
 
 ```bash
-# Copy .env.example to .env and update the variables
-# by creating a new app at https://www.instantdb.com/dash
-cp .env.example .env
+pnpx instant-cli init
 ```
 
-We've provided a schema in `instant.schema.ts` that you can push to your app via
-the CLI:
+We've provided a schema in `instant.schema.ts` that you can push to your app.
+You may have already pushed this during `init` in the previous step. If you
+answered 'no' to the prompt during init, or if you're unsure whether you pushed
+the schema, you can push it now.
 
 ```bash
 pnpx instant-cli push
--- a/client/www/_examples/microblog.md
+++ b/client/www/_examples/microblog.md
@@ -21,17 +21,16 @@ If you haven't already, be sure to log into the Instant CLI
 pnpx instant-cli login
 ```
 
-Create a new app in the [Instant Dashboard](https://www.instantdb.com/dash)
-and copy the `INSTANT_APP_ID` and `INSTANT_APP_ADMIN_TOKEN` into your `.env` file.
+Now let's initialize a new app with the Instant CLI.
 
 ```bash
-# Copy .env.example to .env and update the variables
-# by creating a new app at https://www.instantdb.com/dash
-cp .env.example .env
+pnpx instant-cli init
 ```
 
-We've provided a schema in `instant.schema.ts` that you can push to your app via
-the CLI:
+We've provided a schema in `instant.schema.ts` that you can push to your app.
+You may have already pushed this during `init` in the previous step. If you
+answered 'no' to the prompt during init, or if you're unsure whether you pushed
+the schema, you can push it now.
 
 ```bash
 pnpx instant-cli push
--- a/client/www/_examples/todos.md
+++ b/client/www/_examples/todos.md
@@ -22,17 +22,16 @@ If you haven't already, be sure to log into the Instant CLI
 pnpx instant-cli login
 ```
 
-Create a new app in the [Instant Dashboard](https://www.instantdb.com/dash)
-and copy the `INSTANT_APP_ID` and `INSTANT_APP_ADMIN_TOKEN` into your `.env` file.
+Now let's initialize a new app with the Instant CLI.
 
 ```bash
-# Copy .env.example to .env and update the variables
-# by creating a new app at https://www.instantdb.com/dash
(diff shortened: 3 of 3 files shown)


client/www/lib/examples/data.ts
```diff
@@ -0,0 +1,81 @@
...
+`.trim();
+
+const blogDescription = `
+A microblog app that allows users to create and like posts.
+
+By default anyone can view posts, but only authenticated users can create posts and like
+them. User profiles with unique usernames are also supported.
+
+We also demonstrate how to use the Admin SDK to both bootstrap and reset the
+database. jhis can be super useful for testing and development!
```
dwwoelfel:
```suggestion
database. This can be super useful for testing and development!
```

client/www/_examples/chat.md
```diff
@@ -0,0 +1,50 @@
...
+pnpm i
+```
+
+If you haven't already, be sure to log into the Instant CLI
+
+```bash
+pnpx instant-cli login
+```
+
+Create a new app in the [Instant Dashboard](https://www.instantdb.com/dash)
```
dwwoelfel: For this step, could we have them create the app from the CLI instead?

client/www/_examples/todos.md
```diff
@@ -0,0 +1,299 @@
...
+docs](/docs/presence-and-topics).
+
+## Testing offline mode
+
+Another great feature of Instant is that your app will continue to work even
+when offline! You can test this out like in Chrome devtools like so:
+
+- Open two tabs of the app
+- In one tab, open Chrome devtools, go to the network tab, and select "Offline"
+- om the throttling dropdown
```
dwwoelfel: Something got lost at the beginning of this sentence.

client/www/_examples/chat.md
```diff
@@ -0,0 +1,50 @@
...
+and copy the `INSTANT_APP_ID` and `INSTANT_API_KEY` into your `.env` file.
```
dwwoelfel: I'm not sure what the `INSTANT_API_KEY` is here? Is it the admin token?

client/www/_examples/microblog.md
```diff
@@ -0,0 +1,583 @@
...
+- A $user has one profile, and a profile belongs to one $user.
+- A profile has many posts, and a post belongs to one author (profile).
+- A profile has many likes, and a like belongs to one user (profile).
+- A post has many likes, and a like belongs to one post.
+
+We also define [`cascade` delete](/docs/modeling-data#cascade-delete) behavior for our links between users and their
+profiles, and then between profiles and posts/likes. This means that when a user
+is deleted, their profile, posts, and likes will also be automatically deleted.
+
+You may wonder **why** we have a separate `profiles` entity instead of just
```
dwwoelfel, line 170: Maybe we should update this with @stopachka's new `fields` permissions?

client/www/_examples/microblog.md
```diff
@@ -0,0 +1,583 @@
...
+
+In this app we'll leverage Instant's magic code auth to enable users to sign up
+and log in via email.
+
+We can detect whether a user is logged in by using the `db.useAuth` hook:
+
+<file label="src/app/page.tsx"></file>
+
+```tsx
+const { user, isLoading: userLoading, error: userError } = db.useAuth();
```
dwwoelfel, line 452: Could we use `db.SignedIn` and `db.SignedOut` here instead?

nezaj: I want to show the feed regardless of whether someone is signed in or not. If you're not signed in, then you can't do any writes. But if you are signed in, then the writes you do are associated with your user id.

I think we _must_ use `db.useAuth()` for this case. It feels like for cases where we want to enable some sort of "try before auth" you need to do `useAuth`

client/www/_examples/microblog.md
nezaj: Yea! Was thinking could give the fields a spin and update this!

client/www/_examples/chat.md
nezaj: I think we need to update `init` to not overwrite `instant.schema.ts` and `instant.perms.ts` if it already exists. Right now if you use the CLI it will overwrite those files

client/www/_examples/chat.md
nezaj: This was typo! Should be `INSTANT_APP_ADMIN_TOKEN`

client/www/_examples/todos.md
nezaj: Fixed!

client/www/_examples/chat.md
dwwoelfel: Could we also add instructions on where to find the admin token?

nezaj: Removed info about app_id and admin_token and instead updated to use `init` now that we won't overwrite per https://github.com/instantdb/instant/pull/1924

Should be much easier to follow!

client/www/_examples/microblog.md
nezaj: Will hold off on the fields pattern for now. Once we've done a bit more testing on it and updated our docs elsewhere will update this too!

PR #1901, opened 2025-10-30
Prevent double-submit on oauth start page

dwwoelfel:
Fixes a bug where you might click on the "Grant access" button twice and end up with a "oauth request not found" error because the first submit already claimed the oauth flow.

Now when you click it will put the button into a loading state, preventing multiple clicks.


stopachka (approved):
Yess!

PR #1904, opened 2025-10-31
Sync table

dwwoelfel:
Adds a new `_syncTableExperimental` function to core that takes a query and creates a new event-based subscription to the query.

It is similar to subscribeQuery with some key differences in implementation that make it suitable for syncing large amounts of data. 

1. Instead of sending the full dataset in one large batch, the data for the query is streamed in batches of 100 entities so that UI can be shown while the full data set is loading.
2. When the query is invalidated, we send an update with the added/removed triples instead of re-running the entire query
3. A small amount of state about the subscription is persisted in the database and we keep a log of transactions so that the query does not have to start from scratch on restarts. (the log is currently still a work-in-progress)

### How it works

You call `_syncTableExperimental` with a connection that has an admin token (using the hidden `init({__adminToken: 'token'})`. We require an admin token for now because we haven't implemented permissions yet.

We send a `:start-sync` event though the websocket, which creates a repeatable-read connection to get the current transaction id for the app and all of the matching entities. We use next.jdbc/plan so that we can stream the rows to the client as we fetch them (currently 100 at a time). 

As each batch arrives, the callback to `_syncTableExperimental` will get a new `SyncTableCallbackEventType.InitialSyncBatch` event that contains the entities in the batch and all of the data for the query so far. 

After all of the entities have been sent from the server, we'll call the callback with a `SyncTableCallbackEventType.InitialSyncComplete` event. If the developer wanted to wait for the load to complete before showing any data, then they would wait for this event before showing anything.

Once all of the rows have been sent, we record the tx-id in the reactive store and send a `:refresh-sync-table` event. The handler for that event will check if there are any new transactions for the subscription and send the matching triples for each transaction in a `:sync-update-triples` op. The invalidator will also send a `:refresh-sync-table` when a wal-record invalidates the topics for the subscription.

When the client gets a `:refresh-sync-triples`, it will add the data to the store for the subscription and figure out what was added/removed/updated. Right now we create a store per entity--we'd like a single store, but running the query becomes too slow with a single store. We might be able to mitigate this with some kind of entity cache in the store.

The callback to `_syncTableExperimental` will get a `SyncTableCallbackEventType.SyncTransaction` event for each transaction that contains a list of `added` entities, `removed` entities, and `updated` entities. The update contains the oldValue and newValue for the entity along with a map of old and new values for each field that changed.

If the client goes offline or loses its connection, it will send a `:resync-table` message with the `subscription-id` and `tx-id` that it wants to sync from. We'll look up the subscription from the database and ensure that it's from the same user (we also include a token for the subscription to be extra careful). Then we load that subscription into the store and send a `:refresh-sync-table` to catch up on any missed transactions.

The callback to `_syncTableExperimental` will get a `SyncTableCallbackEventType.LoadFromStorage` event when a subscription is loaded from the store.

All of the callback events contain a `data` field that has all of the data, which makes it easier for a developer to ignore the events and just render all of the data.

### Deployment plan
- [x] Merge PR
- [x] Run `create index concurrently transactions_app_id_id_idx on transactions (app_id, id desc);`
- [x] Run the migrations
- [x] Deploy PR to prod
- [x] Bump client version in a new PR
- [x] PR to remove the old app_id index on transactions

### Still to come
We'll use this to get feedback from users on the interface.

Before it's ready for a wide rollout, we'll need:
1. Transaction log (probably store the tx-data in s3) to replace our `store/sync-table-txes` placeholder
2. Optimistic updates on the client
3. Support for where clauses
4. Support for links (e.g `{:posts {:user {}}`)
5. Support for permissions
6. Change how we put things in the store (all subs in one key slows down when you have lots of subs)

### DIFF
--- a/server/src/instant/reactive/session.clj
+++ b/server/src/instant/reactive/session.clj
@@ -31,7 +31,6 @@
    [instant.reactive.receive-queue :as receive-queue]
    [instant.reactive.store :as rs]
    [instant.util.async :as ua]
-   [instant.util.coll :as ucoll]
    [instant.util.crypt :as crypt-util]
    [instant.util.delay :as delay]
    [instant.util.exception :as ex]


server/src/instant/reactive/topics.clj
```diff
@@ -0,0 +1,212 @@
+(ns instant.reactive.topics
```
dwwoelfel, line 1: This was just moved from the invalidator to avoid a cyclic dependency error.

stopachka (approved):
Beautiful work @dwwoelfel !

PR #1905, opened 2025-10-31
SSR

drew-harris:
The PR adds a utility function exported from core, that allows the user to sync auth state to a cookie that a server can read.

It also creates a `Framework` class in core that makes http requests to get triples and keeps a cache of results. It's designed to be framework agnostic, if we want to add ssr to other frameworks, other cool things.

I created a new init function / suspense provider under a /nextjs import in the react package. 

> [!IMPORTANT]
> I renamed a backend route from `/runtime/query` to `/runtime/framework/query` to make it more clear that the route is meant to only be used with the FrameworkClient class. It also changes so that if an incorrect refresh token is used, it will ignore the refresh token all together and query anonymously. 

Example Repo Readme:

Version to try: 0.22.92-experimental.drewh-ssr.20348370870.1 

# InstantDB Next.JS SSR Example

This repo demonstrates NextJS SSR and cookie sync. Built on top of `create-instant-app`'s NextJS todo example.

## Setup Cookie Sync (Optional)

1.  Create a file at `src/app/api/instant/route.ts` like so:

```typescript
import { createInstantRouteHandler } from "@instantdb/react/nextjs";

export const { POST } = createInstantRouteHandler({
  appId: process.env.NEXT_PUBLIC_INSTANT_APP_ID!,
});
```

2. Add the endpoint to your database config in `src/lib/db.ts`
```typescript
import { init } from "@instantdb/react/nextjs";
import schema from "../instant.schema";

export const db = init({
  appId: process.env.NEXT_PUBLIC_INSTANT_APP_ID!,
  firstPartyPath: "/api/instant",
  schema,
  useDateObjects: true,
});
```

### Create an InstantProvider

```typescript
"use client"

import { InstantSuspenseProvider } from "@instantdb/react/nextjs";
import { type User } from "@instantdb/react";
import { db } from "@/lib/db"

export default function InstantProvider({
  children,
  user,
}: {
  children: React.ReactNode;
  user: User;
}) {
  return (
    <InstantSuspenseProvider user={user} db={db}>
      {children}
    </InstantSuspenseProvider>
  );
}```

### Use InstantProvider in `layout` 

// src/app/layout.tsx 

```typescript
import InstantProvider from "./InstantProvider";
import { getUserOnServer } from "@instantdb/react/nextjs";

export default async function RootLayout({
  children,
}: Readonly<{
  children: React.ReactNode;
}>) {
  const user = await getUserOnServer(process.env.NEXT_PUBLIC_INSTANT_APP_ID!);
  return (
    <html lang="en">
      <body className="antialiased">
        <InstantProvider user={user}>{children}</InstantProvider>
      </body>
    </html>
  );
}

```
## Using SSR

Instead of using `db.useQuery` you can use `db.useSuspenseQuery` instead. 

If a page is rendering on the server, the server will make an HTTP request to Instant to retrieve the results before rendering. 

If the page is running on the client, it will re-use the existing webhook connection.

### SSR and Auth

The `db.useAuth` function works as normal, except that if you provide the user object to InstantSuspenseProvider higher in the component tree, it will skip a loading state.

This also works with <db.SignedIn> and <db.SignedOut> allowing you to skip any flashes of loading spinners or blank pages.

### DIFF
--- a/client/packages/version/src/version.ts
+++ b/client/packages/version/src/version.ts
@@ -2,6 +2,6 @@
 // Update the version here and merge your code to main to
 // publish a new version of all of the packages to npm.
 
-const version = 'v0.22.93';
+const version = 'v0.22.94';
 
 export { version };


client/sandbox/react-nextjs/tsconfig.json
```diff
@@ -8,8 +8,8 @@
     "forceConsistentCasingInFileNames": true,
     "noEmit": true,
     "esModuleInterop": true,
-    "module": "esnext",
-    "moduleResolution": "node",
+    "module": "NodeNext",
+    "moduleResolution": "nodenext",
```
drew-harris: allows importing from "sub exports", moduleResolution: "node" means compiles to support for node v10, node next raises the minimum version of node

client/packages/core/src/createRouteHandler.ts
```diff
@@ -0,0 +1,51 @@
+import { init } from './index.ts';
+
+export const createInstantRouteHandler = (config: {
+  appId: string;
+  apiURI?: string;
+}) => {
+  const admin = init({
```
dwwoelfel: Leftover code from before?

client/packages/core/src/createRouteHandler.ts
```diff
@@ -0,0 +1,51 @@
...
+    apiURI: config.apiURI,
+  });
+
+  async function handleUserSync(req: Request) {
+    const body = await req.json();
+    if (body.user && body.user.refresh_token) {
+      return new Response('sync', {
+        headers: {
+          // 24 hour expiry
+          'Set-Cookie': `instant_refresh_token=${body.user.refresh_token}; Path=/; HttpOnly; Secure; SameSite=Strict; Max-Age=86400`,
```
dwwoelfel: What happens if you're testing on localhost? I would expect the `Secure` field to prevent this from working.

client/packages/core/src/Reactor.js
```diff
@@ -275,7 +275,9 @@ export default class Reactor {
     this._oauthCallbackResponse = this._oauthLoginInit();
 
     // kick off a request to cache it
-    this.getCurrentUser();
+    this.getCurrentUser().then((userInfo) => {
+      this.syncUserToEndpoint(userInfo.user);
```
dwwoelfel, line 353: Since the cookie only lives for 24 hours, should we have something in reactor that periodically refreshes it?

client/packages/core/src/Reactor.js
```diff
@@ -1089,17 +1128,15 @@ export default class Reactor {
       return cached.data;
     }
 
-    const { store, pageInfo, aggregate, processedTxId } = result;
+    let { store, pageInfo, aggregate, processedTxId } = result;
```
dwwoelfel:
```suggestion
    let store = result.store
    const { pageInfo, aggregate, processedTxId } = result;
```

client/packages/core/src/Reactor.js
```diff
@@ -1817,7 +1854,27 @@ export default class Reactor {
     }
   }
 
+  async syncUserToEndpoint(user) {
+    if (this.config.endpointURI) {
```
dwwoelfel: endpointURI seems too generic

client/sandbox/react-nextjs/config.ts
```diff
@@ -1,7 +1,4 @@
-const isProd =
-  typeof window !== 'undefined'
-    ? Boolean(localStorage.getItem('prodBackend'))
-    : false;
+const isProd = true;
```
dwwoelfel: Debug code

client/www/pages/docs/backend.md
```diff
@@ -441,3 +441,122 @@ app.post('/custom_endpoint', async (req, res) => {
...
+  schema,
+  useDateObjects: true,
+});
+```
+
+## Server Side Rendering (Next.js)
+
+The `@instantdb/react` package exports an `InstantSuspenseProvider` that you can use to enable server rendering on client pages via Suspense.
+
+SSR is best used with the cookie sync endpoint. If cookies are not synced, there will a brief flash of the query result from an unauthentcated user on the first render.
```
dwwoelfel:
```suggestion
SSR is best used with the cookie sync endpoint. If cookies are not synced, there will a brief flash of the query result from an unauthenticated user on the first render.
```

client/packages/core/src/Reactor.js
drew-harris: I have it sync in the background when reactor gets init'd, not sure of a better solution? ideas welcome

client/packages/core/src/Reactor.js
drew-harris: i partially intentionally left it generic to account for more possible webhook actions in the future, also can't really think of a good name

client/packages/core/src/createRouteHandler.ts
drew-harris: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Set-Cookie#secure
On localhost addresses, browsers will still send secure cookies according to mdn. Maybe we want to give users more control over cookies anyways though

client/packages/core/__tests__/src/serializeSchema.test.ts
```diff
@@ -0,0 +1,123 @@
...
+type AnySchema = InstantSchemaDef<any, any, any>;
+
+// compare schemas by stringifying them with json and comparing the strings
+const compareSchemas = (schema1: AnySchema, schema2: AnySchema) => {
+  expect(JSON.stringify(schema1, null, 2)).toBe(
+    JSON.stringify(schema2, null, 2),
+  );
+};
+
+test('stuff', () => {
```
stopachka: nit: would add some quick desc

client/packages/core/src/Reactor.js
```diff
@@ -275,7 +275,17 @@ export default class Reactor {
...
+    this.getCurrentUser().then((userInfo) => {
+      this.syncUserToEndpoint(userInfo.user);
+    });
+
+    setInterval(
+      async () => {
+        const currentUser = await this.getCurrentUser();
+        this.syncUserToEndpoint(currentUser.user);
+      },
+      1000 * 60 * 60 * 23,
```
stopachka: nit: maybe should do this every ~20 minutes or something? Main reason: it's hard to know that the interval will actually work at this frequency level.

client/packages/core/src/Reactor.js
```diff
@@ -490,6 +500,43 @@ export default class Reactor {
     }
   }
 
+  /**
+   * Does the same thing as handle-query-add-ok
```
stopachka: nit: would say `add-query-ok` as I think that is the name

client/packages/core/src/Reactor.js
```diff
@@ -275,7 +275,17 @@ export default class Reactor {
     this._oauthCallbackResponse = this._oauthLoginInit();
 
     // kick off a request to cache it
-    this.getCurrentUser();
+    this.getCurrentUser().then((userInfo) => {
+      this.syncUserToEndpoint(userInfo.user);
+    });
+
+    setInterval(
```
stopachka: nit: perhaps we should keep track of the interval, in case we want to cancel it.

client/packages/core/src/Reactor.js
stopachka: Some ideas: 

How about `cookieEndpoint` or `cookieEndpointUI`?

client/packages/core/src/Reactor.js
```diff
@@ -1817,7 +1863,27 @@ export default class Reactor {
     }
   }
 
+  async syncUserToEndpoint(user) {
+    if (this.config.endpointURI) {
```
stopachka: Curious idea: 

Do you guys think there's a way to make it so the user doesn't have to provide this by default. 

Perhaps we could do an attempt on `/instant/sync-auth` for example, and if it's good, we continue on.

client/packages/core/src/createRouteHandler.ts
```diff
@@ -0,0 +1,44 @@
+export const createInstantRouteHandler = (config: {
+  appId: string;
+  apiURI?: string;
+}) => {
+  async function handleUserSync(req: Request) {
+    const body = await req.json();
+    if (body.user && body.user.refresh_token) {
+      return new Response('sync', {
+        headers: {
+          // 24 hour expiry
```
stopachka: imo 24 hours feels a bit too restrictive. Maybe we should do ~7 days? 

main reasoning: if a user comes back in a few days, it would suck if the backend is in a logged out state, while the frontend is in a logged in state.

client/packages/core/src/framework.ts
```diff
@@ -0,0 +1,281 @@
...
+    _query: any,
+    opts?: {
+      ruleParams: RuleParams;
+    },
+  ) => {
+    const { hash } = this.hashQuery(_query, opts);
+    return this.resultMap.get(hash);
+  };
+
+  public completeIsoMorphic = (
```
stopachka: nit: would call this `completeIsomorphic`

client/sandbox/react-nextjs/app/play/ssr/db.ts
```diff
@@ -0,0 +1,13 @@
...
+import { createUseSuspenseQuery } from '@instantdb/react/nextjs';
+
+export const db = init({
+  appId: process.env.NEXT_PUBLIC_INSTANT_APP_ID!,
+  endpointURI: 'https://my-awesome-site.com/api/instant',
+  schema,
+  useDateObjects: true,
+});
+
+export const useSuspenseQuery = createUseSuspenseQuery(db);
```
stopachka: I wonder if we should do something like, 

export an `init` from `@instantdb/react/nextjs`, which has `db.useSuspsenseQuery` 

Main reason: this keeps it consistent with all the other `use` hooks. 

I do think we are fighting against the current a bit with our attachment of hooks on `db`. Perhaps we should evolve to a place where we have hooks in a separate import. But at least for now it would be good if we were consistent.

client/www/pages/docs/backend.md
```diff
@@ -441,3 +441,122 @@ app.post('/custom_endpoint', async (req, res) => {
   // ...
 });
 ```
+
+### Syncing Auth
+
+Instant provides a `createInstantRouteHandler` function that generates a web standard endpoint that can be used to sync the refresh token to a cookie that your server can read.
```
stopachka, line 449: I would start this doc by motivating the problem: 

(Roughly) : 

Sometimes you want to get the logged in user in the backend.  Instant can automatically sync the logged in user for you. Here's how to do it.

client/packages/core/src/Reactor.js
drew-harris: the return type of setInterval can be different depending on if its in node vs the browser. I don't want to leave a hanging variable that makes the types kinda weird unless we're using it.

client/packages/core/src/Reactor.js
drew-harris: I think that we should focus on making the docs clear enough that you need to implement the endpoint *and* add it to the config object. I worry that if we automatically fetch a predefined route the 404 errors in the developer console may confuse people

client/packages/core/src/Reactor.js
stopachka: SG!

client/packages/core/src/Reactor.js
stopachka: Hmm yeah if there is a 404 warning in console would not be good. If we used a HEAD request would it still log it as an issue?

drew-harris: good question ill see

client/sandbox/react-nextjs/app/play/ssr/db.ts
drew-harris: done!

client/packages/core/src/Reactor.js
drew-harris: you still get errors

stopachka: SG!

client/packages/core/src/Reactor.js
```diff
@@ -1817,7 +1863,27 @@ export default class Reactor {
     }
   }
 
+  async syncUserToEndpoint(user) {
+    if (this.config.cookieEndpoint) {
+      try {
+        fetch(this.config.cookieEndpoint + '/sync-auth', {
```
stopachka: nit: if this is called cookieEndpoint, imo we should be probbly include the `sync-auth` in it. 

Reasoning: a user _may_ want to have a specific name for their endpoint.

stopachka (commented):
This is looking good to me!

client/packages/core/src/Reactor.js
drew-harris: It works this way because the cookie sync handler matches on all the routes beneath it in the tree so, the 'sync-auth' isn't seen by the user/in their code at all (except for in the network) tab and allows us to do more webhook stuff in the future without requiring people to make a whole new mounted route handler.

client/packages/react/src/next-ssr/index.tsx
```diff
@@ -0,0 +1,257 @@
...
+          schema: parseSchemaFromJSON(JSON.parse(props.config!.schema)),
+        }),
+  );
+
+  const [trackedKeys] = useState(() => new Set<string>());
+
+  if (!clientRef.current) {
+    if (props.user && !props.user.refresh_token) {
+      throw new Error(
+        'User must have a refresh_token field. Recieved: ' + props.user,
```
nezaj: Nit. I think this should be JSON.stringify(props.user)

client/www/pages/docs/backend.md
```diff
@@ -441,3 +441,117 @@ app.post('/custom_endpoint', async (req, res) => {
...
+## Server Side Rendering (Next.js)
+
+The `@instantdb/react` package exports an `InstantSuspenseProvider` that you can use to enable server rendering on client pages via Suspense.
+
+SSR is best used with the [cookie sync endpoint](/docs/backend#syncing-auth). If cookies are not synced, there will a brief flash of the query result from an unauthenticated user on the first render.
+
+{% callout %}
+To use the suspense query hook and SSR, make sure to update your import for the database object:
+
+`import { init } from '@instantdb/react/nextjs'`
```
nezaj: This callout is a bit confusing to me. In the other examples in this doc we never show importing from nextjs package. Feels like instead of a callout we should have a full example?

client/packages/react/src/next-ssr/index.tsx
```diff
@@ -0,0 +1,257 @@
...
+    if (!entry) {
+      entry = clientRef.current!.query(query, opts);
+    }
+
+    if (entry.status === 'pending') {
+      throw entry.promise;
+    }
+
+    if (entry.status === 'error') {
+      console.log(entry.error);
```
nezaj: Nit. Should we remove this?

client/www/pages/docs/backend.md
drew-harris: Opting to remove mentions of the nextjs ssr from the docs and instead publish an example repo that we can direct people to.

client/packages/react-common/src/useQuery.ts
```diff
@@ -63,6 +63,9 @@ export function useQueryInternal<
       resultCacheRef.current = stateForResult(
         _core._reactor.getPreviousResult(query),
       );
+      if (resultCacheRef.current.data) {
```
nezaj: Curious, why did we need to add this in?

client/packages/react-common/src/InstantReactAbstractDatabase.tsx
```diff
@@ -241,6 +241,10 @@ export default abstract class InstantReactAbstractDatabase<
    *
    */
   useAuth = (): AuthState => {
+    return this._useAuth();
+  };
+
+  protected _useAuth(): AuthState {
```
nezaj, line 247: Also curious why this change?

client/packages/react/src/next-ssr/index.tsx
```diff
@@ -0,0 +1,273 @@
...
+};
+
+/**
+ *
+ * The first step: init your application!
+ *
+ * Visit https://instantdb.com/dash to get your `appId` :)
+ *
+ * @example
+ *  import { init } from "@instantdb/react"
```
nezaj, line 25:
```suggestion
 *  import { init } from "@instantdb/react/nextjs"
```

client/packages/react/src/next-ssr/index.tsx
```diff
@@ -0,0 +1,273 @@
...
+ *
+ *  const db = init({ appId: "my-app-id" })
+ *
+ *  // You can also provide a schema for type safety and editor autocomplete!
+ *
+ *  import { init } from "@instantdb/react"
```
nezaj, line 31:
```suggestion
 *  import { init } from "@instantdb/react/nextjs"
```

client/www/pages/docs/backend.md
```diff
@@ -57,7 +57,7 @@ query once and returns a result.
 
 ### transact
 
-```javascript
+```javascript {% showCopy=true %}
```
nezaj:
```suggestion
```typescript {% showCopy=true %}
```

client/www/pages/docs/backend.md
```diff
@@ -441,3 +441,33 @@ app.post('/custom_endpoint', async (req, res) => {
   // ...
 });
 ```
+
+### Syncing Auth
+
+Sometimes you want to get the logged in user in the backend. Instant can automatically sync the logged in user for you. Here's how to do it.
```
nezaj:
```suggestion
Sometimes you want to get the logged in user in the backend without needing to explicitly pass in the refresh token from the frontend. Here's how you can do it.
```

client/www/pages/docs/backend.md
```diff
@@ -441,3 +441,33 @@ app.post('/custom_endpoint', async (req, res) => {
...
+  appId: process.env.NEXT_PUBLIC_INSTANT_APP_ID!,
+});
+```
+
+The GET and POST functions accept a [Request](https://developer.mozilla.org/en-US/docs/Web/API/Request) and return a [Response](https://developer.mozilla.org/en-US/docs/Web/API/Request) so they should be able to be used in any framework.
+
+Then, provide your mounted api url to the `init` function.
+
+```typescript {% showCopy=true %}
+export const db = init({
```
nezaj, line 469:
```suggestion
import { init } from "@instantdb/react/nextjs";
import schema from "../instant.schema";

export const db = init({
```

client/www/pages/docs/backend.md
```diff
@@ -441,3 +441,33 @@ app.post('/custom_endpoint', async (req, res) => {
   // ...
 });
 ```
+
+### Syncing Auth
```
nezaj, line 445: Should we add another section about SSR and include a link to the demo Repo?

drew-harris: I'll add a section just so that we can have a nice link to the docs. Just a quick blurb and a link to the github repo.

client/packages/react-common/src/InstantReactAbstractDatabase.tsx
drew-harris: We make a subclass of InstantReactAbstractDatabase for ssr that overrides the method and this was the cleanest way to do the inheritance.

client/packages/react-common/src/useQuery.ts
drew-harris: removed this check, there was some point in time when it was necessary for suspense stuff, but not any longer.

nezaj (approved):
Awesome push!

client/packages/core/__tests__/src/serializeSchema.test.ts
```diff
@@ -0,0 +1,123 @@
...
+      },
+    },
+  },
+});
+
+type AnySchema = InstantSchemaDef<any, any, any>;
+
+// compare schemas by stringifying them with json and comparing the strings
+const compareSchemas = (schema1: AnySchema, schema2: AnySchema) => {
+  expect(JSON.stringify(schema1, null, 2)).toBe(
```
stopachka, line 101: Curious question: is this meant to be strictly string equaling? i.e it matters that the keys in objects are the same order?

client/packages/core/src/createRouteHandler.ts
```diff
@@ -0,0 +1,56 @@
+import type { User } from './clientTypes.js';
+
+export const createInstantRouteHandler = (config: { appId: string }) => {
+  function createUserSyncResponse(user: User | null) {
```
stopachka: Nit: I would move these functions outside of this, and have them accept `config` + `user`. Then would curry inside here. 

Reasoning: would reduce the amount of possible shared state between the functions.

client/packages/core/src/framework.ts
```diff
@@ -0,0 +1,295 @@
...
+      triples: any;
+      attrs: any;
+      queryHash: string;
+      pageInfo?: any;
+    }) => void,
+  ) => {
+    this.queryResolvedCallbacks.push(callback);
+  };
+
+  // Runs on the client when ssr gets html script tags
```
stopachka, line 89: I think we should add a lot more docstrings to this particular file. 

Reason: it looks like this is the main thing that goes through the SSR lifecycle. As time goes on we may forget the intricacies of how this all works. 

Normally I am not a fan of large docstrics beside every function. But this particular thing is hard, so imo it makes sense to add lots of extra context here.

client/packages/core/src/Reactor.js
```diff
@@ -348,7 +348,17 @@ export default class Reactor {
...
+    this.getCurrentUser().then((userInfo) => {
+      this.syncUserToEndpoint(userInfo.user);
+    });
+
+    setInterval(
+      async () => {
+        const currentUser = await this.getCurrentUser();
+        this.syncUserToEndpoint(currentUser.user);
+      },
+      1000 * 60 * 20,
```
stopachka: - I realize we may want to cancel this interval, in the case that someone calls `destroy` on this Reactor
- imo we should be running this interval like every minute. Running once every 20 minutes imo _could_ work, but imo it will open us up for hard-to-discover bugs. If we instead push every 1 or 2 minutes, whatever bug we have will be easier to discover. 

cc @dwwoelfel if you have differing thoughts

client/packages/core/src/Reactor.js
```diff
@@ -1989,7 +2047,29 @@ export default class Reactor {
     }
   }
 
+  async syncUserToEndpoint(user) {
+    if (this.config.firstPartyPath) {
```
stopachka: nit: would do if (!this.config.firstPartyPath) return 

Reasoning: preference for guards to reduce nesting.

client/packages/core/src/Reactor.js
```diff
@@ -1989,7 +2047,29 @@ export default class Reactor {
...
+            type: 'sync-user',
+            appId: this.config.appId,
+            user: user,
+          }),
+          headers: {
+            'Content-Type': 'application/json',
+          },
+        });
+      } catch (error) {
+        console.error('Error syncing user with external endpoint', error);
```
stopachka: nit: would use log.error

client/packages/react-common/src/useQuery.ts
```diff
@@ -63,6 +63,7 @@ export function useQueryInternal<
       resultCacheRef.current = stateForResult(
         _core._reactor.getPreviousResult(query),
       );
+      cb();
```
stopachka, line 67: Nit: we may want to add a note about why we are doing this.

client/packages/react/src/next-ssr/HydrationStreamProvider.tsx
```diff
@@ -0,0 +1,193 @@
+'use client';
```
stopachka, line 1: It may make sense to add a top-level comment, (a) describing what a HydrationStream is, (b) describing what this is meant to do.

client/packages/react/src/next-ssr/InstantNextDatabase.tsx
```diff
@@ -0,0 +1,57 @@
+'use client';
```
stopachka: Why is the 'use client' necessary here?

client/packages/react/src/next-ssr/InstantSuspenseProvider.tsx
```diff
@@ -0,0 +1,187 @@
+'use client';
```
stopachka, line 1: Nit: I would see if we could remove this, or explain why it is needed.

client/packages/react/tsconfig.json
```diff
@@ -3,6 +3,7 @@
   "include": ["src"],
   "exclude": ["node_modules", "dist", "src/__types__"],
   "compilerOptions": {
+    "skipLibCheck": true,
```
stopachka, line 6: Curious q: why did we add this?

client/www/pages/docs/backend.md
```diff
@@ -441,3 +441,61 @@ app.post('/custom_endpoint', async (req, res) => {
...
+
+Sometimes you want to get the logged in user in the backend without needing to explicitly pass in the refresh token from the frontend. Here's how you can do it.
+
+Instant provides a `createInstantRouteHandler` function that generates a web standard endpoint that can be used to sync the refresh token to a cookie that your server can read.
+
+To use it in NextJS:
+
+```typescript {% showCopy=true %}
+// src/app/api/instant/[...all]/route.ts
```
stopachka: nit: needs to be updated

client/packages/react/src/next-ssr/getUserOnServer.ts
```diff
@@ -0,0 +1,8 @@
+import { cookies } from 'next/headers.js';
+
+export const getUserOnServer = async (appId: string) => {
```
stopachka: How about we call this `getUserFromInstantCookie`? 

Reasoning: 

`getUserOnServer` feels a bit awkward because: (a) it doesn't specify what kind of user this is. (b) the server part is understandably important, but it's unclear why it has to be on the server.

stopachka: Nit: would type this

drew-harris: It has to be "on the server" because it calls nextjs's `cookies()` method internally. Maybe we could require passing in the return of that function?

client/www/pages/docs/backend.md
```diff
@@ -441,3 +441,61 @@ app.post('/custom_endpoint', async (req, res) => {
...
+        <AuthProvider user={user}>{children}</AuthProvider>
+      </body>
+    </html>
+  );
+}
+```
+
+## NextJS SSR
+
+Instant has built in support for SSR using NextJS. See [this example repo](https://github.com/instantdb/instant-ssr-demo) for more information.
```
nezaj, line 503: Would be good to do a pass in https://github.com/instantdb/instant-ssr-demo to make sure it's in sync with this. For example it currently has the following snippet

```
<InstantSuspenseProvider
  user={user}
  config={db} // this should say db={db}
>
```
 
Looks like we've also changed the convention for the route names, how you get the user `getUser...` so we should be sure to update that as well.

Would also be good to include step-by-step instructions for what someone needs to do to get SSR working (as an idea https://github.com/nezaj/instant-blog-ssr-example?tab=readme-ov-file#making-an-instant-app-ssr-compatible)

client/packages/react/src/next-ssr/getUserOnServer.ts
stopachka: I think the way it's written was good. My main idea here was: if we say `getUserFromInstantCookie`, it 'explains' why this has to be run on the server, without explicitly saying `server` in the name

client/packages/core/__tests__/src/serializeSchema.test.ts
drew-harris: It doesn't have to be strictly ordered, this was just the easiest way to verify that all the data was getting rebuilt correctly. 
We aren't using this code anymore, since I removed the `config` prop. ,, however doing things like sending schemas over json could be useful to things like app builders?

client/packages/react/src/next-ssr/InstantSuspenseProvider.tsx
drew-harris: InstantSuspenseProvider can only be used in a client context so this prevents errors from trying to use it in a server component.

client/packages/core/__tests__/src/serializeSchema.test.ts
stopachka: Yes def! Happy to keep the test if you think it's good

client/packages/core/src/Reactor.js
```diff
@@ -348,7 +348,14 @@ export default class Reactor {
...
     // kick off a request to cache it
-    this.getCurrentUser();
+    this.getCurrentUser().then((userInfo) => {
+      this.syncUserToEndpoint(userInfo.user);
+    });
+
+    setInterval(async () => {
+      const currentUser = await this.getCurrentUser();
+      this.syncUserToEndpoint(currentUser.user);
+    }, 1000 * 60);
```
stopachka: nit: for easy reading would just make this a variable call `oneMinMs` or something

client/packages/core/src/createRouteHandler.ts
```diff
@@ -0,0 +1,63 @@
+import type { User } from './clientTypes.js';
+
+type CreateRouteHandlerConfig = {
+  appId: string;
+};
+
+function createUserSyncResponse(
+  user: User | null,
+  config: CreateRouteHandlerConfig,
```
stopachka, line 8: nit: would move `config` as the first part in the arg. 

Reasoning: 

It's usually nicer to keep the "thing that changes less frequently" higher up in argument lists, to make currying easier.

client/packages/core/src/framework.ts
```diff
@@ -0,0 +1,315 @@
...
+import { createLinkIndex } from './utils/linkIndex.ts';
+
+export const isServer = typeof window === 'undefined' || 'Deno' in globalThis;
+
+export type FrameworkConfig = {
+  token?: string | null;
+  db: InstantCoreDatabase<any, any>;
+};
+
+// represents an eventual result from running a query
```
stopachka, line 22: Nice job with the comments! Spare, but just enough to understand. 

I would add 1 more "high-level comment" somewhere, explaining the chain that would go in a basic "flow" of SSR. 

To give a sense of what I mean: 

https://github.com/jsventures/instant/blob/main/server/src/instant/db/datalog.clj#L2-L29

For some of the "harder" namespaces, it can be good to add a preamble.

client/packages/react/src/next-ssr/HydrationStreamProvider.tsx
```diff
@@ -0,0 +1,198 @@
...
+
+// This allows us to send serialized json from the server to the client using 'useServerInsertedHTML'
```
stopachka, line 3: Maybe one thing that would be nice in this comment: 

Would start with "what is the problem". i.e why do we need to send serialized json from the server to the client? 

What is useServerInsertedHTML?

client/packages/react/src/next-ssr/HydrationStreamProvider.tsx
```diff
@@ -0,0 +1,198 @@
...
+  nonce?: string;
+}
+
+export function createHydrationStreamProvider<TShape>() {
+  const context = React.createContext<HydrationStreamContext<TShape>>(
+    null as any,
+  );
+  /**
+
+   * 1. (Happens on server): `useServerInsertedHTML()` is called **on the server** whenever a `Suspense`-boundary completes
```
stopachka, line 73: Niice!

client/packages/react/src/next-ssr/HydrationStreamProvider.tsx
```diff
@@ -0,0 +1,198 @@
...
+   *    - This means that we might have some new entries in the cache that needs to be flushed
+   *    - We pass these to the client by inserting a `<script>`-tag where we do `window[id].push(serializedVersionOfCache)`
+   * 2. (Happens in browser) In `useEffect()`:
+   *   - We check if `window[id]` is set to an array and call `push()` on all the entries which will call `onEntries()` with the new entries
```
stopachka, line 77: A bit lost with this note. Would get a bit more high level here, and say what the goal is. (I can jump into the code to see the more specific `push` `onEntries` etc.

client/packages/react/src/next-ssr/getUserFromInstantCookie.ts
```diff
@@ -0,0 +1,11 @@
+import { User } from '@instantdb/core';
```
stopachka, line 1: nit: file in this case should be called `getUserFromInstantCookie`

client/www/pages/docs/backend.md
```diff
@@ -441,3 +441,61 @@ app.post('/custom_endpoint', async (req, res) => {
...
+  appId: process.env.NEXT_PUBLIC_INSTANT_APP_ID!,
+});
+```
+
+The GET and POST functions accept a [Request](https://developer.mozilla.org/en-US/docs/Web/API/Request) and return a [Response](https://developer.mozilla.org/en-US/docs/Web/API/Request) so they should be able to be used in any framework.
+
+Then, provide your mounted api url to the `init` function.
+
+```typescript {% showCopy=true %}
+export const db = init({
```
stopachka, line 469: nit: would include the import of `init` here

stopachka (approved):
Ship it sir!!!

PR #1908, opened 2025-10-31
useAuth returns the $users type from schema

drew-harris:
useAuth, useUser, subscribeAuth, getAuth

### DIFF
--- a/client/packages/version/src/version.ts
+++ b/client/packages/version/src/version.ts
@@ -1,6 +1,6 @@
 // This is the shared version for all of the js packages
 // Update the version here and merge your code to main to
 // publish a new version of all of the packages to npm.
-const version = 'v0.22.45';
+const version = 'v0.22.46';
 
 export { version };


client/packages/core/src/clientTypes.ts
```diff
@@ -7,14 +10,29 @@ export type User = {
   isGuest: boolean;
 };
 
-export type AuthResult =
-  | { user: User | undefined; error: undefined }
+export type UserWithSchema<
+  S extends InstantSchemaDef<any, any, any>,
+  UseDates extends boolean,
+> = Expand<ResolveAttrs<S['entities'], '$users', UseDates> & User>;
```
stopachka, line 16: Cool! To make sure, this works even if we (a) don't provide a schema, or (b) we provide a schema without a $users inside the def?

stopachka (approved):
LGTM!

client/packages/core/src/clientTypes.ts
drew-harris: image
It will autocomplete with the standard fields but allow any extra fields too.  Does that work?

stopachka: SGTM!

PR #1909, opened 2025-10-31
cmd/ctrl-s to save sandbox

drew-harris:


client/www/components/dash/Sandbox.tsx
```diff
@@ -87,6 +87,8 @@ export function Sandbox({
   const consoleRef = useRef<HTMLDivElement>(null);
 
   const [selectedSandbox, setSelectedSandbox] = useState<string | null>(null);
+  const selectedSandboxRef = useRef<string | null>(null);
+  const saveDialogRef = useRef<{ onOpen: () => void } | null>(null);
```
drew-harris, line 91: the refs are needed due to the closure for the keybindings :(

PR #1910, opened 2025-11-03
client side limit filter w/ warning

drew-harris:
Allows "limit" in subqueries, but uses `console.warn` to output:
"WARNING: Limits in child queries are only run client-side. Data returned from the server will not have a limit."

Sandbox: 
http://localhost:4000/play/nested-limit

### DIFF
--- a/client/packages/version/src/version.ts
+++ b/client/packages/version/src/version.ts
@@ -1,6 +1,6 @@
 // This is the shared version for all of the js packages
 // Update the version here and merge your code to main to
 // publish a new version of all of the packages to npm.
-const version = 'v0.22.46';
+const version = 'v0.22.47';
 
 export { version };


server/test/instant/db/instaql_test.clj
```diff
@@ -267,11 +267,12 @@
...
-                   "We currently only support `limit`, `offset`, `before`, and `after` clauses on the top-level field.",
-                   :in [:users :bookshelves :$],
-                   :expected supported-options?}
-                 (validation-err ctx {:users {:bookshelves {:$ {:limit 10}}}})))
+          ;; We don't support nested limits on the server, they are ignored so the client can limit manually
+          ;; (is (= '{:message
+          ;;          "We currently only support `limit`, `offset`, `before`, and `after` clauses on the top-level field.",
+          ;;          :in [:users :bookshelves :$],
+          ;;          :expected supported-options?}
+          ;;        (validation-err ctx {:users {:bookshelves {:$ {:limit 10}}}})))
```
drew-harris, line 275: should I just delete it?

stopachka: I would add a test like: 

- Make a query that users an internal limit
- Then assert that the query _does_ not actually use the limit and return all items 

This test makes sure that (a) the limit does not throw an error. (b) It makes it clear that this is intentionally not supported.

client/packages/core/src/instaql.js
```diff
@@ -690,6 +690,12 @@ function resolveObjects(store, { etype, level, form, join, pageInfo }) {
 
   const limit = form.$?.limit || form.$?.first || form.$?.last;
   if (limit != null) {
+    if (level > 0) {
+      console.warn(
+        'WARNING: Limits in child queries are only run client-side. Data returned from the server will not have a limit.',
```
stopachka, line 695: I go back and forth on the warning. I guess let's just start with a warning, as it is the most explicit.

client/packages/core/src/queryValidation.ts
```diff
@@ -379,7 +379,7 @@ const validateDollarObject = (
 
   // Validate that pagination parameters are only used at top-level
   const paginationParams = [
-    'limit',
+    // 'limit', // only supported client side
     'offset',
```
stopachka, line 383: should `offset` work too?

drew-harris: looks to me like offset is implemented in a way that only works server side (I couldn't find the logic for it anywhere in instalql like i did for limit. 
I think if we allow both limit and offset, we are encouraging people to be far too wasteful with the data since each change in pagination will send the whole dataset * amount of pages.

PR #1916, opened 2025-11-04
field-level permissions

stopachka:
## [Try it out](https://www.dropbox.com/scl/fi/o48hr2jnsnyyvz13cchsk/field-perms-dogfooding.paper?rlkey=agld3j3fnhd6h34bahvkfm5la&dl=0)

This PR allows users to set field level permissions for queries. 

For example. Given a database with two users: 

1. alyssa@instantdb.com
2. louis@instantdb.com

With the following rules: 

```js
{
   $users: { 
      allow: { 
         view: "true"
      },
      fields: { 
         email: "auth.id == data.id"
      }
  }
}
```

If an anonymous user were to make a query, they would get _both_ user objects back, but the email will be hidden. 

If a logged in user made this query, they would get both user objects back, but their _own_ user would include the `email`. 

**Notes** 

1. Field-level permissions are meant to act as filters. They won't work with rule wheres

**Quick win follow ons**

I tried to keep this PR focused on this specific problem. I noticed two quick wins that I'll ship as follow-ons though:

1. JSON schema in Sandbox. Right now the sandbox does not use the JSON schema that we have in Perms.tsx. We also don't use json-5, so saving the file is annoying. I will update it so (a) we use the JSON schema, and (b) we are more forgiving with saves
2. Improved view rules UI. This PR _just_ added a simple "view" "label" box for our permission check. But we may want to start grouping rules by the id. I am going to do a separate spike on improving the UI. 

**Harder follow-ons**

I noticed one gotcha, if a user can see an object, but not a field, and uses the field in a where clause. We may still treat that path as valid, when we should actually prune. [[details]](https://discord.com/channels/1031957483243188235/1196584090552512592/1436110071263264798)

One fix could be to "keep track" of which join row causes which child result. If we then filter join-rows, we can quickly filter the corresponding child result. I think we'll have to make the change in nested-sql-result->result in datalog.clj, to keep some extra info and make this feasible. After some playing, I couldn't come up with an elegant solution. I'll keep this on the backburner for later.

### DIFF
--- a/client/packages/version/src/version.ts
+++ b/client/packages/version/src/version.ts
@@ -1,6 +1,6 @@
 // This is the shared version for all of the js packages
 // Update the version here and merge your code to main to
 // publish a new version of all of the packages to npm.
-const version = 'v0.22.49';
+const version = 'v0.22.50';
 
 export { version };


server/src/instant/db/instaql.clj
```diff
@@ -1928,34 +1939,67 @@
...
+                                     (assoc-in acc
+                                               [:no-programs k]
+                                               {:result true}))))
+
+          acc-field-programs (fn [acc etype {:keys [field-programs eid]}]
+                               (reduce-kv
+                                (fn [acc label program]
+                                  (let [k [etype eid label]]
+                                    (if-not program
+                                      (assoc-in acc [:no-programs k] {:result true})
```
stopachka: Right now, I record  a `check` for _every_ field, even if the user never wrote a field program for it. 

I did this to keep things consistent with our `view` rules. For `view` rules, `permissioned-node` expects a `(:check result)` from here.  

I do worry this choice for fields will add a lot of extra objects. Alternatively, I could only include field-programs that exist, and update `permissioned-node` to only check the field program when it's present. 

What do you think @dwwoelfel ?

stopachka: Update: fixed so we don't create unnecessary programs

server/src/instant/db/instaql.clj
```diff
@@ -1815,30 +1825,33 @@
    (tracer/with-span! {:name "extract-permission-helpers"}
      (extract-permission-helpers* acc ctx instaql-res))))
 
+(defn viewable-triple? [attrs etype+eid->check [e a]]
+  (let [[etype label] (-> (attr-model/seek-by-id a attrs)
+                          attr-model/fwd-ident-name)
+        view-check (get etype+eid->check [etype e])
+        field-check (get etype+eid->check [etype e label] {:result true})]
+    (when (= label "handle")
```
dwwoelfel: def-locals

stopachka: 😳

PR #1917, opened 2025-11-05
Use tsc --watch in dev

dwwoelfel:
We run `tshy --dev` in dev mode, which is really slow for two reasons:

1. It builds esm and commonjs, but we only need esm for dev
2. It builds from scratch on each change

This PR adds a new `tsconfig.dev.json` that will let us use `tsc --watch` in incremental mode, which is much faster.

We still have to build the commonjs versions of all of the packages that `react-native` depends on, or else it won't be able to find any of the packages. But those use tsc --watch and they're fast.

As everything builds, you might see some compiler errors about missing dependencies (e.g. missing `@instantdb/version`). Those will quickly resolve as the dependencies build and trigger rebuilds. We could potentially use https://www.typescriptlang.org/docs/handbook/project-references.html to solve that issue.


stopachka (approved):
Yess!!

PR #1921, opened 2025-11-06
Faster validate required query

dwwoelfel:
I noticed the validate required query was very slow when inserting many rows and traced it down to postgres using the triples_created_at index instead of the triples_pkey index.

This PR adds a hint (which we can disable with a flag if it causes issues) to force postgres to use the faster index.

Also modifies the query slightly to take a pair of arrays as input instead of json, which should be a little faster on the postgres side and sends a little bit less data.


stopachka (approved):
heeck yeah!

PR #1925, opened 2025-11-07
Fix metrics

nezaj:
Noticed some discrepancy in our overview and monthly metrics. With these changes the numbers now match and we should feel more confident that we are meeting our goals if we're exceed our trendline!


stopachka (approved):
SGTM!

PR #2274, opened 2026-02-10
Add permissions for rooms

nezaj:
Adds permission checks for room joins, gated by room type. Rules are defined under a `$rooms` key in permissions, keyed by room type, with a single `join` action. The CEL context exposes `data.id` and `auth`.

CleanShot 2026-02-10 at 15 23 55@2x

### Default behavior

- If `$rooms` is **not defined** in rules, all joins succeed (backwards compatible)
- If `$rooms` is defined, room types not listed fall back to `$rooms.$default` if present, otherwise allow
- Admin sessions bypass room permission checks

### Rule syntax

```typescript
const rules = {
  $rooms: {
    chat: {
      allow: {
        join: 'auth.id != null',
      },
      bind: ['isMember', "data.id in auth.ref('$user.groups.id')"],
    },
    $default: {
      allow: {
        join: 'false',
      },
    },
  },
};
```

This pattern also let's us expand to other actions we may want to support like `broadcast` and `setPresence`

### How it works

**Client:** The client now sends `room-type` alongside `room-id` in the `join-room` WebSocket message. The room type is stored in the Reactor's room state so it persists across reconnects.

**Server:** When handling a `join-room` message, the server:

1. Looks up `$rooms` rules for the app
2. If no `$rooms` key exists → allow (backwards compat)
3. If `$rooms` exists but client didn't send `room-type` → deny (old clients must upgrade)
4. Finds the matching CEL program using fallback order: specific room type → `$default`
5. Evaluates the CEL expression with `{data: {id: "..."}, auth: ...}`
6. On denial, sends `join-room-error` instead of `join-room-ok`

### Files changed

| File                                                    | Change                                                                                                 |
| ------------------------------------------------------- | ------------------------------------------------------------------------------------------------------ |
| `client/packages/core/src/Reactor.js`                   | Thread `roomType` through `joinRoom`, `_tryJoinRoom`, reconnect, `subscribePresence`, `subscribeTopic` |
| `client/packages/core/src/index.ts`                     | Pass `roomType` to reactor                                                                             |
| `client/packages/react-common/src/InstantReactRoom.ts`  | Pass `room.type` in hooks                                                                              |
| `client/packages/core/src/rulesTypes.ts`                | Add `$rooms` type to `InstantRules`                                                                    |
| `server/src/instant/db/cel.clj`                         | Add `"join"` to `action->compiler`                                                                     |
| `server/src/instant/model/rule.clj`                     | Add `$rooms` validation, `get-room-program!` with fallback logic                                       |
| `server/src/instant/reactive/session.clj`               | Add `assert-room-permission!`, modify `handle-join-room!`                                              |
| `server/test/instant/model/rule_test.clj`               |   rule validation and program compilation                                                   |
| `server/test/instant/reactive/session_test.clj`         | permission enforcement                                                                     |
| `client/sandbox/react-nextjs/pages/play/room-perms.tsx` | E2E play page                                                                                          |

### DIFF
--- a/client/packages/core/src/Reactor.js
+++ b/client/packages/core/src/Reactor.js
@@ -618,7 +618,8 @@ export default class Reactor {
 
         for (const roomId of Object.keys(this._rooms)) {
           const enqueuedUserPresence = this._presence[roomId]?.result?.user;
-          this._tryJoinRoom(roomId, enqueuedUserPresence);
+          const roomType = this._rooms[roomId]?.roomType;
+          this._tryJoinRoom(roomType, roomId, enqueuedUserPresence);
         }
         break;
       }
@@ -2227,15 +2228,17 @@ export default class Reactor {
   // Rooms
 
   /**
+   * @param {string} roomType
    * @param {string} roomId
    * @param {any | null | undefined} [initialPresence] -- initial presence data to send when joining the room
    * @returns () => void
    */
-  joinRoom(roomId, initialPresence) {
+  joinRoom(roomType, roomId, initialPresence) {
     let needsToSendJoin = false;
     if (!this._rooms[roomId]) {
       needsToSendJoin = true;
       this._rooms[roomId] = {
+        roomType,
         isConnected: false,
         error: undefined,
       };
@@ -2250,7 +2253,7 @@ export default class Reactor {
     }
 
     if (needsToSendJoin) {
-      this._tryJoinRoom(roomId, initialPresence);
+      this._tryJoinRoom(roomType, roomId, initialPresence);
     }
 
     return () => {
@@ -2284,6 +2287,15 @@ export default class Reactor {
   getPresence(roomType, roomId, opts = {}) {
     const room = this._rooms[roomId];
     const presence = this._presence[roomId];
+
+    if (room?.error) {
+      return {
+        ...buildPresenceSlice(presence?.result || {}, opts, this._sessionId),
+        isLoading: false,
+        error: room.error,
+      };
+    }
+
     if (!room || !presence || !presence.result) return null;
 
     return {
@@ -2326,8 +2338,13 @@ export default class Reactor {
     });
   }
 
-  _tryJoinRoom(roomId, data) {
-    this._trySendAuthed(uuid(), { op: 'join-room', 'room-id': roomId, data });
+  _tryJoinRoom(roomType, roomId, data) {
+    this._trySendAuthed(uuid(), {
+      op: 'join-room',
+      'room-type': roomType,
+      'room-id': roomId,
+      data,
+    });
     delete this._roomsPendingLeave[roomId];
   }
 
@@ -2345,6 +2362,7 @@ export default class Reactor {
   // TODO: look into typing again
   subscribePresence(roomType, roomId, opts, cb) {
     const leaveRoom = this.joinRoom(
+      roomType,
       roomId,
       // Oct 28, 2025
       // Note: initialData is deprecated.
@@ -2459,8 +2477,8 @@ export default class Reactor {
     });
   }
 
-  subscribeTopic(roomId, topic, cb) {
-    const leaveRoom = this.joinRoom(roomId);
+  subscribeTopic(roomType, roomId, topic, cb) {
+    const leaveRoom = this.joinRoom(roomType, roomId);
 
     this._broadcastSubs[roomId] = this._broadcastSubs[roomId] || {};
     this._broadcastSubs[roomId][topic] =
--- a/client/packages/core/src/Reactor.js
+++ b/client/packages/core/src/Reactor.js
@@ -251,7 +251,7 @@ export default class Reactor {
   /** @type BroadcastChannel | undefined */
   _broadcastChannel;
 
-  /** @type {Record<string, {isConnected: boolean; error: any}>} */
+  /** @type {Record<string, {roomType: string; isConnected: boolean; error: any}>} */
   _rooms = {};
   /** @type {Record<string, boolean>} */
   _roomsPendingLeave = {};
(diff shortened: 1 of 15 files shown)


client/packages/core/src/Reactor.js
```diff
@@ -2284,6 +2287,15 @@ export default class Reactor {
   getPresence(roomType, roomId, opts = {}) {
     const room = this._rooms[roomId];
```
stopachka, line 2288: What happens if someone joins two rooms, different roomType, but same id? 

I wonder if now that we are using `roomType`, we should create a kind of composite key: roomType+roomId

client/packages/core/src/rulesTypes.ts
```diff
@@ -8,6 +8,11 @@ type InstantRulesAttrsAllowBlock = {
   delete?: string | null | undefined;
 };
 
+type InstantRoomRulesAllowBlock = {
+  $default?: string | null | undefined;
```
stopachka, line 12: Do we need $default here? Maybe we add that in once we have something other than $join?

server/src/instant/db/cel.clj
```diff
@@ -410,6 +410,20 @@
   (-> (runtime-compiler-builder)
       (.build)))
 
+(def ^:private ^CelCompiler cel-join-compiler
+  (-> (CelCompilerFactory/standardCelCompilerBuilder)
```
stopachka, line 414: do we need a custom one? this looks like cel-runtime-compiler. Maybe the missing thing is ruleParams, but perhaps that's fine? (Maybe at some point we'll want to support ruleParams)

nezaj: Yea Claude originally wanted to group this with `view-delete` but I thought conceptually it might be nicer to separate it

server/src/instant/model/rule.clj
```diff
@@ -366,6 +370,7 @@
 
 (defn rule-validation-errors [rules]
   (->> (keys rules)
+       (remove #(= "$rooms" %))
```
stopachka: this feels like a smell. I would destructure before this function gets called, and orchestrate the work there. 

Reason: it's not clear when reasoning locally here why we would _not_ want to look at $rooms. There's an implicit assumption here that somewhere else we do look at rooms.	

Maybe we call this `entity-rule-validation-errors`, or something like this

server/src/instant/model/rule.clj
```diff
@@ -336,7 +340,7 @@
 
                        :else errors)))
              []
-             rules))
+             (dissoc rules "$rooms")))
```
stopachka: would not add a dissoc here. (same reasoning as in rule-validation-errors)

client/packages/core/src/rulesTypes.ts
nezaj: Yea was thinking we may not need it yet but what if LLMs want to use it already? Figured it might be nice to just allow it for now so that it continues working when we add other actions

server/src/instant/model/rule.clj
```diff
@@ -397,10 +402,103 @@
...
+           (catch CelValidationException e
+             (ex/throw-validation-err!
+              :permission
+              (first paths)
+              (->> (.getErrors e)
+                   (map (fn [^CelIssue cel-issue]
+                          {:message (.getMessage cel-issue)})))))))
+       (recur (next paths))))))
+
+(defn get-room-program!
```
stopachka, line 427: It feels like there's a lot of special-purpose code written specifically for `$rooms` here. Why is it necessary to split it out like this?

server/src/instant/reactive/session.clj
```diff
@@ -611,16 +612,60 @@
...
+  [{:keys [app-id room-type room-id current-user admin?]}]
+  (when-not admin?
+    (let [rules (rule-model/get-by-app-id {:app-id app-id})
+          has-room-rules? (some? (get-in rules [:code "$rooms"]))]
+      (when has-room-rules?
+        (when-not room-type
+          (ex/assert-permitted!
+           :has-room-join-permission?
+           ["$rooms" nil "join"]
+           false))
```
stopachka: nit: would pass the room-type conditional where, and remove the when-not

server/src/instant/reactive/session.clj
```diff
@@ -611,16 +612,60 @@
...
+                                :room-type room-type
+                                :room-id room-id
+                                :current-user current-user
+                                :admin? (:admin? auth)})
+      (eph/join-room! app-id sess-id current-user room-id data)
+      (join-room-logger/log-join-room! app-id)
+      (rs/send-event! store app-id sess-id {:op :join-room-ok
+                                            :room-id room-id
+                                            :client-event-id client-event-id})
+      (catch Exception e
```
stopachka, line 657: Is this necessary? I would have expected this exception would have bubbled up to a top-level exception handler

stopachka (commented):
Some ideas!

server/src/instant/reactive/session.clj
nezaj: Without the try/catch, the error would bubble to handle-instant-exception, which sends a generic `{:op :error}` message. On the client, the Reactor handles that in `_handleReceiveError` — but that code doesn't know anything about rooms. It wouldn't set room.error or notify presence subscribers.

  So without it, here's what the user would see:

  1. No join-room-ok sent (permission denied)
  2. No join-room-error sent (bubbled to generic handler)
  3. Generic :error sent, but Reactor doesn't route it to room state
  4. Room stays stuck on isLoading: true forever

The top-level handler includes client-event-id and original-event so Reactor could match it back to the room join. But that would mean adding room-specific routing logic inside the generic error handler

Wdyt?

nezaj:
Will close for now to clean up PRs and put up again later when we re-visit this

PR #2468, opened 2026-04-02
Fix inference in query inspector

dwwoelfel:
The cardinality inference in the query inspector is not working. We create a separate db with the schema in the query inspector component. But then we just end up using the existing db because it has the same reactor cache key.

This updates the reactor cache key so that it changes when inference is disabled/enabled.

### DIFF
--- a/client/packages/core/src/index.ts
+++ b/client/packages/core/src/index.ts
@@ -954,7 +954,11 @@ function init<
   versions?: { [key: string]: string },
   EventSourceImpl?: any,
 ): InstantCoreDatabase<Schema, UseDates> {
-  const cardinalityInference = config.cardinalityInference ?? !!config.schema;
+  const cardinalityInference =
+    Boolean(config.schema) &&
+    ('cardinalityInference' in config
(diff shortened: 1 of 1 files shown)


stopachka (approved):
nice!

client/packages/core/src/index.ts
```diff
@@ -303,7 +304,9 @@ function reactorKey(config: InstantConfig<any, boolean>): string {
     '_' +
     (adminToken || 'client_only') +
     '_' +
-    config.useDateObjects
+    config.useDateObjects +
+    '_' +
+    String(!!config.cardinalityInference)
```
stopachka, line 309: One thing I realize: 

Would this mean, that if a user goes between say "Explorer" and "Inspector" a few times, they will end up with 2 websocket connections?

dwwoelfel: Yeah, they'll get a new connection as soon as they view the query inspector page.

dwwoelfel:
Going to close this--it's drifted too far from the main branch and I haven't solved the single connection problem yet.

PR #2495, opened 2026-04-07
Make self-hostable

drew-harris:
Instructions for both localhost and hetzner setup at:
https://github.com/instantdb/instant/tree/drewh/self-hosting/self-hosting#readme

# Change Summary
 - Create a docker image for the dashboard.
   - Create a method for passing runtime environment variables to the dashboard client/server. Docker container creates a publicSelfHostedVariables.js file in /public that gets loaded before the client.
   - All non dashboard routes redirect to the dashboard. (i.e. you can't self-host our blog/docs)
 - Create a docker image for the server. 
 - On container startup runs the oss-bootstrap task to create a aead-keyset and webhook-keyset which gets stored in a volume to be persisted
 - Refactored s3.clj to allow for any S3 compatible filesystem. 
 - Created 2 docker compose files in self-hosting/
 - Created a github action workflow to publish images. 

# Review Instructions
Visit the instructions above and try to follow the instructions for setting up InstantDB on a fresh VPS. 
Should be able to use `create-instant-app` and `instant-cli` with the `INSTANT_CLI_API_URI` environment variable. 

Also try following the localhost/local dev instructions.

Would also appreciate a thorough review on the clojure code since that is what I'm least confident about. 

# Todos:
 - [x] Publish docker images manually so that the docker-compose files do not need to build them.
 - [x] Publish docker images on push to main. 

# Future Steps:
Make it easier to use create instant app and instant-cli with self hosted instances. 
	Example: create-instant-app should set the apiURI and websocketURI in the init config when scaffolding. 
    Example: instant-cli init should also save the backend api url when creating the .env with the app id and admin token.

### DIFF
--- a/client/Dockerfile.www
+++ b/client/Dockerfile.www
@@ -1,0 +1,96 @@
+ARG NODE_VERSION=22.22.2-slim
+
+FROM node:${NODE_VERSION} AS dependencies
+
+WORKDIR /monorepo
+
+COPY package.json pnpm-workspace.yaml pnpm-lock.yaml ./
+
+
+# Install project dependencies with frozen lockfile for reproducible builds
+RUN --mount=type=cache,target=/root/.npm \
+    --mount=type=cache,target=/usr/local/share/.cache/yarn \
+    --mount=type=cache,target=/root/.local/share/pnpm/store \
+    if [ -f package-lock.json ]; then \
+    npm ci --no-audit --no-fund; \
+    elif [ -f yarn.lock ]; then \
+    corepack enable yarn && yarn install --frozen-lockfile --production=false; \
+    elif [ -f pnpm-lock.yaml ]; then \
+    corepack enable pnpm && pnpm install --frozen-lockfile; \
+    else \
+    echo "No lockfile found." && exit 1; \
+    fi
+
+FROM node:${NODE_VERSION} AS pruner
+
+WORKDIR /monorepo
+
+COPY --from=dependencies /monorepo/node_modules ./node_modules
+
+COPY . .
+RUN corepack enable pnpm;
+RUN pnpm turbo prune instant-www --docker
+
+FROM node:${NODE_VERSION} AS builder
+WORKDIR /monorepo
+ARG NEXT_PUBLIC_SELF_HOSTED=true
+ENV NEXT_PUBLIC_SELF_HOSTED=${NEXT_PUBLIC_SELF_HOSTED}
+COPY --from=pruner /monorepo/out/json/ .
+RUN corepack enable pnpm
+RUN pnpm install --frozen-lockfile
+
+COPY --from=pruner /monorepo/out/full/ .
+COPY tsconfig.base.json ./
+
+RUN pnpm install --frozen-lockfile
+
+RUN pnpm turbo build
+
+# Production runner
+FROM node:${NODE_VERSION} AS runner
+WORKDIR /app
+
+ENV NODE_ENV=production
+ENV NEXT_TELEMETRY_DISABLED=1
+ENV INSTANT_API_URI=
+ENV INSTANT_BACKEND_URL=
+ARG NEXT_PUBLIC_SELF_HOSTED=true
+ENV NEXT_PUBLIC_SELF_HOSTED=${NEXT_PUBLIC_SELF_HOSTED}
+
+RUN addgroup --system --gid 1001 nodejs && \
+    adduser --system --uid 1001 nextjs
+
+# Copy the standalone output
+COPY --from=builder /monorepo/www/.next/standalone ./
+# Copy static assets and public files
+COPY --from=builder /monorepo/www/.next/static ./www/.next/static
+COPY --from=builder /monorepo/www/public ./www/public
+
+COPY <<'EOF' /app/start.sh
+#!/bin/sh
+set -eu
+
+node <<'NODE'
+const fs = require('fs');
+
+const apiURI = process.env.INSTANT_API_URI || process.env.INSTANT_BACKEND_URL;
+const config = apiURI ? { apiURI } : {};
+
+fs.writeFileSync(
+  '/app/www/public/publicSelfHostedVariables.js',
+  `window.__instantConfig = ${JSON.stringify(config)};\n`,
+);
+NODE
+
+exec node www/server.js
+EOF
+
+RUN chmod +x /app/start.sh && chown -R nextjs:nodejs /app/www/public /app/start.sh
+
+USER nextjs
+
+EXPOSE 3000
+ENV PORT=3000
+ENV HOSTNAME="0.0.0.0"
+
+CMD ["/app/start.sh"]
--- a/client/Dockerfile.www
+++ b/client/Dockerfile.www
@@ -11,15 +11,7 @@ COPY package.json pnpm-workspace.yaml pnpm-lock.yaml ./
 RUN --mount=type=cache,target=/root/.npm \
     --mount=type=cache,target=/usr/local/share/.cache/yarn \
     --mount=type=cache,target=/root/.local/share/pnpm/store \
-    if [ -f package-lock.json ]; then \
-    npm ci --no-audit --no-fund; \
-    elif [ -f yarn.lock ]; then \
-    corepack enable yarn && yarn install --frozen-lockfile --production=false; \
-    elif [ -f pnpm-lock.yaml ]; then \
-    corepack enable pnpm && pnpm install --frozen-lockfile; \
-    else \
-    echo "No lockfile found." && exit 1; \
-    fi
+    corepack enable pnpm && pnpm install --frozen-lockfile;
 
 FROM node:${NODE_VERSION} AS pruner
 
--- a/client/www/lib/config.ts
+++ b/client/www/lib/config.ts
@@ -3,6 +3,54 @@ export const isBrowser = typeof window != 'undefined';
 export const isDev = process.env.NODE_ENV === 'development';
 
 const isStaging = process.env.NEXT_PUBLIC_STAGING === 'true';
+export const isSelfHosted = process.env.NEXT_PUBLIC_SELF_HOSTED === 'true';
+
+type DashboardConfig = {
+  apiURI: string;
+  websocketURI: string;
+};
+
+type RuntimeDashboardConfig = Partial<DashboardConfig>;
+
+declare global {
+  interface Window {
+    __instantConfig?: RuntimeDashboardConfig;
+  }
+}
+
+function websocketURIFromApiURI(apiURI: string) {
+  const url = new URL(apiURI);
+  url.protocol = url.protocol === 'https:' ? 'wss:' : 'ws:';
+  url.pathname = '/runtime/session';
+  url.search = '';
+  url.hash = '';
+  return url.toString();
+}
+
+function configFromApiURI(apiURI: string | undefined | null) {
+  if (!apiURI) {
+    return null;
+  }
+
+  return {
+    apiURI,
+    websocketURI: websocketURIFromApiURI(apiURI),
+  };
+}
+
+function getRuntimeConfig() {
+  try {
+    if (isBrowser) {
+      return configFromApiURI(window.__instantConfig?.apiURI);
+    }
+
+    return configFromApiURI(
+      process.env.INSTANT_API_URI || process.env.INSTANT_BACKEND_URL,
+    );
+  } catch (_e) {
+    return null;
+  }
+}
 
 const devBackend = getLocal('devBackend');
 
@@ -15,14 +63,14 @@ if (devBackend && isBrowser) {
   }
 }
 
-const config = {
-  apiURI: devBackend
-    ? `http://localhost:${localPort}`
-    : `https://${isStaging ? 'api-staging' : 'api'}.instantdb.com`,
-  websocketURI: devBackend
-    ? `ws://localhost:${localPort}/runtime/session`
-    : `wss://${isStaging ? 'api-staging' : 'api'}.instantdb.com/runtime/session`,
-};
+const defaultApiURI = devBackend
+  ? `http://localhost:${localPort}`
+  : `https://${isStaging ? 'api-staging' : 'api'}.instantdb.com`;
+
+export const config =
+  isSelfHosted && !devBackend
+    ? getRuntimeConfig() || configFromApiURI(defaultApiURI)!
+    : configFromApiURI(defaultApiURI)!;
 
 // In dev mode, sync the devBackend flag to a cookie so server components
 // can resolve the same apiURI as the client.
@@ -73,7 +121,7 @@ export const stripeCustomerPortalURI = isDev
   ? stripeDevCustomerPortalURI
   : stripeProdCustomerPortalURI;
 
-export function getLocal(k: string) {
+export function getLocal<T = any>(k: string): T | null {
   if (!isBrowser) {
     return null;
   }
@@ -87,7 +135,7 @@ export function getLocal(k: string) {
   }
 }
 
-export function setLocal(k: string, v: any) {
+export function setLocal<T = any>(k: string, v: T) {
   if (!isBrowser) {
     return;
   }
--- a/client/www/next.config.js
+++ b/client/www/next.config.js
@@ -24,6 +24,9 @@ async function fetchStarCount() {
 
 /** @type {import('next').NextConfig} */
 const nextConfig = {
+  ...(process.env.NEXT_PUBLIC_SELF_HOSTED === 'true'
+    ? { output: 'standalone' }
+    : {}),
   reactStrictMode: true,
   pageExtensions: ['js', 'jsx', 'ts', 'tsx', 'md'],
   transpilePackages: ['@instantdb/components'],
@@ -54,7 +57,7 @@ const nextConfig = {
     return config;
   },
   async redirects() {
-    return [
+    const standardRedirects = [
       {
         permanent: false,
         source: '/',
@@ -123,6 +126,21 @@ const nextConfig = {
         destination: 'https://discord.com/invite/VU53p7uQcE',
       },
     ];
+
+    if (process.env.NEXT_PUBLIC_SELF_HOSTED === 'true') {
+      return [
+        ...standardRedirects,
+        // redirect any page that isn't /api, /dash, /docs, or /logout
+        {
+          permanent: false,
+          source:
+            '/((?!api(?:/.*)?|dash(?:/.*)?|docs(?:/.*)?|logout(?:/.*)?|.*\\.[^/]+$).*)',
+          destination: '/dash',
+        },
+      ];
+    }
+
+    return standardRedirects;
   },
   // Proxy to PostHog to avoid ad blockers
   async rewrites() {
--- a/client/www/pages/_document.tsx
+++ b/client/www/pages/_document.tsx
@@ -1,4 +1,6 @@
+import { isSelfHosted } from '@/lib/config';
 import { Html, Head, Main, NextScript } from 'next/document';
+import Script from 'next/script';
 
 export default function Document() {
   return (
@@ -48,6 +50,12 @@ export default function Document() {
         />
       </Head>
       <body>
+        {isSelfHosted && (
+          <Script
+            strategy="beforeInteractive"
+            src="/publicSelfHostedVariables.js"
+          />
+        )}
         <Main />
         <NextScript />
       </body>
--- a/self-hosting/.env.example
+++ b/self-hosting/.env.example
@@ -1,6 +1,7 @@
 # Public URLs. Change these when deploying behind a domain or proxy.
 INSTANT_BACKEND_URL=http://localhost:8888
 INSTANT_DASHBOARD_URL=http://localhost:3000
+S3_PUBLIC_ENDPOINT=http://localhost:9000
 
 # Domains used by the Caddy compose file. For production, point DNS at this host
 # and update the public URLs above to use these domains with https://.
@@ -23,7 +24,6 @@ MINIO_ROOT_USER=pleasegenerateauserhere
 MINIO_ROOT_PASSWORD=pleasegenerateapasswordhere
 S3_BUCKET=instant-bucket
 S3_ENDPOINT=http://minio:9000
-S3_PUBLIC_ENDPOINT=http://localhost:9000
 AWS_REGION=us-east-1
 
 # Optional integrations.
--- a/self-hosting/.env.example
+++ b/self-hosting/.env.example
@@ -1,0 +1,42 @@
+# Public URLs. Change these when deploying behind a domain or proxy.
+INSTANT_BACKEND_URL=http://localhost:8888
+INSTANT_DASHBOARD_URL=http://localhost:3000
+
+# Domains used by the Caddy compose file. For production, point DNS at this host
+# and update the public URLs above to use these domains with https://.
+# These are only required if using docker-compose.with-caddy.yml
+DASHBOARD_DOMAIN=localhost
+BACKEND_DOMAIN=api.localhost
+STORAGE_DOMAIN=files.localhost
+
+# Database. DATABASE_URL is inferred from these values inside docker-compose.
+POSTGRES_USER=instant
+POSTGRES_PASSWORD=pleasegenerateapasswordhere
+POSTGRES_DB=instant
+
+# Object storage. Server AWS credentials are inferred from these MinIO values.
+# Replace these before deploying.
+# If using AWS S3, set MINIO_ROOT_USER to your AWS Access Key ID
+# and MINIO_ROOT_PASSWORD to your AWS Secret Access Key, and
+# leave S3_ENDPOINT and S3_PUBLIC_ENDPOINT blank.
+MINIO_ROOT_USER=pleasegenerateauserhere
+MINIO_ROOT_PASSWORD=pleasegenerateapasswordhere
+S3_BUCKET=instant-bucket
+S3_ENDPOINT=http://minio:9000
+S3_PUBLIC_ENDPOINT=http://localhost:9000
+AWS_REGION=us-east-1
+
+# Optional integrations.
+STRIPE_API_KEY=
+
+# Email. Provide the Postmark token to send login magic codes.
+# If omitted, one-time login codes will be printed to stdout
+# from the server container.
+POSTMARK_TOKEN=
+INSTANT_EMAIL_REPLY_TO=hello@example.com
+INSTANT_DASHBOARD_EMAIL_SENDER_NAME=Instant
+INSTANT_DASHBOARD_EMAIL_SENDER_EMAIL=verify@example.com
+INSTANT_APP_EMAIL_SENDER_NAME=Instant
+INSTANT_APP_EMAIL_SENDER_EMAIL=verify@example.com
+INSTANT_TEAM_EMAIL_SENDER_NAME=Instant
+INSTANT_TEAM_EMAIL_SENDER_EMAIL=teams@example.com
--- a/self-hosting/README.md
+++ b/self-hosting/README.md
@@ -1,0 +1,126 @@
+# Self Hosting InstantDB
+
+An InstantDB self hosted deployment consists of 3 urls (subdomains)
+1. Backend API
+2. Dashboard URL
+3. Files URL
+
+
+## Localhost Setup
+Use this setup if you want to self-host InstantDB while developing on your own machine. In most cases, it's more convenient to use our cloud service to spin up unlimited free apps.
+```bash
+git clone https://github.com/instantdb/instant.git
+cd instant/self-hosting
+docker compose --env-file .env.example up
+```
+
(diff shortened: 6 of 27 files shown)


GrocerPublishAgent:
Hey! Just say thanks for making InstantDB open source in the first place. Really appreciate that you've made that possible.

I saw you HN launch post on Thursday and started fiddling with self-hosting separately. I just today saw your work on this branch.

I took a somewhat different approach than you, so I wanted to share back in case any of my ideas were useful.

My main constraint was that I needed everything to be fully configurable via environment variables — all ports, data directories, etc. The reason is I run multiple worktrees in parallel (each with its own Postgres instance and allocated ports) so that my coding agents can work on different branches simultaneously without stepping on each other. That meant I needed to rely entirely on env vars to make the whole stack relocatable rather than just prod, staging, dev mapped to hardcoded urls.

I also went the MinIO route for storage.  I saw you did not change S3AsyncClient/crtBuilder uploader. Did you get file uploading to work with MinIO?  I ended up building a hacky workaround for it. Would love to know if I just did something wrong and its possible to just use minio without any aws s3 client code changes.

For the dashboard I was able to sidestep needing a `self-hosted` flag threaded through everywhere like I saw you doing.

if you're curious the full diff against upstream: https://peoplesgrocers.com/code/comrades/self-hosted-instantdb/pulls/1/files. It looks a bit big only because I also added in a client library for Qwik and example for Qwik-City ( just ignore that part, its a niche library found in ecommerce).

Anyways, thanks again for making a great codebase. It was really easy to get my app working.

Escapado:
Thanks for working on this! I went through the PR and the code and it seems (but please correct me) that AWS S3 is currently the only way to get storage, right? Would be amazing if other S3 compatible solutions could be used - for example our start up runs on Hetzner due to being EU based and in the education space so AWS is not really an option for us but Hetzner has an S3 compatible Object Storage solution I would love to use.

server/src/instant/config.clj
```diff
@@ -33,6 +34,7 @@
     ;;      override it in the eb env vars
     (= "true" staging-env)                        :staging
     (= "true" prod-env)                           :prod
+    (= "true" self-hosted-env)                    :self-hosted
```
dwwoelfel: Making self-hosted an `env` seems wrong to me. It seems like the `env` should still be prod.

server/src/instant/storage/s3.clj
```diff
@@ -19,17 +21,36 @@
...
+(defn- configure-s3-endpoint-client ^S3ClientBuilder [^S3ClientBuilder builder]
+  (if-let [endpoint (config/s3-endpoint)]
+    (doto builder
+      (.endpointOverride (URI/create endpoint))
+      (.region (Region/of (config/s3-region)))
+      (.forcePathStyle Boolean/TRUE))
+    builder))
+
+(defn- configure-s3-endpoint-async-client ^S3CrtAsyncClientBuilder [^S3CrtAsyncClientBuilder builder]
+  (if-let [endpoint (config/s3-endpoint)]
```
dwwoelfel: This might not work if someone wants to use a directory bucket.

server/src/instant/config.clj
drew-harris: I see what you mean, this is more meant to be a catch all toggle for when we need to change behavior because someone is self hosting
I meant it more as (selfhosted?, prod?) as opposed to (selfhosted | prod)

server/src/instant/storage/s3.clj
drew-harris: Not familiar with those, will research

server/resources/config/self-hosted.edn
```diff
@@ -0,0 +1,23 @@
...
+ ;; :s3-storage-secret-key "your-secret"
+ ;; :postmark-token "your-token"
+ ;; :sendgrid-token "your-token"
+ ;; :stripe-secret "your-secret"
+ ;; :stripe-webhook-secret "your-secret"
+ ;; :honeycomb-api-key "your-key"
+ ;; :posthog-api-key "your-key"
+ ;; :google-oauth-client {:client-id "..." :client-secret "..."}
+
+ :instant-config-app-id #uuid "00000000-0000-0000-0000-000000000000"}
```
dwwoelfel: Are you creating an app with this id? Otherwise, someone could come in and create an app with the id and override the config. You can probably just leave it blank.

self-hosting/README.md
```diff
@@ -0,0 +1,81 @@
+# Self Hosting InstantDB
+
+An InstantDB self hosted deployment consists of 3 urls (subdomains)
+1. Backend API
+2. Dashboard URL
+3. Files URL
+
+## Hetzner Setup
+Create a new server on Hetzner. I used 4vcpu, 8GB of RAM, for more memory constrained environments, the `JAVA_OPTIONS` env can be set to limit the memory the server container uses.
```
dwwoelfel: I wouldn't say "I" here, maybe "We tested"

self-hosting/README.md
```diff
@@ -0,0 +1,81 @@
...
+
+After everything starts up, the dashboard should be available at wherever you set INSTANT_DASHBOARD_URL to. 
+
+
+### Notes
+If the server container quits with an error similar to: 
+```
+FATAL: number of requested standby connections exceeds max_wal_senders (currently 4)
+```
+Edit the docker compose file you are using and increase the `max_wal_senders` argument for the postgres service.
```
dwwoelfel: Why not just make this higher by default?

drew-harris: Ive tested this multiple times and the error only happened once so I'm hesitant to go around adjusting things when I don't understand what they do. If you think its a good idea then we can change it. 
Do you think it should also be changed in the local dev docker compose?

server/src/instant/config.clj
```diff
@@ -200,6 +218,7 @@
    (case env
      :prod "https://www.instantdb.com"
      :staging "https://staging.instantdb.com"
+     :self-hosted (System/getenv "INSTANT_DASHBOARD_URL")
```
stopachka: imo we don't need this self-hosted check. This could we written as 

```
(or (System/getenv ...)
      (case env)
```

server/src/instant/config.clj
stopachka: > need to change behavior because someone is self hosting

Ideally we would not need to do this. 

Instead of detecting the self-hosted state, we should allow an injection for whatever that detection is trying to account for. 

This way, we remove an explicit "branch" for the code

server/src/instant/reactive/ephemeral.clj
```diff
@@ -164,7 +164,7 @@
           (.setProperty "tag-value" (aws-util/get-environment-tag)))
         (.setEnabled metrics-config true))
 
-      :dev
+      (:dev :self-hosted)
```
stopachka: I am a bit worried about this check here. 

Would love to learn more about what will happen to hazelcast in self-hosted instances. i.e say the user has 3 machines running. What happens?

server/resources/config/self-hosted.edn
```diff
@@ -0,0 +1,22 @@
+;; Self-hosted configuration - no AWS dependencies
+;; All secrets are plaintext, optional integrations are omitted
```
stopachka: Maybe for self-hosted setups, we don't have any config.edn at all? 

My reasoning: this way we can standardize to providing everything via env vars

cc @dwwoelfel @drew-harris for thoughts

self-hosting/README.md
```diff
@@ -0,0 +1,73 @@
...
+```
+
+Example:
+```bash
+# Public URLs. Change these when deploying behind a domain or proxy.
+INSTANT_BACKEND_URL=https://api.myinstant.com
+INSTANT_DASHBOARD_URL=https://dash.myinstant.com
+S3_PUBLIC_ENDPOINT=https://files.myinstant.com
+
+# Domains used by the Caddy compose file. For production, point DNS at this host
```
stopachka: I found this comment a bit confusing. 

Idea: Is these 1:1 mirror the vars above, maybe we don't need to set them, and can just remove the https:// 

?

self-hosting/README.md
```diff
@@ -0,0 +1,73 @@
...
+```sh
+cd instant/self-hosting
+cp .env.example .env
+nano .env
+```
+
+Example:
+```bash
+# Public URLs. Change these when deploying behind a domain or proxy.
+INSTANT_BACKEND_URL=https://api.myinstant.com
```
stopachka, line 82: Nit: would be nice to give examples to people who don't have a domain handy. I had to use sslip.io

self-hosting/README.md
```diff
@@ -0,0 +1,73 @@
+# Self Hosting InstantDB
+
+An InstantDB self hosted deployment consists of 3 urls (subdomains)
+1. Backend API
+2. Dashboard URL
+3. Files URL
+
+## Hetzner Setup
+Create a new server on Hetzner. We tested used 4vcpu, 8GB of RAM, for more memory constrained environments, the `JAVA_OPTIONS` env can be set to limit the memory the server container uses.
```
stopachka: Nit: this is reasonably involved (requires setting up firewall, etc etc). we may want to be explicit about the commands

self-hosting/.env.example
```diff
@@ -0,0 +1,28 @@
+# Copy this file or pass it directly to Docker Compose:
+# docker compose --env-file self-hosting/.env.example -f self-hosting/evil-docker-compose.yml up --build
```
stopachka: nit: evil-docker-compose does not exist

stopachka: nit: we should be using .env not .env.example

The `or pass it directly to docker compose` seems a bit off -- the practice is to copy the .env file no? 

Alternatively we can have this be a .env file, and simply let people edit it.

self-hosting/.env.example
```diff
@@ -0,0 +1,28 @@
...
+
+# Domains used by the Caddy compose file. For production, point DNS at this host
+# and update the public URLs above to use these domains with https://.
+DASHBOARD_DOMAIN=localhost
+BACKEND_DOMAIN=api.localhost
+STORAGE_DOMAIN=files.localhost
+
+# Database. DATABASE_URL is inferred from these values inside docker-compose.
+POSTGRES_USER=instant
+POSTGRES_PASSWORD=pass
```
stopachka: we should make this something like pleasegenerateapasswordhere

self-hosting/.env.example
```diff
@@ -0,0 +1,28 @@
...
+POSTGRES_DB=instant
+
+# Object storage. Server AWS credentials are inferred from these MinIO values.
```
stopachka, line 18: What would the user have to do, if they _wanted_ to use S3, no minio?

client/www/next.config.js
```diff
@@ -170,9 +171,11 @@ async function fetchFonts() {
 }
 
 module.exports = async () => {
+  const isSelfHosted = process.env.NEXT_PUBLIC_SELF_HOSTED === 'true';
+
   await fetchFonts();
-  const starCount = await fetchStarCount();
-  if (starCount) {
+  const starCount = isSelfHosted ? null : await fetchStarCount();
+  if (!isSelfHosted && starCount) {
```
stopachka: Why do we disable this for self hosting?

client/www/next.config.js
```diff
@@ -24,6 +24,7 @@ async function fetchStarCount() {
 
 /** @type {import('next').NextConfig} */
 const nextConfig = {
+  output: 'standalone',
```
stopachka: what is this about?

client/www/next.config.js
```diff
@@ -6,10 +6,10 @@ async function fetchStarCount() {
   const { init } = await import('@instantdb/admin');
   const db = init({
     appId:
-      process.env.NEXT_PUBLIC_FEEDBACK_APP_ID ||
+      // process.env.NEXT_PUBLIC_FEEDBACK_APP_ID ||
```
stopachka: nit: is this desired?

client/www/middleware.ts
```diff
@@ -6,10 +6,30 @@ const GETADB_HOSTS = new Set([
...
+
+function isPublicAsset(pathname: string) {
+  return /\.[^/]+$/.test(pathname);
+}
+
 export function middleware(request: NextRequest) {
   const host = (request.headers.get('host') ?? '').split(':')[0];
   const url = request.nextUrl.clone();
 
+  if (isSelfHosted) {
```
stopachka: nit: is this really necessary? imo the user can just write /dash themselves?

client/www/pages/_app.tsx
```diff
@@ -54,19 +55,21 @@ function App({ Component, pageProps }: AppPropsWithLayout) {
...
-          value={{
-            provider: localStorageProvider,
-          }}
-        >
-          <NuqsAdapter>{mainEl}</NuqsAdapter>
-        </SWRConfig>
-      </ErrorBoundary>
-      {isDev ? null : <GoogleScripts />}
-      {isDev ? <Dev /> : null}
+      <DeploymentConfigProvider>
```
stopachka: afaik we should not need this. 

Ideally, what we do is: 

When we start nextjs, let's inject some static file that we can then import from.

client/www/lib/config.ts
```diff
@@ -15,6 +33,12 @@ if (devBackend && isBrowser) {
...
 
+/**
+ * @deprecated Use `getConfig()` from '@/lib/hooks/useDeploymentConfig' instead.
+ * This object is kept for backward compatibility during migration.
+ * For self-hosted deployments, this will have incorrect values until
+ * DeploymentConfigProvider initializes.
+ */
 const config = {
   apiURI: devBackend
     ? `http://localhost:${localPort}`
```
stopachka: imo instead of using `getConfig`, we can _keep_ this a static variable. 

We could create some file during next initializaton, which makes plops these variables into a static file. 

This way also don't need to detect if it's self hosted or not. 

WDYT @drew-harris @dwwoelfel ?

self-hosting/README.md
drew-harris: Are you saying that we should include the standard instructions of setting up things like ufw, or do you mean more in depth instructions for creating a vm on hetzner specifically?

self-hosting/README.md
drew-harris: What would that look like? I feel like sslip.io is very niche and I want to avoid confusing people who think they might need it or something?

client/www/next.config.js
drew-harris: Standalone build output helps reduce the size of the final image. https://nextjs.org/docs/pages/api-reference/config/next-config-js/output

Ive updated it so that this only gets set when using the NEXT_PUBLIC_SELF_HOSTED variable so it will only run for docker builds.

self-hosting/.env.example
drew-harris: You can specify the
MINIO_ROOT_USER
MINIO_ROOT_PASSWORD
S3_BUCKET 
with your s3 credentials
and it will use those with s3 as long as S3_ENDPOINT is unset.

self-hosting/docker-compose.with-caddy.yml
```diff
@@ -0,0 +1,123 @@
...
+      S3_BUCKET: ${S3_BUCKET:-instant-bucket}
+    depends_on:
+      minio:
+        condition: service_healthy
+    entrypoint: >
+      /bin/sh -c "
+      /usr/bin/mc alias set local http://minio:9000 $$MINIO_ROOT_USER $$MINIO_ROOT_PASSWORD;
+      /usr/bin/mc mb local/$$S3_BUCKET;
+      /usr/bin/mc policy set public local/$$S3_BUCKET;
+      exit"
```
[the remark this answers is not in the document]
drew-harris: This is not an issue

self-hosting/README.md
drew-harris: The 2nd paragraph is only for the caddy configuration. The first paragraph is for the logic of the backend itself. 
I updated the .env.example to make this clearer. This is a pretty common self hosting pattern I've seen. I don't know if theres a way to get it to infer http vs https + it feels more flexible. open to ideas.

client/www/middleware.ts
drew-harris: Intention is that we only self host the dashboard itself. Feels strange to for self hosters to also self host our blog + homepage

client/www/lib/config.ts
```diff
@@ -24,6 +78,9 @@ const config = {
     : `wss://${isStaging ? 'api-staging' : 'api'}.instantdb.com/runtime/session`,
 };
 
+const sharedConfig =
```
stopachka: I would make this flow in a single path. 

i.e just export config, but have the values _inside_ config generated.

client/www/pages/_document.tsx
```diff
@@ -48,6 +48,7 @@ export default function Document() {
         />
       </Head>
       <body>
+        <script src="/environment.js" />
```
stopachka: Since we are doing this _only_ for self hosting, maybe we call this `something like `publicSelfHostedVariables` or something.

client/www/middleware.ts
stopachka: Perhaps instead of writing this here, we write this in the home page.tsx ? 

Reasoning: 

It would be nice if we can avoid adding more to middleware. We may end up re-writing and dropping middleware, since it's only used for getadb.com. 

the isPublicAsset also makes me a bit worried here, I think we can make the code simpler by just redirecting the root.

client/www/next.config.js
```diff
@@ -24,6 +24,9 @@ async function fetchStarCount() {
 
 /** @type {import('next').NextConfig} */
 const nextConfig = {
+  ...(process.env.NEXT_PUBLIC_SELF_HOSTED === 'true'
+    ? { output: 'standalone' }
```
stopachka, line 28: What is this about?

self-hosting/.env.example
```diff
@@ -0,0 +1,36 @@
+# Public URLs. Change these when deploying behind a domain or proxy.
+INSTANT_BACKEND_URL=http://localhost:8888
+INSTANT_DASHBOARD_URL=http://localhost:3000
+S3_PUBLIC_ENDPOINT=http://localhost:9000
```
stopachka, line 4: Nit: the pattern seems to INSTANT_BACKEND_URL -> BACKEND_DOMAIN, but for storage we go from S3_PUBLIC_ENDPOINT -> STORAGE_DOMAIN. 

Should this be STORAGE_URL?

self-hosting/.env.example
stopachka: I wonder then if we should make this more generic, since passing in s3 credentials isn't MINIO

self-hosting/.env.example
```diff
@@ -0,0 +1,36 @@
...
+MINIO_ROOT_PASSWORD=minioadmin
+S3_BUCKET=instant-bucket
+S3_ENDPOINT=http://minio:9000
+AWS_REGION=us-east-1
+
+# Optional integrations.
+STRIPE_API_KEY=
+
+# Email. Provide one of these provider API keys to send login magic codes.
+POSTMARK_TOKEN=
```
stopachka, line 35: nice!

self-hosting/.env.example
```diff
@@ -0,0 +1,36 @@
...
+MINIO_ROOT_USER=minioadmin
+MINIO_ROOT_PASSWORD=minioadmin
+S3_BUCKET=instant-bucket
+S3_ENDPOINT=http://minio:9000
+AWS_REGION=us-east-1
+
+# Optional integrations.
+STRIPE_API_KEY=
+
+# Email. Provide one of these provider API keys to send login magic codes.
```
stopachka: `these provider` doesn't make sense here -- imo we just say Postmark?

client/www/next.config.js
drew-harris: Standalone build output helps reduce the size of the final image. [nextjs.org/docs/pages/api-reference/config/next-config-js/output](https://nextjs.org/docs/pages/api-reference/config/next-config-js/output)

Ive updated it so that this only gets set when using the NEXT_PUBLIC_SELF_HOSTED variable so it will only run for docker builds.

self-hosting/README.md
```diff
@@ -0,0 +1,77 @@
+# Self Hosting InstantDB
```
stopachka, line 1: I think we may want to structure this differently. 

https://www.dropbox.com/scl/fi/gvvkvfnn5lswnq87uii6i/Idealized-Instant-Self-Hosting.paper?rlkey=alerpayieal1zupiza1fg74cq&dl=0

Some ideas I had here: 

1. Ideally we start the user off with something they can run in one go, using docker compose up. I got this idea form look at [HyperDX](https://www.hyperdx.io/docs/v2/local#self-hosted-version) 
2. After that we can go into a production guide. There the two ideas I added: (a) a note about using something like terraform (maybe we provide this for the user? It could be a nice win later if not now) (b) telling them about sslip.io 


I asked Codex "soup up" my idealized version and it came up with this: 

https://gist.github.com/stopachka/b0b2820bd23d4ecf84cf7a845ada3cc2

(I don't 100% recommend his version, but wanted to share)

server/src/instant/postmark.clj
```diff
@@ -45,13 +46,15 @@
...
     (if-not (config/postmark-send-enabled?)
-      (tracer/with-span! {:name "postmark/send-disabled"
-                          :attributes body}
-        (tracer/record-info!
-         {:name "postmark-disabled"
-          :attributes
-          {:msg
-           "Postmark is disabled, add postmark-token to config to enable"}}))
+      (do
+        (log/infof "Postmark disabled; would send email: %s" (pr-str body))
```
stopachka: Why did we switch to log/infof?

self-hosting/.env.example
drew-harris: Fixed (comment was written before i realized we didn't use sendgrid)

server/src/instant/postmark.clj
drew-harris: Removed. Had an issue earlier where it wasn't printing the body properly but unable to reproduce.

client/www/pages/_document.tsx
```diff
@@ -48,6 +49,11 @@ export default function Document() {
         />
       </Head>
       <body>
+        <Script
+          onError={() => {}}
```
stopachka: should we add isSelfHosted check here, so no need to load this script in main?

client/Dockerfile.www
```diff
@@ -0,0 +1,119 @@
...
+  websocketURL.hash = '';
+
+  return {
+    apiURI,
+    websocketURI: websocketURL.toString(),
+  };
+}
+
+const config = configFromApiURI(
+  process.env.INSTANT_API_URI || process.env.INSTANT_BACKEND_URL,
```
stopachka: would just store this value rather than the config, so no need for duplicated fn

server/src/instant/dash/routes.clj
```diff
@@ -52,7 +52,7 @@
             [instant.plans :as plans]
             [instant.postmark :as postmark]
             [instant.runtime.magic-code-auth :refer [check-send-rate-limit!
-                                                     check-verify-rate-limit!]]
+                                                      check-verify-rate-limit!]]
```
```suggestion
                                                     check-verify-rate-limit!]]
```

stopachka (approved):
Looks good to me, apart from the nits discussed! 

The parts here on docker will leave to @dwwoelfel to stamp!

server/src/instant/dash/routes.clj
```diff
@@ -204,15 +204,17 @@
 ;; Magic Codes
 
 (defn magic-code-email [{:keys [user magic-code]}]
-  (let [title "Instant"
+  (let [{sender-name :name sender-email :email} (config/dashboard-email-sender)
+        title "Instant"
```
dwwoelfel: Should title be `sender-name`?

server/src/instant/dash/routes.clj
```diff
@@ -247,7 +249,7 @@
                      :code (instant-user-magic-code-model/rand-code)
                      :user-id user-id})]
     (postmark/send-structured!
-     (magic-code-email {:user u :magic-code magic-code}))
+      (magic-code-email {:user u :magic-code magic-code}))
```
dwwoelfel: Do you have cljfmt set up with your editor? Seeing a lot of random whitespace changes.

server/src/instant/storage/s3.clj
```diff
@@ -40,23 +61,27 @@
...
+          creds (if (and access-key secret-key)
+                  {:access-key access-key
+                   :secret-key secret-key
+                   :region region}
+                  (let [creds (.resolveCredentials (.build (DefaultCredentialsProvider/builder)))]
+                    {:access-key (.accessKeyId creds)
+                     :secret-key (.secretAccessKey creds)
+                     :region region}))]
+      (cond-> creds
+        (config/s3-endpoint)
```
dwwoelfel: Should this check for `config/s3-public-endpoint`?

client/Dockerfile.www
```diff
@@ -0,0 +1,96 @@
...
+# Install project dependencies with frozen lockfile for reproducible builds
+RUN --mount=type=cache,target=/root/.npm \
+    --mount=type=cache,target=/usr/local/share/.cache/yarn \
+    --mount=type=cache,target=/root/.local/share/pnpm/store \
+    if [ -f package-lock.json ]; then \
+    npm ci --no-audit --no-fund; \
+    elif [ -f yarn.lock ]; then \
+    corepack enable yarn && yarn install --frozen-lockfile --production=false; \
+    elif [ -f pnpm-lock.yaml ]; then \
+    corepack enable pnpm && pnpm install --frozen-lockfile; \
```
dwwoelfel: We know it's pnpm, so we could skip all of this and just do the pnpm step.

server/src/instant/config.clj
```diff
@@ -93,14 +93,51 @@
...
+      "hello@instantdb.com"))
+
+(defn dashboard-email-sender []
+  {:name (or (System/getenv "INSTANT_DASHBOARD_EMAIL_SENDER_NAME")
+             "Instant")
+   :email (or (System/getenv "INSTANT_DASHBOARD_EMAIL_SENDER_EMAIL")
+              "verify@dash-pm.instantdb.com")})
+
+(defn app-email-sender []
+  {:name (System/getenv "INSTANT_APP_EMAIL_SENDER_NAME")
```
dwwoelfel, line 133: Should this get a default of "Instant"?

server/src/instant/runtime/magic_code_auth.clj
```diff
@@ -125,16 +126,17 @@
...
-        sender-email    (or (:email template) default-sender)
+        sender-email    (or (:email template) default-sender-email)
         email-params    (if template
                           {:sender-email sender-email
                            :sender-name (or (:name template) (:title app))
                            :subject (template-replace (:subject template) template-params)
                            :body (template-replace (:body template) template-params)}
-                          {:sender-name (:title app)
-                           :sender-email default-sender
+                          {:sender-name (or default-sender-name (:title app))
```
dwwoelfel: Why do we put a default sender name here?

server/src/instant/runtime/magic_code_auth.clj
```diff
@@ -146,15 +148,13 @@
                               (invalid-sender? e)
                               (do
                                 (tracer/record-info! {:name "magic-code/unconfirmed-or-unknown-sender" :attributes {:email sender-email :app-id app-id}})
-                                (postmark/send-structured! (magic-code-email email (assoc email-params :sender-email default-sender))))
+                                (postmark/send-structured! (magic-code-email email (assoc email-params :sender-email default-sender-email))))
 
-
-                              ;; Don't throw if it's a test user, even if we can't send email to it
+;; Don't throw if it's a test user, even if we can't send email to it
```
dwwoelfel: The comment got pushed too far to the left

server/src/instant/config.clj
```diff
@@ -93,14 +93,51 @@
 (defn s3-storage-secret-key []
   (some-> @config-map :s3-storage-secret-key crypt-util/secret-value))
 
+(defn s3-endpoint []
```
dwwoelfel: It would be better to make these delays so that we didn't have to evaluate the function every time, e.g.

```clj
(def s3-endpoint 
  (delay
    (or (System/getenv "S3_ENDPOINT")
        (some-> @config-map :s3-endpoint))))
```

And then you would access it as `@s3-endpoint` instead of `(s3-endpoint)`

server/Dockerfile.self-hosted
```diff
@@ -0,0 +1,70 @@
...
+RUN clojure -T:build uber
+
+FROM amazoncorretto:26
+
+WORKDIR /app
+
+# Install dependencies and golang-migrate
+RUN yum -y install tar gzip && \
+    curl -L https://github.com/golang-migrate/migrate/releases/download/v4.17.0/migrate.linux-amd64.tar.gz | tar xvz && \
+    mv migrate /usr/local/bin/migrate && \
```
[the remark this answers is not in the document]
dwwoelfel: This seems like a good idea

dwwoelfel (approved):
Left some comment, but LGTM overall

PR #2526, opened 2026-04-09
Move useLocalStorage to demo component, less noisy useLocalStorage hook

dwwoelfel:
Makes the architecture demos more self-contained. They now keep track of useLocalStorage instead of polluting the essay markdown component.

Also makes `useLocalStorage` only trigger the callback when the key we're watching changes.


stopachka (approved):
Heeck yeah!

PR #2533, opened 2026-04-10
fix dark mode login page

drew-harris:
If a user has previously logged in and toggled the dark mode on the dashboard to be on. When they log out, they will see a bad light mode version when logging back in. Fixing this by adding a background dark mode variant. 

BEFORE:
image

AFTER:
image


dwwoelfel (approved):
LGTM

PR #2534, opened 2026-04-10
fix oauth inputs with password managers

drew-harris:
add new data-1p-ignore and data-bwignore tags for 1Password and Bitwarden

The bitwarden ignore tag still does not work for me


stopachka (approved):
Niice!

PR #2535, opened 2026-04-11
Fix optimistic adds

nezaj:
**Problem**
Optimistic adds were not showing up when a query had an `order` clause and existing data.

**Cause**
`runDataloadAndReturnObjects` in `instaql.ts` unconditionally filtered entities that fell after the server's `end-cursor`.

For leading queries (no pagination), the `start-cursor` filter correctly checked `!isLeadingQuery` first, but the `end-cursor` filter did not. This meant any optimistically-added entity with a newer order value (e.g. `createdAt: Date.now()`) was silently excluded from results.

**Fix**
The fix adds the same `!isLeadingQuery` guard to the `end-cursor` check, matching the `start-cursor` behavior. This was an oversight from #2334 (infinite scroll).

Adds a test too!

### DIFF
--- a/client/packages/core/__tests__/src/instaql.test.ts
+++ b/client/packages/core/__tests__/src/instaql.test.ts
@@ -942,6 +942,61 @@ test('Leading queries should ignore the start cursor', () => {
   expect(usersWithBob).toEqual(['bob', 'nicolegf']);
 });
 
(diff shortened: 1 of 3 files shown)


stopachka (approved):
SGTM!

PR #2537, opened 2026-04-11
Make @instantdb/admin work in the edge runtime

dwwoelfel:
This PR allows @instantdb/admin to work in vercel's edge runtime (should also work in Cloudflare, but I haven't tested).

The edge runtime supports almost everything that `eventsource` relies on, except for `MessageEvent`. And you can't just do a normal polyfill with `globalThis.MessageEvent = globalThis.MessageEvent || PolyFill`, because they define one that just throws at runtime. So I forked `eventsource` to [`@instantdb/eventsource`](https://github.com/instantdb/eventsource) and added a way to pass in our own `MessageEvent`.

I also added `streamId` to the read stream we return. We already had it on the writeStream, but not the readStream. I use that in resumable-stream to return undefined when the stream doesn't exist, matching vercel's behavior.

### DIFF
--- a/client/packages/core/src/Stream.ts
+++ b/client/packages/core/src/Stream.ts
@@ -22,6 +22,10 @@ export interface InstantWritableStream<T> extends WritableStream<T> {
   streamId: () => Promise<string>;
 }
 
+export interface InstantReadableStream<T> extends ReadableStream<T> {
+  streamId: () => Promise<string>;
+}
+
 type WriteStreamStartResult =
   | { type: 'ok'; streamId: string; offset: number }
   | { type: 'disconnect' }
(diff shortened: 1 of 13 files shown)


client/packages/core/src/Stream.ts
```diff
@@ -1005,22 +1075,21 @@ export class InstantStream {
...
 
-    if (msg.files?.length || msg.content) {
-      iterator.push({
-        type: 'append',
-        offset: msg.offset,
-        files: msg.files,
-        content: msg.content,
-      });
-    }
+    iterator.push({
```
dwwoelfel, line 1078: This makes sure that we resolve `streamId` on the enhanced reader, even if the writer hasn't written anything. The server will always send a `stream-append` when we subscribe, even if there is no content.

stopachka (approved):
SGTM!

PR #2540, opened 2026-04-11
Quality: Create permission failure capture system

### LINKED ISSUE #1658: permission browser tool
I was on a call with @nezaj and was discussing my permission-testing process, which involves copying a failing permission, pasting it into the sandbox, recreating the permissions relevant to the query, and messing around.

This workflow is a pain in the ass for a few reasons. It would be amazing to have a browser tool that just pulled the relevant info from a failing query and pasted it into the browser sandbox to make it really easy to test out permissions

**Detail**: I'd also request that the permissions editor is Javascript that returns a permissions object, instead of JSON. That would be really helpful because
- My permissions file is JS not JSON, so copying is a pain
- My permissions file has some string literals that slow me down too
- It's useful to have comments and string literals while testing this stuff

**How?**
- Can you create a list of network requests to instant to find queries that failed for permissions reasons, and allow user to select which one they want to test in Sandbox?
- Or, let user add a debugging flag to the query they are interested in

andoan16:
## Problem

Create a new module that captures permission check failures from API responses, storing the query context, current rules, auth state, and error details. This enables the permission browser tool to display historical failures for debugging.

**Severity**: `high`
**File**: `client/packages/core/src/permissionCapture.ts`

## Solution

|

## Changes

- `client/packages/core/src/permissionCapture.ts` (new)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced

Closes #1658

### DIFF
--- a/client/packages/core/src/permissionCapture.ts
+++ b/client/packages/core/src/permissionCapture.ts
@@ -1,0 +1,49 @@
+import id from './utils/id';
+import type { Query } from './queryTypes';
+import type { AuthState } from './clientTypes';
+import type { InstantRules } from './schemaTypes';
+
+export interface PermissionFailure {
+  id: string;
+  timestamp: number;
+  query: Query;
+  rules: InstantRules;
+  authState: AuthState | null;
+  errorMessage: string;
+  errorCode?: string;
+}
+
+const MAX_CAPTURED_FAILURES = 100;
+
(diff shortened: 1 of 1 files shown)


nezaj:
Capturing permissions/queries/transactions could be really helpful for observability! Marking this as a request for us to discuss more!

Issue #2541, opened 2026-04-11
InvalidStateError: Failed to execute 'transaction' on 'IDBDatabase': The database connection is closing.


kenrick95:
Very low frequency error caught in my app's Sentry project (30 events over the past ~3 months), not sure if worth pursuing. The only similarities are these are only happening in iOS/MacOS Safari.

```
InvalidStateError: Failed to execute 'transaction' on 'IDBDatabase': The database connection is closing.
    at transaction ([native code])
    at Promise (../../../node_modules/.pnpm/@instantdb+core@0.22.179/node_modules/@instantdb/core/dist/esm/IndexedDBStorage.js:243:47)
    at Promise ([native code])
    at tE.multiSet (../../../node_modules/.pnpm/@instantdb+core@0.22.179/node_modules/@instantdb/core/dist/esm/IndexedDBStorage.js:242:20)

```

```
    // Performs all writes in a transaction so that all succeed or none succeed.
    async multiSet(keyValuePairs) {
        const db = await this._dbPromise;
        return new Promise((resolve, reject) => {
            const transaction = db.transaction([this._storeName], 'readwrite'); // <--- this part throws?!
            const objectStore = transaction.objectStore(this._storeName);
            const requests = new Set();
            for (const [k, v] of keyValuePairs) {
                const request = objectStore.put(v, k);
                requests.add(request);
```

nezaj:
Thanks for reporting! This should be fixed in https://github.com/instantdb/instant/pull/2548

You can get this fix by upgrading to 1.0.3 and higher!

PR #2548, opened 2026-04-13
Fix IndexedDB InvalidStateError when browser closes connection

nezaj:
I've seen this error before too and saw a report here https://github.com/instantdb/instant/issues/2541

Talked w/ Claude about it and it proposed adding a retry !

### DIFF
--- a/client/packages/core/src/IndexedDBStorage.ts
+++ b/client/packages/core/src/IndexedDBStorage.ts
@@ -210,6 +210,15 @@ export default class IndexedDBStorage extends StoreInterface {
       request.onsuccess = (event) => {
         const target = event.target as IDBOpenDBRequest;
         const db = target.result;
+        // Browsers can close IndexedDB connections unexpectedly
+        // (e.g. backgrounded tabs, memory pressure, version changes from
+        // other tabs). Re-init so the next operation gets a fresh connection.
+        db.onclose = () => {
+          this._dbPromise = this._init();
+        };
+        db.onversionchange = () => {
+          db.close();
+        };
         if (!requiresUpgrade) {
           const p = upgradePromises.get(this.dbName);
           if (!p) {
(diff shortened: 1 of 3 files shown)


client/packages/core/src/IndexedDBStorage.ts
```diff
@@ -243,98 +252,101 @@ export default class IndexedDBStorage extends StoreInterface {
...
+    try {
+      const db = await this._dbPromise;
+      return await fn(db);
+    } catch (e) {
+      if (e instanceof DOMException && e.name === 'InvalidStateError') {
+        this._dbPromise = this._init();
+        const db = await this._dbPromise;
+        return await fn(db);
+      }
+      throw e;
```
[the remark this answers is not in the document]
nezaj: Not an issue, catching the abort won't help

client/packages/core/src/IndexedDBStorage.ts
```diff
@@ -210,6 +210,15 @@ export default class IndexedDBStorage extends StoreInterface {
       request.onsuccess = (event) => {
         const target = event.target as IDBOpenDBRequest;
         const db = target.result;
+        // Browsers can close IndexedDB connections unexpectedly
+        // (e.g. backgrounded tabs, memory pressure, version changes from
+        // other tabs). Re-init so the next operation gets a fresh connection.
+        db.onclose = () => {
+          this._dbPromise = this._init();
+        };
+        db.onversionchange = () => {
```
dwwoelfel, line 219: Why do we close on version change?

client/packages/core/src/IndexedDBStorage.ts
```diff
@@ -243,98 +252,101 @@ export default class IndexedDBStorage extends StoreInterface {
...
+  // memory pressure, cross-tab version changes, etc.), causing
+  // `db.transaction()` to throw InvalidStateError. This helper catches that
+  // error and retries once with a fresh connection.
+  async _withRetry<T>(fn: (db: IDBDatabase) => Promise<T>): Promise<T> {
+    try {
+      const db = await this._dbPromise;
+      return await fn(db);
+    } catch (e) {
+      if (e instanceof DOMException && e.name === 'InvalidStateError') {
+        this._dbPromise = this._init();
```
dwwoelfel, line 265: Should we close the current db here or check if it's closed? If there is an error that didn't close the db, we might end up with a bunch of indexeddb instances in the background.

client/packages/core/src/IndexedDBStorage.ts
```diff
@@ -243,98 +252,101 @@ export default class IndexedDBStorage extends StoreInterface {
...
-        };
-        request.onsuccess = (_event) => {
-          requests.delete(request);
-          // Last request to finish resolves the transaction
-          if (requests.size === 0) {
-            resolve();
-          }
-        };
-      }
+        transaction.oncomplete = () => resolve();
```
dwwoelfel, line 316: That's handy--good to know you can just wait on the transaction.

client/packages/core/src/IndexedDBStorage.ts
nezaj: We don't need it specifically for this fix. But Claude highlighted it as a good practice for idb if anything ever triggers a version change (e.g. deleteDatabase from devtools, a browser extension, etc.) we'd block the upgrade and the browser could force-close us, causing the same InvalidStateError

Definitely not likely to happen so can remove if we like!

client/packages/core/src/IndexedDBStorage.ts
nezaj: We shouldn't need to, If we're catching InvalidStateError from db.transaction(), it means the connection's closePending flag is already set. The browser will fully close it once any in-flight transactions complete.

dwwoelfel (approved):
LGTM!

PR #2549, opened 2026-04-13
Remove global ISR

dwwoelfel:
We had revalidate set to 3600 on the app layout, which causes everything in the app router to use ISR. I don't think we need it and it should solve our transient `/docs` 404.

https://instant-www-js-no-docs-revalidate-jsv.vercel.app/docs


stopachka (approved):
Cool!

PR #2551, opened 2026-04-13
fix dashboard dark mode auth collapse hover

drew-harris:
BEFORE:
image

AFTER:
image

Applied same change to every provider.


stopachka (approved):
Nice!

PR #2552, opened 2026-04-13
remove dead code in dashboard

drew-harris:
Each oauth provider type had an unused component. I removed them and then renamed the one actually being used to avoid renaming exports.


nezaj (approved):
Nice!

PR #2553, opened 2026-04-13
[dashboard] shared dev credentials for Google

stopachka:
**This PR implements the dashboard for shared dev credentials!**

For the backend PR, check out: https://github.com/instantdb/instant/pull/2588

**Note:** I will have a follow-on PR shortly for the CLI, and for docs

## Testing

1. Load https://instant-www-js-sharedcred-jsv.vercel.app/, and try creating a Google Oauth client!
  1. You can use this to test creating shared oauth clients, and upgrading them!
2. To test E2E: 
   1. `bunx create-instant-app`, and ask your agent to implement Google Web Redirect. Tell them you've made a client aready, and give them your client name (i.e google-web) 

## Product Context

Say you want to add Google Oauth. 

**Problem:** Even to test in development, you have to go through Google Cloud Console and create an Oauth client. Google Cloud Console is quite annoying to navigate, which means poor devex for us.

**Solution:** We will provide default oauth credentials for development!

Users don't have to go through Google Cloud Console. They can Oauth Clients without providing `client-id` and `client-secret` values. In development, Instant will handle those values for them. We'll also default allowed origins to `localhost` and `exp`, so they don't have to create those redirect origins either.

**This has some limits though:**
- **Only works for apps who have less than 100 users.** 
  - This is _strictly_ meant for development. To make sure, **clients with dev credentials will start to throw after 100 sign-ups**. 
  - It's also good note that when the user signs in, the consent screen says "Instant Shared Dev Credentials"
- **Native Flows are not supported**
  - Right now this works for web flows, but not Native IOS, or Google Button. Those experiences require the user to know a `client-id`. 
    - We chose not to share a `client-id` with the user, in case we end up needing to change the oauth client. 
    - Since this is dev only though, perhaps it's not too big of a deal. We may change our decision about this later.

## User Experience

Now the user can select "Use shared credentials". 

This lets them create a google client without a `client-id` or `client-secret`:

image

During runtime, we can fill those values in for them. 

Once they are ready to go to production, they can insert their product `client-id` and `client-secret`: 

image

### DIFF
--- a/client/packages/components/src/components/ui.tsx
+++ b/client/packages/components/src/components/ui.tsx
@@ -1108,7 +1108,7 @@ export function Copytext({ value }: { value: string }) {
   const [showCopied, setShowCopied] = useState(false);
 
   return (
-    <span className="inline-flex items-center rounded-sm bg-gray-500 px-2 text-sm text-white">
+    <span className="inline-flex items-center rounded-sm bg-gray-100 px-2 text-sm text-gray-800 dark:bg-neutral-700 dark:text-neutral-200">
       <code
         className="truncate"
         onClick={(e) => {
--- a/client/www/components/dash/auth/GitHub.tsx
+++ b/client/www/components/dash/auth/GitHub.tsx
@@ -156,7 +156,6 @@ export function AddGitHubClientForm({
         clientId,
         clientSecret,
         redirectTo,
-        meta: { providerName: 'github' },
       });
       onAddClient(resp.client);
     } catch (e) {
--- a/client/www/components/dash/auth/Google.tsx
+++ b/client/www/components/dash/auth/Google.tsx
@@ -1,5 +1,5 @@
 import { FormEventHandler, useState, useContext } from 'react';
-import { errorToast } from '@/lib/toast';
+import { errorToast, successToast } from '@/lib/toast';
 import { TokenContext } from '@/lib/contexts';
 import {
   InstantApp,
@@ -15,6 +15,7 @@ import {
   RedirectUrlInput,
   EditableRedirectUrl,
   TestRedirectButton,
+  updateClient,
 } from './shared';
 import { messageFromInstantError } from '@/lib/errors';
 import {
@@ -42,16 +43,6 @@ import {
 } from '@heroicons/react/24/solid';
 import { useDarkMode } from '../DarkModeToggle';
 
-function NonceCheckNotice() {
-  return (
-    <p className="dark:test-neutral-400 text-sm text-gray-500">
-      This option skips nonce checks for ID tokens. This is useful in iOS
-      environments, because libraries like `react-native-google-signin` do not
-      let you pass a nonce over to google.
-    </p>
-  );
-}
-
 type AppType = 'web' | 'ios' | 'android' | 'button-for-web';
 function isNative(appType: AppType) {
   return appType === 'ios' || appType === 'android';
@@ -71,9 +62,11 @@ export function AddGoogleClientForm({
   usedClientNames: Set<string>;
 }) {
   const token = useContext(TokenContext);
+  const [credentialMode, setCredentialMode] = useState<'dev' | 'custom'>('dev');
   const [appType, setAppType] = useState<
     'web' | 'ios' | 'android' | 'button-for-web'
   >('web');
+  const useSharedCredentials = appType === 'web' && credentialMode === 'dev';
   const [clientName, setClientName] = useState<string>(() =>
     findName(`google-${appType}`, usedClientNames),
   );
@@ -81,14 +74,12 @@ export function AddGoogleClientForm({
   const [clientSecret, setClientSecret] = useState<string>('');
   const [redirectTo, setRedirectTo] = useState<string>('');
   const [updatedRedirectURL, setUpdatedRedirectURL] = useState(false);
-  const [skipNonceChecks, setSkipNonceChecks] = useState(isNative(appType));
   const [isLoading, setIsLoading] = useState<boolean>(false);
 
   const onChangeAppType = (item: { id: string; label: string }) => {
     const newAppType = item.id as 'web' | 'ios' | 'android' | 'button-for-web';
     setAppType(newAppType);
     setClientName(findName(`google-${newAppType}`, usedClientNames));
-    setSkipNonceChecks(isNative(newAppType));
   };
 
   const validationError = () => {
@@ -98,11 +89,13 @@ export function AddGoogleClientForm({
     if (usedClientNames.has(clientName)) {
       return `The unique name '${clientName}' is already in use.`;
     }
-    if (!clientId) {
-      return 'Missing client id';
-    }
-    if (appType === 'web' && !clientSecret) {
-      return 'Missing client secret';
+    if (!useSharedCredentials) {
+      if (!clientId) {
+        return 'Missing client id';
+      }
+      if (appType === 'web' && !clientSecret) {
+        return 'Missing client secret';
+      }
     }
   };
 
@@ -120,15 +113,20 @@ export function AddGoogleClientForm({
         appId: app.id,
         providerId: provider.id,
         clientName,
-        clientId,
-        clientSecret: clientSecret ? clientSecret : undefined,
+        clientId: useSharedCredentials ? undefined : clientId,
+        clientSecret: useSharedCredentials
+          ? undefined
+          : clientSecret
+            ? clientSecret
+            : undefined,
         authorizationEndpoint: 'https://accounts.google.com/o/oauth2/v2/auth',
         tokenEndpoint: 'https://oauth2.googleapis.com/token',
         discoveryEndpoint:
           'https://accounts.google.com/.well-known/openid-configuration',
-        redirectTo,
+        redirectTo: useSharedCredentials ? undefined : redirectTo,
+        useSharedCredentials,
         meta: {
-          skipNonceChecks: skipNonceChecks,
+          skipNonceChecks: isNative(appType),
           appType,
         },
       });
@@ -149,6 +147,9 @@ export function AddGoogleClientForm({
       onSubmit={onSubmit}
       autoComplete="off"
       data-lpignore="true"
+      data-1p-ignore="true"
+      data-bwignore="true"
+      data-form-type="other"
     >
       <SubsectionHeading>Add a new Google client</SubsectionHeading>
       <div className="mb-4">
@@ -167,6 +168,19 @@ export function AddGoogleClientForm({
           ariaLabel="Application type"
         />
       </div>
+      {appType === 'web' && (
+        <div className="mb-2">
+          <ToggleGroup
+            items={[
+              { id: 'dev', label: 'Use dev credentials' },
+              { id: 'custom', label: 'Use my own' },
+            ]}
+            selectedId={credentialMode}
+            onChange={({ id }) => setCredentialMode(id as 'dev' | 'custom')}
+            ariaLabel="Credential mode"
+          />
+        </div>
+      )}
       <TextInput
         tabIndex={1}
         value={clientName}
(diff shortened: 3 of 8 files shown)


server/src/instant/model/app_authorized_redirect_origin.clj
```diff
@@ -68,6 +68,17 @@
   (and (not (contains? reserved-uri-schemes uri-scheme))
        (= scheme uri-scheme)))
 
+(defn shared-credential-match [url]
```
stopachka: Maybe this should be something like `shared-credentials-default-match`. 

We should add a comment block in here, saying why: 

When users choose shared credentials, we automatically allow localhost and expo dev, without them having to set those explicitly in their allowed origins. 

For more custom urls, they can always set them in allowed origins

server/src/instant/model/app_oauth_client.clj
```diff
@@ -119,26 +123,56 @@
...
-      :client-id (:client_id oauth-client)
-      :client-secret (when (:client_secret oauth-client)
-                       (decrypted-client-secret oauth-client))
-      :meta (:meta oauth-client)})
-
-    :else
-    (throw (ex-info "Unsupported OAuth client" {:oauth-client oauth-client}))))
+  (let [provider-name (get (:meta oauth-client) "providerName")
+        use-shared-credential? (get (:meta oauth-client) "useSharedCredentials")
+        oauth-client (if use-shared-credential?
```
stopachka: maybe we do if-not here?

server/src/instant/runtime/routes.clj
```diff
@@ -206,11 +206,14 @@
                                     :let [value (get-in req [:params param])]
                                     :when value]
                                 [param value]))
+        use-shared-credentials? (get (:meta client) "useSharedCredentials")
```
stopachka: small nit, but imo we should have a use-shared-credentials? function in app-oauth-client, feels brittle to be querying meta like this in multiple places

server/src/instant/runtime/routes.clj
```diff
@@ -289,17 +292,17 @@
                        :id (or guest-user-id (random-uuid))
                        :extra-fields extra-fields})
             new-user (app-user-model/create!
```
stopachka, line 309: Let's add a check here. something like when use-shared-credentials, assert less than 100 users. 

we may want this wrapped in a nice function, like assert-shared-credentials-allowed or something. 

I want this abstracted, because we should _also_ include this check when a user is _creating_ a new client. 

it should throw a nice 400 error

server/src/instant/runtime/routes.clj
```diff
@@ -206,11 +206,14 @@
                                     :let [value (get-in req [:params param])]
                                     :when value]
                                 [param value]))
+        use-shared-credentials? (get (:meta client) "useSharedCredentials")
         authorized-origins (app-authorized-redirect-origin-model/get-all-for-app
```
stopachka: I do wonder if we should have a single function like `assert-authorized origin`, which we can re-use here and in our check below.

client/www/components/dash/auth/Google.tsx
```diff
@@ -384,6 +430,51 @@ export function GoogleClient({
...
+      errorToast('Missing client id', { autoClose: 5000 });
+      return;
+    }
+    if (!upgradeClientSecret) {
+      errorToast('Missing client secret', { autoClose: 5000 });
+      return;
+    }
+    try {
+      setIsLoading(true);
+      await deleteClient({
```
stopachka: This is a bad idea. it can lead to a temporary break in service. imo we should add an update-client endpoint, which can take a client name, and update _only_ the clientId and clientSecret.

client/www/components/dash/auth/Google.tsx
```diff
@@ -120,16 +125,20 @@ export function AddGoogleClientForm({
...
         tokenEndpoint: 'https://oauth2.googleapis.com/token',
         discoveryEndpoint:
           'https://accounts.google.com/.well-known/openid-configuration',
-        redirectTo,
+        redirectTo: useSharedCredentials ? undefined : redirectTo,
         meta: {
           skipNonceChecks: skipNonceChecks,
           appType,
+          ...(useSharedCredentials
+            ? { useSharedCredentials: true, providerName: 'google' }
```
stopachka: Curious, why are we passing in `providerName`? That should not be necessary

client/www/components/dash/auth/Google.tsx
```diff
@@ -175,90 +197,108 @@ export function AddGoogleClientForm({
...
-            </a>
-          </>
-        }
-        placeholder=""
-      />
+      {useSharedCredentials ? (
+        <div className="rounded-sm bg-gray-50 p-3 text-sm text-gray-600 dark:bg-neutral-800 dark:text-neutral-400">
+          <p>
+            Instant provides dev credentials so you can test Google sign-in on{' '}
+            <span className="font-medium">localhost</span> without any setup.
```
stopachka: We should say in development, since it can work with Expo too

client/www/components/dash/auth/Google.tsx
```diff
@@ -351,6 +391,9 @@ export function GoogleClient({
   const token = useContext(TokenContext);
   const [open, setOpen] = useState(defaultOpen);
   const [isLoading, setIsLoading] = useState(false);
+  const [showUpgrade, setShowUpgrade] = useState(false);
```
stopachka, line 367: Why do we need a state for this? Can't this be derived from meta?

server/src/instant/dash/routes.clj
```diff
@@ -628,6 +628,9 @@
         discovery-endpoint (when-not (= "github" provider-name)
                              (ex/get-param! req [:body :discovery_endpoint] string-util/coerce-non-blank-str))
 
+        _ (when (get meta "useSharedCredentials")
```
stopachka: we should use a lil helper like (uses-shared-credentials? client) that checks get meta

server/src/instant/dash/routes.clj
```diff
@@ -628,6 +628,9 @@
...
+            (app-oauth-client-model/assert-shared-credentials-allowed! {:app-id app-id}))
```
stopachka: would just pass in app-id as positional arg

server/src/instant/dash/routes.clj
```diff
@@ -644,17 +647,26 @@
...
         redirect-to (-> req :body :redirect_to string-util/coerce-non-blank-str)
+        client-id (ex/get-optional-param! req [:body :client_id] string-util/coerce-non-blank-str)
+        client-secret (ex/get-optional-param! req [:body :client_secret] string-util/coerce-non-blank-str)
         _ (when redirect-to
             (ex/assert-valid!
              :redirect_to
              redirect-to
              (url-util/redirect-url-validation-errors
               redirect-to :allow-localhost? true)))
+        _ (when (get meta "useSharedCredentials")
```
stopachka: same

server/src/instant/model/app_authorized_redirect_origin.clj
```diff
@@ -68,6 +68,23 @@
   (and (not (contains? reserved-uri-schemes uri-scheme))
        (= scheme uri-scheme)))
 
+;; When users choose shared credentials, we automatically allow
+;; localhost (http/https) and expo dev (exp://) as redirect targets,
+;; without them having to set those explicitly in their allowed
+;; origins. For custom domains or schemes, they can still add them to
+;; allowed origins.
```
stopachka: this should be in a "" block for the function def

server/src/instant/model/app_user.clj
```diff
@@ -111,6 +113,20 @@
              (fn [{:keys [get-entity]}]
                (get-entity id)))))
 
+;; Bounded existence check: returns true when the app has at least
```
stopachka: move this into "" block for fn

server/src/instant/model/app_user.clj
```diff
@@ -111,6 +113,20 @@
...
+;; `n` users. Uses the deterministic `$users.id` attr_id, and LIMIT
+;; (inc n) so we don't pay for a full COUNT on large apps.
+(defn users-at-least?
+  ([params] (users-at-least? (aurora/conn-pool :read) params))
+  ([conn {:keys [app-id n]}]
+   (let [attr-id (system-catalog/get-attr-id "$users" "id")
+         rows (sql/select conn
```
stopachka: I wonder if we can use our hsql preformat -- imo it's cooler

client/www/components/dash/auth/Google.tsx
```diff
@@ -120,16 +125,18 @@ export function AddGoogleClientForm({
         appId: app.id,
         providerId: provider.id,
         clientName,
-        clientId,
-        clientSecret: clientSecret ? clientSecret : undefined,
+        clientId: useSharedCredentials ? undefined : clientId,
+        clientSecret:
```
stopachka: let's rewrite this as a triple ternary, useSharedCredentials ? undefined : clientSecret ? clientSecret : undefined

client/packages/components/src/components/ui.tsx
```diff
@@ -231,14 +231,21 @@ export function TextInput({
...
-        // Try to prevent password managers from trying to save
-        // sensitive input, LastPass, 1password, BitWarden
-        autoComplete={type === 'sensitive' ? 'off' : undefined}
-        data-lpignore={type === 'sensitive' ? 'true' : undefined}
-        data-1p-ignore={type === 'sensitive' ? 'true' : undefined}
-        data-bwignore={type === 'sensitive' ? 'true' : undefined}
-        data-form-type={type === 'sensitive' ? 'other' : undefined}
+        // Try to prevent password managers (LastPass, 1Password,
+        // BitWarden) from attaching to or saving sensitive input.
+        autoComplete={type === 'sensitive' ? 'new-password' : undefined}
```
stopachka: can you include a note explaining what 'new-password' is meant to do?

client/www/components/dash/auth/shared.tsx
```diff
@@ -168,6 +168,33 @@ export function updateClientRedirectTo({
...
 
+export function updateClient({
+  token,
+  appId,
+  clientDatabaseId,
+  ...body
+}: {
+  token: string;
+  appId: string;
+  clientDatabaseId: string;
```
stopachka: let's call this oauthClientID

client/www/components/dash/auth/shared.tsx
```diff
@@ -168,6 +168,33 @@ export function updateClientRedirectTo({
...
+  client_id?: string;
```
stopachka: maybe these things can be in a `body` field, not a splice.

server/src/instant/dash/routes.clj
```diff
@@ -622,13 +622,18 @@
              redirect-to
              (url-util/redirect-url-validation-errors
               redirect-to :allow-localhost? true)))
-        provider-name (ex/get-optional-param! meta [:providerName] string-util/coerce-non-blank-str)
+        provider-name (app-oauth-client-model/get-provider-name-by-id!
```
stopachka: I think `get-provider-name-by-id!` feels a bit too overspecificed. Let's have this be `get-provider-by-id!` and just extract that provider-name with destructuring. do this in other places where we use this

server/src/instant/model/app_oauth_client.clj
```diff
@@ -12,6 +15,56 @@
...
+  (when (app-user-model/users-at-least? {:app-id app-id
+                                         :n shared-credentials-user-limit})
+    (ex/throw-validation-err!
+     :shared-credentials
+     app-id
+     [{:message (str "Shared dev credentials are limited to "
+                     shared-credentials-user-limit
+                     " users. Please add your own client_id and client_secret in the dashboard.")}])))
+
+(defn assert-shared-credentials-configured! [provider-name]
```
stopachka: imo do we need this asert? perhaps we can just use get-shared-credential! wherever we call this?

server/src/instant/model/app_user.clj
```diff
@@ -111,6 +114,29 @@
              (fn [{:keys [get-entity]}]
                (get-entity id)))))
 
+(def ^:private users-at-least-q
+  (uhsql/preformat
+   {:select [[[:inline 1]]]
```
stopachka: why not just make this a count count? what is this inline 1?

server/src/instant/runtime/routes.clj
```diff
@@ -184,6 +184,27 @@
...
+  (let [authorized-origins (app-authorized-redirect-origin-model/get-all-for-app
+                            {:app-id (:app_id client)})]
+    (when-not (app-authorized-redirect-origin-model/authorized-origin?
+               authorized-origins
+               redirect-uri
+               (app-oauth-client-model/use-shared-credentials? (:meta client)))
+      (ex/throw-validation-err!
+       :redirect-uri
+       redirect-uri
+       [{:message "Invalid redirect_uri. If you're the developer, make sure to add your website to the list of approved domains from the Dashboard."}]))))
```
stopachka: remove the word "from the Dashboard", as now it can be in the CLI too

client/www/components/dash/auth/GitHub.tsx
```diff
@@ -156,7 +156,6 @@ export function AddGitHubClientForm({
         clientId,
         clientSecret,
         redirectTo,
-        meta: { providerName: 'github' },
```
stopachka, line 159: This is no longer necessary, as we look up the provider name by id in `app_oauth_clients`

client/www/components/dash/auth/Google.tsx
```diff
@@ -71,24 +62,24 @@ export function AddGoogleClientForm({
   usedClientNames: Set<string>;
 }) {
   const token = useContext(TokenContext);
+  const [credentialMode, setCredentialMode] = useState<'dev' | 'custom'>('dev');
```
dwwoelfel, line 65: Are you sure you want to default to the dev credentials?

client/www/components/dash/auth/Google.tsx
```diff
@@ -71,24 +62,24 @@ export function AddGoogleClientForm({
...
   >('web');
+  const useSharedCredentials = appType === 'web' && credentialMode === 'dev';
   const [clientName, setClientName] = useState<string>(() =>
     findName(`google-${appType}`, usedClientNames),
   );
   const [clientId, setClientId] = useState<string>('');
   const [clientSecret, setClientSecret] = useState<string>('');
   const [redirectTo, setRedirectTo] = useState<string>('');
   const [updatedRedirectURL, setUpdatedRedirectURL] = useState(false);
-  const [skipNonceChecks, setSkipNonceChecks] = useState(isNative(appType));
```
dwwoelfel, line 84: If we don't let people disable this, can we make a change on the backend to ignore them? Otherwise, older clients will be stuck with the nonce check on.

dwwoelfel (approved):
Couple of comments, but LGTM

client/www/components/dash/auth/Google.tsx
stopachka: My thinking is yes, mainly to optimize for devex.

PR #2556, opened 2026-04-14
[WWW] Add dark mode for OR

nezaj:
What it says on the tin!

**Before**
before

**After**
after


drew-harris (approved):
would have never spotted that haha

PR #2562, opened 2026-04-16
Fix buffer for gauged-chan

dwwoelfel:
We weren't passing the buffer as an argument to the `gauged-chan`, so it was using the default of 1024.


stopachka (approved):
Nice!

PR #2565, opened 2026-04-16
degit -> tiged to fix node 25+ warning

drew-harris:
Definitely a bad first impression for anyone using node 25+

BEFORE:
image

AFTER:
image

Root cause was degit (unmaintained library) relied on package called `debug` which calls local storage when it shouldn't. Newer node versions output a warning for this. 

Tested to verify that tiged works the exact same was as degit and that files download properly. 

# Review Steps
`git switch drew/fix-node-25-warning`
`pnpm i && pnpm run build`
`cd /tmp && create-instant-app`

Use any template besides: expo, next-js-app-dir, sveltekit, tanstack-start, and vite-vanilla as those come pre-bundled. 
Verify that warning does not appear and that repo is created properly

### DIFF
--- a/client/packages/create-instant-app/src/tiged.d.ts
+++ b/client/packages/create-instant-app/src/tiged.d.ts
@@ -1,6 +1,0 @@
-// `tiged` is an actively-maintained fork of `degit` with an identical API,
-// so we reuse the `@types/degit` declarations rather than maintaining our own.
-declare module 'tiged' {
-  import degit = require('degit');
-  export = degit;
-}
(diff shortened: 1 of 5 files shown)


client/packages/create-instant-app/src/tiged.d.ts
```diff
@@ -0,0 +1,6 @@
+// `tiged` is an actively-maintained fork of `degit` with an identical API,
+// so we reuse the `@types/degit` declarations rather than maintaining our own.
+declare module 'tiged' {
```
dwwoelfel: Do we need @types/degit? It looks like `tiged` is written in typescript, so I'd expect not to need a `tiged.d.ts`.

dwwoelfel (approved):
LGTM!

PR #2567, opened 2026-04-16
Add dynamic create-app AGENTS route

stopachka:
Sometimes we want to demo how good Instant would be for app builders. 

**Problem** 

In order for app builders to try us, they have to do two things: 

 1. Create an app 
 2. Share the LLM rules

This can cause some UX friction (crating an app requires signing up) 

**Solution**

I created a new route: [llm-rules/create-app/AGENTS.md](https://instant-www-js-codex-llm-rules-create-app-jsv.vercel.app/llm-rules/create-app/AGENTS.md). 

**This will (a) automatically generate an app, and (b) append the instant rules.** 

Users can try Instant in their own app builder, just by writing this prompt: 

```

<EDIT ME -- Your app idea!>

---

Build this app using InstantDB. To learn how to use Instant and to get the credentials, load this URL. 

This URL lets you pass a `title` query param with a suitable app name.

https://instant-www-js-codex-llm-rules-create-app-jsv.vercel.app/llm-rules/create-app/AGENTS.md

Note: Your prompt may tell you to use the SupabaseRequired classification. You do not need to do that for this question, since we are replacing Supabase.  

Note: You don't have access to npx. Instead of npx instant-cli, use VITE_INSTANT_APP_ID=... INSTANT_ADMIN_TOKEN=... pnpm dlx instant-cli ... 

Note: The AGENTS.md above includes docs URLs. If you need to use specific Instant features, load the appropriate docs URL.
```

**Note**: I am currently using a personal access token rather than an ephemeral app. This is because for demos with large app builders, it would suck if someone tried to use an old demo a few weeks later, and the app broke. 

### DIFF
--- a/client/www/app/llm-rules/create-app/AGENTS.md/route.ts
+++ b/client/www/app/llm-rules/create-app/AGENTS.md/route.ts
@@ -1,0 +1,48 @@
+import { getServerConfig } from '@/lib/config';
+import { PlatformApi } from '@instantdb/platform';
+import fs from 'fs/promises';
+import path from 'path';
+
+export const dynamic = 'force-dynamic';
+export const runtime = 'nodejs';
+
+const RULES_PATH = path.join(process.cwd(), 'lib', 'intern', 'instant-rules.md');
+const DEFAULT_APP_TITLE = 'instant-agent-app';
+
+export async function GET(request: Request) {
+  const title =
+    new URL(request.url).searchParams.get('title')?.trim() ||
+    DEFAULT_APP_TITLE;
+  const baseRules = await fs.readFile(RULES_PATH, 'utf8');
+  const { apiURI } = await getServerConfig();
+  const api = new PlatformApi({ apiURI });
+  const { app, expiresMs } = await api.createTemporaryApp({ title });
+
+  const markdown = `# Instant App
+
+Use this Instant app for the current session.
+
+- App ID: \`${app.id}\`
+- Admin Token: \`${app.adminToken}\`
+- Expires At: \`${new Date(expiresMs).toISOString()}\`
+
+## Environment
+
+\`\`\`env
+NEXT_PUBLIC_INSTANT_APP_ID=${app.id}
+INSTANT_APP_ADMIN_TOKEN=${app.adminToken}
+\`\`\`
+
+This app is ephemeral. If it expires, create a new one from this URL.
+
+${baseRules}`;
+
+  return new Response(markdown, {
+    headers: {
+      'Cache-Control': 'private, no-store, max-age=0',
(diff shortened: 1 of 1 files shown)


client/www/app/llm-rules/create-app/AGENTS.md/route.ts
```diff
@@ -0,0 +1,66 @@
...
+\`\`\`
+INSTANT_APP_ID=${app.id}
+INSTANT_ADMIN_TOKEN=${app.adminToken}
+\`\`\`
+
+Now you can use this to build a full-stack app. 
+
+What follows are instructions for using Instant. Follow them, and you are golden!
+
+Note: You don't have access to npx. Instead of npx instant-cli, use INSTANT_APP_ID=... INSTANT_ADMIN_TOKEN=... pnpm dlx ...
```
stopachka: Note: I hardcoded this for now, as the customer we are doing this demo for does not expose `npx` in their app builder. However they give access to `pnpm`, and `pnpm dlx` does the trick.

client/www/app/llm-rules/create-app/AGENTS.md/route.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  const { app } = await api.createApp({ title, orgId });
+
+  const markdown = `
+You've just gotten a new Instant app provisioned for you! 
+
+Here's the info:
+
+\`\`\`
+VITE_INSTANT_APP_ID=${app.id}
```
dwwoelfel, line 50: Maybe we should also add 

```
NEXT_PUBLIC_INSTANT_APP_ID=${app.id}
```

Just in case?

client/www/app/llm-rules/create-app/AGENTS.md/route.ts
```diff
@@ -0,0 +1,64 @@
...
+  if (!orgId) {
+    throw new Error('INSTANT_LLM_RULES_CREATE_APP_ORG_ID is not set');
+  }
+
+  const api = new PlatformApi({ auth: { token } });
+
+  const { app } = await api.createApp({ title, orgId });
+
+  const markdown = `
+You've just gotten a new Instant app provisioned for you!
```
dwwoelfel:
```suggestion
A new Instant app has just been provisioned for you!
```

The grammar felt off to me

dwwoelfel (approved):
Couple of comments, but LGTM!

client/www/app/llm-rules/create-app/AGENTS.md/route.ts
stopachka: done!

PR #2568, opened 2026-04-17
Fix the algolia index script

dwwoelfel:
Our docs files moved and the script that indexes our docs for algolia failed.

Updates the script to look in the correct place for docs.

Also runs the script in --dry-run mode on non-main branches to make sure we catch regressions before they make it into main.


stopachka (approved):
Oi thank you for catching this @dwwoelfel !

PR #2569, opened 2026-04-17
increase "renames an attribute" test timeout

drew-harris:
Been seeing this test be flaky in CI/CD for my Oauth branches.


stopachka (approved):
sgtm!

PR #2570, opened 2026-04-17
Link to rate limits from permissions

dwwoelfel:
Tested with claude and it was able to correctly implement permissions with this change. Without it, it didn't discover the rate limit doc.

### DIFF
--- a/client/www/components/docs/Heading.js
+++ b/client/www/components/docs/Heading.js
@@ -32,7 +32,7 @@ export function Heading({ level = 1, id, children, ...props }) {
   // Use the id from collectHeadings if provided, otherwise generate one
   const slugId =
     id ||
-    (typeof children === 'string' ? slugifyWithCounter(children) : undefined);
+    (typeof children === 'string' ? slugifyWithCounter()(children) : undefined);
 
   return (
     <Tag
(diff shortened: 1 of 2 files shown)


stopachka (approved):
Nice!

client/www/components/docs/Heading.js
```diff
@@ -31,8 +31,7 @@ export function Heading({ level = 1, id, children, ...props }) {
 
   // Use the id from collectHeadings if provided, otherwise generate one
   const slugId =
-    id ||
-    (typeof children === 'string' ? slugifyWithCounter(children) : undefined);
+    id || (typeof children === 'string' ? slugify(children) : undefined);
```
nezaj, line 34: Curious, why this change?

dwwoelfel: I had some hydration errors and claude thought this was the culprit.

I don't think it was, but it's still a good change to make. It doesn't make sense to call `slugifyWithCounter` that way. First, `slugifyWithCounter` returns a `slugify` function when you call it and then you're supposed to use that function to create your slugs., so it wouldn't have worked. And second, even if you use it correctly, it's very hard to get consistent function call orders in react.

PR #2571, opened 2026-04-17
Add LinkedIn OAuth Clients via CLI (#2571) [bump to v1.0.12]

drew-harris:
Adds linkedin as an oauth provider to `instant-cli auth clients add`
Also:
- Fixes formatting inconsistencies with other providers. 
- Unifies consts where they are the same (DEFAULT_OAUTH_CALLBACK_URL) instead of the same variable 5 times with different names. 
- Cleans up formatting for `instant-cli auth clients add --help`
- Uses the same consts in the dashboard. 

```bash
cd sandbox/cli-nodejs;
# Works Properly
INSTANT_CLI_DEV=true instant-cli auth client add --type linkedin --name test1 --client-id asdf --client-secret asdf -y

# Missing Name
INSTANT_CLI_DEV=true instant-cli auth client add --type linkedin --client-id asdf --client-secret asdf -y

# All Missing Info
INSTANT_CLI_DEV=true instant-cli auth client add --type linkedin -y
```


client/packages/cli/src/commands/auth/client/add.ts
```diff
@@ -154,7 +156,7 @@ const handleGoogleClient = Effect.fn(function* (opts: Record<string, unknown>) {
     required: true,
     skipIf: false,
     prompt: {
-      prompt: `Client ID ${chalk.dim('(from https://console.developers.google.com/apis/credentials)')}`,
+      prompt: `Client ID: ${chalk.dim('(from https://console.developers.google.com/apis/credentials)')}`,
```
dwwoelfel, line 159: Can we make this (and the other urls for other clients) clickable?

dwwoelfel (approved):
One suggestion, but LGTM

client/packages/cli/src/commands/auth/client/add.ts
drew-harris: Already started on the next prs so I will include this in the next pr to avoid having to deal with merge conflicts.

PR #2577, opened 2026-04-21
Switch to wal-entry/parse-buffer permanently

dwwoelfel:
Follow up to https://github.com/instantdb/instant/pull/2576

We've been running this in production for a few hours now with no issues, so it should be safe to switch over permanently.


stopachka (approved):
Let's do it!

PR #2579, opened 2026-04-21
[server] get-a-db apps

stopachka:
For getadb.com, we need a way to:

1. Spin up apps that never expire
2. Have these apps be claimable

We couldn't use ephemeral apps to satisfy 1., since they expire. 

To solve for this, 

(1) I added separate `get-a-db` endpoints for creating these apps. They are quite similar to the ephemeral-app endpoints, with one caveat: this _requires_ an access token to create apps. These apps never expire

(2) I updated our claim flow, so we accept _both_ ephemeral apps and get-a-db apps.

**Notes** 

The only way to claim ephemeral apps and get-a-db apps currently is through the cli. 

**You may wonder: why not support the dashboard, by passing in ?appId=...&adminToken=... ?** 

This would expose the admin token in url history, which could be a security vulnerability. We could create a special-purpose UI in the dashboard that does this, or create some kind of ticketing system, but for now I think just the CLI is fine. 

### DIFF
--- a/server/src/instant/dash/get_a_db.clj
+++ b/server/src/instant/dash/get_a_db.clj
@@ -10,9 +10,7 @@
    [instant.util.string :as string-util]
    [instant.util.uuid :as uuid-util]
    [instant.superadmin.routes :refer [req->superadmin-user!]]
-   [ring.util.http-response :as response])
-  (:import
-   (java.util UUID)))
+   [ring.util.http-response :as response]))
 
 (def get-a-db-creator-email "hello+getadbapps@instantdb.com")
 
--- a/server/src/instant/dash/get_a_db.clj
+++ b/server/src/instant/dash/get_a_db.clj
@@ -1,0 +1,72 @@
+(ns instant.dash.get-a-db
+  (:require
+   [clojure.walk :as w]
+   [instant.model.app :as app-model]
+   [instant.model.instant-user :as instant-user-model]
+   [instant.model.rule :as rule-model]
+   [instant.model.schema :as schema-model]
+   [instant.util.exception :as ex]
+   [instant.util.posthog :as posthog]
+   [instant.util.string :as string-util]
+   [instant.util.uuid :as uuid-util]
+   [instant.superadmin.routes :refer [req->superadmin-user!]]
+   [ring.util.http-response :as response])
+  (:import
+   (java.util UUID)))
+
+(def get-a-db-creator-email "hello+getadbapps@instantdb.com")
+
+(def get-a-db-creator
+  (delay
+    (instant-user-model/get-by-email {:email get-a-db-creator-email})))
+
+(defn create!
+  [{:keys [title]}]
+  (app-model/create!
+   {:id (UUID/randomUUID)
+    :title title
+    :creator-id (:id @get-a-db-creator)
+    :admin-token (UUID/randomUUID)}))
+
+;; -----------
+;; HTTP Handler
+
+(defn http-post-handler [req]
+  (let [{user-id :id} (req->superadmin-user! :apps/write req)
+        _ (ex/assert-permitted!
+           :get-a-db-user?
+           user-id
+           (= user-id (:id @get-a-db-creator)))
+        title (ex/get-param! req [:body :title] string-util/coerce-non-blank-str)
+        schema (get-in req [:body :schema])
+        rules-code (ex/get-optional-param! req [:body :rules :code] w/stringify-keys)
+        _ (when rules-code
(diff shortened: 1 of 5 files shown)


server/src/instant/dash/get_a_db.clj
```diff
@@ -0,0 +1,72 @@
...
+(def get-a-db-creator-email "hello+getadbapps@instantdb.com")
+
+(def get-a-db-creator
+  (delay
+    (instant-user-model/get-by-email {:email get-a-db-creator-email})))
+
+(defn create!
+  [{:keys [title]}]
+  (app-model/create!
+   {:id (UUID/randomUUID)
```
dwwoelfel:
```suggestion
   {:id (random-uuid)
```

server/src/instant/dash/get_a_db.clj
```diff
@@ -0,0 +1,72 @@
...
+    :title title
+    :creator-id (:id @get-a-db-creator)
+    :admin-token (UUID/randomUUID)}))
```
dwwoelfel:
```suggestion
    :admin-token (random-uuid)}))
```

server/test/instant/dash/routes_test.clj
```diff
@@ -1057,3 +1059,94 @@
...
+
+(deftest get-a-db-create-allowed-with-creator-pat
+  (with-pat @get-a-db/get-a-db-creator
+    (fn [token]
+      (let [resp (http/post (str config/server-origin "/dash/apps/get_a_db")
+                            {:headers {:Authorization (str "Bearer " token)
+                                       :Content-Type "application/json"}
+                             :as :json
+                             :body (->json {:title "test-get-a-db"})})
+            app-id (UUID/fromString (get-in resp [:body :app :id]))]
```
dwwoelfel:
```suggestion
            app-id (parse-uuid (get-in resp [:body :app :id]))]
```

server/test/instant/dash/routes_test.clj
```diff
@@ -4,17 +4,22 @@
...
    [instant.model.instant-stripe-customer :as stripe-customer-model]
    [instant.model.org-members :as org-members]
    [instant.model.app-members :as app-members]
    [instant.stripe :as stripe]
    [instant.util.crypt :as crypt-util]
    [instant.util.json :refer [->json <-json]]
-   [instant.util.tracer :as tracer]))
+   [instant.util.tracer :as tracer])
+  (:import
+   (java.util UUID)))
```
dwwoelfel:
```suggestion
   [instant.util.tracer :as tracer]))
```

dwwoelfel (approved):
LGTM. One suggestion about using uuid fns from clojure.core instead of importing java.util.UUID

server/src/instant/dash/get_a_db.clj
```diff
@@ -0,0 +1,72 @@
...
+   [instant.model.rule :as rule-model]
+   [instant.model.schema :as schema-model]
+   [instant.util.exception :as ex]
+   [instant.util.posthog :as posthog]
+   [instant.util.string :as string-util]
+   [instant.util.uuid :as uuid-util]
+   [instant.superadmin.routes :refer [req->superadmin-user!]]
+   [ring.util.http-response :as response])
+  (:import
+   (java.util UUID)))
```
stopachka:
```suggestion
(ns instant.dash.get-a-db
  (:require
   [clojure.walk :as w]
   [instant.model.app :as app-model]
   [instant.model.instant-user :as instant-user-model]
   [instant.model.rule :as rule-model]
   [instant.model.schema :as schema-model]
   [instant.util.exception :as ex]
   [instant.util.posthog :as posthog]
   [instant.util.string :as string-util]
   [instant.util.uuid :as uuid-util]
   [instant.superadmin.routes :refer [req->superadmin-user!]]
   [ring.util.http-response :as response]))
```

server/src/instant/dash/get_a_db.clj
```diff
@@ -0,0 +1,70 @@
...
+   [instant.util.string :as string-util]
+   [instant.util.uuid :as uuid-util]
+   [instant.superadmin.routes :refer [req->superadmin-user!]]
+   [ring.util.http-response :as response]))
+
+(def get-a-db-creator-email "hello+getadbapps@instantdb.com")
+
+(def get-a-db-creator
+  (delay
+    (instant-user-model/get-by-email {:email get-a-db-creator-email})))
```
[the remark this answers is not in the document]
stopachka: This is fine. It would end up throwing in runtime, and would be caught by our tests.

PR #2580, opened 2026-04-21
Fix attr name change timeouts

dwwoelfel:
Fixes a timeout where changing the attr name would cause a timeout on large namespaces.

When you update the attr, we update the index fields on the triples. But when you change the name, none of the index fields should be affected. This PR skips updating the triples if none of the relevant attr columns change.


stopachka (approved):
Heeck yeah!

nezaj (approved):
Nice!

PR #2581, opened 2026-04-21
[CLI] Make abort actions cancel isntead of confirm

nezaj:
We got a report that hitting `Esc` when pushing with `instant-cli` would confirm instead of abort. This is because we would use `defaultValue` in the abort/error case which is often set to true.

This updates behavior so aborts return `false`, making it so pressing escape works as expected


stopachka:
Landed an approach that bubbled up errors instead: https://github.com/instantdb/instant/pull/2582

PR #2582, opened 2026-04-21
Let errors bubble up in promptOk

stopachka:
**Problem** 

In promptOk we wrote: 

```
  const ok = yield* Effect.tryPromise(() =>
    renderUnwrap(
      new UI.Confirmation({
        ...props,
        defaultValue,
      }),
    ),
  ).pipe(Effect.orElseSucceed(() => defaultValue));
```

The `orElseSucceed` swallowed legitimate issues: 

1. For example, if a user presses 'esc', orElseSucceed would treat that as a choice for default value

But really, 'esc' should stop the flow and exit 

**Solution** 

I removed this `orElseSucceed` pipe, and just let the error bubble up. 

In printRedErrors, I do two things: 
1. If unhandled promises throw in Effect, an `UnknownException` type will throw. I went ahead and unwrapped that
2. This then goes through the standard machinery, and prints out: 

```
[error] Prompt was aborted
```

CleanShot 2026-04-21 at 11 48 23@2x

### Testing

```
cd client/sandbox/cli-nodejs
# make some changes in schema
pnpm exec instant-cli push
# press esc
```


nezaj (approved):
LGTM!

PR #2583, opened 2026-04-21
getadb.com v0

stopachka:
https://instant-www-js-getadb-jsv.vercel.app/getadb

This is an outline implementation of getadb.com!

When you load getadb.com, we directly load the AGENTS.md file with our spliced in token. We use our new backend /get-a-db endpoint to create apps that never expire.

**Where is getadb.com hosted?** 

I chose to host us in our existing www nextjs app. I thought this best for maintanbility. 

To detect what the host is, we use nextjs middleware 

**Why use nextjs middleware?** 

We could have rewrite rules, like we do with Posthog. That method is simplest, but it opens up a problem: what if users load getadb.com/pricing? This would load our pricing page on www. I do wonder if I could get around this in other ways, but I wanted to push something up.

**What is next** 

In a follow on PR I will: 

1. Add detections and tests for Claude Code, Figma, Replit, Figma Make
2. Finalize how we do redirects: do we use rewrite rules or middleware?
3. Play with a /human page


client/www/app/getadb/markdown.ts
```diff
@@ -21,25 +19,43 @@ async function loadBaseRules(): Promise<string> {
...
+  if (!res.ok) {
+    throw new Error(
+      `Failed to create get-a-db app: ${res.status} ${await res.text()}`,
+    );
+  }
+  const body = (await res.json()) as {
+    app: { id: string; 'admin-token': string };
+  };
+  return { id: body.app.id, adminToken: body.app['admin-token'] };
+}
```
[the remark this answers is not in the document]
stopachka: no need I think

client/www/app/getadb/markdown.ts
```diff
@@ -21,25 +19,43 @@ async function loadBaseRules(): Promise<string> {
   return contents;
 }
 
-export async function GET(request: Request) {
+async function createGetADbApp(
+  token: string,
+  title: string,
+): Promise<{ id: string; adminToken: string }> {
+  const { apiURI } = await getServerConfig();
+  const res = await fetch(`${apiURI}/dash/apps/get_a_db`, {
```
stopachka, line 27: follow-up: I should use jsonFetch

client/www/middleware.ts
```diff
@@ -0,0 +1,20 @@
...
+  'getadb.com',
+  'www.getadb.com',
+  'getadb.localhost',
+]);
+
+export function middleware(request: NextRequest) {
+  const host = (request.headers.get('host') ?? '').split(':')[0];
+  if (!GETADB_HOSTS.has(host)) return NextResponse.next();
+
+  const url = request.nextUrl.clone();
```
stopachka, line 13: follow-up, I should wrap this in a functon

nezaj (approved):
Let's do it!

PR #2586, opened 2026-04-22
Add Svelte client JS SDK to AGENTS.md

laurentpayot:
Added missing JS SDK @instantdb/svelte


nezaj:
Thanks for pointing this out! Including this in another rules update here https://github.com/instantdb/instant/pull/2621

PR #2597, opened 2026-04-24
fix components package build

drew-harris:
image

Fixes an error when trying to build the components package. Interestingly enough, this error doesn't cause the build to fail which is why CI has been passing but it is annoying to see pop up.


stopachka (approved):
SGTM

PR #2599, opened 2026-04-24
Strengthen no-cache headers on /getadb

stopachka:
- Some fetchers (e.g. v0's WebFetch) appear to cache `getadb.com` responses globally despite `Cache-Control: no-store`
- Belt-and-suspenders: added the legacy HTTP/1.0 fallbacks (`Pragma: no-cache`, `Expires: 0`), `must-revalidate`, `proxy-revalidate`, and `Vary: *`

@dwwoelfel @nezaj @drew-harris

PR #2600, opened 2026-04-24
Redirect /getadb with a unique n param to bust upstream caches

stopachka:
## Summary
- Follow-up to #2599. Even with hard no-cache headers, upstream fetchers (notably v0's WebFetch) were still serving a cached `/getadb` response to different users.
- The cache key is the URL, so we now 302 bare `/getadb` → `/getadb?n=<base58 nanoid>`. Each new visitor gets a unique URL and therefore a unique cache entry.
- The `n` uses Unkey's base58 alphabet via `nanoid` (16 chars, omits confusable `0/O`, `I/l`) so the URL stays readable. See https://www.unkey.com/blog/uuid-ux.
- The redirect response itself is marked no-store so the 302 can't be cached either.

## Test plan
- [ ] `curl -i https://instantdb.com/getadb` returns a 302 with `Location: /getadb?n=<16 chars>` and no-store headers.
- [ ] Following the redirect returns the AGENTS.md markdown.
- [ ] Visiting `/getadb` twice in a row produces two different `n` values in the `Location`.
- [ ] `?title=Foo` is preserved through the redirect.
- [ ] Asking v0 to fetch `/getadb` no longer returns a stale / shared app.

### DIFF
--- a/client/www/app/getadb/route.ts
+++ b/client/www/app/getadb/route.ts
@@ -1,17 +1,44 @@
 import { randomUUID } from 'crypto';
+import { customAlphabet } from 'nanoid';
 import { getServerConfig } from '@/lib/config';
 import generateMarkdown from './generateMarkdown';
+
+// Base58 alphabet (omits confusable 0/O, I/l). Shorter and prettier than a
+// UUID in the URL bar. See https://www.unkey.com/blog/uuid-ux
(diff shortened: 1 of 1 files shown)


nezaj (approved):
Let's try!

PR #2604, opened 2026-04-26
[Auth] Fix missing app title in magic code emails

nezaj:
Got a report that app title's were missing from magic code emails. Was able to verify too

CleanShot 2026-04-25 at 18 20 05@2x

**Cause**
This was introduced in #2406 when default-body switched from positional args to a destructured map without renaming the key to match template-params.

**Fix**
title -> app_title in default-body. Also added a test asserting the rendered body includes the title, code, and expiration


stopachka (approved):
Nice!

PR #2607, opened 2026-04-27
Redirect getadb new aliases in middleware

stopachka:
**Problem**

Claude Design only allows URLs in the web_fetch tool if the user has explicitly listed them. 

This breaks our system where the agent itself generates the random uuid. 

**Solution** 

I added a new getadb.com/new redirect to getadb.com/provision/new

Claude Design doesn't cache like the others, so this is usable there

### DIFF
--- a/client/www/middleware.ts
+++ b/client/www/middleware.ts
@@ -8,9 +8,20 @@ const GETADB_HOSTS = new Set([
 
 export function middleware(request: NextRequest) {
   const host = (request.headers.get('host') ?? '').split(':')[0];
+  const url = request.nextUrl.clone();
+
+  if (url.pathname === '/getadb/new') {
+    url.pathname = '/getadb/provision/new';
+    return NextResponse.redirect(url, 307);
+  }
+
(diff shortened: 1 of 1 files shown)


client/www/middleware.ts
```diff
@@ -8,9 +8,20 @@ const GETADB_HOSTS = new Set([
 
 export function middleware(request: NextRequest) {
   const host = (request.headers.get('host') ?? '').split(':')[0];
+  const url = request.nextUrl.clone();
+
+  if (url.pathname === '/getadb/new') {
+    url.pathname = '/getadb/provision/new';
```
stopachka, line 14: This is to allow localhost:3000/getadb/new and instant.ngrok.dev/getadb/new

dwwoelfel (approved):
LGTM, but maybe it would be easier if instead of saying `"fetch https://getadb.com/provision/<generate-a-random-16-char-token>"`, we said `"fetch https://getadb.com?n=<generate-a-random-uuid>"`, that way it's always the same url. Or does it not allow different query params?

PR #2608, opened 2026-04-27
Print app ID and add --temp flag to CIA (#2608) [bump to v1.0.19]

drew-harris:
Adds a print statement after creating/importing an app in create instant app.

Also added a --temp flag so that temporary apps can be created if the user is logged in and using --yes. 
If --temp is used without --yes (interactive mode) the user will be prompted for a title without being given any other options for creating/connecting the app. 

image
image

How to test
`cd /tmp && create-instant-app --yes` and verify that an app id is printed
`cd /tmp && create-instant-app --yes --temp` and verify that a temporary app id is printed
`cd /tmp && create-instant-app --yes --app <app_id>` and verify that an imported app id is printed
`cd /tmp && create-instant-app` import/create an app and verify the id is printed.

### DIFF
--- a/client/packages/create-instant-app/src/cli.ts
+++ b/client/packages/create-instant-app/src/cli.ts
@@ -32,6 +32,7 @@ export type Project = {
   app: string | null;
   token: string | null;
   yes: boolean;
+  temporary: boolean;
 };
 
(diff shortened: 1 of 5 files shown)


dwwoelfel (approved):
LGTM!

PR #2611, opened 2026-04-28
Increase grpc max message size for invalidator records

dwwoelfel:
Increases the maximum size for grpc invalidator records.

This is just a short-term fix--the longer-term fix is to skip putting the wal-logs into the channel and have the server read them from the db if it needs them, but we can wait for that until after webhooks are shipped.


stopachka (approved):
Cool!

PR #2617, opened 2026-04-29
Update from jdk 22 to jdk 26

dwwoelfel:
This should prevent `locking` from causing pinning on virtual threads.

I didn't see any changes between 22 and 26 that would cause issues for us.

Also adds `-Djdk.tracePinnedThreads=full` to dev mode so that it will be easier for us to catch pinned threads.

We should look into using compact object headers, which is supposed to help with the size of the heap in clojure since it creates lots of small objects and this saves 4 bytes per object (enabled with the `-XX:+UseCompactObjectHeaders` flag, which we can add later).

Tested locally and in staging.

### DIFF
--- a/.github/workflows/clojure.yml
+++ b/.github/workflows/clojure.yml
@@ -123,7 +123,7 @@ jobs:
       uses: actions/setup-java@v4
       with:
         distribution: 'corretto'
-        java-version: '22'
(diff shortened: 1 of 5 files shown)


nezaj (approved):
🚢

PR #2624, opened 2026-04-30
[cli] auth client update

stopachka:
This PR adds `npx instant-cli auth client update`, so users can change OAuth client credentials after a client has already been created.

The main use case is Google shared dev credentials: start fast with Instant dev credentials, then switch to your own Google OAuth credentials when you are ready. It also supports rotating credentials and updating provider-specific settings for the rest of our OAuth client types.

## shared helpers

This also extracts shared OAuth client helpers that are now used by `add`, `update`, `list`, and `delete`:

- Flag helpers for checking provided CLI args.
- Shared provider prompts for client IDs, client secrets, redirect URIs, Apple fields, Clerk publishable keys, and Firebase project IDs.
- Shared provider utilities for Apple private key files, Clerk discovery domains, Firebase discovery endpoints, OAuth client lookup by ID or name, and redirect setup success messages.
- `optOrPromptWhenNeeded`, which lets update flows only prompt for the fields selected by flags or by the interactive action.

This keeps `add` and `update` using the same prompt copy and validation instead of duplicating provider-specific details across commands.

## update

**Select a client by name or ID**

You can update a client by `--name` or `--id`:

```
npx instant-cli auth client update --name google-dev --client-id test-google-client-id --client-secret test-google-client-secret --yes
```

If you do not pass `--name` or `--id` in interactive mode, the CLI asks which client to update. Dev-credential clients are labeled in the picker.

**Google dev credentials can move to custom credentials**

A Google web client that was created with Instant dev credentials can be updated to use your own Google OAuth credentials:

```
npx instant-cli auth client update \
  --name google-dev \
  --client-id test-google-client-id \
  --client-secret test-google-client-secret \
  --yes
```

When the client moves to custom credentials, the CLI prints the Google Console redirect URI again:

```txt
┌─────────────────────────────────────────────────────┐
│ Google OAuth client updated: google-dev             │
│ Credentials: custom                                 │
│ ID: 33b53917-bc51-486f-af84-413f2b52e603            │
│                                                     │
│ This client no longer uses Instant dev credentials. │
│                                                     │
│ Add this redirect URI in Google Console:            │
│ https://api.instantdb.com/runtime/oauth/callback    │
└─────────────────────────────────────────────────────┘
```

**Google custom credentials can move back to dev credentials**

You can also switch a Google web client back to Instant dev credentials:

```
npx instant-cli auth client update --name google-custom --dev-credentials --yes
```

The update clears the redirect URI and marks the client as using shared credentials again:

```txt
┌─────────────────────────────────────────────────┐
│ Google OAuth client updated: google-dev         │
│ Credentials: Instant dev credentials            │
│ ID: 33b53917-bc51-486f-af84-413f2b52e603        │
│                                                 │
│ No Google Console setup required.               │
│ Works on localhost and Expo during development. │
└─────────────────────────────────────────────────┘
```

**Custom credential providers can rotate values**

GitHub, LinkedIn, and Google custom clients can rotate `--client-id`, `--client-secret`, and `--custom-redirect-uri` without recreating the client.

Apple can update `--services-id`, and the web redirect settings: `--team-id`, `--key-id`, `--private-key-file`, and `--custom-redirect-uri`.

Clerk and Firebase can update their source settings:

```
npx instant-cli auth client update \
  --name clerk \
  --publishable-key pk_test_Z3VpZGluZy1wZWdhc3VzLTkzLmNsZXJrLmFjY291bnRzLmRldiQ \
  --yes

npx instant-cli auth client update \
  --name firebase \
  --project-id my-app-123 \
  --yes
```

For Clerk and Firebase, the CLI also refreshes the stored discovery endpoint.

**Interactive mode asks what you want to update**

For custom credential clients, interactive mode asks whether to rotate credentials or update the redirect URI. For Google web clients, it can also switch back to Instant dev credentials. When a redirect URI changes, the success output mirrors the provider-specific setup guidance from `auth client add`.

## add

The Google dev-credentials success box now points users at the new update command:

```txt
Google OAuth client created: google-dev
App type: web
Credentials: Instant dev credentials
ID: ae5ac75d-6840-42c9-a8d5-2881f630300e

No Google Console setup required.
Works on localhost and Expo during development.

Ready for production? Run:
  instant-cli auth client update --name google-dev \
    --client-id <id> \
    --client-secret <secret>
```

The rest of the add flow should behave the same as before, but now uses the shared helpers described above for provider prompts and validation.

## Testing

Update flow:

```bash
# 0. Make sure to re-run `make dev`, so your backend is up to date.

# 1. Come on into cli-nodejs and build a throwaway app.

export INSTANT_CLI_DEV=true
cd client/sandbox/cli-nodejs
rm -f .env
pnpm exec instant-cli init --yes --title testing-app

# 2. Create a Google web client with Instant dev credentials.

pnpm exec instant-cli auth client add \
  --type google \
  --app-type web \
  --name google-dev \
  --dev-credentials

## Notice the success box says it is using Instant dev credentials and points at auth client update.

# 3. Update that Google client to custom credentials.

pnpm exec instant-cli auth client update \
  --name google-dev \
  --client-id test-google-client-id \
  --client-secret test-google-client-secret \
  --yes

## Notice the output says the client no longer uses Instant dev credentials and shows the Google Console redirect URI.

# 4. List clients and confirm google-dev now shows Credentials: custom.

pnpm exec instant-cli auth client list

# 5. Move the same Google client back to Instant dev credentials.

pnpm exec instant-cli auth client update \
  --name google-dev \
  --dev-credentials \
  --yes

## Notice the output says Credentials: Instant dev credentials.

# 6. List clients and confirm google-dev is back to dev credentials.

pnpm exec instant-cli auth client list

# 7. Create a custom Google web client and update only its redirect URI.

pnpm exec instant-cli auth client add \
  --type google \
  --app-type web \
  --name google-custom \
  --client-id old-google-client-id \
  --client-secret old-google-client-secret \
  --yes

pnpm exec instant-cli auth client update \
  --name google-custom \
  --custom-redirect-uri https://example.com/oauth/callback \
  --yes

## Notice this does not ask for a new client ID or client secret. It prints the Google Console redirect guidance and custom redirect test URL.

# 8. Test the interactive Google path.

pnpm exec instant-cli auth client update --name google-custom

## See the CLI ask what you want to update. Try both switching to dev credentials and updating the redirect URI.

# 9. Test validation: --dev-credentials cannot be combined with custom credential flags.

pnpm exec instant-cli auth client update \
  --name google-custom \
  --dev-credentials \
  --client-id test-google-client-id \
  --yes

# Notice this fails with a message saying --dev-credentials cannot be combined with --client-id.

# 10. Test validation: --dev-credentials is rejected for native Google clients.

pnpm exec instant-cli auth client add \
  --type google \
  --app-type ios \
  --name google-ios \
  --client-id test-ios-google-client-id \
  --yes

pnpm exec instant-cli auth client update \
  --name google-ios \
  --dev-credentials \
  --yes

# Notice this fails with a message saying --dev-credentials is only supported for Google web clients.

# 11. Create and update a GitHub client.

pnpm exec instant-cli auth client add \
  --type github \
  --name github-custom \
  --client-id old-github-client-id \
  --client-secret old-github-client-secret \
  --yes

pnpm exec instant-cli auth client update \
  --name github-custom \
  --client-secret new-github-client-secret \
  --yes

## Notice this updates without requiring a new client ID.

# 12. Test a GitHub interactive redirect-only update.

pnpm exec instant-cli auth client update --name github-custom

## Choose Update redirect URI and enter https://example.com/oauth/callback. Notice the output says to add this callback URL in your GitHub OAuth App settings.

# 13. Create and update a LinkedIn redirect URI.

pnpm exec instant-cli auth client add \
  --type linkedin \
  --name linkedin-custom \
  --client-id old-linkedin-client-id \
  --client-secret old-linkedin-client-secret \
  --yes

pnpm exec instant-cli auth client update \
  --name linkedin-custom \
  --custom-redirect-uri https://example.com/linkedin/callback \
  --yes

## Notice the output says to add this redirect URI in your LinkedIn app settings.

# 14. Create and update an Apple redirect URI.

pnpm exec instant-cli auth client add \
  --type apple \
  --name apple-custom \
  --services-id com.example.apple \
  --yes

pnpm exec instant-cli auth client update \
  --name apple-custom \
  --custom-redirect-uri https://example.com/apple/callback \
  --yes

## Notice the output says to add this return URL under your Apple Services ID.

# 15. Create and update Clerk.

pnpm exec instant-cli auth client add \
  --type clerk \
  --name clerk \
  --publishable-key pk_test_Z3VpZGluZy1wZWdhc3VzLTkzLmNsZXJrLmFjY291bnRzLmRldiQ \
  --yes

pnpm exec instant-cli auth client update \
  --name clerk \
  --publishable-key pk_test_Z3VpZGluZy1wZWdhc3VzLTkzLmNsZXJrLmFjY291bnRzLmRldiQ \
  --yes

## Notice the update refreshes the Clerk discovery endpoint.

# 16. Create and update Firebase.

pnpm exec instant-cli auth client add \
  --type firebase \
  --name firebase \
  --project-id old-app-123 \
  --yes

pnpm exec instant-cli auth client update \
  --name firebase \
  --project-id my-app-123 \
  --yes

## Notice the update refreshes the Firebase discovery endpoint.

# 17. Test --yes validation.

pnpm exec instant-cli auth client update --client-secret new-secret --yes

# Notice this fails because --yes requires --id or --name.

pnpm exec instant-cli auth client update --name github-custom --yes

# Notice this fails because --yes requires at least one update field.
```

Add and shared-helper regressions:

```bash
# 1. Reuse the same sandbox app from the first block.

export INSTANT_CLI_DEV=true
cd client/sandbox/cli-nodejs

# 2. Create a Google web client with explicit dev credentials.

pnpm exec instant-cli auth client add \
  --type google \
  --app-type web \
  --name google-add-dev \
  --dev-credentials

## Notice the success box says Credentials: Instant dev credentials and includes the auth client update command.

# 3. Create a Google web client with custom credentials.

pnpm exec instant-cli auth client add \
  --type google \
  --app-type web \
  --name google-add-custom \
  --client-id test-google-client-id \
  --client-secret test-google-client-secret \
  --yes

## Notice the success box still shows the Google Console redirect URI.

# 4. Run an interactive GitHub add to check the shared client ID, client secret, and redirect URI prompts.

pnpm exec instant-cli auth client add \
  --type github \
  --name github-add-interactive

## Enter a client ID, client secret, and optional redirect URI. Notice the prompt copy and output still match the previous add flow.

# 5. Exercise the shared Apple private key reader and Apple web fields.

printf '%s\n' \
  '-----BEGIN PRIVATE KEY-----' \
  'test-private-key' \
  '-----END PRIVATE KEY-----' \
  > /tmp/instant-apple-test-key.p8

pnpm exec instant-cli auth client add \
  --type apple \
  --name apple-add-web \
  --services-id com.example.apple.add \
  --team-id TEAM123 \
  --key-id KEY123 \
  --private-key-file /tmp/instant-apple-test-key.p8 \
  --yes

## Notice the Apple client is created and the .p8 file path is accepted.

# 6. Exercise the shared Firebase project ID validation.

pnpm exec instant-cli auth client add \
  --type firebase \
  --name firebase-invalid \
  --project-id BAD \
  --yes

# Notice this fails with the Firebase project ID validation message.

# 7. Create a valid Firebase client and list everything.

pnpm exec instant-cli auth client add \
  --type firebase \
  --name firebase-add-valid \
  --project-id valid-app-123 \
  --yes

pnpm exec instant-cli auth client list

## Confirm the Google dev client is labeled as Instant dev credentials, the Google custom client is custom, and the provider-specific clients are present.
```


client/packages/cli/src/commands/auth/client/update.ts
```diff
@@ -0,0 +1,786 @@
...
+  heading: 'Custom redirect URI (optional):',
+});
+const newRedirectPrompt = redirectUriPrompt({ heading: 'New redirect URI:' });
+const googleConsoleUrl =
+  'https://console.developers.google.com/apis/credentials';
+
+const resolveClient = Effect.fn(function* (params: {
+  id?: string;
+  name?: string;
+  yes: boolean;
```
drew-harris: yield yes from global opts instead of using param

client/packages/cli/src/commands/auth/client/update.ts
```diff
@@ -0,0 +1,786 @@
...
+});
+
+const selectUpdateAction = Effect.fn(function* <T extends string>(
+  options: { label: string; value: T }[],
+) {
+  return yield* runUIEffect(
+    new UI.Select({
+      options,
+      promptText: 'What do you want to update?',
+      modifyOutput: UI.modifiers.piped([UI.modifiers.dimOnComplete]),
```
drew-harris: No need to use UI.modifiers.piped if you are just using 1 modifier

client/packages/cli/src/commands/auth/client/update.ts
```diff
@@ -0,0 +1,786 @@
...
+const resolveGoogleUpdateMode = Effect.fn(function* ({
+  opts,
+  isWeb,
+  switchingFromShared,
+  yes,
+}: {
+  opts: Record<string, unknown>;
+  isWeb: boolean;
+  switchingFromShared: boolean;
+  yes: boolean;
```
drew-harris: prefer yield yes

client/packages/cli/src/commands/auth/client/update.ts
```diff
@@ -0,0 +1,786 @@
...
+  );
+});
+
+const handleClerkUpdate = Effect.fn(function* (
+  opts: Record<string, unknown>,
+  client: OAuthClientRow,
+) {
+  const { yes } = yield* GlobalOpts;
+  const publishableKey = yield* optOrPromptIf(opts, 'publishable-key', {
+    promptIf: !yes && !hasFlag(opts, 'publishable-key'),
```
drew-harris, line 723: What's the reasoning for the optOrPromptIf, is it just an inverted case of optOrPrompt's skipIf?

client/packages/cli/src/commands/auth/client/update.ts
```diff
@@ -0,0 +1,786 @@
...
+  );
+});
+
+export const authClientUpdateCmd = Effect.fn(
+  function* (
+    opts: OptsFromCommand<typeof authClientUpdateDef> & Record<string, unknown>,
+  ) {
+    const { yes } = yield* GlobalOpts;
+    const id = yield* getOptionalStringFlag(getFlag(opts, 'id'), '--id');
+    const name = yield* getOptionalStringFlag(getFlag(opts, 'name'), '--name');
```
drew-harris: the `opts` object should already have these optional string flags parsed, and already typesafe

drew-harris: Image

client/packages/cli/src/lib/oauth.ts
```diff
@@ -157,6 +157,35 @@ export const addOAuthClient = Effect.fn(function* (params: {
     );
 });
 
+export const updateOAuthClient = Effect.fn(function* (params: {
+  appId?: string;
```
drew-harris: No reason to pass appId since we should always yield it from context.
This parameter is unused currently anyways.

drew-harris (commented):
Image
I think in the dev mode case, this menu should still be displayed with grayed out options that say (can't do this in dev mode)

client/packages/cli/src/commands/auth/client/update.ts
stopachka: Done!

client/packages/cli/src/commands/auth/client/update.ts
stopachka: Done!

client/packages/cli/src/commands/auth/client/update.ts
stopachka: done!

client/packages/cli/src/lib/oauth.ts
stopachka: Done!

client/packages/cli/src/commands/auth/client/update.ts
stopachka: Nice!

client/packages/cli/src/commands/auth/client/update.ts
stopachka: Not quite. The reason is that `update` has a third case that `skipIf` doesn't model.

`optOrPrompt` handles:

1. Parse the flag if it is present
2. If the flag is missing, prompt for it

`skipIf` means the option is not allowed at all. If `skipIf` is true, we should reject the flag and also avoid prompting.

For update, we need a different behavior:

1. Parse the flag if it is present
2. If the flag is missing, prompt only when this update path needs the value
3. If the flag is missing and this update path does not need the value, leave it unset

For example:

```
instant-cli auth client update --name github-prod --custom-redirect-uri https://example.com/callback
```

This should update only the redirect URI. We still want to allow `--client-id` and `--client-secret` if the user provides them, but we don't want to prompt for them when they're omitted.

--- 

I do wonder if we need to re-think the abstraction though. It's a bit hard to reason about. I changed the name to `optOrPromptWhenNeeded`, but perhaps I can think of something a bit different. Let me mull on this

stopachka: Update: I will keep `optOrPromptWhenNeeded` in this PR, but propose a refactor in a separate PR. 

My thinking is that we can 'decompose' optOrPromptIf, into a series of pipe operators: 

```
 const name = yield* parseStringFlag(opts.name, '--name').pipe(
   orAsk(clientNamePrompt),
   orFail('--name'),
 )
```

This could let us be more explicit. 

I don't want to make the change here though, as it would be quite involved.

drew-harris: Good idea

client/packages/cli/src/commands/auth/client/update.ts
```diff
@@ -0,0 +1,853 @@
...
+
+type ProviderRow = {
+  id: string;
+  provider_name: string;
+};
+
+const redirectPrompt = redirectUriPrompt({
+  heading: 'Custom redirect URI (optional):',
+});
+const newRedirectPrompt = redirectUriPrompt({ heading: 'New redirect URI:' });
```
drew-harris, line 51:
```
Checking for an Instant SDK...
Found @instantdb/core in your package.json.

Select a client to update:
● jlaksdfj (4eee42ad-9f77-4149-bc7e-0a972fa972e8)

Current mode: custom credentials
What do you want to update?
● Update redirect URI

New redirect URI:

[error] Missing required value for --custom-redirect-uri
hint: run `instant-cli auth client update --help` for available arguments

Press Enter to close
```
Pressing "update redirect uri" and then hitting enter without typing it, gives an error that looks like a flag is missing.

client/packages/cli/src/commands/auth/client/update.ts
```diff
@@ -0,0 +1,853 @@
...
+import { link } from '../../../logging.ts';
+
+type OAuthClientRow = typeof OAuthClient.Type;
+
+type ProviderRow = {
+  id: string;
+  provider_name: string;
+};
+
+const redirectPrompt = redirectUriPrompt({
```
drew-harris, line 48: Image

Might be nice for this to have a "skipped" indicator in the "answered" state

client/packages/platform/src/clerk.ts
```diff
@@ -0,0 +1,27 @@
...
+  return new TextDecoder().decode(bytes);
+}
+
+export function clerkDomainFromPublishableKey(key: string): string | null {
+  try {
+    const match = key.match(/^pk_[^_]+_(.+)$/);
+    const domainPartB64 = match?.[1];
+    if (!domainPartB64) return null;
+    const domainPart = base64DecodeUtf8(domainPartB64);
+    return domainPart.replace(/\$$/, '');
```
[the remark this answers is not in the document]
stopachka: I think this is fine.

drew-harris (approved):
functionality works

client/packages/cli/src/commands/auth/client/update.ts
stopachka: I think it's ~somewhat reasonable to expect an error like a flag is missing in this case. Here we are explicitly going through "Update redirect URI", but then it is not provided. 

It does expose a problem: what if someone wants to _remove_ the redirect uri? 

Right now we don't allow this. 

Perhaps we should, but I worry it could lead to weird edge cases (like accidentally removing a redirect uri, when the user did not intend it)

client/packages/cli/src/commands/auth/client/update.ts
stopachka: image

fixed!

PR #2627, opened 2026-05-04
code fence fixes

drew-harris:
BEFORE:
image
image

AFTER:
image
image

Also fixes an issue where the copy button scrolls with the content. 

Also fixes an issue where the auth page firebase code snippet starts with a blank line.


nezaj (approved):
LGTM.

As a nit, I think it would be nicer if the copy button wasn't all the way in the corner.

As an example you can see our copy buttons from the docs

Image

drew-harris:
re: spacing

image

Fixed to match the others

PR #2628, opened 2026-05-05
Instrument Apple email claim parsing

stopachka:
A user reported that some apple emails were coming in with a null email fields. 

One theory is that apple can sometimes send `email_verified` as the string "true", rather than a boolean. 

To see that this is the case, I am adding a quick trace. Perhaps this can help us detect the issue. 


dwwoelfel (approved):
LGTM!

PR #2631, opened 2026-05-06
Faster db startup

dwwoelfel:
Two small changes that eliminate a couple of round-trips to the db on connection startup:

1. Set `assumeMinServerVersion` in the jdbc config to `9.0`, which lets jdbc put the application name in the initial startup packet
2. Put the idle timeout in the options flag, saving a round-trip to do a `SET`.

I'm also exploring pre-warming the type-cache. jdbc makes metadata queries to the database to map oid -> typename. It's not shared across the pool, each connection fetches it individually. I have some wip code that caches that data, but it still needs a bit of work to prevent breaking things if a type name gets mapped to a new oid (e.g. drop and then add the same field).

### DIFF
--- a/server/src/instant/jdbc/aurora.clj
+++ b/server/src/instant/jdbc/aurora.clj
@@ -201,17 +201,29 @@
                       (catch Exception _e nil)))
      :get-config (fn [] @current-config)}))
 
+(defn build-startup-options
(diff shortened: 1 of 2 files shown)


stopachka (approved):
Cool!

PR #2633, opened 2026-05-06
CLI email template commands (3/3) (#2633) [bump to v1.0.47]

drew-harris:
Adds `instant-cli auth email` commands to push and pull the email template to/from an instant.email.ts file. 

Added commands
`instant-cli auth email status` - Shows the on-server config for the custom email template. also shows the verification status if using a custom sender.

`instant-cli auth email push` - Push the content from
`instant-cli auth email pull` - Pulls the current state of the custom email template info the `instant.email.ts` file. 
`instant-cli auth email verify <code>` - Verifies the custom email sender magic code. 
`instant-cli auth email reset` - Equivalent to clicking "delete template" on the dashboard, remove the email template. 
`instant-cli auth email resend` - Resend the verification email

https://asciinema.org/a/KiMVaBTKaBSFuQ5T

### DIFF
--- a/client/packages/cli/src/commands/auth/email/pull.ts
+++ b/client/packages/cli/src/commands/auth/email/pull.ts
@@ -86,10 +86,6 @@ const DefaultEmailTemplateSchema = Schema.Struct({
   'sender-email': Schema.String.pipe(Schema.optional),
 });
 
-class MissingDefaultTemplateError extends TaggedError<MissingDefaultTemplateError>(
-  'MissingDefaultTemplateError',
-)('MissingDefaultTemplateError', {}) {}
-
 export const getDefaultEmailTemplate = Effect.gen(function* () {
   const http = yield* InstantHttp;
   const template = yield* http
@@ -101,10 +97,6 @@ export const getDefaultEmailTemplate = Effect.gen(function* () {
     );
 
   const appName = yield* getAppName;
-
-  if (!template) {
-    return yield* MissingDefaultTemplateError.make({});
-  }
 
   return {
     subject: template.subject,
--- a/client/packages/cli/src/commands/auth/email/pull.ts
+++ b/client/packages/cli/src/commands/auth/email/pull.ts
@@ -12,7 +12,6 @@ import type { authEmailPullDef, OptsFromCommand } from '../../../index.ts';
 import { InstantHttp } from '../../../lib/http.ts';
 import { getAppName } from '../../../util/getAppName.ts';
 import type { EmailConfig } from '../../../lib/email.ts';
-import { TaggedError } from 'effect/Schema';
 
 export const authEmailPullCmd = Effect.fn(function* (
   opts: OptsFromCommand<typeof authEmailPullDef>,
--- a/client/packages/cli/src/commands/auth/email/pull.ts
+++ b/client/packages/cli/src/commands/auth/email/pull.ts
@@ -1,5 +1,4 @@
 import { HttpClientResponse, Path } from '@effect/platform';
-import { defaultMagicCodeEmailConfig } from '@instantdb/platform';
 import { Effect, Schema } from 'effect';
 import { ProjectInfo } from '../../../context/projectInfo.ts';
 import { BadArgsError } from '../../../errors.ts';
@@ -13,6 +12,7 @@ import type { authEmailPullDef, OptsFromCommand } from '../../../index.ts';
 import { InstantHttp } from '../../../lib/http.ts';
 import { getAppName } from '../../../util/getAppName.ts';
 import type { EmailConfig } from '../../../lib/email.ts';
+import { TaggedError } from 'effect/Schema';
 
 export const authEmailPullCmd = Effect.fn(function* (
   opts: OptsFromCommand<typeof authEmailPullDef>,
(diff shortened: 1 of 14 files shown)


client/packages/cli/src/commands/auth/email/status.ts
```diff
@@ -0,0 +1,107 @@
...
+  return app.info;
+});
+
+export const emailStatusCmd = Effect.fn(function* (
+  opts: OptsFromCommand<typeof authEmailStatusDef>,
+) {
+  const info = yield* getEmailTemplateStatus;
+
+  if (!info) {
+    yield* Effect.log('No custom magic code email associated with this app.');
```
stopachka: I wonder if we should have something more helpful here. Like, "to add one, do X"

WDYT @drew-harris ?

client/packages/platform/src/email.ts
```diff
@@ -0,0 +1,35 @@
...
+          Note: This code will expire in 10 minutes, and can only be used once. If
+          you didn't request this code, please reply to this email.
+        </p>
+      </div>
+    </div>
+  </div>
+</div>
+`;
+
+export const defaultMagicCodeEmailSubject =
```
stopachka: Maybe it would be better if this stuff came directly from the server? 

Otherwise this can get easy to drift

client/packages/platform/src/email.ts
```diff
@@ -0,0 +1,35 @@
...
+  '{code} is your code for {app_title}';
+
+export const defaultMagicCodeEmailConfig = {
+  authEmail: {
+    subject: defaultMagicCodeEmailSubject,
+    from: '',
+    fromAddress: undefined,
```
stopachka: maybe we should call this customFromAddress, so it's a it more clear that you don't _have_ to provide it.

client/packages/cli/src/commands/auth/email/pull.ts
```diff
@@ -0,0 +1,115 @@
+import { Path } from '@effect/platform';
```
stopachka: For the pull, I wonder if initially we should add a little comment block, which explains the variables that are available. (like we do in the dashboard)

client/packages/cli/src/commands/auth/email/pull.ts
```diff
@@ -0,0 +1,115 @@
...
+  file?: string;
+};
+
+export const authEmailPullCmd = Effect.fn(function* (opts: AuthEmailPullOpts) {
+  yield* pullEmail(opts.file);
+});
+
+export type EmailConfig = {
+  subject: string;
+  from: string;
```
stopachka: Maybe we call this `fromName` or something?

stopachka: I realized this doesn't work this way in the email template on dash, but imo we should say what the default value is

client/packages/cli/src/commands/auth/email/push.ts
```diff
@@ -0,0 +1,10 @@
+import { Effect } from 'effect';
+import { pushEmail } from '../../../lib/email.ts';
+
+type AuthEmailPushOpts = {
+  file?: string;
+};
+
+export const authEmailPushCmd = Effect.fn(function* (opts: AuthEmailPushOpts) {
+  yield* pushEmail(opts.file);
```
stopachka: Image

I pushed a custom email, and was expected I would get some notice like "you need to verify it", but I didn't see that.

client/packages/cli/src/lib/email.ts
```diff
@@ -0,0 +1,150 @@
...
+        boxen(
+          "We've sent a confirmation email containing a six digit code to verify the sender email address.\nUse instant-cli auth email verify <code> to complete verification.",
+          {
+            borderColor: 'yellow',
+            padding: { right: 1, left: 1 },
+          },
+        ),
+      );
+      return;
+    }
```
stopachka: I also wonder if we should put in somewhere here the details for the MX records they should set, etc. 

WDYT @drew-harris ?

client/packages/platform/src/email.ts
drew-harris: {:sender-name (:title app)
                           :sender-email default-sender-email
                           :subject (str code " is your verification code for " (:title app))
                           :body (default-body template-params)})

its hardcoded like this in the send! function, would need to add a new route just to get the subject line and body. Not sure if thats worth it?

client/packages/cli/src/lib/email.ts
drew-harris: good idea

client/packages/cli/src/commands/auth/email/push.ts
drew-harris: not able to reproduce this

client/packages/platform/src/email.ts
stopachka: Let's call this `sendName` and `sendEmail`? Maybe easier to understand

stopachka:
- Naming: `sendName`, `senderEmail` DONE
- Let's make a route for default sender name, email, and html~~ DONE
- One more naming pass with Claude DONE

stopachka (approved):
Ship!!

PR #2635, opened 2026-05-06
[getadb] use Sec-Fetch-Mode to detect browser navigation

stopachka:
- Replace the `Accept: text/html` heuristic with `Sec-Fetch-Mode` for routing the getadb.com root between the HTML page and the markdown guide.
- The old check misclassified Claude's fetch tool (and other agents that accept `text/html`) as humans, so they got the HTML page instead of the markdown guide.
- `Sec-Fetch-Mode` is only sent by browsers on top-level navigation (Chrome 76+, Firefox 90+, Safari 16.4+). JS `fetch()` sends `cors`, and agents/curl don't send the header at all — both fall through to the guide.

### DIFF
--- a/client/www/middleware.ts
+++ b/client/www/middleware.ts
@@ -19,8 +19,12 @@ export function middleware(request: NextRequest) {
 
   if (url.pathname === '/') {
     // Browser → human page; curl/agent → markdown guide.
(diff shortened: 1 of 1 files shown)


dwwoelfel (approved):
LGTM!

PR #2638, opened 2026-05-08
Fix delay being used as a function

dwwoelfel:
Fixes a bug where a delay was being used as a function. 

I just made it a function again. I thought it was a good idea to make it a delay, since it will never change, but it's only used once when we create the s3 clients, so it doesn't really matter.


stopachka (approved):
Thank you!

PR #2639, opened 2026-05-08
Better trace propagation

dwwoelfel:
Includes traceId in the `error` we return from `useQuery` to make it easier for users to report errors.

Also makes sure that we pass the traceId through any transact errors.

Finally, we include the trace-id from the original message (separated by a `,`) to make it easier to look up the event that caused the error. In the backend, we'll also annotate the `handle-receive` with the trace-id of the original message for errors.


stopachka (approved):
SGTM!

PR #2644, opened 2026-05-11
Move self hosting docs to www/

drew-harris:
- moves self hosting docs to www/
 - small changes to make look better with formatting / more professional
 - adds link to the docs site in the old readme to ensure single source of truth.


stopachka (approved):
Ship it!

PR #2645, opened 2026-05-11
Use www docs URLs in getadb rules

stopachka:
**Context** 

We had links in our AGENTS.md that pointed to `https://instantdb.com/*` for docs. 

**Problem** 

If an agent curls that url, it will get a "Redirecting..." message back. After that agents generally do `-sL` to follow the redirect, but that still makes it two tool calls

**Solution** 

I updated to make include `www` in all the rules, so curls would not need a redirect. I also ran `pnpm exec tsx scripts/gen-llm-rules.ts` to update all the rule files, as afaik create-instant-app still needs this to run manually


dwwoelfel (approved):
Cool!

PR #2648, opened 2026-05-11
Fix instant-cli startup with strict pnpm peers

stopachka:
## Problem

```sh
mkdir /tmp/cli-repro && cd /tmp/cli-repro
pnpx instant-cli@1.0.26 --version
```

`instant-cli@1.0.26` crashes on startup, with the error:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package '@effect/cluster'
```

## Root cause

`instant-cli` had this import:

```ts
import { NodeContext, NodeHttpClient } from '@effect/platform-node';
```

That hits the `@effect/platform-node` **top-level barrel**, which eagerly re-exports cluster-aware modules like `NodeClusterHttp`. ESM module linking evaluates those re-exports at import time, so resolving the barrel forces Node to also resolve `@effect/cluster`, `@effect/rpc`, `@effect/sql`, and `@effect/workflow`.

None of those are declared in the CLI's `dependencies`. They're peer-style deps of `@effect/platform-node` itself, not of us. In a strict pnpm install (no auto-install-peers, no hoisting) they aren't on disk, so resolution fails before the CLI's `main` even runs.

This was easy to miss in the Instant monorepo because the workspace lockfile already contains those packages, so the barrel resolves fine locally. The bug only surfaces in a clean consumer install.

### Fix

Import only the specific submodules the CLI uses:

```ts
import * as NodeContext from '@effect/platform-node/NodeContext';
import * as NodeHttpClient from '@effect/platform-node/NodeHttpClient';
```

Both subpaths are real entries in `@effect/platform-node`'s `exports` map, so `tsc --noEmit` validates the paths at typecheck time. This avoids evaluating the barrel and therefore avoids loading any cluster/RPC code during module linking.

Same swap applied in 5 unit tests and `__tests__/effectHelpers.ts` so the test runtime matches production.


dwwoelfel (approved):
Looks good to me! 

I'm surprised effect would do that--you'd think many more people would hit this error...

PR #2650, opened 2026-05-12
docs: fix manual setup guide to use import.meta.env for a public env on a vite project

bryanprimus:
the issue

> Something went wrong!
Instant must be initialized with an appId.

the fix:
in tanstack start which is a vite project, public env with prefix `VITE_` should be read with `import.meta.env`.

more on this here

https://tanstack.com/start/v0/docs/framework/react/guide/environment-variables

this will reduce confusion/issue on manual setup.


nezaj (approved):
Thank you!

PR #2651, opened 2026-05-12
backfill email verifications migration

drew-harris:
Creates a new migration that fills existing app <-> sender_id pairs that were created via the user-level email sender authentication into the new `app_email_verifications` table. 

# Important!
Do not run migration until https://github.com/instantdb/instant/pull/2647 has been fully deployed.


dwwoelfel (approved):
LGTM!

PR #2653, opened 2026-05-12
Add useInfiniteQuery to Solid and Svelte SDKs

nezaj:
The Solid and Svelte SDKs were missing `useInfiniteQuery`. #2334 added `subscribeInfiniteQuery` to core and `useInfiniteQuery` to React, but the same hook hasn't been wired up to the other framework SDKs. This brings all three to parity.

**API**

Both implementations mirror React's signature: pass a query (or a function returning a query) plus optional opts, get back `data` / `error` / `isLoading` / `canLoadNextPage` / `loadNextPage()`. Function-form queries re-subscribe when their reactive dependencies change, and subscriptions clean up on unmount.

```ts
// Solid
const state = db.useInfiniteQuery({
  posts: { $: { limit: 20, order: { createdAt: 'desc' } } },
});
// state().data, state().canLoadNextPage, state().loadNextPage()

// Svelte
const state = db.useInfiniteQuery({
  posts: { $: { limit: 20, order: { createdAt: 'desc' } } },
});
// state.data, state.canLoadNextPage, state.loadNextPage()
```

**Tests**

Added some Svelte tests for the new hook in the existing mocked test suite. Solid has no wrapper test infrastructure in this package yet so not tests there. Instead in #2654 I add sandbox apps to verify both work as expected

### DIFF
--- a/client/packages/solidjs/src/InstantSolidDatabase.ts
+++ b/client/packages/solidjs/src/InstantSolidDatabase.ts
@@ -10,6 +10,7 @@ import {
   type PageInfoResponse,
   type InstaQLLifecycleState,
   type InstaQLResponse,
+  type InfiniteQuerySubscription,
   type ValidQuery,
   // classes
   Auth,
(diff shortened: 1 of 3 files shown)


stopachka (approved):
SGTM!

nezaj:
Merged https://github.com/instantdb/instant/pull/2654 which contained these changes as well!

PR #2672, opened 2026-05-14
fix dark mode webhook ui colors

drew-harris:
Use "neutral" color palette in dark mode + value tweaks. 

BEFORE:
image
image
image

AFTER
image
image
image


nezaj (approved):
Nice!

PR #2673, opened 2026-05-14
Rename etype to namespace for user-facing webhooks code

dwwoelfel:
Uses `namespaces` instead of `etypes` for webhooks, since we don't ever call them `etypes` anywhere public-facing.

Technically a breaking change, but we haven't announced webhooks and nobody is using them yet.

After this is merged, I'll update the docs PR and the cli PR


stopachka (approved):
Niice!

PR #2675, opened 2026-05-15
fix info command

drew-harris:
Was incorrectly trying to use the admin token even if we already had it + route wouldn't accept admin token.

Simplified CLI code + allowed GET /dash/app/:app-id to accept an admin token. 

# includes backend change, deploy required

### DIFF
--- a/client/packages/cli/src/commands/info.ts
+++ b/client/packages/cli/src/commands/info.ts
@@ -1,10 +1,6 @@
-import {
-  HttpClient,
-  HttpClientRequest,
-  HttpClientResponse,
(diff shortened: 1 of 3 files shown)


nezaj (approved):
LGTM with one question.

Saw you marked the code rabbit as resolved but didn't make the change. Was just curious why? My understanding is that it thought it would be nice to give the app id in the error message instead of the regular http error. But maybe not worth it?

drew-harris:
RE joe: The previous code didn't show any errors at all, even if we knew there was a valid app context but somehow failed to fetch.

PR #2676, opened 2026-05-15
Validate fields is an array

dwwoelfel:
Fixes a bug where throw an unknown error if we get a `map` for fields, e.g.

```js
{books: {$: {fields: {}}}}
```

Saw this error pop up in production early this morning.


nezaj (approved):
Nice!

stopachka (approved):
Cool!

PR #2679, opened 2026-05-15
Move custom email verification to app-level (1/3)

drew-harris:
This PR migrates the current user-based system for custom oauth email senders to use app ids instead. This adds another on top of the postmark email to verify sender. We will also send a magic code to the user so that they can link the sender's identity with the app. 

This is gated behind the boolean feature flag: `use-app-email-verification?`

Added Routes
POST /dash/apps/:id/sender-verification/send-magic-code
POST /dash/apps/:id/sender-verification/verify-magic-code

I also updated the frontend to include the ui needed for seeing the status of both verification steps and submitting the magic code / resending. It is backwards compatible with the backend feature flag. 

### Important! Includes database migrations which must be run before turning on the feature flag. 

# Deploy Plan
1. merge pr
3. run migrations included in pr
4. deploy backend
5. toggle `use-app-email-verification?` feature flag to true

How to test:
image
Turn on the feature flag. 

Try adding / changing an oauth email template with a custom sender address on the dashboard.

image

### DIFF
--- a/client/www/components/dash/auth/Email.tsx
+++ b/client/www/components/dash/auth/Email.tsx
@@ -21,6 +21,7 @@ import { errorToast, successToast } from '@/lib/toast';
 import clsx from 'clsx';
 import { useFetchedDash } from '../MainDashLayout';
 import { useDarkMode } from '../DarkModeToggle';
+import { useCustomSenderVerification } from '@/lib/hooks/useCustomSenderVerification';
 
 export type EmailValues = {
   from: string;
@@ -41,58 +42,13 @@ export type SenderVerificationInfo = {
   ReturnPathDomainCNAMEValue: string;
 };
 
-export function getSenderVerification({
-  token,
-  appId,
-}: {
-  token: string;
-  appId: string;
-}): Promise<{
-  senderEmail: string;
-  verification: SenderVerificationInfo | null;
-}> {
-  return jsonFetch(`${config.apiURI}/dash/apps/${appId}/sender-verification`, {
-    method: 'GET',
-    headers: {
-      authorization: `Bearer ${token}`,
-      'content-type': 'application/json',
-    },
-  });
-}
-
 export function Email({ app }: { app: InstantApp }) {
   const dashResponse = useFetchedDash();
   const template = app.magic_code_email_template;
   const token = useContext(TokenContext);
   const [isEditing, setIsEditing] = useState(Boolean(template) ?? false);
-  const [{ isVerifying, verification }, setVerification] = useState<{
-    isVerifying: boolean;
-    verification: SenderVerificationInfo | null;
-  }>({
-    isVerifying: false,
-    verification: null,
-  });
-
+  const verification = useCustomSenderVerification(app);
   const { darkMode } = useDarkMode();
-
-  const checkVerification = async () => {
-    setVerification((prev) => ({ ...prev, isVerifying: true }));
-    try {
-      const response = await getSenderVerification({
-        token,
-        appId: app.id,
-      });
-      setVerification((prev) => ({
-        ...prev,
-        verification: response.verification,
-      }));
-    } catch (error) {
-      console.error('Failed to check verification:', error);
-      errorToast('Failed to check verification status');
-    } finally {
-      setVerification((prev) => ({ ...prev, isVerifying: false }));
-    }
-  };
 
   async function onSubmit(values: EmailValues) {
     return dashResponse
@@ -112,10 +68,13 @@ export function Email({ app }: { app: InstantApp }) {
         ),
       )
       .then(
-        () => {
+        async () => {
           successToast('Email template saved!');
           if (values.senderEmail) {
-            checkVerification();
+            const res = await verification.refetch();
+            if (!res?.instant?.['verified?']) {
+              verification.sendCode();
+            }
           }
         },
         (errorRes) =>
@@ -146,7 +105,7 @@ export function Email({ app }: { app: InstantApp }) {
 
   useEffect(() => {
     if (template?.email) {
-      checkVerification();
+      verification.refetch();
     }
   }, []);
 
@@ -242,16 +201,16 @@ export function Email({ app }: { app: InstantApp }) {
           />
         </div>
 
-        {verification && (
+        {verification.raw.data?.verification && (
           <div className="flex flex-col gap-2 rounded-sm border bg-gray-50 p-3 dark:border-neutral-700 dark:bg-neutral-800">
             <div className="flex items-center justify-between">
               <SubsectionHeading>
-                Verify {verification.EmailAddress}
+                Verify {verification.raw.data?.verification?.EmailAddress}
               </SubsectionHeading>
               <Button
                 type="button"
-                onClick={checkVerification}
-                loading={isVerifying}
+                onClick={() => verification.refetch()}
+                loading={verification.loading}
                 variant="primary"
                 size="mini"
               >
@@ -261,13 +220,15 @@ export function Email({ app }: { app: InstantApp }) {
 
             <div className="rounded-sm border bg-white p-4 dark:border-neutral-700 dark:bg-neutral-700/60">
               <div className="mb-2 flex items-center justify-between">
-                <div className="text-sm font-medium">Email Confirmation</div>
+                <div className="text-sm font-medium">
+                  Postmark Email Confirmation
+                </div>
                 <div className="flex items-center gap-2">
                   <StatusCircle
-                    isLoading={isVerifying}
-                    isSuccess={verification.Confirmed}
+                    isLoading={verification.loading}
+                    isSuccess={verification.postmarkVerified}
                   />
-                  {verification.Confirmed ? (
+                  {verification.postmarkVerified ? (
                     <div className="text-xs font-medium text-green-600">
                       Confirmed
                     </div>
@@ -279,10 +240,33 @@ export function Email({ app }: { app: InstantApp }) {
                 </div>
               </div>
               <Content className="text-sm text-gray-600">
-                {verification.Confirmed
-                  ? `Great! You've confirmed ${verification.EmailAddress} and can now send emails from this address.`
-                  : `We've sent a confirmation email to ${verification.EmailAddress}. Please click the link in that email to confirm ownership.`}
+                {verification.postmarkVerified
+                  ? `Great! You've confirmed ${verification.raw.data.verification?.EmailAddress} with Postmark.`
+                  : `We've sent a confirmation email to ${verification.raw.data.verification?.EmailAddress}. Please click the link in that email to confirm ownership.`}
               </Content>
+              <div className="mt-4 flex items-center justify-between">
+                <div className="text-sm font-medium">
+                  Instant Email Confirmation
+                </div>
+                <div className="flex items-center gap-2">
+                  <StatusCircle
+                    isLoading={verification.loading}
+                    isSuccess={verification.instantVerified}
+                  />
+                  {verification.instantVerified ? (
+                    <div className="text-xs font-medium text-green-600">
+                      Confirmed
+                    </div>
+                  ) : (
+                    <div className="text-xs text-gray-500 dark:text-neutral-400">
+                      Pending confirmation
+                    </div>
+                  )}
+                </div>
+              </div>
+              {!verification.instantVerified ? (
+                <SenderOtpVerification verification={verification} />
+              ) : null}
             </div>
 
             {/* Domain Verification */}
@@ -317,7 +301,8 @@ export function Email({ app }: { app: InstantApp }) {
                         Hostname:
                       </div>
                       <code className="block rounded-sm bg-gray-100 px-2 py-1 text-xs break-all select-all dark:bg-neutral-700">
-                        {verification.DKIMPendingHost || verification.DKIMHost}
+                        {verification.raw.data?.verification?.DKIMPendingHost ||
+                          verification.raw.data?.verification?.DKIMHost}
                       </code>
                     </div>
                     <div>
@@ -325,8 +310,9 @@ export function Email({ app }: { app: InstantApp }) {
                         Value:
                       </div>
                       <code className="block rounded-sm bg-gray-100 px-2 py-1 text-xs break-all select-all dark:bg-neutral-700">
-                        {verification.DKIMPendingTextValue ||
-                          verification.DKIMTextValue}
+                        {verification.raw.data?.verification
+                          ?.DKIMPendingTextValue ||
+                          verification.raw.data?.verification?.DKIMTextValue}
                       </code>
                     </div>
                   </div>
@@ -344,7 +330,7 @@ export function Email({ app }: { app: InstantApp }) {
                         Hostname:
                       </div>
                       <code className="block rounded-sm bg-gray-100 px-2 py-1 text-xs break-all select-all dark:bg-neutral-700">
-                        {verification.ReturnPathDomain}
+                        {verification.raw.data?.verification?.ReturnPathDomain}
                       </code>
                     </div>
                     <div>
@@ -352,7 +338,10 @@ export function Email({ app }: { app: InstantApp }) {
                         Value:
                       </div>
                       <code className="block rounded-sm bg-gray-100 px-2 py-1 text-xs break-all select-all dark:bg-neutral-700">
-                        {verification.ReturnPathDomainCNAMEValue}
+                        {
+                          verification.raw.data?.verification
+                            ?.ReturnPathDomainCNAMEValue
+                        }
                       </code>
                     </div>
                   </div>
@@ -391,6 +380,54 @@ export function Email({ app }: { app: InstantApp }) {
       </form>
 
       <MagicCodeExpirationSection app={app} />
+    </div>
+  );
+}
+
+function SenderOtpVerification({
+  verification,
+}: {
+  verification: ReturnType<typeof useCustomSenderVerification>;
+}) {
+  const [code, setCode] = useState('');
+  const address = verification.raw.data?.verification?.EmailAddress;
+
+  return (
+    <div className="mt-2 rounded-sm border bg-gray-50 p-3 dark:border-neutral-600 dark:bg-neutral-800/60">
+      <Content className="mb-3 text-sm text-gray-600 dark:text-neutral-300">
+        {verification.justSentCode
+          ? `We emailed ${address} a six-digit code to verify the email address with this app. Check your inbox for the code.`
+          : `Send a one-time passcode to ${address}, then enter the 6 digits here to confirm you can receive email at this address.`}
+      </Content>
+      <div className="flex flex-col gap-2 sm:flex-row sm:items-end">
+        <Button
+          type="button"
+          variant="secondary"
+          loading={verification.verifyCode.isMutating}
+          onClick={() => verification.sendCode()}
+        >
+          {verification.justSentCode ? 'Resend code' : 'Send code'}
+        </Button>
+        <div className="flex flex-1 gap-2">
+          <TextInput
+            label="Verification code"
+            value={code}
+            onChange={(e) => setCode(e)}
+            placeholder="123456"
+            inputMode="numeric"
+          />
+          <Button
+            type="button"
+            variant="primary"
+            loading={verification.verifyCode.isMutating}
+            disabled={code.length !== 6}
+            onClick={() => verification.verifyCode.trigger(code)}
+            className="self-end"
+          >
+            Verify
+          </Button>
+        </div>
+      </div>
     </div>
   );
 }
--- a/client/www/components/dash/auth/Email.tsx
+++ b/client/www/components/dash/auth/Email.tsx
@@ -421,7 +421,24 @@ function SenderOtpVerification({
           type="button"
           variant="secondary"
           loading={verification.verifyCode.isMutating}
-          onClick={() => verification.sendCode()}
+          onClick={() =>
+            verification
+              .sendCode()
+              .then(() => {
+                successToast(`Verification code sent to ${address}`);
+              })
+              .catch((error) => {
+                if (error?.body?.message) {
+                  errorToast(
+                    `Failed to send verification code: ${error.body.message}`,
+                  );
+                } else {
+                  errorToast(
+                    `Failed to send verification code: ${error instanceof Error ? error.message : String(error)}`,
+                  );
+                }
+              })
+          }
         >
           {verification.justSentCode ? 'Resend code' : 'Send code'}
         </Button>
--- a/client/www/components/dash/auth/Email.tsx
+++ b/client/www/components/dash/auth/Email.tsx
@@ -258,7 +258,7 @@ export function Email({ app }: { app: InstantApp }) {
               </div>
               <Content className="text-sm text-gray-600">
                 {!verification.postmarkVerified &&
-                  `We've sent a confirmation email to ${verification.raw.data.verification?.EmailAddress}. Please click the link in that email to confirm ownership.`}
+                  `Postmark has sent a confirmation email to ${verification.raw.data.verification?.EmailAddress}. Please click the link in that email to confirm ownership.`}
               </Content>
               <div className="mt-4 flex items-center justify-between">
                 <div className="text-sm font-medium">
--- a/client/www/components/dash/auth/Email.tsx
+++ b/client/www/components/dash/auth/Email.tsx
@@ -73,7 +73,24 @@ export function Email({ app }: { app: InstantApp }) {
           if (values.senderEmail) {
             const res = await verification.refetch();
             if (!res?.instant?.['verified?']) {
-              verification.sendCode();
+              verification
+                .sendCode()
+                .then(() => {
+                  successToast(
+                    `Verification code sent to ${values.senderEmail}`,
+                  );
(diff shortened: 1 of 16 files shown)


server/resources/migrations/111_add_app_email_verification_codes.up.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE app_email_verification_codes (
+    id UUID PRIMARY KEY,
+    code text NOT NULL,
+    verification_id UUID NOT NULL references app_email_verifications(id) on delete cascade,
```
dwwoelfel, line 5: You might want to put an index on `verification_id` so that the foreign key reference doesn't slow down deleting an app_email_verification.

server/src/instant/app_email_verification.clj
```diff
@@ -0,0 +1,19 @@
+(ns instant.app-email-verification
+  (:require
+   [instant.config :as config]
+   [instant.postmark :as postmark]
+   [instant.util.email :as email]))
+
+(defn send-code-email! [{:keys [code sender-email]}]
```
stopachka: There's two small nits here: 

**1. the better abstraction here is the email structure, not the effect**

I think the function to really extract is the internal email here, something like `sender-verification-email` 

Reasoning: 
1. Most of the logic here is about the email
2. The email is stateless, and easy to run in a repl 
3. But by sticking the postmark/send-structured! here, the only way to test the email code is to run it 

**2. This namespace is likely unnecessary** 

Afaik this is only used in the routes file. We can likely just have a function there that generates the email, similar to what we have for magic codes. 

The reason I wouldn't move this into a namespace, because the names are a bit confusing. This is the namespace `app_email_verification` with only 1 fn, and another one called `app_email_verification_code`

If we did want it in an ns I would likely just move it into app_email_verification_code

server/src/instant/model/app_email_verification.clj
```diff
@@ -6,7 +6,49 @@
...
+          ON CONFLICT (app_id, sender_id) DO UPDATE SET sender_id = EXCLUDED.sender_id
+          RETURNING id, verified"
+                           (random-uuid) app-id sender-id verified])))
+
+(defn get-from-app-and-sender
+  "gets the email verification for the given app and sender"
+  [app-id sender-id]
+  (sql/execute-one! (aurora/conn-pool :read) ["SELECT * FROM app_email_verifications WHERE app_id = ? AND sender_id = ?" app-id sender-id]))
+
+(defn verify!
```
stopachka: nit: I would make this function `mark-verified!`

Mainly because the word `verify!` can be a bit confusing -- it may imply that we are doing the act of verification inside this function.

server/src/instant/model/app_email_verification.clj
```diff
@@ -6,7 +6,49 @@
...
-     (id, app_id, sender_id, verified)
-     VALUES (?::uuid, ?, ?, ?) ON CONFLICT DO NOTHING" (random-uuid) app-id sender-id verified])))
+   (sql/execute-one! conn ["INSERT INTO app_email_verifications
+          (id, app_id, sender_id, verified)
+          VALUES (?::uuid, ?, ?, ?)
+          ON CONFLICT (app_id, sender_id) DO UPDATE SET sender_id = EXCLUDED.sender_id
+          RETURNING id, verified"
+                           (random-uuid) app-id sender-id verified])))
+
+(defn get-from-app-and-sender
```
stopachka: nit: I would use the word `by`, it's not really getting it _from_

server/src/instant/model/app_email_verification.clj
```diff
@@ -6,7 +6,49 @@
 (defn put!
   ([params] (put! (aurora/conn-pool :write) params))
   ([conn {:keys [app-id sender-id verified]}]
-   (sql/execute-one! conn ["INSERT INTO
-     app_email_verifications
-     (id, app_id, sender_id, verified)
-     VALUES (?::uuid, ?, ?, ?) ON CONFLICT DO NOTHING" (random-uuid) app-id sender-id verified])))
+   (sql/execute-one! conn ["INSERT INTO app_email_verifications
```
stopachka: May be nice to re-format this a bit, so the sql is easier to read.

server/resources/migrations/111_add_app_email_verification_codes.up.sql
```diff
@@ -0,0 +1,11 @@
...
+    created_at TIMESTAMP NOT NULL DEFAULT NOW()
+);
+
+create trigger update_updated_at_trigger
```
stopachka: nit: there is not `updated_at` column above.

server/src/instant/dash/routes.clj
```diff
@@ -1364,20 +1371,65 @@
...
+                                      :email-type "magic-code"}]})
+        code (app-user-magic-code-model/rand-code)
+        _ (app-email-verification-code/put!
+           {:code code
+            :verification-id verification-id})]
+    (app-email-verification/send-code-email! {:code code
+                                              :sender-email sender-email})
+    (response/ok {:sent true})))
+
+(defn email-sender-verification-verify
```
stopachka: cc @dwwoelfel too, as I am a bit light on rate limits, but afaik we should have them here?

server/src/instant/model/app_email_verification.clj
drew-harris: how would you write it? this feels like the clearest way i could get it, factoring in code formatting.

server/src/instant/app_email_verification.clj
drew-harris: 👍 choosing to put the formatting fn in app_email_verification code

server/src/instant/dash/routes.clj
drew-harris: still working on those

stopachka (commented):
Overall LGTM! 

One higher level issue I had was with the naming. 

## Frontend

- The world `verification` is stacked with the word `verify`. 
  - `verifySenderVerificationCode`
- The other names are long and slightly different, which makes it hard to understand the data model
 - For example, consider  `getSenderVerification` and `sendSenderVerificationCode`
   - I guess `SenderVerification` is the object, while `SenderVerificationCode` is just the code linked to the object. But this can be hard to parse out 

**Suggestion**

Maybe we cross out the word verification from the codes:

- `getSenderVerification` stays 
- The rest are `sendSenderMagicCode` and `verifySenderMagicCode`

## HTTP

- The HTTP endpoint has the same issue of overloading `verify` with `verification`. 

**Suggestion** 

Maybe we follow the same idea: 

```
  GET    /dash/apps/:app_id/sender-verification
  POST   /dash/apps/:app_id/sender-verification/send-magic-code
  POST   /dash/apps/:app_id/sender-verification/verify-magic-code
```

This drops some of the pretense on REST, which imo is fine for the sake of clarity.

You can then change the handler names too, to fns like `sender-verification-send-magic-code` etc

The response structure is also confusing: 
- `sender-verification` is a boolean, when the expectation would be an object

**Suggestion** 

Maybe we return a response like: 

```
:sender { email }
:postmark { id, ... }
:instant { 'verified?' }
```

client/www/lib/hooks/useCustomSenderVerification.ts
```diff
@@ -0,0 +1,134 @@
...
+  const sendCode = async () => {
+    const rep = await sendSenderVerificationCode({ token, appId: app.id });
+    if (!rep.sent) {
+      throw new Error('Failed to send verification code');
+    }
+    setJustSentCode(true);
+    successToast(
+      `Verification code sent to ${resp.data?.verification?.EmailAddress}`,
+    );
+  };
```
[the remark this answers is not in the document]
stopachka: Seems reasonable @drew-harris

server/src/instant/dash/routes.clj
```diff
@@ -1,85 +1,91 @@
...
-            [instant.util.posthog :as posthog]
-            [instant.util.tracer :as tracer]
-            [instant.util.url :as url-util]
-            [instant.util.uuid :as uuid-util]
-            [instant.webhook-processor :as webhook-processor]
-            [medley.core :as medley]
-            [next.jdbc :as next-jdbc]
-            [ring.middleware.cookies :refer [wrap-cookies]]
-            [ring.util.http-response :as response])
+  (:require
```
stopachka: nit: would move [clj-http .. up to :require, so this doesn't auto-format and change so much

server/src/instant/dash/routes.clj
```diff
@@ -1,85 +1,91 @@
...
+   [instant.isn :as isn]
+   [instant.jdbc.aurora :as aurora]
+   [instant.lib.ring.websocket :as ws]
+   [instant.machine-summaries :as machine-summaries]
+   [instant.model.app :as app-model]
+   [instant.model.app-admin-token :as app-admin-token-model]
+   [instant.model.app-authorized-redirect-origin :as app-authorized-redirect-origin-model]
+   [instant.model.app-email-sender :as app-email-sender-model]
+   [instant.model.app-email-template :as app-email-template-model]
+   [instant.model.app-email-verification :as verification]
```
stopachka: would call this app-email-verification, as `verification` is too general

server/src/instant/dash/routes.clj
```diff
@@ -1337,10 +1343,11 @@
 
 (defn sender-verification-get [req]
   (let [{{app-id :id} :app} (req->app-and-user! :admin req)
-        {postmark-id :postmark_id}
-        (app-email-template-model/get-by-app-id-and-email-type
+        {postmark-id :postmark_id verification :verification_verified}
```
stopachka: would call this `instant-verified?` or something, as it is a bool

server/src/instant/dash/routes.clj
```diff
@@ -1364,20 +1371,69 @@
         sender-email (email/coerce (get-in req [:body :sender-email])) ;; optional
         custom-sender-name (string-util/coerce-non-blank-str (get-in req [:body :sender-name])) ;; optional
         sender-name (or custom-sender-name (:title app))
-        sender (when sender-email
-                 (app-email-sender-model/sync-sender!
-                  {:app-id (:id app)
-                   :user-id (:id user)
-                   :email sender-email
-                   :name sender-name}))
+        {sender :sender needs-verify :needs-verify} (when sender-email
```
stopachka: would call this `needs-verification?` or `needs-verify?`

server/src/instant/dash/routes.clj
```diff
@@ -1364,20 +1371,69 @@
...
+    (response/ok {:id (:id template) :needs-verify needs-verify})))
+
+(defn sender-verification-send-magic-code
+  "sends the email containing a code to verify a custom sender domain with an app id"
+  [req]
+  (let [app-id (:id (:app (req->app-accepting-superadmin-or-ref-token! :admin :apps/write req)))
+        verification-info (verification/get-by-app-id-and-email-type-with-template
+                           {:app-id app-id :email-type "magic-code"})
+        verification-id (:verification_id verification-info)
+        _ (ex/assert-record! verification-id
```
stopachka: would write a `get-by-app-id-and-email-type-with-template!` function in the samespace where `get-by-app-id-and-email-type-with-template` is defined, and include the assert record there.

server/src/instant/dash/routes.clj
```diff
@@ -1364,20 +1371,69 @@
...
+                             :app-email-verification
+                             {:args [{:app-id app-id
+                                      :email-type "magic-code"}]})
+        sender-email (:email verification-info)
+        _ (check-custom-sender-rate-limit! {:email sender-email})
+        verified (next-jdbc/with-transaction [tx-conn (aurora/conn-pool :write)]
```
stopachka: nit: would make this `verified?`

server/src/instant/model/app_email_verification.clj
stopachka: One rough example: 

```
(defn put!
  ([params] (put! (aurora/conn-pool :write) params))
  ([conn {:keys [app-id sender-id verified]}]
   (sql/execute-one! 
      conn 
      ["INSERT INTO app_email_verifications
        (id, app_id, sender_id, verified)
         VALUES (?::uuid, ?, ?, ?)
         ON CONFLICT (app_id, sender_id) DO UPDATE SET sender_id = EXCLUDED.sender_id
         RETURNING id, verified"
       (random-uuid) app-id sender-id verified])))
```

server/src/instant/runtime/routes.clj
```diff
@@ -372,19 +372,18 @@
...
    :headers {"content-type" "text/html"}
    :body (str (h/html (h/raw "<!DOCTYPE html>")
-                [:html {:lang "en"}
-                 [:head
-                  [:meta {:charset "UTF-8"}]
-                  [:meta {:name "viewport"
-                          :content "width=device-width, initial-scale=1.0"}]
-                  [:title "Test redirect"]
-                  [:style "
+                      [:html {:lang "en"}
```
stopachka, line 380: weird change in spacing

stopachka (approved):
Overall LGTM!

server/resources/migrations/111_add_app_email_verification_codes.up.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE app_email_verification_codes (
+    id UUID PRIMARY KEY,
+    code text NOT NULL,
+    app_id UUID NOT NULL references apps(id) on delete cascade,
+    verification_id UUID NOT NULL references app_email_verifications(id) on delete cascade,
+    created_at TIMESTAMP NOT NULL DEFAULT NOW()
```
dwwoelfel: We usually use timestamp with time zone (it's nice as a default in case someone connects to the db with a different tz setting).

server/resources/migrations/111_add_app_email_verification_codes.up.sql
```diff
@@ -0,0 +1,10 @@
...
+);
+
+CREATE INDEX ON app_email_verification_codes (app_id);
+CREATE INDEX ON app_email_verification_codes (verification_id);
```
dwwoelfel, line 11: Let's put an index on created_at so that we can quickly find the older ones and delete them later.

server/src/instant/model/app_email_verification_code.clj
```diff
@@ -0,0 +1,45 @@
...
+   [instant.config :as config]
+   [instant.jdbc.aurora :as aurora]
+   [instant.jdbc.sql :as sql]
+   [instant.util.email :as email]))
+
+(defn put!
+  ([params] (put! (aurora/conn-pool :write) params))
+  ([conn {:keys [code app-id verification-id]}]
+   (sql/execute-one! conn ["INSERT INTO app_email_verification_codes
+          (id, code, app_id, verification_id)
```
dwwoelfel, line 23: The indentation of the sql queries in this file doesn't match the indentation in the app_email_verification.clj file

server/src/instant/model/app_email_verification_code.clj
```diff
@@ -0,0 +1,45 @@
...
+          VALUES (?::uuid, ?, ?, ?)
+          RETURNING *"
+                           (random-uuid) code app-id verification-id])))
+
+(defn get-all-by-verification-id
```
dwwoelfel: When do we use this function?

server/src/instant/model/app_email_verification_code.clj
```diff
@@ -0,0 +1,45 @@
...
+          AND created_at >= NOW() - (? * INTERVAL '1 minute')
+          RETURNING *"
+                           code verification-id expiry-minutes])))
+
+(defn format-email [{:keys [code sender-email]}]
+  (let [{sender-name :name from-email :email} (config/dashboard-email-sender)]
+    {:from {:name sender-name
+            :email from-email}
+     :to [{:email sender-email}]
+     :subject (str code " is your Instant sender verification code")
```
dwwoelfel: Maybe "is your verification code for your custom sender email address on Instant" would be easier to recognize? It's easy to misread "Instant sender" as fast sender.

server/src/instant/runtime/magic_code_auth.clj
```diff
@@ -125,10 +134,18 @@
                          :code code
                          :app_title (:title app)
                          :expiration (friendly-expiration app)}
-
         {default-sender-email :email} (config/app-email-sender)
```
dwwoelfel, line 137: Could we get the `verified` info in this query so that we don't have to make an additional round-trip to the database below?

server/src/instant/rate_limit.clj
```diff
@@ -201,6 +201,14 @@
         ^Bucket bucket (get-bucket key)]
     (.tryConsume bucket 1)))
 
+(defn try-consume-custom-sender
+  "Takes (hz/rate-limit) and email, will return false if the rate-limit is exceeded.
+   Otherwise, will return true and increment the bucket counter."
+  [{:keys [get-bucket]} {:keys [email]}]
+  (let [key (magic-code-key-hash "custom-sender-verify" #uuid "00000000-0000-0000-0000-000000000000" email)
```
dwwoelfel, line 210: Might be nice to include a comment about why we use an all-zeroes uuid here.

dwwoelfel, line 210: I'd also rename this to just `custom-sender`, since we use the same key for both `send` and `verify`.

server/src/instant/model/app_email_verification_code.clj
drew-harris: we dont -> removed

dwwoelfel:
Could you add a deploy plan to the PR body with all of the steps?

client/www/components/dash/auth/Email.tsx
```diff
@@ -279,10 +241,32 @@ export function Email({ app }: { app: InstantApp }) {
...
+                    </div>
+                  ) : (
+                    <div className="text-xs text-gray-500 dark:text-neutral-400">
+                      Pending confirmation
+                    </div>
+                  )}
+                </div>
+              </div>
+              {!verification.instantVerified ? (
+                <SenderOtpVerification verification={verification} />
```
dwwoelfel, line 284: If I hit `enter` when my cursor is in the "Verification code" input, it submits the full form

dwwoelfel, line 284: The error message is see if I use the wrong code is "Failed to verify code: undefined"

client/www/components/dash/auth/Email.tsx
```diff
@@ -279,10 +241,32 @@ export function Email({ app }: { app: InstantApp }) {
...
+              ) : null}
```
dwwoelfel, line 285: I get this error when I click "resend-code", but only if I open the console and look at the network request. There's no indication in the UI that anything went wrong:

```
{
    "type": "validation-failed",
    "message": "Validation failed for email: This email address has been marked inactive.",
    "hint": {
        "data-type": "email",
        "input": " <dwwoelfel+new-sender2@instantdb.com>",
        "errors": [
            {
                "message": "This email address has been marked inactive."
            }
        ]
    },
    "trace-id": "e7b3a655f7892825e0d8312821610c54"
}
```

dwwoelfel: It seems weird that I'm getting that error, because if I submit the correct code, everything on the page is green.

drew-harris: Image
Fixed

server/src/instant/model/app_email_verification.clj
```diff
@@ -1,12 +1,74 @@
...
+     (random-uuid) app-id sender-id verified])))
+
+(defn get-by-app-and-sender
+  "gets the email verification for the given app and sender"
+  [app-id sender-id]
+  (sql/execute-one! (aurora/conn-pool :read)
+                    ["SELECT * FROM app_email_verifications
+                      WHERE app_id = ? AND sender_id = ?" app-id sender-id]))
+
+(defn mark-verified!
```
dwwoelfel, line 26: I would put an `app_id` in here also as an extra permissions guard and to future-proof ourselves against database sharding on app_id later.

server/src/instant/model/app_email_verification_code.clj
```diff
@@ -0,0 +1,40 @@
...
+    conn ["INSERT INTO app_email_verification_codes
+          (id, code, app_id, verification_id)
+          VALUES (?::uuid, ?, ?, ?)
+          RETURNING *"
+          (random-uuid) code app-id verification-id])))
+
+(defn consume!
+  ([params] (consume! (aurora/conn-pool :write) params))
+  ([conn {:keys [code verification-id expiry-minutes]}]
+   (sql/execute-one!
```
dwwoelfel: I'd also put an app_id in this query for the same reason as above.

server/src/instant/model/app_email_verification_code.clj
```diff
@@ -0,0 +1,40 @@
...
+(defn format-email [{:keys [code sender-email app-title]}]
+  (let [{sender-name :name from-email :email} (config/dashboard-email-sender)]
+    {:from {:name sender-name
+            :email from-email}
+     :to [{:email sender-email}]
+     :subject (str code " is your Instant sender verification code for app: " app-title)
+     :reply-to from-email
+     :html
+     (email/standard-body
+      "<p>Use this code to verify your custom sender email address on Instant for the app: <strong>" app-title "</strong></p>
```
dwwoelfel: Use hiccup to construct the email string, it will protect against html injection. You can see how it works in `instant.dash.routes/team-member-invite-email`.

server/src/instant/model/app_email_sender.clj
```diff
@@ -87,14 +88,20 @@
...
+                      :postmark-id postmark-id})
+        verification (if (flags/use-app-email-verification?)
+                       (verification/put! {:app-id app-id
+                                           :sender-id (:id sender)
+                                           :verified false})
+                       (verification/put! {:app-id app-id
+                                           :sender-id (:id sender)
+                                           :verified true}))
+        needs-verify (not (:verified verification))]
+    {:sender sender :needs-verify needs-verify}))
```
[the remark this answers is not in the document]
drew-harris: Fixed by removing needs-verify entirely.

server/src/instant/model/app_email_sender.clj
```diff
@@ -87,14 +88,19 @@
...
-                        :verified true})
-    sender))
+                      :postmark-id postmark-id})
+        _ (if (flags/use-app-email-verification?)
+            (verification/put! {:app-id app-id
+                                :sender-id (:id sender)
+                                :verified false})
+            (verification/put! {:app-id app-id
+                                :sender-id (:id sender)
+                                :verified true}))]
```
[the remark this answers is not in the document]
drew-harris: This is a downright lie! on conflicting app and sender id pair it does a no op

client/www/components/dash/auth/Email.tsx
```diff
@@ -112,10 +68,30 @@ export function Email({ app }: { app: InstantApp }) {
...
+                      `Failed to send verification code: ${error.body.message}`,
+                    );
+                  } else {
+                    errorToast(
+                      `Failed to send verification code: ${error instanceof Error ? error.message : String(error)}`,
+                    );
+                  }
+                });
+            }
           }
```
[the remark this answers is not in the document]
drew-harris: This is only a factor in the case that the flag was previously on, then turned off.

server/src/instant/dash/routes.clj
```diff
@@ -1379,6 +1385,55 @@
                    :body body})]
     (response/ok {:id (:id template)})))
 
+(defn sender-verification-send-magic-code
+  "sends the email containing a code to verify a custom sender domain with an app id"
+  [req]
+  (when-not (flags/use-app-email-verification?)
+    (ex/throw+ {::type ::permission-denied
+                ::message "Permission denied: app-level sender verification not enabled"}))
```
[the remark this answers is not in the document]
dwwoelfel: @drew-harris `::message` here will still be `:instant.dash.routes/message` instead of `instant.util.exception/message`.

server/src/instant/model/app_email_verification_code.clj
```diff
@@ -0,0 +1,45 @@
...
+
+(defn consume!
+  ([params] (consume! (aurora/conn-pool :write) params))
+  ([conn {:keys [code verification-id expiry-minutes app-id]}]
+   (sql/execute-one!
+    conn ["DELETE FROM app_email_verification_codes
+          WHERE code = ?
+          AND verification_id = ?::uuid
+          AND app_id = ?
+          AND created_at >= NOW() - (? * INTERVAL '1 minute')
```
dwwoelfel: It would be better to run the expired check separately, so that you could throw a better error when the code exists, but is expired. That's how we do it in `instant_user_magic_code.clj`.

client/www/components/dash/auth/Email.tsx
```diff
@@ -395,6 +400,77 @@ export function Email({ app }: { app: InstantApp }) {
...
+      <div className="flex flex-col gap-2 sm:flex-row sm:items-end">
+        <Button
+          type="button"
+          variant="secondary"
+          loading={verification.verifyCode.isMutating}
+          onClick={() =>
+            verification
+              .sendCode()
+              .then(() => {
+                successToast(`Verification code sent to ${address}`);
```
dwwoelfel, line 428: I never see this success toast when I click the `(re)send code` button.

dwwoelfel (approved):
LGTM, one comment about the success toast not showing up for me.

PR #2683, opened 2026-05-18
Set explicit max-entity-size for undertow

nezaj:
Bumping `io.undertow/undertow-core` from 2.3.18 to 2.3.23 in #2668 silently introduced a 2 MiB cap on request bodies. This meant things like uploads above the cap would fail.

Setting this config explicitly now to restore previous behavior


dwwoelfel (approved):
LGTM!

PR #2685, opened 2026-05-18
Use URLSearchParams for admin SDK getUser/deleteUser query strings

nezaj:
## Summary

`deleteUser` built its query string with raw template interpolation, missing two things `getUser` already did:

```ts
const qs = Object.entries(params).map(([k, v]) => `${k}=${v}`);
```

- **No URI encoding.** An email like `alyssa+test@example.com` is sent as `email=alyssa+test@example.com`, where the server parses `+` as a space, so the lookup misses the intended user.
- **No `.join('&')`.** `map` returns an array; the outer template literal joins it with commas, not `&`. Harmless today because `deleteUser` is always called with exactly one key, but still wrong.

Rather than copy `getUser`'s hand-rolled `encodeURIComponent + join('&')` into `deleteUser`, this PR switches both to `URLSearchParams`:

```ts
const qs = new URLSearchParams(Object.entries(params)).toString();
```

Encoding and joining are now handled by the platform. Caught by CodeRabbit and @drew-harris on the original PR thread.

### Note on encoding behavior

`URLSearchParams` encodes spaces as `+` (form-urlencoded) where `encodeURIComponent` uses `%20`. Both are valid in query strings and Ring decodes them identically, so no server-side change. In practice none of the values passed here (emails, UUIDs) can contain spaces.

### DIFF
--- a/client/packages/admin/src/index.ts
+++ b/client/packages/admin/src/index.ts
@@ -684,9 +684,7 @@ class Auth<Schema extends InstantSchemaDef<any, any, any>> {
   getUser = async (
     params: { email: string } | { id: string } | { refresh_token: string },
   ): Promise<Omit<User, 'refresh_token'> | null> => {
-    const qs = Object.entries(params)
-      .map(([k, v]) => `${k}=${encodeURIComponent(v)}`)
-      .join('&');
+    const qs = new URLSearchParams(Object.entries(params)).toString();
 
     const response: { user: User | null } = await jsonFetch(
       `${this.config.apiURI}/admin/users?app_id=${this.config.appId}&${qs}`,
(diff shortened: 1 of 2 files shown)


stopachka (approved):
SGTM!

drew-harris:
Why are we looping over the object with .entries? Could we do something like this:
```typescript
const paramsObj = { foo: "bar", baz: "bar" };
const searchParams = new URLSearchParams(paramsObj);

console.log(searchParams.toString()); // "foo=bar&baz=bar"
```

nezaj:
> Why are we looping over the object with .entries? Could we do something like this:
> 
> ```ts
> const paramsObj = { foo: "bar", baz: "bar" };
> const searchParams = new URLSearchParams(paramsObj);
> 
> console.log(searchParams.toString()); // "foo=bar&baz=bar"
> ```

We could! Was doing this mostly to match what was already in getUser, updated to do it like this!

PR #2686, opened 2026-05-18
Remove noisy console log in reactor

### LINKED ISSUE #2674: db.streams: noisy unconditional console.log on subscribe-stream race + no way to silence
`db.streams.createReadStream()` against a stream the server hasn't created yet logs a noisy error every time, with no way to silence it. Common when the read stream is opened before the corresponding `createWriteStream` lands (we hit it in an LLM-streaming chat app — once per message send).

The error is benign — caller can catch and retry — but `Reactor._handleReceiveError` logs it before any handler runs:

```js
// @instantdb/core 1.0.33  Reactor.js:785
_handleReceiveError(msg) {
    console.log('error', msg);   // unconditional
    ...
}
```

```
LOG  error {... "message": "Validation failed for subscribe-stream: Stream is missing." ...}
```

`setInstantWarningsEnabled(false)` and `verbose: false` don't gate it.

Related to #164 — same class of transient-race error log. That one was addressed by reducing the race server-side (#361), not by gating the log. A similar approach would also help here, but a logger gate would cover any future cases.

Asks (any one):

1. Gate the log (`if (this._log.isEnabled) console.log(...)`).
2. Special-case `Stream is missing` and don't log it (it's recoverable by design).
3. Expose a `db.streams.has(clientId)` so consumers can coordinate without polling/retry.

Happy to PR option 1 if you're open to it.

dwwoelfel:
Replace it with a this._log instead.

Fixes https://github.com/instantdb/instant/issues/2674

### DIFF
--- a/client/packages/core/src/Reactor.js
+++ b/client/packages/core/src/Reactor.js
@@ -956,7 +956,7 @@ export default class Reactor {
   }
 
   _handleReceiveError(msg) {
-    console.log('error', msg);
+    this._log.info('error', msg);
     const eventId = msg['client-event-id'];
     // This might not be a mutation, but it can't hurt to delete it
     this._inFlightMutationEventIds.delete(eventId);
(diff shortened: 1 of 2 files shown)


nezaj (approved):
SGTM!

drew-harris (approved):

PR #2687, opened 2026-05-18
Pick migrate binary by arch in self-hosted Dockerfile

nezaj:
## Summary

The arm64 variant of `ghcr.io/instantdb/server:latest` was being built with the amd64 `migrate` binary baked into an otherwise-arm64 image. This was causing self host to fail on arm-based machines

The fix is for the Dockerfile to pick the arm variant of migrate. This updates to match the pattern we already used in `Dockerfile-dev`.

Tested this on my local machine and got it working!


dwwoelfel (approved):
LGTM!

PR #2690, opened 2026-05-20
Address weakHash query collisions

stopachka:
**Problem** 

Our previous `weakHash` algorithm could produce real collisions. 

For example, the following queries would make the same weakHash: 

```ts
const q = (propertyId) => ({
  pro_search_properties: {
    $: { where: { pro_searches: 'b14fae2f-ce9b-4677-b6a9-6dddd81914d0', propertyId } },
    pro_searches: {},
  },
});

weakHash(q(936)) ==  weakHash(q(27140)) // collides! 
```

**Reason**

There's a few reasons for this: 

- **The mixer lost precision.** The object branch did `hash *= 16777619; hash = hash >>> 0`. JS numbers are floats: once `hash * 16777619` exceeds `2^53`, the low bits are silently lost before the truncate. So the effective entropy was less than the nominal 32 bits. 
- **Numbers could produce linear hash sequences.** The number branch was `(input * 2654435761) >>> 0` with no avalanche, so sequential ids produced near-linear hash sequences. Combined with the lossy multiply above, queries that differed only by an integer field had very little differential entropy.

**Fix**

We could have fixed our 32-bit hash, but I thought it would be better to go for 64-bits.

Replaced with a cyrb53-style hash: two 32-bit state words mixed with `Math.imul`, xxhash-style final avalanche, 64-bit output. Object keys are still sorted, and `undefined` values are still skipped in objects.


client/packages/core/src/utils/weakHash.ts
```diff
@@ -1,81 +1,73 @@
...
-    return hash.toString(16);
+    return out + ']';
   }
 
-  // For objects, hash keys and values
-  if (typeof input === 'object') {
-    let hash = 0x811c9dc5;
-    const keys = Object.keys(input).sort(); // Sort for consistency
-
+  if (input && typeof input === 'object') {
```
dwwoelfel, line 52: One suggestion here: if someone puts a date object, we'll encode this as `{}`, so 

`{created: {$gt: new Date(1)}}` will be the same as `{created: {$gt: new Date(2)}}`. 

Two ideas: 

1. round-trip the whole input through JSON.stringify before we call stableStringify
2. see if it has a `.toJSON` function--if it does, call that and treat it as a string

client/packages/core/__tests__/src/utils/weakHash.test.ts
```diff
@@ -0,0 +1,50 @@
...
+        firstCollisionAt = i;
+        break;
+      }
+      hashes.add(hash);
+    }
+    expect(firstCollisionAt, `collision at i=${firstCollisionAt}`).toBeNull();
+  }
+});
+
+test('is stable across object key order and undefined values', () => {
```
dwwoelfel, line 35: Could we get one smoke test where we say `weakHash(somePredefinedQuery) == "actual-hash-value"`? Might help us catch an accidental regression that changes the weakhash function.

client/packages/core/src/utils/weakHash.ts
stopachka: I like the toJSON idea! Added

client/packages/core/__tests__/src/utils/weakHash.test.ts
stopachka: Good idea! Added this test to both weakHash and weakHashLegacy!

client/packages/core/src/Reactor.js
```diff
@@ -1070,14 +1071,47 @@ export default class Reactor {
...
+        const legacyHash = weakHashLegacy(q);
+        if (legacyHash !== hash && prev[legacyHash]) {
+          prev[hash] = prev[legacyHash];
+          delete prev[legacyHash];
+        }
+      }
       prev[hash] = prev[hash] || { q, result: null, eventId };
       prev[hash].lastAccessed = Date.now();
     });
+    this._tryMigrateLegacyQuerySub(q, hash);
```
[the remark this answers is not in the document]
stopachka: I don't think this is necessary.

dwwoelfel (approved):
Beautiful!

PR #2691, opened 2026-05-20
Faster in-room? check

dwwoelfel:
Small change to help with the thundering herd in hazelcast rooms.

Before we would fetch the full value of the map to check if you're in the room, which would require hazelcast to deserialize the full value and possibly transfer it to the machine. Now we just look at our local room-maps atom.

Part of a larger change to fix the thundering herd problem. Putting this in its own PR so that we can roll back to this low-risk change if the larger hazelcast changes cause errors (those will be in the next couple of PRs).


stopachka (approved):
Woohoo!

PR #2698, opened 2026-05-22
Cleanup hazelcast code after flipping flag

dwwoelfel:
I flipped the `batch-hz-room-map-actions` flag yesterday.

Everything is looking good in production, so we can clean up this old code.


stopachka (approved):
Niice!

PR #2701, opened 2026-05-24
feat: improve explorer table tooltips, cell editing, and edit row dialog UX

Kaori2499:
From discord thread: https://discord.com/channels/1031957483243188235/1506882312351256616

**Description**

This PR improves the dashboard explorer’s table and edit-row dialog UX: clearer tooltips, faster in-cell editing, and more reliable form validation/layout.

### Changes

**Tooltips**
- Added a **400ms hover delay** on cell tooltips to reduce accidental popups
- Made long tooltip content **scrollable and word-wrapped**, with responsive max width/height
- Fixed overflow detection so truncated text (not just JSON) shows tooltips again

**Table interactions**
- Added **hover edit buttons** on editable cells; opens the row dialog with that field focused and scrolled into view, a feature requested in this [discord thread](https://discord.com/channels/1031957483243188235/1504028488078725172).
- Unified edit button sizing; **disabled copy** for empty/null/meaningless values
- Focuses the **main value input** in the edit dialog (not the null toggle)

**Edit row dialog**
- Added **client-side validation** for required blob and ref fields
- Pinned **Reset/Save footer** with a single inner scroll area so that saving data does not necessitate scrolling all the way down to the bottom.

### How to Test

1. **Tooltip delay** — Hover a truncated cell; tooltip should appear after ~400ms, instead of 100ms.
2. **Tooltip overflow** — Hover a cell with long text or JSON; content should wrap and scroll within the tooltip.
3. **Truncated text tooltips** — Hover a truncated string cell (non-JSON); tooltip should appear with full content.
4. **Cell edit button** — Hover an editable cell; click the pencil icon; dialog opens with that field focused and centered.
5. **Empty copy** — Click empty/null cells; copy should not trigger (no pointer cursor / no “Copied!”).
6. **Required validation** — Add/edit a row and leave a required field empty; Save should block with an inline error.
7. **Dialog layout** — Open a row with many fields; only the body scrolls, Reset/Save stay visible at the bottom.

### Screenshots

1. Table cell with edit button and scrollable tooltip:
CleanShot 2026-05-24 at 14 58 07

2. Edit dialog when client validation failed:
CleanShot 2026-05-24 at 14 56 58

### DIFF
--- a/client/packages/components/src/components/explorer/edit-row-dialog.tsx
+++ b/client/packages/components/src/components/explorer/edit-row-dialog.tsx
@@ -3,6 +3,7 @@ import { id, InstantReactWebDatabase, tx } from '@instantdb/react';
 import {
   TextareaHTMLAttributes,
   useEffect,
+  useLayoutEffect,
   useMemo,
   useRef,
   useState,
@@ -37,7 +38,6 @@ import { validate } from 'uuid';
 import clsx from 'clsx';
 import { ClockIcon } from '@heroicons/react/24/outline';
 import { useExplorerProps } from '.';
-import { useShadowRoot } from '@lib/components/StyleMe';
 
 type FieldType = 'string' | 'number' | 'boolean' | 'json';
 type FieldTypeOption = { value: FieldType; label: string };
@@ -525,7 +525,6 @@ function RefItem({
   refUpdates,
   handleLinkRef,
   handleUnlinkRef,
-  autoOpenSearch,
 }: {
   db: InstantReactWebDatabase<any>;
   item: Record<string, any>;
@@ -534,13 +533,8 @@ function RefItem({
   refUpdates: null | Record<string, { action: 'link' | 'unlink'; item: any }>;
   handleLinkRef: (attr: SchemaAttr, item: any) => void;
   handleUnlinkRef: (attr: SchemaAttr, id: string) => void;
-  autoOpenSearch?: boolean;
 }) {
-  const [showAddLink, setShowAddLink] = useState(autoOpenSearch ?? false);
-
-  useEffect(() => {
-    setShowAddLink(autoOpenSearch ?? false);
-  }, [autoOpenSearch]);
+  const [showAddLink, setShowAddLink] = useState(false);
   const searchIgnoreIds = useMemo(() => {
     const res: Set<string> = new Set();
     for (const [k] of Object.entries(refUpdates || {})) {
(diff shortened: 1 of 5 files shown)


client/packages/components/src/components/explorer/edit-row-dialog.tsx
```diff
@@ -867,15 +955,129 @@ export function EditRowDialog({
...
+        const root = shadowRoot ?? document;
+        const section = root.querySelector(
+          `[data-explorer-field="${focusAttr}"]`,
+        );
+        if (!section) {
+          return;
+        }
+        focusExplorerField(section);
+      });
+    });
```
drew-harris: whats the reasoning for double requestAnimationFrame?

client/packages/components/src/components/explorer/edit-row-dialog.tsx
```diff
@@ -867,15 +955,129 @@ export function EditRowDialog({
...
+    const addLink = section.querySelector<HTMLElement>(
+      '[data-explorer-add-link]',
+    );
+    if (addLink) {
+      addLink.click();
+      requestAnimationFrame(() => {
+        section.querySelector<HTMLElement>('input')?.focus();
+        scrollFieldSectionIntoView(section);
+      });
+    }
```
drew-harris: I think this could be simplified. We are already passing the focusAttr prop to the EditRowDialog component so the EditRowDialog should just set the autoFocus html attribute to true for whatever field should be focused.

drew-harris (changes requested):
I think we should not include an edit button for links that auto-opens the search box. I think its convenient for editing the blob values but the story for links felt too cluttered. 
Functionality is working but the reliance on css selectors makes this feel a little more fragile than a more straightforward react-based solution

drew-harris:
Also please make sure you have prettier setup or run `pnpm run format` from `client/` so that the code formatting passes ci :)

client/packages/components/src/components/explorer/edit-row-dialog.tsx
Kaori2499: This is a timing hack for working around shadow-DOM/portal mount order with querySelector, indeed it looks a little awkward. I’ve removed this with the whole DOM-query focus path.

client/packages/components/src/components/explorer/edit-row-dialog.tsx
Kaori2499: Agreed. Blob fields now read autoFocus={focusAttr === attr.name} directly (input / textarea / null-state button), JSON uses CodeEditor’s onMount to focus the Monaco instance, and a small ref per field section handles scroll-into-view. The boolean Select is the only one that focuses imperatively (no native autoFocus on the trigger), but it’s scoped to the section ref rather than a global query.

Kaori2499:
New changes include:

1. hide edit button for link column;
2. autofocus code simplification;
3. prettier format.

Sorry for the problems in previous commits, I think these changes fixed them :)

drew-harris:
awesome thanks will re-review soon

drew-harris (approved):

drew-harris:
@Kaori2499 sorry for delay in review, merged now. Thanks for the contribution!

PR #2702, opened 2026-05-24
fix: clear search filter when namespace changes

kenrick95:
When I'm in the explorer, after I added a search filter and search, it applies the search filter to the current query of data:

Screenshot 2026-05-24 200546

(At this point, the data shown is still correct as expected)

if I navigate away to other namespace (via the namespace list at the sidebar), the new page will always show empty data

Screenshot 2026-05-24 200552

After debugging using React dev tools, it seems like due to 'Inner Explorer' component maintaining its own search filter state, and this search filter state wasn't cleared if the associcated namespace id has been changed

Screenshot 2026-05-24 200351

So the PR here set the search filter to empty array if it detects the namespace id has changed

### DIFF
--- a/client/packages/components/src/components/explorer/inner-explorer.tsx
+++ b/client/packages/components/src/components/explorer/inner-explorer.tsx
@@ -627,9 +627,10 @@ export const InnerExplorer: React.FC<{
     {},
   );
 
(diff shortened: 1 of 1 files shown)


stopachka (approved):
Thank you!

PR #2706, opened 2026-05-27
Add route handler protocol body helper

mareszhar:
[Reference Discord Thread](https://discord.com/channels/1031957483243188235/1508597081122799626)

This PR introduces a unified protocol for utils working with `firstPartyPath`.
## Benefits

**Avoids client/server drift** — adds a shared `routeHandlerProtocol.ts` contract used by both `Reactor` when posting auth sync updates and `createInstantRouteHandler` when receiving them.

**Keeps future payload types easy to add** — models request bodies through a payload-by-type map, so new route-handler request types can extend the map and automatically flow into the derived body union.

**Provides a consistent body factory** — adds `createInstantRouteHandlerBody('sync-user', { appId, user })` so protocol bodies are constructed from one helper instead of hand-written object literals.

**Improves custom handler DX** — exports `InstantRouteHandlerBody`, `InstantRouteHandlerRawBody`, `InstantRouteHandlerType`, and `InstantRouteHandlerPayloadByType` so custom route handlers can type the payload they receive without duplicating the SDK's internal shape.

**Documents intended behavior through tests** — adds runtime and type-level coverage for the helper, the existing route handler behavior, unknown request types, and the exported type contract.

## Validation

```
pnpm --dir client/packages/core run check
pnpm --dir client/packages/core run test:types
pnpm --dir client/packages/core exec vitest run __tests__/src/createRouteHandler.test.ts
```


drew-harris (changes requested):
Thanks for taking initiative to submit this PR. I think currently this diff does a bit too much and instead we would prefer to export the bare minimum typescript type if we can get away with that. 

```typescript
export type InstantRouteHandlerPayloadByType = {
  'sync-user': {
    appId: string;
    user: User | null;
  };
};


/** A valid request body for Instant's first-party route handler protocol. */
export type InstantRouteHandlerBody<
  Type extends InstantRouteHandlerType = InstantRouteHandlerType,
> = {
  [KnownType in Type]: {
    type: KnownType;
  } & InstantRouteHandlerPayloadByType[KnownType];
}[Type];
```

Would exporting these 2 types from all packages be enough for your specific usecase / reasoning for opening the PR? 

We don't feel the need for any tests on this code and we prefer to use the `satisfies` keyword over a constructor method. For the raw body type, we think end users creating a zod / validator schema for a body is a better alternative.

drew-harris:
Closing in favor of https://github.com/instantdb/instant/pull/2756 due to merge conflicts. Thanks for your contribution!

Issue #2707, opened 2026-05-27
useInfiniteQuery: how to reliably distinguish initial load from confirmed empty result?


uriva:
When using `useInfiniteQuery` on a page that may legitimately have zero results, I'm finding it tricky to render an "empty state" without flashing it briefly during the initial load.

The natural pattern would be:

```tsx
const { data, isLoading } = useInfiniteQuery({ items: { $: { limit: 20 } } });
const items = data?.items ?? [];

if (isLoading) return <Spinner />;
if (items.length === 0) return <EmptyState />;
return <List items={items} />;
```

In practice, on first mount I sometimes see the empty state flash for a moment before the real results appear. It looks like `isLoading` flips to `false` (or `data` becomes a defined object with an empty array) before the actual page of results has arrived from the server, so `items.length === 0` is briefly true even though there are items on the server.

A few questions:

1. Is there a recommended way to distinguish "we haven't received the first page yet" from "first page arrived and is genuinely empty"?
2. Is there a field on the result (something like `pageInfo`, an `endCursor`, a `hasLoaded` flag, etc.) that's intended for this?
3. What's the idiomatic pattern people use for empty states with `useInfiniteQuery`?

Currently I'm working around it by gating the empty state behind a `setTimeout`, but that feels like a hack. Would love to know if there's a cleaner approach, or if this is something the hook could expose more directly.

Thanks!

uriva:
For reference, I'm on `@instantdb/react@1.0.21`.

nezaj:
@uriva we just pushed up v1.0.41, does this work for you?

uriva:
gemini 3.5:

Thanks @nezaj! I took a look under the hood at `@instantdb/react-common` and `@instantdb/core` in `v1.0.41`.

While `v1.0.41` beautifully resolves the flicker when query dependencies/parameters change (by leveraging `getInfiniteQueryInitialSnapshot` to preserve cached state), it unfortunately **still flashes empty state on cold starts (first mount with an empty local cache)**.

Here is the exact code path causing this:

### 1. `subscribeInfiniteQuery` starter sub fires synchronously
When `useInfiniteQuery` is mounted, `useInfiniteQuerySubscription.js` sets up the subscription:
```javascript
const sub = core.subscribeInfiniteQuery(query, (resp) => {
    stateCacheRef.current = {
        ...resp,
        isLoading: false, // <-- Unconditionally forces isLoading: false
    };
    cb();
}, opts);
```

[...]

PR #2710, opened 2026-05-27
fix snapshot resetting in useInfiniteQuery (#2710) [bump to v1.0.41]

drew-harris:
This PR fixes an issue where if the client has saved data for an initial starting snapshot the react hook for useInifiniteQuery will briefly clear the data and set isLoading to true when not necessary.

At the start of the useCallback for subscribe we reset any current data we have and then start the subscription. In local dev this issue wasn't a factor because the callback for `this.core.subscribeInfiniteQuery` would set the data correctly before react had a chance to render the state where data=undefined and loading=true. 

Notes: I think there still should be a good way to figure out if the data comes from a snapshot or if its from the subscription but since the basic useQuery doesn't include this, I won't bother in this PR. 

Before:
image
image
console log statement placed at end of infinite scroll sandbox. 

after:
image

Note for #2707:  The user is saying that the hook would "accidentally flash the empty state" before loading. This is actually correct behavior because the state truly is empty but we were showing a loading state when we should not have been at all. i.e. the issue wasn't a flashing "has data" state prior to loading, the issue was that the pre-loading and post-loading states were correct but there shouldn't have been a loading state since we had snapshotted data in local memory

patch to test:
```patch
diff --git a/client/packages/react-common/src/useInfiniteQuerySubscription.ts b/client/packages/react-common/src/useInfiniteQuerySubscription.ts
index 108a249741..a8e98fb1cc 100644
--- a/client/packages/react-common/src/useInfiniteQuerySubscription.ts
+++ b/client/packages/react-common/src/useInfiniteQuerySubscription.ts
@@ -82,11 +82,14 @@
       const sub = core.subscribeInfiniteQuery(
         query,
         (resp) => {
-          stateCacheRef.current = {
-            ...resp,
-            isLoading: false,
-          };
-          cb();
+          // wait a second
+          new Promise((resolve) => setTimeout(resolve, 1000)).then(() => {
+            stateCacheRef.current = {
+              ...resp,
+              isLoading: false,
+            };
+            cb();
+          });
         },
         opts,
       );
diff --git a/client/sandbox/react-nextjs/pages/play/infinite-scroll.tsx b/client/sandbox/react-nextjs/pages/play/infinite-scroll.tsx
index ebb690048b..c689edc816 100644
--- a/client/sandbox/react-nextjs/pages/play/infinite-scroll.tsx
+++ b/client/sandbox/react-nextjs/pages/play/infinite-scroll.tsx
@@ -107,8 +107,15 @@
     [allNumbersResult.data],
   );
 
+  console.log(
+    'IS LOADING!',
+    scrollResult.isLoading,
+    scrollResult?.data?.items?.length,
+  );
+
   return (
     <div>
+      {scrollResult.isLoading && <div>Loading...</div>}
       <div>
         <input
           className="m-2 border border-gray-400 p-2"
```

### DIFF
--- a/client/packages/react-common/src/useInfiniteQuerySubscription.ts
+++ b/client/packages/react-common/src/useInfiniteQuerySubscription.ts
@@ -73,12 +73,6 @@ export function useInfiniteQuerySubscription<
   const subscribe = useCallback(
     (cb: () => void) => {
       subRef.current = null;
-      stateCacheRef.current = {
-        error: undefined,
-        data: undefined,
-        isLoading: true,
-        canLoadNextPage: false,
-      };
       cb();
 
       if (!query) {
--- a/client/packages/react-common/src/useInfiniteQuerySubscription.ts
+++ b/client/packages/react-common/src/useInfiniteQuerySubscription.ts
@@ -73,6 +73,13 @@ export function useInfiniteQuerySubscription<
   const subscribe = useCallback(
     (cb: () => void) => {
       subRef.current = null;
+
+      const nextSnapshot = getInfiniteQueryInitialSnapshot(core, query, opts);
+      stateCacheRef.current = {
+        ...nextSnapshot,
+        isLoading: !nextSnapshot.data && !nextSnapshot.error,
+      };
+
       cb();
 
       if (!query) {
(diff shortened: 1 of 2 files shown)


nezaj (approved):
LGTM!

PR #2713, opened 2026-05-28
Sync user cookie less often (#2713) [bump to v1.0.43]

drew-harris:
Stop calling cookie sync with firstPartyPath every time reactor starts up and every minute. Should now only be called on state changes. 

Includes a helpful snippet in the docs for when a user wants to manually sync.
 Ex: the user could put their own setInterval in their _app.tsx that wouldn't run on page navigations. 

### Question:
Should we include a function on db.auth to run a manual sync?

### DIFF
--- a/client/packages/core/__tests__/src/cookieSync.e2e.test.ts
+++ b/client/packages/core/__tests__/src/cookieSync.e2e.test.ts
@@ -1,32 +1,163 @@
-import { vi, expect } from 'vitest';
-import { http, HttpResponse } from 'msw';
-import { i } from '../../src';
-import { makeE2ETest } from './utils/e2e';
+import { expect, type TestContext, vi } from 'vitest';
+import { http, HttpResponse, ws } from 'msw';
+import { i, init, type User } from '../../src';
+import { apiUrl, makeE2ETest } from './utils/e2e';
 import { COOKIE_SYNC_LAST_UPDATED_KEY } from '../../src/Reactor';
 
-const test = makeE2ETest({
-  rules: {
-    code: {},
-  },
-  schema: i.schema({
-    entities: {
-      animal: i.entity({}),
-    },
-  }),
-  config: {
-    firstPartyPath: 'https://example.com',
+const FIRST_PARTY_PATH = 'https://example.com';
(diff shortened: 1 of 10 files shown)


client/packages/core/src/Reactor.js
```diff
@@ -371,16 +371,6 @@ export default class Reactor {
 
     this._oauthCallbackResponse = this._oauthLoginInit();
 
-    // kick off a request to cache it
-    this.getCurrentUser().then((userInfo) => {
```
stopachka, line 375: SGTM, but I do wonder: how long will these cookies last? 

Perhaps we should do something like once every 7 days. It begs the question: how will we know? I guess we could store something in local storage. 

WDYT @dwwoelfel @drew-harris @nezaj ?

drew-harris: oh good point  the cookies do expire so we'd probably want a local storage to re-sync before they expire

client/packages/core/__tests__/src/cookieSync.e2e.test.ts
```diff
@@ -0,0 +1,22 @@
...
+});
+
+test('does things', async ({ db }) => {
+  const keys = await db._reactor._userSyncStorage?.getAllKeys();
+  console.log();
+  console.log();
+  console.log();
+  console.log();
+  console.log();
+});
```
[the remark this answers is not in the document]
drew-harris: from an old commit?

client/www/app/docs/backend/page.md
```diff
@@ -507,3 +507,11 @@ export default async function RootLayout({
   );
 }
 ```
+
+The first party path will be called whenever a user's `user` information is changed. To sync the user on demand, you can use:
```
stopachka: I wouldn't include a call to _reactor, as that is meant to be internal. 

We _could_ have a core level `syncUserToEndpoint` function. 

For now though maybe we just remove it from the docs. WDYT @drew-harris @dwwoelfel ?

stopachka (approved):
One note but SGTM!

client/www/app/docs/backend/page.md
drew-harris: just removing it for now

PR #2716, opened 2026-05-29
Update invalidator `like` regex to match postgres on newline matching

dwwoelfel:
Fixes our `like` matcher on the backend to support newlines in the same way that postgres does. 

It's the same fix as the sdk fix at https://github.com/instantdb/instant/pull/2714


nezaj (approved):
Nice!

PR #2718, opened 2026-06-01
Skip no-op triple updates

stopachka:
## Summary
- skip object triple conflict updates when the value md5 is unchanged
- keep id triples on the update path so invalidation still sees them
- preserve overwrite-t behavior for stream machineId writes

## Testing
- scripts/nrepl-eval --port 6005 targeted test-var for insert-multi-skips-same-value-updates-except-id-triples
- clj-kondo --lint src/instant/db/model/triple.clj test/instant/db/model/triple_test.clj


dwwoelfel (commented):
LGTM!

PR #2719, opened 2026-06-01
server changes for email cli (2/3)

drew-harris:
# BEFORE MERGING: use-app-email-verification? must be togged on. 

Includes migration to drop user_id field from custom senders table. 
Need to run migration before deploying.

Purpose of this pr is to remove the restriction against superuser tokens for auth email routes.

### DIFF
--- a/server/resources/migrations/112_allow_null_sender_user_id.down.sql
+++ b/server/resources/migrations/112_allow_null_sender_user_id.down.sql
@@ -1,0 +1,1 @@
+ALTER TABLE app_email_senders ALTER COLUMN user_id SET NOT NULL;
--- a/server/resources/migrations/112_allow_null_sender_user_id.down.sql
+++ b/server/resources/migrations/112_allow_null_sender_user_id.down.sql
@@ -1,1 +1,5 @@
+-- WARNING: This rollback will fail if any app_email_senders rows have user_id = NULL.
+-- Before running this migration, manually clean up or backfill null user_ids:
+--   UPDATE app_email_senders SET user_id = <sentinel_user_id> WHERE user_id IS NULL;
+-- Rollback is only safe if deployed before any null user_ids were written.
 ALTER TABLE app_email_senders ALTER COLUMN user_id SET NOT NULL;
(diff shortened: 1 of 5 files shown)


client/packages/platform/src/email.ts
```diff
@@ -0,0 +1,35 @@
...
+</div>
+`;
+
+export const defaultMagicCodeEmailSubject =
+  '{code} is your code for {app_title}';
+
+export const defaultMagicCodeEmailConfig = {
+  authEmail: {
+    subject: defaultMagicCodeEmailSubject,
+    senderName: undefined,
```
stopachka: Should we have senderName and senderEmail filled in too?

drew-harris: This is only used for the dashboard after (3/3) gets merged. Sender name and sender email are dynamic

server/src/instant/dash/routes.clj
```diff
@@ -1336,9 +1337,19 @@
 
 ;; ---------------
 ;; Email templates
+;; ---------------
+
+(defn get-default-email-template [_req]
+  (let [params (magic-code-auth/default-email-template-params)]
+    (response/ok {:email-type "magic-code"
+                  :sender-email (:sender-email params)
```
stopachka, line 1346: I would simplify this, less re-direction of variable names. also not sure if email and reply-to are needed

drew-harris: Done

server/resources/migrations/112_allow_null_sender_user_id.down.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE app_email_senders ALTER COLUMN user_id SET NOT NULL;
```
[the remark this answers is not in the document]
stopachka: Adding a note here imo is not a bad idea

stopachka (approved):
SGTM!

server/resources/migrations/112_allow_null_sender_user_id.down.sql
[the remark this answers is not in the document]
drew-harris: note added from coderabbit suggestion.

PR #2722, opened 2026-06-01
Redesign the Admin dashboard

stopachka:
**Preview:** https://instant-www-js-adminredesign1-jsv.vercel.app/dash?s=main&t=admin

**Before** 

The Admin Tab had a few design issues. Mainly, there was a lack of hierarchy (all the things you could do felt "at the same level"). And lack of uniformity: some buttons were in the center, some on the right, some were full width

CleanShot 2026-06-01 at 15 36 42@2x

**After** 

I re-organized the UI around a hierarchy. I also tried to make the sections more uniform, with text on the left and buttons generally on the right. 

CleanShot 2026-06-01 at 15 39 41@2x


client/www/components/admin/AdminPage.tsx
```diff
@@ -144,34 +305,215 @@ export function Admin({
...
+            this token a secret.{' '}
+            {isMinRole('admin', role) ? (
+              <>
+                If need be, you can regenerate it by{' '}
+                <a
+                  className="hover:cursor-pointer dark:text-white"
+                  onClick={regenerateDialog.onOpen}
+                >
+                  clicking here
+                </a>
```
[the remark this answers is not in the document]
stopachka: This is how we had it before. Keeping it as is for now

client/www/components/admin/AdminPage.tsx
```diff
@@ -144,34 +305,215 @@ export function Admin({
...
+              </RowList>
+            ) : canAddMembers ? (
+              <Callout>
+                <FieldHint>
+                  Invite teammates to collaborate on this app.
+                </FieldHint>
+                <Button variant="primary" onClick={inviteDialog.onOpen}>
+                  <PlusIcon height={14} /> Invite a team member
+                </Button>
+              </Callout>
```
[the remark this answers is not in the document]
stopachka: The admin tab itself is already gated.

client/www/components/admin/AdminPage.tsx
```diff
@@ -39,6 +39,146 @@ import Link from 'next/link';
...
+        className="inline-flex shrink-0 cursor-pointer items-center gap-1 rounded-sm border border-red-300 px-3 py-1.5 text-sm text-red-600 hover:bg-red-50 dark:border-red-900/60 dark:text-red-400 dark:hover:bg-red-950/40"
+      >
+        <TrashIcon height={'1rem'} />
+        {actionLabel}
+      </button>
+    </div>
+  );
+}
+
 export function Admin({
```
drew-harris, line 182: admin page content should take the same width / top padding as the webhooks page content, for consistency

client/www/components/admin/AdminPage.tsx
```diff
@@ -144,34 +305,215 @@ export function Admin({
...
+              </>
+            ) : null}
+          </Content>
+          <Copyable
+            onChangeHideValue={() => setHideAdminToken(!hideAdminToken)}
+            hideValue={hideAdminToken}
+            label="Admin token"
+            value={app.admin_token}
+          />
+          <HomeButton href="/docs/backend" title="Instant and your backend">
```
drew-harris: would be nice to have target=_blank for docs links

client/www/components/admin/AdminPage.tsx
```diff
@@ -144,34 +305,215 @@ export function Admin({
...
-          </>
+    <TabContent className="mx-auto h-full w-full">
+      <div className="flex flex-col gap-8">
+        {/* App settings: name + admin token */}
+        <div className="flex flex-col gap-4">
+          <SectionHeader title="App settings" />
+
+          {isMinRole('owner', role) ? (
+            <div className="flex flex-col gap-2">
+              <Label>App name</Label>
```
drew-harris, line 320: just my opinion but feels like the font weight on the label here is really strong.

Image

client/www/components/admin/AdminPage.tsx
```diff
@@ -144,34 +305,215 @@ export function Admin({
...
+              <form
+                className="flex items-start gap-2"
+                {...appNameForm.formProps()}
+              >
+                <div className="flex-1">
+                  <TextInput
```
drew-harris, line 326: should disallow " " as an app name

client/www/components/admin/AdminPage.tsx
```diff
@@ -144,34 +305,215 @@ export function Admin({
...
+          </div>
+        </div>
+
+        {isMinRole(app.org ? 'admin' : 'owner', role) && (
+          <TransferApp app={app} />
+        )}
+
+        {isMinRole(app.org ? 'admin' : 'owner', role) ? (
+          <div className="flex flex-col gap-3 pb-4">
+            <SectionHeader
```
drew-harris, line 501: Image
Maybe the danger zone should have a dotted outline like the team members section? Just looks like we forgot the background for the section.

client/www/components/admin/AdminPage.tsx
```diff
@@ -144,34 +305,215 @@ export function Admin({
...
+              title="Danger zone"
+              description="These actions are irreversible and permanently delete data."
+            />
+            <div className="divide-y overflow-hidden rounded-sm border border-red-200 dark:divide-red-900/50 dark:border-red-900/50">
+              {isMinRole('owner', role) && (
+                <DangerRow
```
drew-harris, line 507: Image
I think this shouldn't be done in this PR but I think a dashboard pass over our buttons is needed. The main <Button /> component ("Transfer" in screenshot) feels weirdly scaled wherever its used on the dashboard but it looks fine on the main site. 
Would be nice to make a new button component thats more flexible and start migrating all the one-off buttons to it eventually.

client/www/components/admin/AdminPage.tsx
```diff
@@ -144,34 +305,215 @@ export function Admin({
...
-            >
-              clicking here
-            </a>
-            .
-          </>
+    <TabContent className="mx-auto h-full w-full">
+      <div className="flex flex-col gap-8">
+        {/* App settings: name + admin token */}
+        <div className="flex flex-col gap-4">
+          <SectionHeader title="App settings" />
```
drew-harris, line 316: Image

I feel like "app settings" is a much better name for this page, and we should move it to a more sensible place on the sidebar.

Whenever I see "Admin" my head goes to more auth type of stuff. We really only call it that because its the place you can find the admin token. 

Image
Maybe we could move the admin token to be on the home page next to the public app id? I think it would give us a nice place to explain the difference side-by-side. We can still keep it on this page too since people are used to finding it there. 

Thoughts? @stopachka @nezaj @dwwoelfel

client/www/components/admin/AdminPage.tsx
stopachka: I updated webhooks in main to be the same width. Updated top padding to match too, thanks!

client/www/components/admin/AdminPage.tsx
stopachka: Agreed! I think yeah in general we need to revamp our 'Design system'. Buttons, heading weights, and colors. 

I wanted us to do that in a separate PR though, since imo the most important thing to get right is the layout

client/www/components/admin/AdminPage.tsx
stopachka: Good idea! I think we may want to explore this in a separate PR, as I am not sure how the admin token should look on the home page.

client/www/components/admin/AdminPage.tsx
stopachka: done!

client/www/components/admin/AdminPage.tsx
stopachka: done!

client/www/components/admin/AdminPage.tsx
stopachka: Done!

client/www/components/admin/AdminPage.tsx
```diff
@@ -243,287 +585,102 @@ export function Admin({
...
+                  authorization: `Bearer ${token}`,
+                  'content-type': 'application/json',
+                },
+              });
+
+              setIsClearingApp(false);
+              clearDialog.onClose();
+              dashResponse.mutate();
+              successToast('App cleared!');
             }}
```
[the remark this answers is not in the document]
stopachka: Already fixed!

client/www/components/admin/AdminPage.tsx
stopachka: Done!

PR #2723, opened 2026-06-01
show instant verification status as true

drew-harris:
www client side code expects shape like:
```json
{
  "instant": {
    "verified?": false
  },
  "verification": {
    "ID": 5971037,
    "Confirmed": true,
    "EmailAddress": "baba1@drewh.cloud",
    "DKIMHost": "REDACTED",
    "ReturnPathDomain": "pm-bounces.drewh.cloud",
    "DKIMPendingTextValue": "",
    "ReturnPathDomainCNAMEValue": "pm.mtasv.net",
    "DKIMTextValue": "READCTED",
    "DKIMPendingHost": ""
  }
}
```

but in current backend the shape is {verification: {..}} without instant top level key. 

current code defaults lack of key to be equal to not verified when it should be opposite

### DIFF
--- a/client/www/lib/hooks/useCustomSenderVerification.ts
+++ b/client/www/lib/hooks/useCustomSenderVerification.ts
@@ -119,7 +119,7 @@ export const useCustomSenderVerification = (app: InstantApp) => {
 
   return {
     raw: resp,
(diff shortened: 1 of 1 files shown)


stopachka (approved):
SGTM!

PR #2732, opened 2026-06-04
Add a flag to fallback to sendgrid

dwwoelfel:
Adds a flag that falls back to sendgrid that we can toggle when postmark goes down.

Will only work for emails that don't have a custom sender.

The flag is `send-with-sendgrid` and we check that we have a valid sendgrid email before we try to send with it.


stopachka (approved):
Niice!

PR #2734, opened 2026-06-04
[Flags] Use string instead of UUID for sampled apps

nezaj:
What it says on the tin. Before we weren't coercing this as a uuid which was causes the lookup to fail in our config map


dwwoelfel (approved):
LGTM!

PR #2737, opened 2026-06-05
Remove sendTestEmail flag now that the backend is live

stopachka:
The `POST /dash/apps/:id/send-test-email` endpoint is now deployed, so the "Send a test email" button in the magic-code email editor no longer needs to be gated behind the `sendTestEmail` flag.

- Removed the `sendTestEmail` flag from `lib/flags.ts`
- Removed the `useFlag('sendTestEmail')` check in `Email.tsx` and render the button unconditionally
- Dropped the now-unused `useFlag` import


dwwoelfel (approved):
LGTM!

PR #2742, opened 2026-06-07
Log error message as JSON string to actually see what `instaql-queries` is

clarkg:
Right now `instantql-queries` is printed as unhelpful `[[Object]]`


dwwoelfel:
@clarkg would this work for you? https://github.com/instantdb/instant/pull/2745

clarkg:
> @clarkg would this work for you? #2745

@dwwoelfel that would work tyvm! very kind

PR #2748, opened 2026-06-09
Bump slot nums

dwwoelfel:
To be deployed after migrating the slots via `(migrate-invalidator-to-replica)` and `(migrate-aggregator-to-replica)` from https://github.com/instantdb/instant/pull/2747

### DIFF
--- a/server/src/instant/isn.clj
+++ b/server/src/instant/isn.clj
@@ -15,7 +15,6 @@
    slot_num in front of the LSN. Any time we migrate to a new db, we increment
    the slot_num."
   (:require
-   [instant.config :as config]
    [instant.util.json :as json])
   (:import
    (java.lang Comparable)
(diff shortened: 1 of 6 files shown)


server/src/instant/isn.clj
```diff
@@ -100,7 +99,7 @@
 (defn test-isn
   "Generates an isn. (test-isn i) will be less than (test-isn (inc i))"
   [^long i]
-  (->ISN config/invalidator-slot-num
+  (->ISN 0
```
nezaj, line 102: Just curious why do we set this to 0 now?

nezaj (approved):

server/src/instant/isn.clj
dwwoelfel: So that we don't have to update the tests every time we bump the slot num.

PR #2751, opened 2026-06-10
Fix count estimates on unique fields

dwwoelfel:
Fixes count estimates on unique fields. 

If you have a query like `{:books {:$ {:where {:id {:$in <list-of-ids>}}}}` and the sketch is overloaded, we might overestimate the number of books. If you put a limit on that query and we overestimate too much, then we'll try to use the triples_created_at to get better sorting.

This change will cap the count for each id in the list to 1. 

In case it causes a regression, there is a `disable-hint-plan-av-fix` flag we can set to `true`.

### DIFF
--- a/server/src/instant/db/datalog.clj
+++ b/server/src/instant/db/datalog.clj
@@ -1174,10 +1174,17 @@
                    0
                    (case (:type vs)
                      :total (:total sketch)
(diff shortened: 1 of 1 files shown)


stopachka (approved):
Niice!

PR #2752, opened 2026-06-10
[WWW] Add discord to footer

nezaj:
I noticed we no longer have a discord link on our front page. We used to put it as part of our top-bar but then removed it when we did the redesign.

Adding it back in to our footer so it's at least easier to find via ctrl+f. I also slightly re-organized the links

**Before**
CleanShot 2026-06-10 at 09 34 34@2x

**After**
CleanShot 2026-06-10 at 09 34 20@2x


dwwoelfel (approved):
LGTM!

stopachka (approved):
sgtm!

PR #2753, opened 2026-06-10
remove feature flag for app email verification

drew-harris:
removes the `use-app-email-verification?` flag. done by claude checked by me

tested a full custom email verification flow

Also removes setting user_id for app_email_senders in the tests so that we can drop the user_id field in the next pr


server/resources/migrations/113_drop_user_id_email_senders.up.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE IF EXISTS app_email_senders
```
dwwoelfel: Can you take out the `if exists` checks? If the table or column doesn't exist, then something weird has happened and it should probably fail.

dwwoelfel (approved):
LGTM!

PR #2754, opened 2026-06-10
drop user_id column from app_email_senders

drew-harris:
Drops the user_id column from 'app_email_senders'

image

## Deploy Plan
Wait for https://github.com/instantdb/instant/pull/2753 to fully deploy
Run migration


dwwoelfel (approved):
LGTM!

PR #2755, opened 2026-06-10
Add row estimate hints

dwwoelfel:
Adds hints for number of rows returned by a join.

We're getting some queries where the indexes are good, but postgres chooses a bad join method.

This PR will add row count hints, which causes postgres to choose the appropriate join method. 

In my testing, it was almost always better, although there were a few small regressions. I made a couple of fixes to address the regressions:

1. We cap the row estimate to the `limit` if the pattern has a `limit` field.
2. We skip the row estimates on ref joins if one side of the join has a large row-count in the symbol-map

There are flags for turning the whole thing off (`enable-row-hints-globally`), turning it off for an individual query (`disable-row-hints-query-hashes`), and enabling it for a single query (`enable-row-hints-query-hashes`).

There is also a flag for disabling the `limit` fix for page patterns (`disable-cap-symbol-map-by-page-limit`).

### DIFF
--- a/server/src/instant/db/datalog.clj
+++ b/server/src/instant/db/datalog.clj
@@ -1713,7 +1713,7 @@
 (defn annotate-pattern-with-hints
   "Annotates the pattern with best-index and adds updates the symbol-map."
   [ctx symbol-map pattern]
-  (let [row-estimate (estimate-rows ctx pattern)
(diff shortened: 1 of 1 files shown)


stopachka (approved):
Aweesome!

dwwoelfel:
@coderabbitai review

PR #2756, opened 2026-06-10
export route handler types (#2756) [bump to v1.0.48]

drew-harris:


client/packages/core/src/routeHandlerProtocol.ts
```diff
@@ -0,0 +1,24 @@
...
+    appId: string;
+    user: User | null;
+  };
+};
+
+/** Known `type` values for Instant route handler request bodies. */
+export type InstantRouteHandlerType = keyof InstantRouteHandlerPayloadByType;
+
+/** A valid request body for Instant's first-party route handler protocol. */
+export type InstantRouteHandlerBody<
```
nezaj, line 18: Should we attach this to `createInstantRouteHandler` in `createRouteHandler.ts`

nezaj (approved):
LGTM w/ one nit!

PR #2757, opened 2026-06-10
Don't throw errors when we shut down the loadbalancer watcher

dwwoelfel:
This catches the interrupted exceptions that get thrown when we stop the loadbalancer watcher. 

Should makes deploys a little less noisy.

### DIFF
--- a/server/src/instant/loadbalancer.clj
+++ b/server/src/instant/loadbalancer.clj
@@ -11,7 +11,6 @@
   (:import
    (java.lang InterruptedException)
    (java.time Duration Instant)
-   (software.amazon.awssdk.core.exception AbortedException)
    (software.amazon.awssdk.services.sns SnsClient)
    (software.amazon.awssdk.services.sns.model ListSubscriptionsByTopicRequest SubscribeRequest Subscription UnsubscribeRequest)
    (software.amazon.awssdk.services.sqs SqsClient)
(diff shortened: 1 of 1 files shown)


server/src/instant/loadbalancer.clj
```diff
@@ -99,24 +101,31 @@
...
+                              (delete-message queue-url message))
+                            :continue)
+                          (catch Throwable t
+                            (when-not (or (instance? InterruptedException t)
+                                          (and (instance? AbortedException t)
+                                               (instance? InterruptedException (.getCause t))))
+                              (tracer/record-exception-span! t {:name "loadbalancer/get-messages-error"}))
+                            :backoff))]
+        (case next-action
+          :quit :quit
```
nezaj: Claude points out this is dead code

nezaj (approved):
LGTM!

server/src/instant/loadbalancer.clj
```diff
@@ -99,24 +101,31 @@
...
+                            (doseq [^Message message messages]
+                              (on-message (json/<-json (.body message) true))
+                              (delete-message queue-url message))
+                            :continue)
+                          (catch Throwable t
+                            (when-not (or (instance? InterruptedException t)
+                                          (and (instance? AbortedException t)
+                                               (instance? InterruptedException (.getCause t))))
+                              (tracer/record-exception-span! t {:name "loadbalancer/get-messages-error"}))
+                            :backoff))]
```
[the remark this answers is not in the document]
dwwoelfel: Fixed!

server/src/instant/loadbalancer.clj
dwwoelfel: Fixed! And improved shutdown logic by passing in a function to check if we're supposed to shutdown.

PR #2759, opened 2026-06-11
Update our ECR publish so that we can gc

dwwoelfel:
Changes our ECR publishing so that we can set up lifecycle rules to delete old containers.

Two changes to how we publish:
1. Use a separate repo for the build cache. This lets us more easily expire the build cache images without worrying about accidentally removing images we care about.
2. Tag the latest image with a `prod` tag, so that we definitely don't delete the version we rely on in prod

Then the other change is to add a lifecycle rule that will expire old images. I put that in the `config/ecr/README.md` file.

After merging this in, I'll add the lifecycle rules for the ecr repo and preview what it deletes to confirm we're not deleting stuff we need.

### DIFF
--- a/.github/workflows/clojure.yml
+++ b/.github/workflows/clojure.yml
@@ -347,15 +347,29 @@ jobs:
       id: login-ecr
       uses: aws-actions/amazon-ecr-login@v2
 
+    - name: Determine deploy target
+      id: target
+      env:
+        COMMIT_MSG: ${{ github.event.head_commit.message }}
+      run: |
+        if [[ "$COMMIT_MSG" == *"[staging]"* ]]; then
+          echo "target=staging" >> "$GITHUB_OUTPUT"
+        else
+          echo "target=prod" >> "$GITHUB_OUTPUT"
+        fi
+
     - name: Build, tag, and push docker image to Amazon ECR
       working-directory: server
       env:
         IMAGE_TAG: ${{ github.sha }}
+        DEPLOY_TARGET: ${{ steps.target.outputs.target }}
       run: |
         docker buildx build --push \
-          --cache-to mode=max,image-manifest=true,oci-mediatypes=true,compression=zstd,type=registry,ref=597134865416.dkr.ecr.us-east-1.amazonaws.com/instant-prod-ecr:cache \
+          --cache-to mode=max,image-manifest=true,oci-mediatypes=true,compression=zstd,type=registry,ref=597134865416.dkr.ecr.us-east-1.amazonaws.com/instant-prod-ecr-buildcache:cache \
+          --cache-from type=registry,ref=597134865416.dkr.ecr.us-east-1.amazonaws.com/instant-prod-ecr-buildcache:cache \
           --cache-from type=registry,ref=597134865416.dkr.ecr.us-east-1.amazonaws.com/instant-prod-ecr:cache \
-          -t 597134865416.dkr.ecr.us-east-1.amazonaws.com/instant-prod-ecr:$IMAGE_TAG .
+          -t 597134865416.dkr.ecr.us-east-1.amazonaws.com/instant-prod-ecr:$IMAGE_TAG \
+          -t 597134865416.dkr.ecr.us-east-1.amazonaws.com/instant-prod-ecr:$DEPLOY_TARGET .
 
   publish-eb:
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


.github/workflows/clojure.yml
```diff
@@ -347,15 +347,28 @@ jobs:
...
+        DEPLOY_TARGET: ${{ steps.target.outputs.target }}
       run: |
         docker buildx build --push \
-          --cache-to mode=max,image-manifest=true,oci-mediatypes=true,compression=zstd,type=registry,ref=597134865416.dkr.ecr.us-east-1.amazonaws.com/instant-prod-ecr:cache \
-          --cache-from type=registry,ref=597134865416.dkr.ecr.us-east-1.amazonaws.com/instant-prod-ecr:cache \
-          -t 597134865416.dkr.ecr.us-east-1.amazonaws.com/instant-prod-ecr:$IMAGE_TAG .
+          --cache-to mode=max,image-manifest=true,oci-mediatypes=true,compression=zstd,type=registry,ref=597134865416.dkr.ecr.us-east-1.amazonaws.com/instant-prod-ecr-buildcache:cache \
+          --cache-from type=registry,ref=597134865416.dkr.ecr.us-east-1.amazonaws.com/instant-prod-ecr-buildcache:cache \
+          -t 597134865416.dkr.ecr.us-east-1.amazonaws.com/instant-prod-ecr:$IMAGE_TAG \
+          -t 597134865416.dkr.ecr.us-east-1.amazonaws.com/instant-prod-ecr:$DEPLOY_TARGET .
```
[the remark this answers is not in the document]
dwwoelfel: > clojure.yml tags/pushes the image with DEPLOY_TARGET (staging/prod), but publish-eb (lines 373-408) doesn’t reference DEPLOY_TARGET, so the EB deployment target won’t change based on the tag.

It references IMAGE_TAG, which is what we want to pull from. We don't want to pull from DEPLOY_TARGET since that tag moves.

> The repo search for staging.*env and Instant-docker-prod-env found no matches; update publish-eb/EB config to route by DEPLOY_TARGET, or document that staging is only an ECR lifecycle tag and EB staging deploys are handled manually.

That's not how this works. We have a paper doc with instructions for deploying to staging.

server/.elasticbeanstalk/config.yml
```diff
@@ -1,6 +1,6 @@
 branch-defaults:
   default:
-    environment: instant-docker-prod-env
+    environment: Instant-docker-prod-env-2
```
nezaj, line 3: Is this intentional?

nezaj: Ah I guess we have a new env in AWS?

dwwoelfel: Yep! It's been like that for a while, but nobody uses `eb` directly any more since we have `make deploy`.

server/config/ecr/README.md
```diff
@@ -0,0 +1,119 @@
...
+
+```json
+{
+  "rules": [
+    {
+      "rulePriority": 1,
+      "description": "Keep the moving prod and staging tags forever",
+      "selection": {
+        "tagStatus": "tagged",
+        "tagPatternList": ["prod", "staging"],
```
nezaj: Question, was this meant to select on images that have both "prod" and "staging" ?

Claude pointed out from the ECR docs having multiple tags acts as an AND

https://docs.aws.amazon.com/AmazonECR/latest/userguide/lifecycle_policy_parameters.html#lp_tag_pattern_list

nezaj (approved):
LGTM but one question

server/config/ecr/README.md
dwwoelfel: Good catch! Fixed.

server/config/ecr/README.md
```diff
@@ -0,0 +1,130 @@
...
+        "countUnit": "days",
+        "countNumber": 7
+      },
+      "action": {"type": "expire"}
+    }
+  ]
+}
+```
+
+ECR is aware of image indexes, a tagged manifest list protects its referenced platform manifests even when those children appear untagged. So rule 3 only expires true orphans.
```
nezaj:
```suggestion
ECR is aware of image indexes, a tagged manifest list protects its referenced platform manifests even when those children appear untagged. So rule 4 only expires true orphans.
```

PR #2761, opened 2026-06-11
email cli auth docs

drew-harris:
Docs explaining how to use instant-cli to manage the email templates

### DIFF
--- a/client/www/app/docs/cli/page.md
+++ b/client/www/app/docs/cli/page.md
@@ -261,11 +261,17 @@ npx instant-cli@latest auth email pull
 
 It will create a file like this:
 ```typescript
+// We provide a few dynamic variables for you to use in your email:
+// {code}, the magic code e.g. 123456
+// {app_title}, your app's title, i.e. test-fresh
+// {user_email}, the user's email address, e.g. happyuser@gmail.com
+// {expiration}, the magic code expiration, e.g. 10 minutes
+// Note: {code} is required in both the subject and body.
 const email = {
   authEmail: {
     subject: "{code} is your verification code for {app_title}",
-    senderName: "My App",
-    senderEmail: "verify@myapp.com",
+    senderName: "test-fresh",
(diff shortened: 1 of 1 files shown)


client/www/app/docs/cli/page.md
```diff
@@ -249,3 +249,76 @@ npx instant-cli@latest auth client update \
...
+npx instant-cli@latest auth email pull
+```
+
+It will create a file like this:
+```typescript
+const email = {
+  authEmail: {
+    subject: "{code} is your verification code for {app_title}",
+    senderName: "My App",
+    senderEmail: "verify@myapp.com",
```
nezaj: Nit. perhaps we should put what the default value is

client/www/app/docs/cli/page.md
```diff
@@ -249,3 +249,76 @@ npx instant-cli@latest auth client update \
...
+        </p>
+      </div>
+    </div>`,
+  },
+};
+
+export default email;
+````
+
+We provide a few dynamic variables for you to use in your email:
```
nezaj: Instead of listing these out we should link to the variables in the custom email doc. That way we won't have two update in two places if we make changes

client/www/app/docs/cli/page.md
```diff
@@ -249,3 +249,76 @@ npx instant-cli@latest auth client update \
...
+
+{app_title}: your app's title, i.e. test-fresh
+
+{user_email}: the user's email address, e.g. happyuser@gmail.com
+
+{expiration}: the magic code expiration, e.g. 10 minutes
+
+
+{% callout type="note" %}
+Note: {code} is required in both the subject and body.
```
nezaj, line 301: Instead of a callout, we can just leave a small note about this here -- effectively this whole section can just be condensed to linking to the email section + containing this note about including the code

nezaj (approved):
LGTM w/ nits

PR #2763, opened 2026-06-12
[Metrics] Speed up new activity query

nezaj:
I gave Claude the output of `EXPLAIN` on the current query and it identified that finding the earliest tx date for an app was the most expensive part (huge scan)

Instead we can do two things:

* Restrict transaction scan to only new txs since our last entry for our daily_app_transactions table
* Use active_date from daily_app_transactions table to get earliest date for existing apps

We'll still get all new transactions and look at a lot less data. Tested this for todays job and the query time wen from 6.5 min -> ~50 seconds


dwwoelfel (approved):
Nice!

PR #2765, opened 2026-06-19
[WWW] Add padding to bottom of explorer form

nezaj:
I remember we mentioned this during our weekly meeting and I encountered it today. Thought it would be nice to clean up!

**Before**
CleanShot 2026-06-19 at 12 51 18@2x

**After**
CleanShot 2026-06-19 at 12 51 04@2x


stopachka (approved):
SGTM!

drew-harris (approved):

PR #2770, opened 2026-06-26
[WWW] Fix mobile recipes

nezaj:
What it says on the tin!

## **Before**
CleanShot 2026-06-26 at 07 52 41@2x

## **After**
CleanShot 2026-06-26 at 07 52 07@2x


dwwoelfel (approved):
LGTM!

PR #2772, opened 2026-06-29
[Docs] Fix recommended workflow link

nezaj:
Saw a user left feedback about the link being broken -- fix is here!


stopachka (approved):
SGTM!

PR #2773, opened 2026-07-01
Disable 1password icon on non-sensitive fields

nezaj:
1password will try to infer which fields may be sensitive and add in an icon. It will commonly do this when there is only one input for a form. Got complaints about this for things like our create namespace modal so asked Claude to do a sweep to find where we can fix this.

This updates our create + rename namespace / org modals, as well as our OAuth config forms to hide 1password and bitwarden icons.


dwwoelfel (approved):
LGTM!

stopachka (approved):
SGTM!

PR #2774, opened 2026-07-06
[Storage] Add index on file size

nezaj:
Noticed our internal storage usage query has been consistently timing out. Talked with Claude and it suggested adding this index which seemed reasonable to me!


stopachka (approved):
SGTM!

nezaj:
Discussed offline, even with this large apps still seem to be a bottle-neck. will close for now and dig in more later

PR #2776, opened 2026-07-06
[WWW] Add explicit width to sandbox icons

nezaj:
Got a user report that icons save/trash icons weren't appearing in the sandbox page on Safari. Was able to confirm and pushing up a quick fix here!

**Before**
CleanShot 2026-07-06 at 14 36 05@2x

**After**
CleanShot 2026-07-06 at 14 36 16@2x


stopachka (approved):
SGTM!

PR #2782, opened 2026-07-20
resolve storage adapter native modules from the sandbox app

nezaj:
We got a PR to add a expo-sqlite package for react native https://github.com/instantdb/instant/pull/2780

Was able to verify it worked but hit one snag with dependency resolution for our sandbox. Putting up this PR to fix that so we can also update #2780 with a sandbox example using expo-sqlite


stopachka (approved):
SGTM!

PR #2783, opened 2026-07-20
set startup plan by default for self-hosted

drew-harris:
- Make stripe-secret in config.clj return nil for empty strings (codex)
 - Allow stripe related columns to be null in subscriptions table
 - Create routes that let you upgrade apps to pro and orgs to startup for free with one click. (codex)
 - Wire routes into dashboard

 Other changes:
  - adds a docker-compose.local.yml file for use testing future self hosting changes locally
  - Remove `deploy: replicas: 2` from standard docker compose (should be a no-op anyways) 

image
image
image
image


dwwoelfel (approved):
LGTM!

PR #2785, opened 2026-07-20
Add per-app maintenance mode

stopachka:
**Try it locally:** http://localhost:4000/play/app-status

**Problem** 

Users may want to do migrations, or perform changes that require _disabling all writes_ 

**Solution** 

Added a "Read Only Mode" Toggle. 

Users can set it in the dashboard: 

CleanShot 2026-07-20 at 16 49 23@2x

And all writes will start to fail with a descriptive error. 

If the developer wants to show some custom UI when in maintenance mode, they can subscribe to a hook: 

```typescript
function App() {
  const { isLoading, isReadOnly } = db.useAppStatus()
  if (isLoading) return null
  if (isReadOnly) return <MaintenanceBanner />
  // ...
}
```

## Implementation

**Data Model**

Apps now have a new `status` column, with values `active | read-only | disabled`. 

I added a `disabled` option, which also allows us to disable _both_ reads and writes. I don't currently make `disabled` part of the public featureset though.

**Reactivity** 

We listen to app changes, and send out an invalidation for apps that are listening, for `status`. 

**Why not use a system table?** 

I thought about making a `$settings` system table. This would let us hook into the existing reactive machinery. But this added more complications. We would have to special-case that query in the "disabled" case for example, and the initial message for the status would be delayed (instead of init-ok, it would come from a standard add-query-ok). I thought the app status column with a new invalidation message would be best.


client/packages/core/src/index.ts
```diff
@@ -827,6 +828,32 @@ class InstantCoreDatabase<
...
+  /**
+   * The app's maintenance-mode state. While read-only, reads and live
+   * queries keep working but writes are rejected. The state arrives on the
+   * connection handshake, so pair it with `subscribeConnectionStatus` if
+   * you need to know whether it's fresh.
+   *
+   * @example
+   *   const { isLoading, isReadOnly } = db.getAppStatus();
+   */
+  getAppStatus(): AppStatusState {
```
stopachka: imo this is not a public API we really want to offer to users. the place we use it is internal, so we can use the underlying reactor thing.

client/packages/core/src/index.ts
```diff
@@ -827,6 +828,32 @@ class InstantCoreDatabase<
...
+    return this._reactor.getAppStatusState();
+  }
+
+  /**
+   * Listen for changes to the app's maintenance-mode state. The callback
+   * fires immediately with the current state.
```
stopachka: imo in this doc we should explain the isReadOnly thing we had in getAppStatus

client/packages/core/src/Reactor.js
```diff
@@ -888,6 +894,16 @@ export default class Reactor {
         this._trySetRoomConnected(roomId, false);
         break;
       }
+      case 'app-status-changed': {
+        const prevStatus = this._appStatus;
+        this._setAppStatus(msg.status);
+        if (prevStatus === 'disabled' && msg.status !== 'disabled') {
+          // Leaving `disabled`: refreshes were paused server-side, so restart
+          // the socket to re-add every query and get fresh results.
+          this._startSocket();
```
stopachka, line 917: q: will _startSocket _stop_ this existing socket? i.e we need a full on restart no?

client/packages/core/src/Reactor.js
```diff
@@ -2081,6 +2100,34 @@ export default class Reactor {
...
+  /**
+   * @param {'active' | 'read-only' | 'disabled' | undefined} status
+   */
+  _setAppStatus(status) {
+    if (!status || status === this._appStatus) return;
+    this._appStatus = status;
+    // `disabled` is an operator-level state we don't surface to apps;
+    // clients only see whether writes are paused.
+    const isReadOnly = status !== 'active';
+    if (this._appStatusState.isReadOnly !== isReadOnly) {
```
stopachka: do we need this "dual" cache? one for appStatus, and one for appStatusState? 

Perhaps at this level we should only be caching appStatus? 

Perhaps the Core abstraction can handle this collapse

server/src/instant/admin/routes.clj
```diff
@@ -144,6 +145,7 @@
   (let [query (ex/get-param! req [:body :query] #(when (map? %) %))
         inference? (-> req :body :inference? boolean)
         {:keys [app-id] :as perms} (get-perms! req :data/read)
+        _ (app-status-model/assert-read-allowed! app-id)
```
stopachka: perhaps this assert-read-allowed should be pushed _down_ into instaql? that way all callers will have this check on by default?

server/src/instant/db/transaction.clj
```diff
@@ -569,6 +570,7 @@
                                        :num-tx-steps (count tx-step-vecs)
                                        :detailed-tx-steps (pr-str tx-step-vecs)}}
         (prevent-system-catalog-updates! app-id opts)
+        (app-status-model/assert-write-allowed! app-id tx-step-maps)
```
stopachka: imo we should not be "detecting" sign out in this way. perhaps we should have a bypass-app-status-disabled? opt that we pass in, _only_ from the signout handler?

server/src/instant/jdbc/cache_evict.clj
```diff
@@ -70,6 +71,17 @@
                               parse-json-uuid)]
       (app-stream-model/notify-machine-id-changed stream-id machine-id))))
 
+(defn notify-app-status-changed
+  "Pushes the app's status to this machine's sessions when an apps row
```
stopachka: Why _not_ diff it? imo we already have this data no? or we don't?

server/src/instant/model/app_status.clj
```diff
@@ -0,0 +1,81 @@
+(ns instant.model.app-status
+  "Operational status (maintenance mode) for an app.
+
+   `active`    - everything works
+   `read-only` - reads keep working, durable writes are rejected
+   `disabled`  - reads are rejected too; init and sign-out stay available
+
+   Status lives on the `apps` row (control plane), so the toggle itself never
```
stopachka: No need to add this para in the comment, it does not add much

server/src/instant/model/app_status.clj
```diff
@@ -0,0 +1,81 @@
...
+   enters the Instant transaction machinery and can't lock itself out.
+   Reads go through the WAL-evicted app cache, so enforcement adds no SQL to
+   hot paths."
+  (:require
+   [instant.jdbc.aurora :as aurora]
+   [instant.jdbc.sql :as sql]
+   [instant.util.exception :as ex]))
+
+;; Resolved lazily to avoid a circular dependency:
```
stopachka: hmm, any way we can avoid these circual dependencies? i.e can we rewrite this osmehow? would be awesome if we could. -- i.e perhaps this can _be_ inside model.app?

stopachka: hmm, any way we can avoid these circual dependencies? i.e can we rewrite this osmehow? would be awesome if we could. -- i.e perhaps this can _be_ inside model.app?

server/src/instant/reactive/session.clj
```diff
@@ -467,7 +470,12 @@
         current-user (-> auth :user)
         admin? (-> auth :admin?)
         {:keys [attrs table-info]} (get-attrs (:app auth))
-        stale-queries (rs/get-stale-instaql-queries store app-id sess-id)
+        ;; While disabled we refuse to recompute or send data. Queries stay
```
stopachka: is this the right way to think about this? maybe we can just do assert-read-allowed! in the beginning? wdyt?

client/packages/core/src/index.ts
stopachka: Done: removed getAppStatus from the public API. subscribeAppStatus is the one public surface now, and the react hook consumes it internally with a ref for snapshot stability.

client/packages/core/src/index.ts
stopachka: Done: folded the isReadOnly explanation into the subscribeAppStatus doc (writes rejected server-side, reads and live queries keep working, isLoading until the first handshake).

client/packages/core/src/Reactor.js
stopachka: Yes, it is a full restart: _startSocket creates the new transport and explicitly closes the previous one if it is open (the prevTransport.close() branch), and the old socket's onclose is ignored because it is no longer the current transport. New socket goes through init and re-adds every query.

client/packages/core/src/Reactor.js
stopachka: Done: the reactor now only tracks the raw status and notifies with it. The collapse to {isLoading, isReadOnly} lives in InstantCoreDatabase.subscribeAppStatus, and the react hook dedupes with a ref.

server/src/instant/admin/routes.clj
stopachka: Agree with the direction (default-on for every caller). The wrinkle is that instaql also serves paths that must keep working while disabled: init's refresh-token validation, the internal system-catalog reads that auth flows do, and control-plane reads like dash auth data. So pushing it into instaql means adding a ctx opt-out for those callers. Leaving that as a follow-up; for now the gates are at add-query, refresh, admin query-post, and join-room.

server/src/instant/db/transaction.clj
stopachka: Done: removed the shape detection. transact-without-tx-conn-impl! now does (when-not (:bypass-app-status? opts) (assert-write-allowed! app-id)), and only the sign-out fns in app-user-refresh-token pass the opt. A raw $userRefreshTokens delete without the opt is now rejected like any other write (tested), and /admin/sign_out verified working while read-only.

server/src/instant/jdbc/cache_evict.clj
stopachka: We do not have the old value: apps uses the default replica identity, so the WAL update record only carries the new row plus the PK. Updated the comment to say this. If we ever want a real diff we can set REPLICA IDENTITY FULL on apps (tiny table) or compare against the app cache before evicting; clients already no-op on an unchanged status, so I left it.

server/src/instant/model/app_status.clj
stopachka: Dropped; the whole namespace is gone now (see the circular-dependency thread).

server/src/instant/model/app_status.clj
stopachka: Done. The cycle was model.app -> system-catalog-ops -> db.transaction, and the only reason model.app needed system-catalog-ops was get-dash-auth-data, which is really dash-route logic. Moved it into dash/routes.clj, folded the status fns directly into instant.model.app, and deleted app_status.clj. db.transaction now requires model.app cleanly; no more requiring-resolve.

server/src/instant/reactive/session.clj
stopachka: Yeah, simpler. Done: handle-refresh! now calls assert-read-allowed! up front. The client error handler just console.errors an error for an op with no listener, so throwing here is benign for old clients, and refreshes for a disabled app mostly stop arriving anyway since writes are blocked.

client/packages/core/src/Reactor.js
```diff
@@ -637,6 +642,7 @@ export default class Reactor {
       case 'init-ok': {
         this._setStatus(STATUS.AUTHENTICATED);
         this._reconnectTimeoutMs = 0;
+        this._setAppStatus(msg['app-status']?.status);
```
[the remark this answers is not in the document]
stopachka: Applied: init-ok now defaults to 'active' when the server does not send app-status.

client/packages/core/src/Reactor.js
```diff
@@ -2081,6 +2100,34 @@ export default class Reactor {
...
+  }
+
+  subscribeAppStatus(cb) {
+    this.appStatusCbs.push(cb);
+    cb(this._appStatusState);
+
+    return () => {
+      this.appStatusCbs = this.appStatusCbs.filter((x) => x !== cb);
+    };
+  }
```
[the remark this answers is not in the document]
stopachka: Keeping the immediate callback intentionally: getAppStatus was removed from the public API, so subscribing is the only way to read the current state, and the react hook dedupes via a ref before notifying useSyncExternalStore.

client/packages/core/src/index.ts
```diff
@@ -827,6 +828,31 @@ class InstantCoreDatabase<
     return this._reactor.subscribeConnectionStatus(cb);
   }
 
+  /**
+   * Listen for changes to the app's maintenance-mode state. While
+   * `isReadOnly` is true, writes are rejected server-side but reads and
+   * live queries keep working. `isLoading` is true until the first
```
stopachka: you can remove the sentence starting with `isLoading` is true until... (imo not necc)

client/packages/core/src/index.ts
```diff
@@ -827,6 +828,31 @@ class InstantCoreDatabase<
...
+   * connection handshake delivers the state; the callback fires
+   * immediately with the current state and again on every change.
+   *
+   * @example
+   *   const unsub = db.subscribeAppStatus(({ isLoading, isReadOnly }) => {
+   *     if (!isLoading && isReadOnly) showMaintenanceBanner();
```
stopachka: imo simpler would be to write 

if (isLoading) return
if (isReadOnly) showMaintenanceBanner()

client/packages/react-common/src/InstantReactAbstractDatabase.tsx
```diff
@@ -371,6 +377,44 @@ export default abstract class InstantReactAbstractDatabase<
...
   };
 
+  /**
+   * Observe the app's maintenance-mode state. While read-only, reads and
+   * live queries keep working but writes are rejected. Returns a loading
+   * state until the first connection handshake completes.
+   *
+   * @example
+   *  function App() {
+   *    const { isLoading, isReadOnly } = db.useAppStatus()
```
stopachka, line 386: would do the `isLoading` return null pattern instead

server/src/instant/admin/routes.clj
```diff
@@ -144,6 +144,7 @@
   (let [query (ex/get-param! req [:body :query] #(when (map? %) %))
         inference? (-> req :body :inference? boolean)
         {:keys [app-id] :as perms} (get-perms! req :data/read)
+        _ (app-model/assert-read-allowed! app-id)
```
stopachka: is this even necc? esp if we push this down into instaql?

server/src/instant/model/app.clj
```diff
@@ -388,6 +331,53 @@
...
+
+(defn assert-read-allowed!
+  "Rejects data reads while the app is disabled."
+  [app-id]
+  (when (= :disabled (get-status app-id))
+    (ex/throw-app-disabled!)))
+
+(defn set-status!
+  "Plain UPDATE on apps; never an Instant transaction, so it works in every
+   status. The WAL evicts the app cache on every other machine."
```
stopachka: no need to say never an Instant transaction, so it works in every status.

server/src/instant/reactive/session.clj
```diff
@@ -464,6 +466,9 @@
                                           :attributes {:session-id sess-id}})
   (let [auth (get-auth! store sess-id)
         app-id (-> auth :app :id)
+        ;; While disabled we refuse to recompute or send data. Queries stay
```
stopachka: no need for comment

server/src/instant/reactive/session.clj
```diff
@@ -227,6 +228,7 @@
 (defn- handle-add-query! [store sess-id {:keys [q client-event-id return-type] :as _event}]
   (let [{:keys [app user admin?]} (get-auth! store sess-id)
         {app-id :id} app
+        _ (app-model/assert-read-allowed! app-id)
```
stopachka: is this necc? esp if we push this down into instaql?

client/packages/core/src/index.ts
stopachka: Removed.

client/packages/core/src/index.ts
stopachka: Done, example uses the early-return style now.

client/packages/react-common/src/InstantReactAbstractDatabase.tsx
stopachka: Done, example is the isLoading return null pattern now.

server/src/instant/admin/routes.clj
stopachka: Pushed it down: instaql/query now does the check itself, with :skip-app-status-check? set by the system-catalog read helpers (auth flows and control-plane reads must work in every status). Removed this per-caller gate and the ws add-query one; verified while disabled that admin query returns app-disabled via the instaql gate and that refresh-token validation still works.

server/src/instant/model/app.clj
stopachka: Trimmed.

server/src/instant/reactive/session.clj
stopachka: Removed, along with the assert itself: the refresh recompute goes through instaql now, so the gate there covers it.

server/src/instant/reactive/session.clj
stopachka: Yep, removed. The instaql-level check covers add-query now (the join-room gate stays since presence does not go through instaql).

client/packages/core/src/index.ts
```diff
@@ -827,6 +828,31 @@ class InstantCoreDatabase<
...
+    ): AppStatusState =>
+      status === undefined
+        ? { isLoading: true, isReadOnly: undefined }
+        : { isLoading: false, isReadOnly: status !== 'active' };
+    cb(toState(this._reactor._appStatus));
+    return this._reactor.subscribeAppStatus(
+      (status: 'active' | 'read-only' | 'disabled') => cb(toState(status)),
+    );
+  }
+
```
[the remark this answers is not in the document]
stopachka: This is as intended.

server/src/instant/dash/routes.clj
```diff
@@ -562,9 +564,63 @@
...
+                           "redirect_to" (get client "redirectTo")
+                           "use_shared_credentials" (boolean (get client "useSharedCredentials"))})
+                        $oauthClients)]
+       {:data {"oauth_service_providers" providers
+               "oauth_clients" clients
+               "authorized_redirect_origins" redirect-origins}}))))
+
 (defn dash-apps-auth-get [req]
   (let [{{app-id :id} :app} (req->app-accepting-superadmin-or-ref-token! :collaborator :apps/read req)
-        {:keys [data]} (app-model/get-dash-auth-data {:app-id app-id})]
```
stopachka, line 567: curious, why did we move this out? if this was in app-model, and we _pushed_ this logic in there, there would be no circular deps, and we could ahve had it there?

server/src/instant/db/instaql.clj
```diff
@@ -1813,6 +1814,10 @@
     result))
 
 (defn query [ctx o]
+  ;; System-catalog reads (auth flows, control plane) set
+  ;; :skip-app-status-check? -- they must work in every status.
+  (when-not (:skip-app-status-check? ctx)
```
stopachka: the names here are off. maybe we can do skip-app-status-read-check? and in the other place we have skip-status-write-check? 

not sure, but the different naming across this and tx is weird

stopachka: Agreed it was inconsistent. Unified on one keyword for both: :skip-app-status-check? is now what the tx layer checks in opts and what instaql checks in ctx (sign-out passes it for writes, the system-catalog read helpers pass it for reads). Went with one name instead of read/write variants since it is the same concept: this operation is exempt from status enforcement.

server/src/instant/dash/routes.clj
stopachka: It is load-bearing for the cycle: db.transaction needs the status check, so it requires model.app; get-dash-auth-data needs query-op, so it pulls system-catalog-ops, which requires db.transaction. Keeping it in model.app closes the triangle transaction -> app -> system-catalog-ops -> transaction. One of the two had to leave model.app: the status fns (the separate app-status ns you asked to fold in) or get-dash-auth-data. I moved the latter since it is dash response shaping with exactly one caller. If you would rather keep it in model.app, the alternative is going back to a separate status namespace.

stopachka: Moved it to a fresh instant.model.app-auth-data ns (requires only aurora, sql, and system-catalog-ops; nothing requires it except dash routes, so no cycle). Also switched apps to REPLICA IDENTITY FULL (migration 115) so cache-evict diffs old vs new status from the WAL record and only notifies sessions on a real status change; verified a rename sends nothing and a toggle sends exactly one event per session.

client/packages/core/src/index.ts
```diff
@@ -827,6 +828,31 @@ class InstantCoreDatabase<
     return this._reactor.subscribeConnectionStatus(cb);
   }
 
+  /**
+   * Listen for changes to the app's maintenance-mode state. While
```
stopachka: imo the desc in the react v is better: 

```
   * Observe the app's maintenance-mode state. While read-only, reads and
   * live queries keep working but writes are rejected.
```

Make this something similar

server/src/instant/db/instaql.clj
stopachka: imo this should be like skip-app-status-read-check?  

no?

server/src/instant/db/transaction.clj
```diff
@@ -569,6 +570,9 @@
                                        :num-tx-steps (count tx-step-vecs)
                                        :detailed-tx-steps (pr-str tx-step-vecs)}}
         (prevent-system-catalog-updates! app-id opts)
+        ;; Sign-out passes :skip-app-status-check? -- it must work in every status
+        (when-not (:skip-app-status-check? opts)
```
stopachka: imo skip-app-status-write-check?

client/packages/core/src/index.ts
stopachka: Done, matched the react phrasing.

server/src/instant/db/instaql.clj
stopachka: Renamed to :skip-app-status-read-check? (here and in the system-catalog read helpers).

server/src/instant/db/transaction.clj
stopachka: Renamed to :skip-app-status-write-check? (here and at the sign-out call sites). Re-verified while disabled: sign-out still works and auth reads still execute.

client/packages/react-common/src/InstantReactAbstractDatabase.tsx
```diff
@@ -371,6 +377,42 @@ export default abstract class InstantReactAbstractDatabase<
...
+   *    if (isLoading) return null
+   *    if (isReadOnly) return <MaintenanceBanner />
+   *    // ...
+   *  }
+   */
+  useAppStatus = (): AppStatusState => {
+    const stateRef = useRef<AppStatusState>(defaultAppStatusState);
```
stopachka: Q: Shold we be doing defaultAppStatusState, or effectively "getting" whatever the current value is in reactor? i.e this way if we re-mount a component we dont' get into a loading state, see what i mean?

stopachka: Good catch, fixed. The reactor now derives the collapsed state on demand (getAppStatusState, a getter off the single raw _appStatus, not a second cache), core's subscribeAppStatus uses it for the immediate emission, and the hook seeds its ref from it. A re-mounted component renders with the current state on the first frame; defaultAppStatusState only remains as the SSR snapshot.

server/src/instant/model/app_auth_data.clj
```diff
@@ -0,0 +1,59 @@
+(ns instant.model.app-auth-data
+  (:require
+   [instant.jdbc.aurora :as aurora]
+   [instant.jdbc.sql :as sql]
+   [instant.system-catalog-ops :refer [query-op]]))
+
+(defn get-dash-auth-data [{:keys [app-id]}]
```
stopachka, line 7: Moved this to it's own ns to avoid a cyclic dependency: 

We started requiring `instant.model.app` inside `instant.db.transaction`. 

And `instant.model.app` required `system-catalog-ops`

But `system-catalog-ops` requires `instant.db.transaction`.

client/www/components/admin/AdminPage.tsx
```diff
@@ -804,6 +805,109 @@ function InviteTeamMemberDialog({
...
+        ))}
+      <Dialog title="Pause writes" {...confirmDialog}>
+        <div className="flex flex-col gap-2">
+          <SubsectionHeading>Pause writes</SubsectionHeading>
+          <Content>
+            Writes from your users will be rejected while paused, including
+            offline writes queued on their devices. Reads, live queries, and
+            presence keep working.
+          </Content>
+          <Button
```
drew-harris, line 893: image
I think this should just be a toggle switch with a confirmation dialog whenever you flip it to "ON"

dwwoelfel (approved):
LGTM!

PR #2788, opened 2026-07-21
Use tls 1.2 for copy connection

dwwoelfel:
I found this issue while bootstrapping the triples size (#2764). It will also bite us for backups, so I'm fixing it now.

After a certain amount of data has passed through the postgres copy pipe, the jdk's tls 1.3 stack will trigger a key update before it will accept more data. Postgres is only sending data, it's not checking for incoming data, so it never responds. We get stuck in a traffic jam where the jdk is waiting for the response to the key update and postgres is waiting for the jdk to take the data it sent last.

For the copy command, it happened when postgres had processed ~218gb of data. 

Java 27 should fix this issue (https://bugs.openjdk.org/browse/JDK-8329548). For now we can just downgrade to tls 1.2 for the connection we'll use for copy.

This PR creates a socket wrapper that will force a downgrade to 1.2. We set the ssl factory class when we create a new connection for copy-ing.


server/src/instant/jdbc/wal.clj
```diff
@@ -87,6 +87,28 @@
                                           props)]
     (.unwrap conn PgConnection)))
 
+(defn- get-pg-copy-ready-conn
+  "Given a db-spec, return a PGConnection that is suitable for running a
+   long-running COPY command."
+  ^PgConnection [db-spec]
+  (let [db-spec (if-let [secret-arn (:secret-arn db-spec)]
+                  (-> db-spec
+                      (dissoc db-spec :secret-arn)
```
nezaj: Nit. `db-spec` is already being threaded in so this expands to (dissoc db-spec db-spec :secret-arn)

Same issue in get-pg-replication-conn

nezaj (approved):
Aside from one nit LGTM!

server/src/instant/jdbc/wal.clj
dwwoelfel: Updated both!

PR #2792, opened 2026-07-22
allow google oauth environment variables for self hosted

drew-harris:
- modifies config.clj to allow for google oauth client id and secret
 - slightly more complicated diff because in the .edn file the google config is a nested object

tested e2e with a fresh google oauth pair and the docker-compose.local.yml file


dwwoelfel (approved):
LGTM

PR #2793, opened 2026-07-22
allow setting apiUri in instant.config.ts (#2793) [bump v1.0.52]

drew-harris:
Allows you to set apiURI in the instant.config.ts file to make it easier to use self hosted installations of instantdb


stopachka (approved):
SGTM! Maybe we should update docs to explain this?

PR #2794, opened 2026-07-22
keep swarm.yml version pinned to latest instead of test tag

drew-harris:
The image tag for the server service in the swarm.yml file was pinned to test-swarm when I made the swarm pr. Now it should be pinned to latest so that it keeps receiving updates.


stopachka (approved):
Nice!

PR #2796, opened 2026-07-24
use unique ids for hazelcast in swarm mode

drew-harris:
In self hosted mode with 2 instances of the backend server, hazelcast was using "dev" as instance ids which tricked the servers into seeing each other's sessions as orphaned and deleting them. 

This fixes it so that if docker swarm mode is enabled, the hazelcast instance will use random uuids as machine ids.


dwwoelfel (approved):
LGTM!

PR #2799, opened 2026-07-28
catch and log persistedObject GC errors (#2799) [bump to v1.0.53]

### LINKED ISSUE #2795: Floating `_gc()` promise causes unhandled rejection storm every minute when IndexedDB fails
**Describe the bug**

`PersistedObject.gc()` fires `_gc()` as a floating promise (no `await`, no `.catch`). When IndexedDB is unavailable or failing (blocked site data/cookies, privacy extensions, enterprise policies, some private-browsing modes), `_gc()` rejects every time it runs — about once per minute per reactor — producing a steady stream of unhandled promise rejections in the host app.

**To Reproduce**

1. Init an Instant client in a browser where IndexedDB is broken (e.g. block all site data for the origin, or make `indexedDB.open` fire `onerror`).
2. Keep the tab open.
3. Observe `unhandledrejection` events every ~60s (+ jitter), deferred to `window visible` in backgrounded tabs (idle callbacks don't run while hidden).

**Root cause (two bugs that combine)**

1. `client/packages/core/src/utils/PersistedObject.ts` — `gc()` schedules `_gc()` without handling its promise:
```ts
this._nextGc = setTimeout(
  () => {
    safeIdleCallback(() => {
      this._nextGc = null;
      this._gc(); // <-- floating: rejections here are unhandled
    }, 30 * 1000);
  },
  1000 * 60 + Math.random() * 500,
);
```
`gc()` re-arms after every `_writeToStorage`, so it runs ~every minute for the life of the page.

2. `client/packages/core/src/IndexedDBStorage.ts` — `_init()` rejects `_dbPromise` with the raw DOM `Event` (`request.onerror = (event) => reject(event)`). Since `_dbPromise` is cached, every subsequent storage op rejects with that same `Event` forever. `_gc()`'s first await is `this._persister.getAllKeys()`, which rethrows it.

**Observed impact**

In production (an app with two reactors on one page), affected users generate 2–3 unhandled rejections per minute for as long as they keep the tab open — one multi-hour session produced 300+ events and repeatedly triggered anomaly alerts (e.g. 372–523 events/hour spikes). PostHog captures the reason as `Event captured as exception with keys: isTrusted`. Confirmed present in `@instantdb/core` 1.0.41 through 1.0.52 (current latest).

**Expected behavior**

`_gc()` failures should be caught and logged (e.g. `this._gc().catch(e => this._log.info('gc failed', e))`), and ideally `_init()` should reject with a real `Error` (`request.error`) instead of the raw event so error reporting can classify it.

**Workaround**

Host apps can swallow `unhandledrejection`s whose `reason instanceof Event`, but a fix in the client would help everyone.

drew-harris:
fixes: #2795 
If indexeddb was unavailable for whatever reason, the scheduled 60s garbage collection would throw an uncaught error. This catches the error and logs it instead.

### DIFF
--- a/client/packages/core/src/utils/PersistedObject.ts
+++ b/client/packages/core/src/utils/PersistedObject.ts
@@ -485,7 +485,9 @@ export class PersistedObject<K extends string, T, SerializedT> {
       () => {
         safeIdleCallback(() => {
           this._nextGc = null;
-          this._gc();
+          this._gc().catch((e) =>
+            this._log.error('Persisted Object GC Failed:', e),
+          );
         }, 30 * 1000);
       },
       // 1 minute + some jitter to keep multiple tabs from running at same time
--- a/client/packages/version/src/version.ts
+++ b/client/packages/version/src/version.ts
@@ -2,6 +2,6 @@
 // Update the version here and merge your code to main to
 // publish a new version of all of the packages to npm.
 
-const version = 'v1.0.52';
+const version = 'v1.0.53';
 
 export { version };


dwwoelfel (approved):
LGTM!

PR #2800, opened 2026-07-29
Include app_id in replica identity

dwwoelfel:
Updates the replica identity for a few tables to include the app_id so that we can filter out a single app when migrating them to a new database.


stopachka (approved):
SGTM!

PR #2803, opened 2026-07-30
[CRA] Configure create-instant-app for self-hosted backends

nezaj:
This PR does two things

1) Make create-instant-app less of a schlep for self-hosted apps
2) Centralize our logic for project bases -- there was a bit of duplication and making this template update work would have required copying things around again so figured was a good time to centralize

**Self-hosting**

Previously, `INSTANT_CLI_API_URI` pointed `create-instant-app` at a self-hosted backend, but generated apps still required manually adding `apiURI` and `websocketURI` to `init`.

Generated apps now include the self-hosted API and derived WebSocket configuration automatically.

**Project bases**

Consolidated duplicated project base declarations into one registry. CLI choices, environment variables, bundled templates, and backend configuration files now derive from the same source.

### DIFF
--- a/client/packages/create-instant-app/scripts/copyExamples.ts
+++ b/client/packages/create-instant-app/scripts/copyExamples.ts
@@ -2,17 +2,7 @@ import path from 'path';
 import { fileURLToPath } from 'url';
 import fs from 'fs-extra';
 import { copyRespectingGitignore } from '../src/scaffold.js';
-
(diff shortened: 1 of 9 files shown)


drew-harris (approved):
Tested and working!
https://gist.github.com/drew-harris/16f6ba91dc8fc26a37f412d4b512fb35

PR #2804, opened 2026-07-30
Keep dashboard routes out of app proxy

stopachka:
## Summary

- Keep `/dash` and `/dash/*` requests on the Instant backend when an app proxy target is configured
- Apply the same dashboard boundary to the Undertow router and the Ring fail-closed guard
- Cover app IDs from both dashboard paths and headers, including the `/dashboard` segment-boundary case

## Test plan

- `clojure.test/run-tests instant.app-proxy-test` (6 tests, 39 assertions)
- `clj-kondo --lint src/instant/app_proxy.clj test/instant/app_proxy_test.clj`
- `git diff --check`

### DIFF
--- a/server/src/instant/app_proxy.clj
+++ b/server/src/instant/app_proxy.clj
@@ -218,6 +218,10 @@
 ;; ----------
 ;; Request routing
 
+(defn- dashboard-path? [path]
+  (or (= "/dash" path)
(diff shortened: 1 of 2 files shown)


dwwoelfel (approved):
LGTM!

PR #2805, opened 2026-07-30
Add self-hosted deployment settings

stopachka:
We had two problems: 

1. Self-hosted deployments don't bootstrap the instant-config app. This app can be really useful though.
2. One case where it's useful: it would be _awesome_ if we could restrict signups in self-hosted apps

So this PR does the following:

1. We now accept an `INSTANT_SUPERUSER_EMAIL` env var
2. In bootstrap, we now _always_ create the instant-config app 
3. If a superuser email is provided, we make sure that instant-config app's creator _is_ the superuser 
4. We added a "Deployment Settings" page that shows up for the superuser
5. There the user can now _set_ configuration for restricting signups!

Curated Deployment Settings


self-hosting/deploy-swarm.sh
```diff
@@ -11,6 +11,7 @@ POSTGRES_PASSWORD=changeme
 MINIO_ROOT_PASSWORD=changeme
 INSTANT_DASHBOARD_GOOGLE_OAUTH_CLIENT_ID=
 INSTANT_DASHBOARD_GOOGLE_OAUTH_CLIENT_SECRET=
+INSTANT_BOOTSTRAP_SUPERUSER_EMAIL=admin@example.com
```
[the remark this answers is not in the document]
stopachka: Fixed?

client/www/app/docs/self-hosting/page.md
```diff
@@ -108,6 +108,26 @@ The `_DOMAIN` variables are only used by the Caddy reverse proxy, so if you are
...
 
+### Configure the Deployment Superuser
+
+The deployment superuser can manage settings for your entire self-hosted deployment. Set `INSTANT_BOOTSTRAP_SUPERUSER_EMAIL` to the email address of the person who should administer it:
+
+```shell
+INSTANT_BOOTSTRAP_SUPERUSER_EMAIL=admin@example.com
+```
+
+Instant creates this dashboard user on the first startup. Log in with this email and open **Deployment Settings** from the account menu to manage your deployment.
```
dwwoelfel: Will it try to create the user if the user already exists?

client/www/components/dash/ProfilePanel.tsx
```diff
@@ -97,6 +97,16 @@ export const ProfilePanel = () => {
                   <TooltipContent side="right">User Settings</TooltipContent>
                 </Tooltip>
               </div>
+              {dashResponse.data.superuser && (
+                <Link
+                  href="/dash/deployment-settings"
```
dwwoelfel: Maybe we should put this in `/intern`?

server/src/instant/rate_limit.clj
```diff
@@ -160,7 +160,9 @@
                        (tracer/with-span! {:name "rate-limit/change-magic-code-rate-limit-per-hour"
                                            :attributes {:old-value old-value
                                                         :new-value new-value}}
-                         (let [new-config (create-rate-limit-config new-value)]
+                         (let [new-config
+                               (create-rate-limit-config
+                                (flags/magic-code-rate-limit-per-hour))]
```
dwwoelfel: Why this change? It should work with `new-value`.

client/www/app/docs/self-hosting/page.md
stopachka: [From Codex] No. Bootstrap first looks up the dashboard user by email and reuses it if it exists. It only creates the user when that lookup returns nothing, then creates Instant Config under that user.

server/src/instant/rate_limit.clj
stopachka: You’re right. The listener already receives the updated flag value, so the extra lookup was unnecessary. I switched this back to `new-value` in 0d1492a8a.

server/src/tasks.clj
```diff
@@ -82,23 +97,30 @@
...
   (let [path (or (System/getenv "OVERRIDE_CONFIG_PATH")
-                 "resources/config/override.edn")]
+                 override-config-path)]
     (io/make-parents path)
     (crypt-util/register-aead)
     (crypt-util/register-signature)
     (println "Writing config file to" path)
     (spit path
           (pr-str
-           {:aead-keyset {:encrypted? false
```
dwwoelfel, line 92: If they don't set the `INSTANT_BOOTSTRAP_SUPERUSER_EMAIL`, we won't create the config app for them, but we'll still put the config-app-id in the `override.edn`. I think that will cause errors on startup (or allow someone else to create the config app and mess with the config).

client/www/pages/intern/deployment-settings.tsx
```diff
@@ -0,0 +1,892 @@
...
+
+  if (!configApp) {
+    return (
+      <p className="text-sm text-red-600">
+        Instant Config is not available for this account.
+      </p>
+    );
+  }
+
+  return <InstantConfigSettings key={configApp.id} app={configApp} />;
```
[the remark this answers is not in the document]
stopachka: This is not a real issue. That would just cause an error

client/www/components/dash/ProfilePanel.tsx
stopachka: Done. The page now lives at `/intern/deployment-settings`, outside the dashboard shell, while the account-menu link remains in the dashboard.

server/src/tasks.clj
stopachka: Good catch. I changed the design so this state cannot occur: the Instant Config app ID is now a fixed constant in `config.clj`, it is no longer written to `override.edn`, and OSS bootstrap always creates the app. Without `INSTANT_SUPERUSER_EMAIL` it belongs to the seeded test user; when that variable is set or changed, bootstrap creates the user if needed and transfers the existing app. I verified fresh boot, two ownership transfers, and an idempotent restart in Docker.

server/src/tasks.clj
```diff
@@ -111,9 +124,91 @@
...
+    (when (seq steps)
+      (tx/transact-without-tx-conn!
+       conn
+       current-attrs
+       app-id
+       steps
+       {:skip-app-status-write-check? true}))
+    (when-not (rule-model/get-by-app-id conn {:app-id app-id})
+      (rule-model/put! conn {:app-id app-id
+                             :code config-app/rules}))))
```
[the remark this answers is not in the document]
stopachka: Reviewed against the final bootstrap flow and intentionally not applied. The config app is created once with deny-all rules, then operators may manage it through the admin-token Explorer. Rewriting rules on every server start would overwrite operator-managed rules. Bootstrap now only creates the missing app or reconciles its owner, which matches the intended self-hosted contract. The Swarm E2E verified that a fresh app receives the deny-all rules.

server/src/instant/model/webhook.clj
```diff
@@ -335,11 +335,12 @@
...
-                    attrs
-                    config-app-id
-                    [[:add-triple [setting-aid "enable-wal-entity-log-apps-map"] id-aid [setting-aid "enable-wal-entity-log-apps-map"]]
-                     [:deep-merge-triple [setting-aid "enable-wal-entity-log-apps-map"] value-aid {(str app-id) true}]]))))
+      (when (every? some? [id-aid setting-aid value-aid])
+        (tx/transact! (aurora/conn-pool :write)
+                      attrs
+                      config-app-id
+                      [[:add-triple [setting-aid "enable-wal-entity-log-apps-map"] id-aid [setting-aid "enable-wal-entity-log-apps-map"]]
+                       [:deep-merge-triple [setting-aid "enable-wal-entity-log-apps-map"] value-aid {(str app-id) true}]])))))
```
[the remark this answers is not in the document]
stopachka: Reconsidered after review: no extra guard is needed here. `enable-webhooks-config-flag` unconditionally calls `tx/transact!` with the derived attribute IDs, so a missing ID makes that transaction fail. `create!` invokes this before its webhook insert transaction, which already propagates the failure and prevents persistence. I removed the redundant explicit check in b760bc1f6. The focused webhook suite passes with 18 tests and 204 assertions.

client/www/pages/dash/deployment-settings.tsx
```diff
@@ -0,0 +1,765 @@
...
+    setMagicCodeRateLimitInput(String(magicCodeRateLimit));
+  }, [magicCodeRateLimit]);
+
+  useEffect(() => {
+    setMagicCodeExpiryInput(String(magicCodeExpiry));
+  }, [magicCodeExpiry]);
+
+  useEffect(() => {
+    setRefreshThrottleInput(String(refreshThrottleMs));
+  }, [refreshThrottleMs]);
```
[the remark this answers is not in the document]
stopachka: Fixed in 8cc06bed5 after the page moved to `/intern/deployment-settings`. The authentication and refresh-throttle inputs now track whether the local form is being edited, and live-query synchronization only updates inputs while they are clean. Saving clears the edit state so later remote updates resume syncing. Type checking and the production dashboard build pass.

client/www/app/docs/self-hosting/page.md
```diff
@@ -108,6 +108,26 @@ The `_DOMAIN` variables are only used by the Caddy reverse proxy, so if you are
...
 
+### Configure the Deployment Superuser
+
+The deployment superuser can manage settings for your entire self-hosted deployment. Set `INSTANT_SUPERUSER_EMAIL` to the email address of the person who should administer it:
+
+```shell
+INSTANT_SUPERUSER_EMAIL=admin@example.com
+```
+
+Instant creates this dashboard user when the server starts. Log in with this email and open **Deployment Settings** from the account menu to manage your deployment. If you change the email, restart the server to transfer superuser access.
```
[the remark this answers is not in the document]
stopachka: This is stale against the final contract. The variable was intentionally renamed to `INSTANT_SUPERUSER_EMAIL` because it is no longer bootstrap-only: the config app is always created, and every server start reconciles its owner to the configured superuser. The Docker Swarm E2E changed this value during a rolling update and verified ownership moved, then moved it back successfully. The documentation is correct as written.

server/src/instant/model/webhook.clj
```diff
@@ -330,16 +330,16 @@
...
+  (let [config-app-id config/instant-config-app-id
+        attrs (attr-model/get-by-app-id config-app-id)
+        id-aid (:id (attr-model/seek-by-fwd-ident-name ["flags" "id"] attrs))
+        setting-aid (:id (attr-model/seek-by-fwd-ident-name ["flags" "setting"] attrs))
+        value-aid (:id (attr-model/seek-by-fwd-ident-name ["flags" "value"] attrs))]
+    (tx/transact! (aurora/conn-pool :write)
+                  attrs
+                  config-app-id
+                  [[:add-triple [setting-aid "enable-wal-entity-log-apps-map"] id-aid [setting-aid "enable-wal-entity-log-apps-map"]]
+                   [:deep-merge-triple [setting-aid "enable-wal-entity-log-apps-map"] value-aid {(str app-id) true}]])))
```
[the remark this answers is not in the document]
stopachka: Reconsidered after review: the existing ordering already enforces this invariant without an extra guard. `enable-webhooks-config-flag` unconditionally calls `tx/transact!`, so missing or invalid attribute IDs fail that transaction. `create!` invokes it before the webhook insert transaction, so no webhook is persisted on failure. I removed the redundant explicit check in b760bc1f6. The focused webhook suite passes with 18 tests and 204 assertions.

dwwoelfel (approved):
LGTM!

PR #2806, opened 2026-07-30
log url on instant-cli errors (#2806) [bump to v1.0.55]

drew-harris:
Shows the url on errors for the info command. For any InstantHttpError, the url is already part of the error message. 

image

### DIFF
--- a/client/packages/cli/src/commands/info.ts
+++ b/client/packages/cli/src/commands/info.ts
@@ -1,6 +1,6 @@
 import { HttpClientResponse } from '@effect/platform';
 import { Effect, Schema, Option } from 'effect';
-import { InstantHttpAuthed } from '../lib/http.ts';
+import { getBaseUrl, InstantHttpAuthed } from '../lib/http.ts';
 import { version } from '@instantdb/version';
 import { CurrentApp } from '../context/currentApp.ts';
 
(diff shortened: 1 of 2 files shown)


client/packages/cli/src/commands/info.ts
```diff
@@ -32,7 +33,10 @@ export const infoCommand = () =>
       const meData = yield* authedHttp.get('/dash/me').pipe(
         Effect.flatMap(HttpClientResponse.schemaBodyJson(DashMeResponse)),
         Effect.mapError(
-          (e) => new Error("Couldn't get user information.", { cause: e }),
+          (e) =>
+            new Error(`Couldn't get user information (${baseUrl}).`, {
+              cause: e,
+            }),
```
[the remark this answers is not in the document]
drew-harris: the later /dash/apps request can fail but ti will fail as an InstantHttpError which will print the url.

stopachka (approved):
SGTM!

client/packages/cli/src/commands/info.ts
```diff
@@ -32,7 +33,10 @@ export const infoCommand = () =>
       const meData = yield* authedHttp.get('/dash/me').pipe(
         Effect.flatMap(HttpClientResponse.schemaBodyJson(DashMeResponse)),
         Effect.mapError(
-          (e) => new Error("Couldn't get user information.", { cause: e }),
+          (e) =>
+            new Error(`Couldn't get user information (${baseUrl}).`, {
```
nezaj, line 37: Curious can you include a screenshot of how this looks?

client/packages/version/src/version.ts
nezaj:
```suggestion
const version = 'v1.0.55';
```

PR #2807, opened 2026-07-30
[Dash] Use configured default sender email

nezaj:
On the auth tab we would hardcode the default sender email as `verify@auth-pm.instantdb.com` but for self-hosted this wouldn't make sense.

We'll now actually include the configured default sender email to the dashboard!

**Before**
CleanShot 2026-07-30 at 13 39 00@2x

**After**
CleanShot 2026-07-30 at 13 39 29@2x

### DIFF
--- a/client/www/components/dash/AppAuth.tsx
+++ b/client/www/components/dash/AppAuth.tsx
@@ -887,7 +887,9 @@ export function AppAuth({
         <Email
           key={emailFormKey}
           app={app}
-          defaultSenderEmail={data.default_sender_email}
+          defaultSenderEmail={
+            data.default_sender_email ?? 'verify@auth-pm.instantdb.com'
+          }
         />
       </AuthDetailLayout>
     );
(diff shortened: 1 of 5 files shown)


client/www/components/dash/AppAuth.tsx
```diff
@@ -884,7 +884,11 @@ export function AppAuth({
         title="Magic code email"
         description="The email your users receive with their sign-in code."
       >
-        <Email key={emailFormKey} app={app} />
+        <Email
+          key={emailFormKey}
+          app={app}
+          defaultSenderEmail={data.default_sender_email}
```
dwwoelfel: You might want to do `?? verify@auth-pm.instantdb.com` just until this is fully deployed.

dwwoelfel (approved):
One suggestion, but LGTM!

client/www/components/dash/AppAuth.tsx
nezaj: Done!

PR #2808, opened 2026-07-30
separate cli auth keys by host

drew-harris:
Transitions the api key storage for the instant-cli to use JSON with a map between api hosts and the key itself. 
People who are upgrading will have their previous key tested against their current apiURI, then production, on first initial read. 

Also updated create-instant-app to work with the new format. Uses a new export in package.json from cli for auth specific code that cia can borrow

Logging out now only removes the corresponding value from the map, rather than deleting the entire file.


nezaj:
Talked about this with the LLM and it suggested instead of having one file for all tokens we would be better of having a separate file per backend.

The benefit here is it would be fully backward compatible (if someone upgrades instant-cli and now the file type changes to JSON, if they go to a project using an older instant-cli that will break since our auth file changed).

Put up a PR for that here https://github.com/instantdb/instant/pull/2813

PR #2816, opened 2026-08-03
Serialize concurrent webhook config updates

stopachka:
**Problem**

add-triple and deep-merge-triple can race and cause deadlocks. This breaks our webhook config flag workflow in tests, since we concurrently try to update this setting. 

As a workaround for now, I added a pg advisory lock for the webhook transaction. 

@dwwoelfel @nezaj


dwwoelfel (approved):
LGTM!

PR #2821, opened 2026-08-03
[Self-host] Fail dashboard on missing configuration

nezaj:
Previously, if a self-hosted dashboard was missing INSTANT_API_URI or INSTANT_BACKEND_URL, it silently fell back to api.instantdb.com.

This change requires INSTANT_API_URI to be set and fails with a clear configuration error instead of connecting to Instant Cloud.


stopachka (approved):
SGTM!

PR #2822, opened 2026-08-03
Use configured backend for OAuth callbacks

nezaj:
Previously, OAuth setup in the dashboard and CLI _always_ used Instant Cloud’s callback URL.

This change derives the callback from the configured API URL so self-hosted deployments use their own backend!


stopachka (approved):
SGTM!

PR #2823, opened 2026-08-03
[Self-host] Add docs for updating an existing deployment

nezaj:
What it says on the tin!

CleanShot 2026-08-03 at 14 23 22@2x


stopachka (approved):
SGTM!

PR #2825, opened 2026-08-03
Enable hints by default

dwwoelfel:
Enables hints by default. This will help to resolve reports of slower queries on self-hosted instances.

This matches the setting we use in production.


nezaj (approved):
LGTM!

stopachka (approved):
Nice!

PR #2826, opened 2026-08-03
[Docs] Restrict dashboard signups

nezaj:
Promoting this section to it's own h3 since the h4 made it easy to miss.

**Before**
CleanShot 2026-08-03 at 15 11 10@2x

**After**
CleanShot 2026-08-03 at 15 12 18@2x


stopachka (approved):
SGTM!

PR #2827, opened 2026-08-03
[Self-host] Gate ephemeral app creation

nezaj:
Adds a flag to the config app to enable/disable un-authenticated ephemeral app creation. By default this is set to true (matching our open signups pattern).

**Deployment settings**
CleanShot 2026-08-03 at 15 56 18@2x

**Docs**
CleanShot 2026-08-03 at 15 56 45@2x

### DIFF
--- a/server/src/instant/flags.clj
+++ b/server/src/instant/flags.clj
@@ -429,6 +429,9 @@
          (get key not-found))
      (get-in (query-result) [:flags key] not-found))))
 
+(defn ephemeral-apps-enabled? []
+  (not (false? (flag :ephemeral-apps-enabled true))))
+
 (defn app-proxy-targets
   "The app proxy routing table. Emptying the flag turns off all routing
    without a deploy."
--- a/server/src/instant/flags.clj
+++ b/server/src/instant/flags.clj
@@ -430,7 +430,7 @@
      (get-in (query-result) [:flags key] not-found))))
 
 (defn ephemeral-apps-enabled? []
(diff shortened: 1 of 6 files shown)


stopachka (approved):
SGTM!

server/src/instant/flags.clj
```diff
@@ -429,6 +429,9 @@
          (get key not-found))
      (get-in (query-result) [:flags key] not-found))))
 
+(defn ephemeral-apps-enabled? []
+  (not (false? (flag :ephemeral-apps-enabled true))))
```
dwwoelfel: Does it really need `not false?` here?

server/test/instant/dash/ephemeral_app_test.clj
```diff
@@ -0,0 +1,15 @@
+(ns instant.dash.ephemeral-app-test
+  (:require
+   [clojure.test :refer [deftest is]]
+   [instant.dash.ephemeral-app :as ephemeral-app]
+   [instant.flags :as flags]
+   [instant.util.exception :as ex]))
+
+(deftest create-is-denied-when-ephemeral-apps-are-disabled
+  (with-redefs [flags/ephemeral-apps-enabled? (constantly false)]
```
dwwoelfel: There's a `flags/*toggle-overrides` that might make a nicer way to test this.

dwwoelfel (approved):
Couple of comments, but LGTM!

PR #2828, opened 2026-08-03
[Storage] Schedule files for deletion

nezaj:
Deprecated storage delete paths immediately removed S3 objects, bypassing the 32-day sweeper grace period that keeps files available while backups are valid.

This change leaves S3 cleanup to the existing storage sweeper. Deleted `$files` records are still removed immediately and their objects are deleted after the grace period.


dwwoelfel (approved):
LGTM!

PR #2832, opened 2026-08-04
[Self-host] Preserve path for custom s3 public endpoints

nezaj:
Got a user report that `$files.url` was dropping path prefixes from S3_PUBLIC_ENDPOINT.

Generated URLs now preserve the configured path without changing existing signing behavior!


dwwoelfel (approved):
LGTM!

PR #2835, opened 2026-08-04
[Self-host] Revert path for custom s3 endpoints

nezaj:
Looks like hosting MinIO at a subpath [will not work with S3](https://github.com/minio/docs/blob/main/source/integrations/setup-nginx-proxy-with-minio.rst?plain=1#L117))

This reverts the earlier change in https://github.com/instantdb/instant/pull/2832 and adds a comment to let folks know they can't use a subpath


stopachka (approved):
Nice!

PR #2838, opened 2026-08-05
Backups through the CLI

stopachka:
Adds `instant-cli backup list` and `instant-cli backup download`, built on a new `BackupsManager` in `@instantdb/platform` that the dashboard's download dialog now consumes too.

list: 

CleanShot 2026-08-05 at 14 54 12@2x

download: 

CleanShot 2026-08-05 at 14 54 41@2x

**Platform (`@instantdb/platform`)**

- `api.backups(appId)` returns a `BackupsManager` with `list`, `listFiles`, `getFileUrl`, `streamStorageFiles` (an async generator over the NDJSON storage-files stream that honors the `{done: true}` sentinel and throws on truncation), and `downloadArchive`.
- `downloadArchive` is the whole download pipeline, shared by the dashboard and the CLI: canonical entry order, eager draining of the storage-files discovery so the NDJSON connection isn't held open while multi-GB blobs download, progress reporting, backpressure into the caller's sink, and teardown. 
  - The runtime-specific pieces are injected: `fetchBody` (fetch and decompress a presigned URL), `sink` (where the bytes go), and `createWriter` (a small `BackupArchiveWriter` interface that zip.js's `ZipWriter` satisfies structurally, so the platform package doesn't depend on a zip implementation).
- The manager's docs are the canonical statement of the archive ordering contract: `config.json` first, then every `entities/*.jsonl` shard, then `files/<locationId>`, with all entity files before any storage file so restore can process the archive in one streaming pass. Since the pipeline lives behind `downloadArchive`, that order is enforced in one place instead of once per consumer.
- Also shared: `estimateZipSize` (the min/max zip size estimate both UIs show), `formatFileSize` (decimal/SI byte formatting that matches how Finder reports file sizes, so both surfaces print identical numbers), and `toAppBackup` (parses a server backup row).
- The dashboard's `BackupDownloadDialog` swaps its hand-rolled pipeline for `downloadArchive` (a large net deletion); the browser-specific pieces (save picker, service worker, browser fetch, React progress UI) stay in www.

**Dashboard fix**

- A 404 from the storage-files endpoint was treated as "no storage files" and produced a complete-looking zip. The server signals the empty case with a 200 + `{done: true}` sentinel (it sends the sentinel even when there's no `$files` shard), so a 404 can only mean the backup record is missing or expired (e.g. it expired mid-download). The download now fails instead of silently omitting every storage file; this behavior lives in the shared `BackupsManager`.

**Testing**

- 28 new tests: NDJSON sentinel/truncation semantics and pipeline-level tests (cancellation race, counter timing, entry order, locationId validation) in the platform package, CLI command tests mocked at the SDK boundary, and an integration test that runs the real shared pipeline against a local HTTP server serving zstd-encoded bodies, then reads the zip back and asserts entry order and contents.
- Real-world run against production with a ~4.9 GB backup: 65 namespaces -> 1.54 GB zip in ~3 minutes, memory flat at ~133 MB the whole run, the sum of uncompressed entry sizes exactly equal to the recorded `uncompressed_size`, and a clean `unzip -t`.

@dwwoelfel @nezaj


dwwoelfel (approved):
LGTM!

PR #2840, opened 2026-08-05
Copy s3 files with storage user

dwwoelfel:
Follow-up to https://github.com/instantdb/instant/pull/2814

Updates `copy` to use the storage user. We override the creds for each copy command.

This makes it easier for the destination bucket to accept requests from our user.


stopachka (approved):
SGTM!

Issue #2841, opened 2026-08-06
Storage file uploads fail behind Caddy ≥ 2.11.4 — SDK sends app_id as an underscore header that modern proxies drop


saimon24:
### Summary

`storage.uploadFile` sends the app ID as an HTTP request header named `app_id`
(underscore) on `PUT /storage/upload`. Reverse proxies increasingly drop request
headers whose names contain underscores by default, so on a self-hosted stack the
header never reaches the server and every upload fails with:

    [InstantAPIError: Missing parameter: ["app_id"]]

This makes file/avatar uploads the one subsystem that breaks out of the box on the
official self-hosting compose, while auth, queries, transactions and realtime all
work — so it's easy to ship a self-hosted instance and only discover it when a user
tries to set a profile photo.

### Root cause

- Client: `@instantdb/core` `StorageAPI.uploadFile` sets `headers = { app_id, path,
  authorization, 'content-type' }`. `app_id` is the only header name with an underscore.
- Server: `server/src/instant/storage/routes.clj` `upload-put` reads params from
  `(:headers req)` and validates `app_id` first, so a missing header throws
  `param-missing` for `app_id`.
- Proxy: the official `self-hosting/docker-compose.with-caddy.yml` pins the floating
  tag `caddy:2-alpine`. **Caddy v2.11.4** (commit `3eb8e48`) silently drops all
  request headers containing underscores as an httpoxy/CGI hardening, with no opt-out
  and only a DEBUG-level log (caddyserver/caddy#7808). This is standard behavior for
  nginx (`underscores_in_headers off` by default), AWS ALB, and Apache too.

### Reproduction (official self-hosting compose)

Through Caddy — fails:

    curl -s -X PUT "https://api-stg.<DOMAIN>/storage/upload" \
      -H "app_id: 11111111-1111-1111-1111-111111111111" \
      -H "path: probe/test" -H "authorization: Bearer bogus" --data-binary "hi"
    # {"type":"param-missing","message":"Missing parameter: [\"app_id\"]"}

Straight to the backend on the compose network (bypassing Caddy) — header arrives,
gets past the param check to the permission check:

    docker run --rm --network self-hosting_default curlimages/curl -s -X PUT \
      http://server:8888/storage/upload \
      -H "app_id: 11111111-1111-1111-1111-111111111111" \
      -H "path: probe/test" -H "authorization: Bearer bogus" --data-binary "hi"
[...]

PR #2842, opened 2026-08-06
Fix storage uploads behind proxies that drop underscore headers

stopachka:
**Context**

Most admin endpoints use the `app-id` header, but storage uploads used `app_id`.

**Problem**

Caddy drops headers with underscores: #2841

**Solution**

- send both `app-id` and `app_id` from `StorageAPI.uploadFile`
- send both headers from the dashboard Explorer upload
- accept either header on the storage backend
- bump the shared JS package version to `v1.0.62`


client/packages/core/src/StorageAPI.ts
```diff
@@ -30,7 +30,7 @@ export async function uploadFile({
   contentDisposition?: string;
 }): Promise<UploadFileResponse> {
   const headers = {
-    app_id: appId,
```
dwwoelfel, line 33: Could we set both headers? Then we don't have to worry about sequencing of the backend deploy and frontend deploy.

stopachka: Updated. Both upload paths now send `app-id` and `app_id`, and this PR bumps the shared JS version to `v1.0.62`.

dwwoelfel (approved):
LGTM!

PR #2844, opened 2026-08-07
Add instant.config.ts for self-hosted create-instant-app

nezaj:
I noticed when using `create-instant-app` in a self-hosted setup the LLM would first try to use `instant-cli` without specifying `INSTANT_CLI_API_URI` which cause an error since the app doesn't exist in production. The LLM would be able to inspect the app and figure out it needs to add this env variable but I thought it would be nicer to remove that schlep.

This PR makes it so when you use `create-instant-app` with a custom `INSTANT_CLI_API_URI` we'll create an `instant.config.ts` file that our CLI will pick up and avoid the error from before.

As a bonus, in the self-hosted dashboard we'll also update the `create-instant-app` command to include both self hosted api and dash uris

CleanShot 2026-08-06 at 18 24 20@2x

### DIFF
--- a/client/packages/create-instant-app/src/backendConfig.test.ts
+++ b/client/packages/create-instant-app/src/backendConfig.test.ts
@@ -67,6 +67,33 @@ describe('applyBackendConfig', () => {
         '  appId: "app-id",\n' +
         '});\n',
     );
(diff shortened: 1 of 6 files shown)


stopachka (approved):
SGTM!

PR #2845, opened 2026-08-07
Custodian: a sweeper to delete apps in smaller chunks

dwwoelfel:
Creates a new `custodian` process that allows us to delete large apps. Previously, we had to set a very long timeout to delete large apps. That put pressure on the invalidator and would fail for very large apps.

This PR will delete the triples and transactions in batches first, then finally delete the app.

Each machine runs one deletion process that polls for new work in the background. We only have one because my main concern is preventing too much load on the database. If the database can handle it and the replication lag isn't affected, we could increase the number of processes.

Each process is essentially a linked list in the database. It looks like this:

Delete app:
`Delete app` `->` `Delete attrs` `->` `Delete triples` `->` `Delete transactions`

Delete attr
`Delete attr` `->` `Delete triples`

The worker grabs the tail of the list, performs the work, then deletes the item so the next worker can pick up the new tail of the list.

`Delete transactions` and `Delete triples` both run a long-running process that deletes the rows 1000 at a time (configurable with a flag). Each time it deletes, it 1. checks to make sure the app/attr has `deletion_marked_at` and 2. updates a heartbeat on the job. There's a separate cron process that frees stuck jobs.

In the background, we poll the replication slots to make sure that the lag doesn't get out of hand. If it does, we automatically pause the workers until replication catches up. We can raise or lower the lag threshold with a flag.

If a job fails, we'll try it two more times, then mark it failed. Then we'll try it again after 24 hours.


server/src/instant/custodian.clj
```diff
@@ -0,0 +1,352 @@
...
+             from pg_replication_slots
+            where active and confirmed_flush_lsn is not null"])))
+
+(defn sample-lag! [backing-off?]
+  (try
+    (let [threshold (flags/flag :custodian-max-replication-lag-bytes (* 100 1024 1024))
+          lag (or (max-replication-lag-bytes (aurora/conn-pool :read)) 0)]
+      (reset! backing-off? (> lag threshold)))
+    (catch Throwable t
+      (tracer/record-exception-span! t {:name "custodian/sample-lag-error"}))))
```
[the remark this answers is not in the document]
dwwoelfel: That is the primary

server/src/instant/custodian.clj
```diff
@@ -0,0 +1,352 @@
...
+    (catch Throwable t
+      (tracer/record-exception-span! t {:name "custodian/reap-error"}))))
+
+(defn start-worker []
+  (let [stop? (atom false)
+        backing-off? (atom false)
+        fut (ua/vfuture (run-worker stop? backing-off?))
+        reaper (chime-core/chime-at
+                (chime-core/periodic-seq (Instant/now) (Duration/ofHours 1))
+                reap-tick)
```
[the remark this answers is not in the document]
dwwoelfel: That's okay--we're fine with a delay here. Deleting isn't an operation that has to happen immediately.

server/src/instant/hard_deletion_sweeper.clj
```diff
@@ -74,14 +74,10 @@
...
-             (format "delete-attr-%s-%s" id (date-util/numeric-date-str maximum-marked-date))
-             (fn [] (straight-jacket-delete-attr! attr))))
-          (doseq [{:keys [id] :as app} apps-to-delete]
-            (grab/run-once!
-             (format "delete-app-%s-%s" id (date-util/numeric-date-str maximum-marked-date))
-             (fn [] (straight-jacket-delete-app! app)))))))))
+          (doseq [{:keys [id app_id]} attrs-to-delete]
+            (custodian/enqueue-attr-deletion! {:app-id app_id :attr-id id}))
+          (doseq [{:keys [id]} apps-to-delete]
+            (custodian/enqueue-app-deletion! {:app-id id})))))))
```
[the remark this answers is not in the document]
dwwoelfel: If this fails, I'd rather it bubble up here rather than suppressing the error.

server/src/instant/core.clj
```diff
@@ -359,6 +360,9 @@
       (future
         (tracer/with-span! {:name "stop-join-room-logger"}
           (join-room-logger/stop)))
+      (future
+        (tracer/with-span! {:name "stop-custodian"}
+          (custodian/stop)))
```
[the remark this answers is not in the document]
dwwoelfel: Will fix

server/src/instant/reactive/aggregator.clj
```diff
@@ -151,7 +151,7 @@
                    (config/get-aurora-config)
                    slot-name))]
       (with-open [connection ^PgConnection connection]
-        (doseq [sketches (partition-all 1000 (initial-sketch-seq connection copy-sql))]
+        (doseq [sketches (partition-all 100 (initial-sketch-seq connection copy-sql))]
```
dwwoelfel, line 154: This is just to reduce memory usage when creating the aggregator from scratch.

server/src/instant/reactive/invalidator.clj
```diff
@@ -120,7 +120,8 @@
 
         some-changes (or (seq idents)
                          (seq triples)
-                         (seq attrs))
+                         (seq attrs)
+                         (seq transactions))
```
dwwoelfel, line 124: This allows the invalidator slot to make progress if we import transactions with no triples, e.g. with replication.

server/src/instant/custodian.clj
```diff
@@ -0,0 +1,544 @@
...
+                      "triples" (if attr_id
+                                  (drain! ::drain-attr-triples stop? backing-off? conn id delete-attr-triples-q {:app-id app_id :attr-id attr_id})
+                                  (drain! ::drain-app-triples stop? backing-off? conn id delete-app-triples-q {:app-id app_id}))
+                      "transactions" (drain! ::drain-app-transactions stop? backing-off? conn id delete-app-transactions-q {:app-id app_id})
+                      ;; Delete the app's attrs one at a time so each attr's triples
+                      ;; cascade is its own bounded, committed statement.
+                      "attrs" (drain! ::drain-app-attrs stop? backing-off? conn id delete-app-attrs-q {:app-id app_id} (attr-batch-size))
+                      ;; Terminal steps run to completion; they aren't interrupted mid-way.
+                      "attr" (do (attr-model/hard-delete-multi! conn app_id #{attr_id}) ::completed)
+                      "app" (do (app-model/delete-immediately-by-id! conn {:id app_id}) ::completed))]
```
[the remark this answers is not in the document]
dwwoelfel: Fixed

stopachka (approved):
Suuper cool!

server/src/instant/custodian.clj
```diff
@@ -0,0 +1,557 @@
...
+   we don't hand the invalidator one giant WAL transaction.
+
+   The plan for a deletion is stored as a chain of `custodian` rows linked by
+   `depends_on`:
+
+     app:  triples <- transactions <- attrs <- app
+     attr: triples <- attr
+
+   `depends_on` points at the step that must finish first, so a row is runnable
+   once the row it depends on is `done_at`. Each worker claims one runnable row
```
nezaj: Nit. I'm guessing `done_at` existed before, but it no longer does?

server/resources/migrations/124_add_custodian.up.sql
```diff
@@ -0,0 +1,52 @@
+create table custodian (
+  id uuid primary key default gen_random_uuid(),
+  -- The app being deleted. Cascade so the terminal `app` delete cleans up its
+  -- own plan rows for free.
+  app_id uuid not null references apps(id) on delete cascade,
+  -- Set when a unit of work is scoped to a single attr (e.g. deleting one
+  -- attr's triples). Null means the whole app.
+  attr_id uuid references attrs(id) on delete cascade,
+  -- What this row deletes: 'triples' | 'transactions' | 'attr' | 'app'
```
nezaj: Nit, also `attrs` is a valid type?

server/src/instant/custodian.clj
dwwoelfel: Updated docstring

server/resources/migrations/124_add_custodian.up.sql
dwwoelfel: Updated

server/src/instant/hard_deletion_sweeper.clj
nezaj, line 34: Also can delete this

nezaj: Nit, I think we can delete `straight-jacket-delete-app!` and  `straight-jacket-delete-attr!` now

dwwoelfel: Deleted!

dwwoelfel: Deleted!

PR #2846, opened 2026-08-07
Display backups on dashboard

dwwoelfel:
Enables the backups page on the dashboard for everyone.

I confirmed in production that we could both create backups on-demand and download backups.


nezaj (approved):
Woohoo!

PR #2851, opened 2026-08-08
Filter intern tools for self-hosted deployments

dwwoelfel:
Filters the internal tools on `/intern` for self-hosted deployments to just the `Restore App`, `Storage Usage`, and `Deployment Settings` cards.


stopachka (approved):
Awesome!

stopachka (approved):
Awesome!

PR #2853, opened 2026-08-10
Experiment 1: Explicitly commit IndexedDB write transactions

stopachka:
# Problem 

If you background a tab in android, sometimes it causes active tabs to freeze

# Reason 

We open a transaction on IndexedDB when we save pending mutations. If Chrome decides to _freeze_ while the tx is open, all other transactions to IndexedDB will be blocked, waiting for that transaction 

# Solution 

Unfortunately there is no easy way to _tell_ IndexedDB to time out when we are sending a tx. 

But one thing we can do is to commit the transaction faster: 

We call `IDBTransaction.commit()` after `setItem`, `multiSet`, and `removeItem` right away. 

This lets IDB commit the transaction, before having to send back a "done" result to our callback. 

When running tests, this reduced the number of noticed blocks from 4/6 to 1/6 

Detailed explainer: https://instant-frozen-tab-idb.stopa.chatgpt.site

# Future Work 

This still may not be enough. It's still possible that we try to do a very large write (with lots of pending mutations), and just at that moment Android freezes us. 

To avoid this, we would have to start saving each transaction as a separate row, so we could start to bound writes to pendingMutations. 

This may be a bit more involved, so I wanted to first ship this version, before experimenting with a full on rewrite.

@dwwoelfel @nezaj

### DIFF
--- a/client/packages/core/__tests__/src/utils/PersistedObject.test.ts
+++ b/client/packages/core/__tests__/src/utils/PersistedObject.test.ts
@@ -1,5 +1,5 @@
 import 'fake-indexeddb/auto';
-import { test, expect, describe } from 'vitest';
+import { test, expect, describe, vi } from 'vitest';
 import { PersistedObject } from '../../../src/utils/PersistedObject';
 import { IndexedDBStorage } from '../../../src';
 import { randomUUID } from 'crypto';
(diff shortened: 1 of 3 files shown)


dwwoelfel (approved):
Cool!

PR #2857, opened 2026-08-11
[Docs] Add links for superuser and restore

nezaj:
What it says on the tin!

CleanShot 2026-08-11 at 15 01 50@2x


stopachka (approved):
SGTM!

PR #2859, opened 2026-08-12
Lightweight check for aggregator slot

dwwoelfel:
When we built the singleton invalidator, we introduced a lighter-weight check to see if the slot was active. This PR applies that same strategy to the aggregator slot.

Each machine will periodically try to take the aggregator slot, unless it's the machine currently holding the slot.

Before this PR, the machine would try to create replication stream. That requires creating a new postgres connection issuing a `START REPLICATION` command. When it fails (which it will do most of the time), it logs a noisy error into the postgres log. 

After the PR, we'll just issue a query with the existing connection pool to check if the slot is active. If it's not, then we'll go to the trouble of creating the postgres connection and starting replication.

### DIFF
--- a/server/src/instant/custodian.clj
+++ b/server/src/instant/custodian.clj
@@ -39,7 +39,9 @@
 ;; Config
 
 (defn- batch-size [] (flags/flag :custodian-batch-size 1000))
-(defn- idle-sleep-ms [] (flags/flag :custodian-idle-sleep-ms 5000))
+(defn- idle-sleep-ms [] (flags/flag :custodian-idle-sleep-ms (if (config/dev?)
+                                                               (* 1000 60 10)
+                                                               5000)))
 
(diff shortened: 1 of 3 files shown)


server/src/instant/custodian.clj
```diff
@@ -39,7 +39,9 @@
 ;; Config
 
 (defn- batch-size [] (flags/flag :custodian-batch-size 1000))
-(defn- idle-sleep-ms [] (flags/flag :custodian-idle-sleep-ms 5000))
+(defn- idle-sleep-ms [] (flags/flag :custodian-idle-sleep-ms (if (config/dev?)
+                                                               (* 1000 60 10)
```
dwwoelfel, line 43: This was getting a little noisy in dev, so I reduced the frequency.

stopachka (approved):
Awesome!

PR #2861, opened 2026-08-12
Enable size sweeper by default, prevent deadlocks

dwwoelfel:
Enables the triples_size_updates sweeper by default (we use it to show storage and db usage).

Also sorts the updates to the aggregate table to prevent deadlocks.


nezaj (approved):
🚢

PR #2865, opened 2026-08-17
Fix error messages in self-hosted

dwwoelfel:
There are a lot of error messages that say things like "ping Joe and Stopa" or "ping us on discord".

This PR updates the messages in self-hosted to say something like "contact your deployment administrator".

Also enables the `debug-uri` route in self-hosted. If you're an admin, it will show the traceId and include a message that they should search for the traceId in the logs.


stopachka (approved):
Niice!

PR #2877, opened 2026-08-25
Set next database cluster

dwwoelfel:
Set the next database cluster in preparation for a database upgrade.

We'll run slot failover and the next PR will update the slot-num in code.

Then we can run failover and deploy a PR to promote the replica to the primary.


stopachka (approved):
niice!

PR #2878, opened 2026-08-25
Bump slot nums

dwwoelfel:
To be deployed after running through `migrate-invalidator-to-replica` and `migrate-aggregator-to-replica`.


nezaj (approved):
🚢

PR #2880, opened 2026-08-25
Promote replica to primary

dwwoelfel:
Final step in the migration, to be deployed after running `do-failover-to-new-db`.


nezaj (approved):
LGTM!

PR #2883, opened 2026-08-28
Smaller backup clone instance

dwwoelfel:
Reduce the instance from an `r8gd` to an `r8g`. Removes the ephemeral disks, which don't help us for the copy scan.


stopachka (approved):
Awesome

stopachka (commented):
heck yeah!

PR #2884, opened 2026-08-28
Increase idle_in_transaction timeout for snapshot copy

dwwoelfel:
Increases the idle_in_transaction timeout for the snapshot copy so that we can run the copy on the prod db instead of creating a clone.


stopachka (approved):
Big timeout!

PR #2888, opened 2026-08-31
Add devtool support for self-hosted

nezaj:
Got a report that devtool doesn't play nice with self-hosted Instant. This PR makes it so!

We introduce an optional `dashURI` to the existing `devtool` config. When this is set we'll connect the devtool to the provided dash url instead of localhost or production instant.

Updated `create-instant-app` to inject this config too when invoked with `INSTANT_CLI_DASH_URI` so you don't need to manually edit `init` each time you create an app.

### DIFF
--- a/client/www/app/docs/self-hosting/page.md
+++ b/client/www/app/docs/self-hosting/page.md
@@ -144,27 +144,7 @@ INSTANT_CLI_DASH_URI=https://dash.myinstant.com \
 npx instant-cli@latest login
 ```
 
-After authenticating, you can use the auth token associated with your self-hosted
-Instant in `create-instant-app` by setting `INSTANT_CLI_API_URI`:
-
-```shell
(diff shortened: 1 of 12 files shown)


dwwoelfel (approved):
LGTM!

client/www/app/docs/self-hosting/page.md
```diff
@@ -144,27 +144,7 @@ INSTANT_CLI_DASH_URI=https://dash.myinstant.com \
 npx instant-cli@latest login
 ```
 
-After authenticating, you can use the auth token associated with your self-hosted
```
nezaj, line 147: Simplified documentation here. You will most likely want to supply both `INSTANT_CLI_API_URI` and `INSTANT_CLI_DASH_URI` when using `create-instant-app` with self-hosted. So we should just talk about that

PR #2919, opened 2026-09-17
Give the entity delete's reverse-ref lookup an indexable predicate

stopachka:
`delete-entity-multi!` finds the refs that point at the deleted entities by joining `triples` to the `id_etypes` CTE on `json_uuid_to_uuid(value)`. The planner assumes `jsonb_array_elements` returns 100 rows, so for an app with many refs it picks a hash join and reads every `vae` triple in the app instead of probing `vae_uuid_index` per entity.

After the upsert fixes in #2915 and #2918 this statement became the largest write-side item in prod: 8.5% of database execution time and 16% of physical reads on 2026-09-17, with deletes from one app running about 2 s each with no wait event.

Read-only `EXPLAIN (ANALYZE, BUFFERS)` of the reverse-ref lookup in prod, for an app with 1.36 M ref triples and two entity ids:

| | plan | shared buffer hits | time |
| --- | --- | ---: | ---: |
| current | Parallel Bitmap Heap Scan of all the app's `vae` triples, hash join | 646,030 | 1,100 ms on 3 workers |
| with the predicate | Nested Loop, `Index Scan using vae_uuid_index` per entity | 10 | 0.2 ms |

The change, behind a `delete-entity-indexed-ref-lookup` toggle that defaults off, adds a predicate the join already implies:

```sql
AND json_uuid_to_uuid(triples.value) = ANY(ARRAY(SELECT entity_id FROM id_etypes))
```

Every row the join can produce already satisfies it, so results are unchanged. I tried pg_hint_plan first (aliases plus `Leading`/`NestLoop`), but the CTE alias is inlined and the join hints did not apply; the redundant predicate does not depend on alias resolution.

Validation: a new test runs the same deletes with the toggle on and off and asserts identical remaining triples, including a delete with an etype that does not own the refs. `instant.db.transaction-test` passes (105 tests, 832 assertions). CI clj-kondo config is clean on the changed files.


stopachka:
Deployed to prod on 2026-09-17 at 07:03 UTC as `app-delete-ref-lookup-20260917-b9cf815bd266`; `delete-entity-indexed-ref-lookup` toggle enabled at 07:11 UTC and kept on after a one-hour watch.

| entity delete statement | calls | ms per call | shared buffer hits per call | max ms |
| --- | ---: | ---: | ---: | ---: |
| old text, since instance start | 92,204 | 62.3 | 10,419 | 29,293 |
| new text, first hour | 6,223 | 1.2 | 77 | 369 |

The delete class went from about 7% of database execution time to under 1%. No 5xx. The only transact exceptions in the hour were one app's pre-existing permission-rule error (`live_users.delete`, 209,736 of them in an hour the day before, thrown before any delete SQL runs) and 18 ordinary permission denials.

PR #2920, opened 2026-09-17
Let a paged numeric range stop at the page instead of sorting every match

stopachka:
After the write-side fixes, reads are about 90% of the database's non-IO time. The single largest statement is a sync query that three apps share (a15fca0e, f5d067f2, 26f1cf25): a numeric range on one attribute, ordered by that attribute, 500 rows per page with a cursor.

```clojure
{:prices {:$ {:where {:and [{:modified {:$gt a}} {:modified {:$lte b}}]}
              :order {:modified "asc"} :first 500 :after cursor}}}
```

Each page scans every row above the lower bound, looks each row up twice more (`t1`, `t2`), sorts all of them, and keeps 500. Over 8.6 hours on 2026-09-17 that one statement was 14.6% of all shared buffer hits and 6.6% of non-IO execution time, at 1.5 calls per second.

The filters, the order, and the cursor all read the same single-valued attribute, so `t0`, `t1` and `t2` are the same triple. For apps enabled in a new `scoped-query-plans` flag (plan name `numeric-range-page`), the plan:

- adds the upper bound to the first scan, like the existing `numeric-range` plan does for one app
- makes the page query and the previous-page check order by the first scan's columns, so the planner keeps `triples_number_type_idx` order and the limit stops the scan early (incremental sort on the entity id for ties)
- for a required attribute, also bounds the first scan with the cursor value, inside the page query only. `m-0` is shared with the previous-page check, so its definition does not get the cursor bound.
- lets the child join reuse the page's entity ids, which `reuse-bound-child-entities?` already does for one other app

Read-only in prod with real parameters, both queries inside one repeatable-read snapshot:

| | shared buffer hits | execution time |
| --- | ---: | ---: |
| current | 223,024 | 197 ms |
| with the plan | 14,141 | about 30 ms |

The 500 page rows are identical and in the same order, `has-next` and `has-previous` match, and the 3,000 child triples match as a multiset (their order changes, the same as the existing child-entity reuse).

Flag shape, same as `scoped-write-plans`:

```json
{"a15fca0e-3517-40ba-b286-774ae9a46c22": {"numeric-range-page": true}}
```

`disable-scoped-query-plans` and `disable-pg-hints` turn it off. The plan only applies when the scans, CTE layout, predicates and attribute (cardinality one, checked number) match; anything else keeps the normal SQL.

Validation: the new `numeric-range-page-plan-test` checks the rewritten CTEs, the gates, nearby shapes that must not change, and pages through real Postgres data with ties across page boundaries, empty and inverted ranges, comparing every page and its page info with the plan on and off. It passes together with the existing scoped plan tests (37 tests, 369 assertions) and `instaql-test` plus `datalog-test` (72 tests, 593 assertions). CI clj-kondo config is clean.


stopachka:
Prod results, 2026-09-17. Deployed 17:41 UTC and enabled through `scoped-query-plans` for the six apps that run this query (a15fca0e, f5d067f2, 26f1cf25, c0045256, c8a71cbe, 83e2c074).

| statement | before | after |
| --- | ---: | ---: |
| page with cursor, per call | 91.5 ms, 89k buffer hits | 25.2 ms, 18k |
| first page, per call | 31.3 ms, 31k | 5.5 ms, 4.2k |
| a15fca0e page latency in app logs | 141 ms | 53 ms |
| c0045256 page (explain) | about 1,030 ms | 20 ms |

No query errors for these apps in the hour after enabling.

PR #2922, opened 2026-09-17
Let a flag put selective where keys first

stopachka:
The SQL we generate starts from the first key of a `where` map and joins the other conditions onto it. Nothing reorders conditions by how selective they are, so the order a client happens to write its keys in decides how much work a query does.

Two prod queries, run through `instaql/explain` with the same keys in a different order:

| app | query | as written | selective key first |
| --- | --- | ---: | ---: |
| 41467407 | `taskInstances` by `record.organization.id`, `record.asset.id`, status, key | 3,793 ms, 4.73 M buffer hits | 1.2 ms, 263 hits |
| 1c436238 | `Prices` by `pairId` and `timestamp`, limit 1 | 84 ms warm and up to 9 s cold, 124 k hits | 26 ms cold, 290 hits |

The first one starts from the organization (182 k records) and only filters down to the one asset at the end. Its two statements are about 4.5% of the database's non-IO time.

This adds a `where-order` flag that lists, per app and namespace, the where keys to move to the front:

```json
{"41467407-d2cf-4e9a-8509-19659a7e2885": {"taskInstances": ["record.asset.id"]},
 "1c436238-c543-44d0-9a6b-51f7e5b840e3": {"Prices": ["timestamp"]}}
```

Where conditions are ANDed, so every key order is a query a client could already send, and the results are the same. Only the top-level `where` map of the listed namespace is reordered. Nested `and`/`or` clauses, other namespaces and other apps are untouched. The query hash and normalized form do not depend on key order, so existing per-hash flags keep working. It is keyed by namespace instead of query hash because hashes change with the length of each `$in` list.

Tests cover the reordering, string keys, queries that do not match, and that reordered queries return the same rows in Postgres.


stopachka:
Prod results, 2026-09-17. Deployed 21:06 UTC, flag set at 21:07 for four apps.

| query | before | after |
| --- | ---: | ---: |
| 1c436238 `Prices` by pairId and timestamp (app logs) | 3,407 ms | 18 ms |
| 41467407 `taskInstances` (app logs) | 3,469 ms | 84 ms |
| 543ea45b `externalLink` by localId (app logs) | 5,003 ms | 16 ms |
| 8f3bebac `typingStates` (app logs) | 35 ms | 5 ms |

Some reorders depend on the value and were left alone: `messages` with `time` first is 78 ms to 5 ms for a recent bound but scans every chat for an old one, and `cronRuns` with `status` first is 35x worse. A sketch-driven ordering of where conditions would be the general fix.

